Issue #9, opened 2016-08-10
Downward facing sprites for Tajarn are larger than others.


Brimcon:
Picture related, also left/right is bigger at the feet oddly.
![image](https://cloud.githubusercontent.com/assets/7198283/17565928/4197c7de-5eee-11e6-991f-ca4e248b42dc.png)

TalkingCactus:
Does this happen in-game, too? Or is it just in the preview?

Brimcon:
In game as well.
![image](https://cloud.githubusercontent.com/assets/7198283/17566580/df0187f6-5ef0-11e6-9ce7-1952b6ae34a9.png)

Brimcon:
Update on Tajarn sprite issue for your reference 
![image](https://cloud.githubusercontent.com/assets/7198283/17568587/51f8d572-5ef9-11e6-8ae5-ea7ac1fa9333.png)

TalkingCactus:
Fixed on my end, will be in the next update.

PR #10, opened 2016-08-10
Sprites

TalkingCactus:
Adds a lot of missing sprites from the previous port, more to come.

Fixes (maybe) Tajarans and Huskies having displaced body parts.


TalkingCactus:
Fixes #9

PR #13, opened 2016-08-11
Lobby/Title screen music

ghost:
Some new tunes for the lobby
1: System Shock 2 - Operations 2
2: David Bowie: - Space Oddity Instrumental
3: Transistor OST - Old Friends


Zeldazackman:
Why not keep the other title songs in while adding these to the list at the same rarity as the clown title song?

TalkingCactus:
> Why not keep the other title songs in while adding these to the list at the same rarity as the clown title song?

I agree, sort of. Just have them all be the same rarity. Clown one can stay at 7, though.

ghost:
Thanks.

PR #14, opened 2016-08-12
Changes Species to Mutant Races & MORE

TalkingCactus:
- [x] Races converted to new system
- [x] Three mutant race colors
- [x] Eyes, underwear, lips, and mutant bodyparts render correctly at their appropriate layers.
- [x] Several holodeck layouts now function correctly #11 
- [x] Ears/Tails usable for all appropriate races
- [x] Some gigantic amount of bugs smashed
- [x] Framework for future species
- [x] Game balancing, too much to cover
- [x] Removes vore

Character Creation Screen
Species List

PR #15, opened 2016-08-12
Re added races

thingpony:
:cl: Re added races
Re added : Fennec, Skunk, Squirrel, and Porkupine 
/:cl:


TalkingCactus:
Both changed files need to be in their correct folders or they will not work and may cause issues in game.

Furthermore, all species need to be moved to the mutant races system. See species.dm, species_type.dm, sprite_accessories.dm, and other files I changed in #14

PR #28, opened 2016-08-31
Science nerfs

Poojawa:
Removes firing pins from RD locker. Seriously no.

Removes assault rifle ammo from protolathe.


code/modules/cargo/packs.dm
```diff
@@ -205,7 +205,7 @@
 
 /datum/supply_pack/security/laser
 	name = "Lasers Crate"
-	cost = 1500
+	cost = 3500
```
TalkingCactus: I don't think we really need to punish cargo like this. Punish the players who abuse it. Having affordable emergency equipment is often times necessary. Not to mention plenty of antags can clear an armory in a matter of minutes before they go loud.

Poojawa:
It's so easy to make money though. since with goofnomics you can pretty much just sell everything botany makes for infinite cash. One good haul from mining sets up science for the round if they update the ore box. then you just sell off stuff as the stock market computer says to.

TalkingCactus:
Can't agree with these changes until you revert the cargo prices.

Poojawa:
Cargo can purchase guns and firing pins for less than a bunch of pizzas. Totally balanced.

Whatever.

TalkingCactus:
We can debate a solution to the problem in another PR, but I think this one is ready like this. Still on the fence about the 550 magazines since you can't get those anywhere else, but we can do that later. Double check if it compiles then merge.

PR #30, opened 2016-09-02
Bob Barker lessons

Poojawa:
https://www.youtube.com/watch?v=ZytpnElPI_E

It compiled, just ommitted cat family cloning and Lisa's baby factory ability


TalkingCactus:
So does it work?

Poojawa:
Yarp

PR #36, opened 2016-09-06
shuttle auto call

LetterJay:
adds the auto call back from 5th. pretty basic, doesn't need much explaining. 3 hours is the auto call.


LetterJay:
Self-commit.

PR #37, opened 2016-09-06
Latest useful refractors

Poojawa:
Might as well attack lag at the source.

Compiles just fine, no errors.


Poojawa:
Will double check that everything needed for this is actually included and running. because Murphy is probably lurking.

Poojawa:
Gunna just roll updates from TG when we've got everything finished

PR #53, opened 2016-09-23
Admin logging

Poojawa:
Porting Admin PM ticket system from Hippie Station

Needs testing due to local host auto-grant of administration powers.


Poojawa:
Tested and working. Totally knew I could just.

y'know

_Deadmin myself_

hue.

PR #33, opened 2016-09-04
Races Continued

TalkingCactus:
This adds support for species that are locked to certain players. Also included is another batch of species which are playable by everyone, full list coming soon. Snout/Jawline color options will soon be selectable for canids/felids.
- [x] Framework for donator races
- [x] Exotic species framework
- [x] Race specific scream/roar sounds. Code refactored and made not shit. See emote.dm
- [ ] Organize species.dm and species_types.dm to make them readable to non-coders
- [ ] Birds
- [ ] Sharks/Carp
- [ ] Snout/Jawline color options (see 5th port wolves, foxes, etc.)

PR #58, opened 2016-09-25
New preference system

Poojawa:
Hoooooly Shit boy-os and G.i.r.l.s

DAT Polaris preferences system. It's happening.

Gunna be TG-ifying most of the code from RBG to hex value colors, etc. y'know, normal stuff.
- [ ] Ensure all files and code is in place
- [ ] Squash Error codes.
- [ ] Rage about it in coderbus
- [ ] Separate Mutant Race to it's own thing if needed
- [ ] Scream and rage over byond code again
- [ ] Test it again after smashing face into keyboard
- [ ] Admit defeat for like, a day or two.
- [ ] Actually finish it


TalkingCactus:
This will conflict with #33 pretty hard. I'll push it soon so we don't end up in conflict hell later.

Poojawa:
do this later.

PR #65, opened 2016-10-28
Ports n' Stuff

ghost:
- Adds powertools

image
- Adds bacon

image
- Adds spookier lavaland music
- Adds a desertion sound that is played for every player who is abandoned on the station when the shuttle leaves

Powertools, bacon, and lavaland music change are from /tg/station. All credit for those goes to the original creators.


.gitignore
```diff
@@ -181,3 +181,5 @@ Temporary Items
 .apdisk
 
 *.before
+
+.atom-build.json
```
TalkingCactus, line 185: What exactly is this file?

ghost: It's a configuration file for an Atom package _(I've now switched to Github's Atom editor because it's infinitely better than Dream Maker)_ that basically tells it what to compile and with what compiler, instead of compiling from Dream Maker.
http://i.imgur.com/WtmmwFO.png
Github is making me either commit the file itself or add it in the .gitignore and commit that.

PR #64, opened 2016-10-28
Everything I Absolutely Will Not Discuss With Anyone Under Any Circumstances

TalkingCactus:
This PR will include the framework for the thing everyone does but doesn't talk about. The system will attempt to be as modular as possible (before I go insane) so others can incorporate their own things.

Built from the ground up without any of that garbage written by that garbage coder that was around a while ago or something.

Also rebuilds tgui for Pooj since I was having a few bugs that might be related to that. May or may not fix it. Should at least fix medihound sleeper ui.
Also adds:

- [x] Maintenance Bar
- [x] KinkMate
- [x] Door buttons are fixed, will open all linked doors at the same time
- [x] A bunch of bug fixes, re-balances, etc
- [x] Stuff I won't talk about 
- [ ] Overlays for junk
- [ ] Character preferences

PR #69, opened 2016-11-07
Space Parallax Port

ghost:
This ports /vg/'s Space Parallax system.
Now tested and working.

Space parallax at default speed, with space dust:
https://www.dropbox.com/s/62k04npb8ynzmzr/2016-11-07_06-52-02.mp4?dl=0

As you can see, it doesn't work while ghosted, but that's probably for the better IMO.

~I'll need someone to make a few preferences in character setup. The initial values for the parallax, parallax speed, and space dust should already be in client/preferences.dm~

This also makes space totally black rather than an gray because it looks better that way.

I'm really surprised this works at all. If all goes well, I may try to port planes as well.


TalkingCactus:
How does this handle sloped walls on shuttles?

TalkingCactus:
Also for the savefile go here and just copy the style that's used for those.
`S["space_parallax"]            >> space_parallax`
`S["space_dust"]            >> space_dust`
`S["parallax_speed"]            >> parallax_speed`
Insert ^that into this proc where all the other lines like this are.
https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/code/modules/client/preferences_savefile.dm#L218

Then do the same thing except reversed essentially in this proc https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/code/modules/client/preferences_savefile.dm#L280
`S["space_parallax"]            << space_parallax`
`S["space_dust"]            << space_dust`
`S["parallax_speed"]            << parallax_speed`

If they need to be sanitized then add that as well around line 265

ghost:
This last commit will made the empty side of sloped shuttles empty, which shows as black. There isn't really any way to make it work perfectly with parallax as far as I know, so this is the best solution. It blends in fairly well and looks better than if there was static space on the other side.

ghost:
And for your suggestion on the savefile, that won't actually show anything in character setup, will it? Because that's what we want to have.

TalkingCactus:
No, that will just make sure it can be saved/loaded.

ghost:
Well, not sure what to do then because I've never screwed around with that stuff

Just added those pref values in the style of the rest, space_parallax and space_dust should be toggles while parallax_speed is a number between 0 & 5

TalkingCactus:
That's preferences.dm, I edited it in #64 if you need an idea for how it works. This setting would likely be on the game settings page of the character creator whatevermabober.

ghost:
Should be good now.

code/modules/client/preferences.dm
```diff
@@ -448,6 +453,9 @@ var/list/preferences_datums = list()
 			dat += "<b>Ghost pda:</b> <a href='?_src=prefs;preference=ghost_pda'>[(chat_toggles & CHAT_GHOSTPDA) ? "All Messages" : "Nearest Creatures"]</a><br>"
 			dat += "<b>Pull requests:</b> <a href='?_src_=prefs;preference=pull_requests'>[(chat_toggles & CHAT_PULLR) ? "Yes" : "No"]</a><br>"
 			dat += "<b>Midround Antagonist:</b> <a href='?_src_=prefs;preference=allow_midround_antag'>[(toggles & MIDROUND_ANTAG) ? "Yes" : "No"]</a><br>"
+			dat += "<b>Space Parallax:</b> <a href='?_src_=prefs;preference=parallax'>[space_parallax ? "Enabled" : "Disabled"]</a><br>"
+			dat += "<b>Parallax Speed:</b> <a href='?_src_=prefs;preference=p_speed'>[parallax_speed]</a><br>"
```
TalkingCactus: Maybe run an `if(space_parallax)` before this and the next thing so they only show up if you're using parallax.

ghost:
Done.

ghost:
~~oh hey it doesn't work 1 second~~

ghost:
actually nevermind i didn't compile yet i'm stupid

TalkingCactus:
I'll merge this soon, just want to finish something on my PR to make merge conflicts easier to deal with.

PR #70, opened 2016-11-15
Cautery update

katnep:
you can now cauterize with a laser gun now
just seemed like a good idea


ghost:
this would be an incredibly bad idea realistically because you'd be firing an entire laser round through someone by trying to cauterize something on them

katnep:
the justification that the TG threads made was that it's basically a heated gas burst directly to the wound to seal it shut

ghost:
fair enough

PR #73, opened 2016-11-21
Throwing weapons price rebalance

katnep:
changed throwing weapons price from 5 to 3 since while it's decent as it is, you could get an energy sword for 3 more and do ALOT more damage as well as the fact that throwing weapons are really only one time use because once you've thrown them your enemy has them


katnep:
hmm it added the changes from my other branches :/ one sec going to fix this

katnep:
time to try again

PR #74, opened 2016-11-21
throwing weapons price edit

katnep:
changed throwing weapons price from 5 to 3 since while it's decent as it is, you could get an energy sword for 3 more and do ALOT more damage as well as the fact that throwing weapons are really only one time use because once you've thrown them your enemy has them


TalkingCactus:
# P O W E R C R E E P

TalkingCactus:
5 TC seems like a decent price for throwing stars. In the right hands they're better than an esword.

katnep:
i can't recall from memory but i believe there's been bolas added to to it, which is something anyone can make in 6th so it sort of devalues the kit as a whole in my opinion

PR #75, opened 2016-11-21
added syndrones

katnep:
can now purchase syndrones for 15 TC, probably gonna find a lot of use

### DIFF
--- a/code/modules/uplink/uplink_item.dm
+++ b/code/modules/uplink/uplink_item.dm
@@ -509,6 +509,13 @@ var/list/uplink_items = list() // Global list so we only initialize this once.
 	item = /obj/mecha/combat/marauder/mauler/loaded
 	cost = 140
 
+/datum/uplink_item/support/syndrone
+	name="syndrone"
+	desc="A drone that when summoned will be able to listen to carry out your orders, or just sow destruction (results may vary)"
+	item = /obj/item/drone_shell/syndrone
+	cost = 15
+	surplus = 25
+
 // Stealthy Weapons
 /datum/uplink_item/stealthy_weapons
 	category = "Stealthy and Inconspicuous Weapons"


code/modules/uplink/uplink_item.dm
```diff
@@ -509,6 +509,13 @@ var/list/uplink_items = list() // Global list so we only initialize this once.
 	item = /obj/mecha/combat/marauder/mauler/loaded
 	cost = 140
 
+/datum/uplink_item/support/syndrone
+	name="syndrone"
+	desc="A drone that when summoned will be able to listen to carry out your orders, or just sow destruction (results may vary)"
```
ghost, line 514: Fix this sentence.

ghost (commented):
A ghost will have to click on the drone or else it won't do anything, just so you know.

TalkingCactus:
I don't think you fully grasp how powerful a syndrone is if you're going to put the power to create one in the hands of the player. This will not be merged in it's current state.

code/modules/uplink/uplink_item.dm
katnep: do you suggest i add in the fact that it requires a ghost to function as well?

katnep:
a few balance ideas to prevent it from being the go to choice and to make it worth considering other options instead
A. boost the price to around the same as a borg or higher
B. make it so it can only be used at a certain population thresh hold, same way that nuclear war is

TalkingCactus:
Price won't solve the problem. Lazy traitors will always purchase them just for shits and giggles.

katnep:
also forgot to add this option
C. restrict it to newcops

TalkingCactus:
Syndrones do not have an appropriate lawset to work with anyone or anything else.

katnep:
i could put in a little more work and make a copy that has laws that mean it has to work with the nuke ops maybe also nerf it slightly from the regular version

TalkingCactus:
I really don't think we need this. Leave syndrones in the bin of admin shitlery tools.

PR #82, opened 2016-11-29
Thermite reskin

katnep:
I reskined the thermite becuase crow challenged me to do a better job will include a link in a sec


katnep:
image

TalkingCactus:
What if you sprited it completely different from a normal spray can and made it auto-light the thermite on the wall after spraying. Or made it just a normal spray bottle so it can be used as a stealth item.

katnep:
i mean i could i was literally just doing this because crow challenged me to and it turned out alright

TalkingCactus:
2 missing pixels, though

katnep:
those dots were remenents of the line across the bottle one sec

katnep:
image

katnep:
i think the point of it is to go loud with a hole in the wall though so making it pretty much undetectable would make it a bit overpowered

ghost:
the bottle itself is supposed to be a knockoff of the space cleaner bottle as sort of a joke which this still achieves

katnep:
syndicatebrand- TOTALLY SPACE CLEANER

Axolotl123:
The spritzer of it should atleast be a different color, like the color of an actual space cleaner bottle.

PR #89, opened 2016-12-10
Wizard Pontiac Firebird AKA Wizmobile port

ghost:
Originally from /vg/.
Adds the "Chariot" item to the wizard spellbook, in the mobility section. When bought, it will summon a Pontiac Firebird Trans Am along with the keys at the player's location.

The Firebird can travel very quickly and drive through space. It destroys any tables or racks it hits., and leaves behind a trail of exhaust fumes as it moves. There's also an engine sound loop and startup sound.

In action:
https://gfycat.com/FamousGiantBarnacle

***GAS GAS GAS***
**I'M GONNA STEP ON THE GAS**


code/modules/vehicles/firebird.dm
```diff
@@ -0,0 +1,81 @@
...
+	duration = 10
+	randomdir = 0
+
+/obj/effect/overlay/temp/firebird_trail/New(loc,move_dir)
+	..()
+	setDir(move_dir)
+
+/obj/vehicle/firebird/Move(newloc,move_dir)
+	if(has_buckled_mobs())
+		PoolOrNew(/obj/effect/overlay/temp/firebird_trail,list(loc,move_dir))
```
TalkingCactus: I almost wish this was fire, but that thing moves fast as fuck

ghost: I can add a small first trail that only lags behind like 2 tiles
made of fire ofc

TalkingCactus: aw yis

ghost:
this /datum/action is not working correctly and i can't figure out why

ghost:
finally fixed it, should be good now

PR #92, opened 2016-12-11
Makes sideways fox tails a bit more pointy in their neutral state

ghost:
http://i.imgur.com/nxbOdzH.png
2 pixel difference at the end of the tail to make it slightly more pointy and a little less like a beaver's tail


TalkingCactus:
y tho
(before and after would be 👌 )

ghost:
doesn't make the difference i thought it would
they just look too round :v

TalkingCactus:
You were in the wrong file anyway.

PR #93, opened 2016-12-16
New supermatter ambience

ghost:
Adds an ambient sound whenever the supermatter is glowing (when it's powered and using a decent amount of oxygen) using the same loop in the firebird engine sound.

Sound:
https://my.mixtape.moe/cavkdg.ogg

It's played every 1.2 deciseconds so that there won't be much of a delay between the position of the sound as the client hears it when you're moving around it, and it loops just fine.

That probably sounds laggy but BYOND is super robust with sound playing and it seems to work just fine.


ghost:
whoops lmao

PR #94, opened 2016-12-16
Various lighting fixes.

### LINKED ISSUE #67: Lighting Bugs
Here's a list of anomalies that I've noticed so far.

1. Shuttles: shuttle
     Shuttles have always been awful for lighting, even in the last system. The solution use to just be to make shuttles always illuminated. There must be a better solution than this, though.

2. Roundstart: When the round starts, the lighting builds itself. This should happen in the pregame lobby during initialization.

**rest of issues have been resolved**

PJB3005:
Fixes #67

### DIFF
--- a/code/modules/lighting/lighting_area.dm
+++ b/code/modules/lighting/lighting_area.dm
@@ -18,12 +18,12 @@
 	dynamic_lighting = new_dynamic_lighting
 
 	if (IS_DYNAMIC_LIGHTING(src))
-		for (var/turf/T in (area_contents()))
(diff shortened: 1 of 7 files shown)


PJB3005:
Shuttles *seem* to be fine with lighting so you can turn that on if you want, though it might shit itself later.

TalkingCactus:
Does this allow fixed wall lights to function correctly as well?

TalkingCactus:
Regardless, I tested it and it seems to work fine. Thanks for the help.

PR #97, opened 2016-12-17
Strangling buff

ghost:
Makes strangling cause you to suffocate a bit more, stutter for a few seconds, and then go mute shortly after.

"but this is too op"

No, it isn't, because as long as you're conscious and not stunned or paralyzed in some way, you can resist any grip to try and break free, not to mention it takes a few moments to even start strangling someone. 
This still gives you a few seconds to cry for help as well, though you'll be stuttering.
It's also more realistic tbh.

Also this makes you go mute if you're suffocating badly in any case.

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -56,8 +56,13 @@
 	//Suffocate
 	if(losebreath > 0)
 		losebreath--
+		if(losebreath >= 3) //can't talk from lack of air
+			silent = max(silent, 2)
 		if(prob(10))
 			emote("gasp")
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/life.dm
```diff
@@ -56,8 +56,13 @@
 	//Suffocate
 	if(losebreath > 0)
 		losebreath--
+		if(losebreath >= 3) //can't talk from lack of air
+			silent = max(silent, 2)
```
Iamgoofball, line 60: nice buff to all martial arts

Iamgoofball: this is also going to have a lot of wide-range balance effects you may not have forseen

PR #109, opened 2016-12-25
change null to 0 in lighting/lighting_source.dm

LetterJay:
Changes null values to 0. Seems to have fixed the runtime flooding, but I can't confirm without a large playerbase to test with.

### DIFF
--- a/code/modules/lighting/lighting_corner.dm
+++ b/code/modules/lighting/lighting_corner.dm
@@ -53,7 +53,7 @@
 	T = get_step(new_turf, diagonal)
 	if (T) // In case we're on the map's border.
 		if (!T.corners)
-			T.corners = list(null, null, null, null)
+			T.corners = list(0,0,0,0)
 
 		masters[T]   = diagonal
 		i            = LIGHTING_CORNER_DIAGONAL.Find(turn(diagonal, 180))
@@ -63,7 +63,7 @@
 	T = get_step(new_turf, horizontal)
 	if (T) // Ditto.
 		if (!T.corners)
-			T.corners = list(null, null, null, null)
+			T.corners = list(0,0,0,0)
 
 		masters[T]   = ((T.x > x) ? EAST : WEST) | ((T.y > y) ? NORTH : SOUTH) // Get the dir based on coordinates.
 		i            = LIGHTING_CORNER_DIAGONAL.Find(turn(masters[T], 180))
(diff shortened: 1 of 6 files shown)


PJB3005:
The zeroes won't accomplish anything.
The nullcheck is in the wrong spot and won't do anything either.

So after actual months of this runtime throwing on /vg/ (and being surpressed by the error handler) I think I finally figured out why it could happen.

I didn't want to fix it like this, because I thought the problem lied somewhere else, but I was wrong so here goes:

nullcheck `S.source_turf` here: https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/code/modules/lighting/lighting_turf.dm#L49

LetterJay:
Sorry, I'm gonna sound dumb asking this. But, you want me to place ```if(!S.source_turf)``` on line 49? I can't remember how to do a nullcheck and no one that's online can either.

PJB3005:
Yes, exactly.

LetterJay:
Alright, running it. I'm still unsure of what triggers this. The runtimes comes randomly.

LetterJay:
Should I run this live with a ```return``` or ```continue```?

PJB3005:
Continue.

PJB3005:
Well then

LetterJay:
Didn't mean to close that; I need sleep.

LetterJay:
So I don't screw something up due to lack of sleep, I'm gonna let you look over if you want and make sure I didn't do something I wasn't supposed to in this PR.

PJB3005:
Only change that should actually do anything is the nullcheck for `if (!S.source_turf)`, rest can't make a difference.

LetterJay:
Yea, a few of these I don't actually remember doing. Anyway, merging. Updating, we'll see what happens from here.

PJB3005:
Just chiming in, I found the source of the runtime proper, finally.

I'll have a fix up soon:tm:

LetterJay:
The crash loop?

PJB3005:
the `null.x` runtime

LetterJay:
That's usually what floods right before the bad list runtime flood. It only locks up players. Observers and ghosts are unaffected

PJB3005:
uh.

Weird.

PJB3005:
https://github.com/d3athrow/vgstation13/pull/13291

PR #114, opened 2016-12-26
table SLAM

LetterJay:
make a very loud point.

still not ready. need to add a sleep on it so it's not spammed to shit and double check damage works when slamming glass tables.

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -186,13 +186,13 @@
 	H.adjustBrainLoss(CLONE_INITIAL_DAMAGE)
 	H.Paralyse(4)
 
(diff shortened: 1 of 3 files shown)


TalkingCactus:
don't use sleep()

LetterJay:
which is used, then?

TalkingCactus:
use a timer with `addtimer()` or something

(pretty sure that's the proc, might be a little different)

ghost:
ye, just look up add_timer or whatever it is in the code to see how it's used, make a proc with everything that needs to be run after the delay and set that as the proc for the timer to run after

LetterJay:
Fucked up something, closing.

PR #116, opened 2016-12-26
Fix lumcount issues - let's see if it works now edition

ghost:
Hopefully fixes lumcount issues with podmen, plants, etc.


TalkingCactus:
couldn't you just edit the `get_lumcount()` proc?

ghost:
does it really matter that much

PR #117, opened 2016-12-26
Gateway-stuff

ghost:
Makes gateway work.


TalkingCactus:
This is currently on the backburner until we square away some performance issues and the like.

PR #118, opened 2016-12-27
clone fix

LetterJay:
Something to do with the vore bit. Disabled it, tested a furry-based character and it both cloned with the right text as well as outputting the right species.

Tested vore after cloning, also seemed to work.


LetterJay:
Self-merging: minor tweaks, fixes important game mechanic.

PR #119, opened 2016-12-27
gamemode players required tweaks

LetterJay:
Reduces most of the gamemode's required player count a bit further down. If people have any suggestions on what to tweak more, comment here.


code/game/gamemodes/cult/cult.dm
```diff
@@ -41,7 +41,7 @@
 	antag_flag = ROLE_CULTIST
 	restricted_jobs = list("Chaplain","AI", "Cyborg", "Security Officer", "Warden", "Detective", "Head of Security", "Captain", "Head of Personnel")
 	protected_jobs = list()
-	required_players = 24
+	required_players = 20
```
Roland410, line 44: Might want to change this to 15 since clockwork is 15 too.

ghost: I don't think so, clockcult and regular cult are much different.

PR #120, opened 2016-12-27
Reduce duration of radstorm as requested by players

ghost:
Reduces duration from 60 - 150 seconds to 30 - 45 seconds
you guys are so unrobust

also i don't understand git at all so this took me like 20 minutes to properly commit without changes from another PR

### DIFF
--- a/code/datums/weather/weather_types.dm
+++ b/code/datums/weather/weather_types.dm
@@ -123,8 +123,8 @@
 
 	weather_message = "<span class='userdanger'><i>You feel waves of heat wash over you! Find shelter!</i></span>"
 	weather_overlay = "radiation"
-	weather_duration_lower = 600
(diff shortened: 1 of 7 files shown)


code/datums/weather/weather_types.dm
```diff
@@ -162,4 +162,4 @@
 	if(..())
 		return
 	priority_announce("The radiation threat has passed. Please return to your workplaces.", "Anomaly Alert")
-	spawn(300) revoke_maint_all_access()
\ No newline at end of file
+	spawn(300) revoke_maint_all_access()
```
Iamgoofball, line 165: please update to use callbacks/addtimer

ghost: this is actually legit advice but i didn't add that part it was already there

PR #123, opened 2016-12-28
Explosions Tweak

ghost:
Explosions will only shake the screen and be more obvious when devastation range is above 1.
(it should be a sign that something really bad is going down, not just a couple pieces of plating ripped off the floor)

### DIFF
--- a/code/game/objects/explosion.dm
+++ b/code/game/objects/explosion.dm
@@ -72,7 +72,7 @@
 							skip_shake = 1
 
 					if(!explosion_shake_message_cooldown && devastation_range > 1 && !skip_shake && M.stat == CONSCIOUS && !istype(get_turf(M), /turf/open/space))
-						if(istype(get_turf(M), /turf/open/floor/plasteel/asteroid))
+						if(istype(get_turf(M), /turf/open/floor/plasteel/asteroid | /turf/open/floor/plating/ash))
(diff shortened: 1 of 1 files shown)


LetterJay:
I can't see anything wrong with this.

code/game/objects/explosion.dm
```diff
@@ -64,14 +64,14 @@
...
+						if(devastation_range > 2)
 							M.playsound_local(epicenter, 'sound/effects/explosionfar.ogg', far_volume, 1, frequency, falloff = 5)
 							shake_camera(M, 3, 1)
 						else
 							M.playsound_local(epicenter, 'sound/effects/explosionsmallfar.ogg', far_volume, 1, frequency, falloff = 5)
 							skip_shake = 1
 
-					if(!explosion_shake_message_cooldown && devastation_range > 0 && !skip_shake)
+					if(!explosion_shake_message_cooldown && devastation_range > 2 && !skip_shake)
 						M << "<span class='danger'>You feel the station's structure shaking all around you.</span>"
```
TalkingCactus: This needs a  check for if you're actually in the same station/ship, conscious, etc. It also needs to be more generic (or more specific, like for planet tiles).

ghost:
Changed from 2 to 1 since it seems like a lot of bad explosions are 2 devastation range explosions stacked on top of eachother, and added something for asteroid/lavaland turf explosions

PR #127, opened 2017-01-02
port from cactus did

LetterJay:
A dump of most everything Cactus did. Species and vore still needs to be moved over, but it's getting there. This shouldn't even take a fraction of the time it did to get to 6th. this is just a fast forward. with this, all ruins can be enabled with minimal lag.


ghost:
make sure you test before you merge

LetterJay:
``Species and vore still needs to be moved over, but it's getting there. ``

ghost:
regardless
when it's done make sure it works before we break more things

PR #131, opened 2017-01-03
unfuck

LetterJay:
this unfucks what happened earlier. github crashed during a commit and everything went to shit. I lost all my work, this just reverts it back to the way it was originally.


LetterJay:
I've tested it, it works as before. Now I get to start all the fuck over.

PR #134, opened 2017-01-06
gamemode tweaks/minimap fix

LetterJay:
Drastically reduces the amount required for other gamemodes. I'll tweak this based on feedback.

Also tosses in the minimap fix from https://github.com/tgstation/tgstation/pull/20620/

Tested; works.


Roland410:
Might want to adjust the probability for the modes, since we pretty much always have traitorling, but this is a much needed change too, hopefully will be merged soon.

PR #145, opened 2017-01-13
Revert "Wizard Pontiac Firebird AKA Wizmobile port"

LetterJay:
Reverts Citadel-Station-13/Citadel-Station-13#89

compile errors


LetterJay:
@unusualcrow

ghost:
revert-[the name] contains the code prior to the revert right because in that case i'll fix it

PR #151, opened 2017-01-16
Mentor system, Tickets, and discord

Poojawa:
Ported from both Hippie for mentor system and tickets, and VGstation for discord bot framework. 

Compiles and works just fine. 

Mentor system doesn't show up in Who list, but I think that's a database problem more than anything.

Discord bot needs to be set up, so technically also a database issue I couldn't test on here.


Poojawa:
Oh that's nice, web editing probably fucked the indentations, but it only say 32 files changed instead of 2,322 lmao

Issue #152, opened 2017-01-16
something is causing memory leaks


LetterJay:
This goes on forever. Seems to be due to the poor optimization. This should be fixed in the branch update, which will have it optimized better... we hope. This isn't my project, though. I've been told it's between @unusualcrow and @Poojawa

PJB3005:
But, lighting never uses pick???

LetterJay:
I wasn't paying attention when I linked the runtime logs. I looked around and the only related thing I could find was an excess amount of memory being used which was subsequently breaking shit. Primarily things that picked from a list. @unusualcrow said something about him porting it to the wrong subsystem or something to me in a DM a while back.

ghost:
No, I said I made a subsystem but PJB has a better one now that vg switched to StonedMC

LetterJay:
@PJB3005 here's the runtime logs right before it flooded the server and crashed.

```runtime error: 
[06:22:43]Simple animal being instantiated in nullspace
[06:22:43]proc name: stack trace (/proc/stack_trace)
[06:22:43]  source file: unsorted.dm,1379
runtime error: 
[06:22:43]Simple animal being instantiated in nullspace
[06:22:43]proc name: stack trace (/proc/stack_trace)
[06:22:43]  source file: unsorted.dm,1379
runtime error: 
[06:22:44]Simple animal being instantiated in nullspace
[06:22:44]proc name: stack trace (/proc/stack_trace)
[06:22:44]  source file: unsorted.dm,1379
runtime error: 
[06:28:09]Simple animal being instantiated in nullspace
[06:28:09]proc name: stack trace (/proc/stack_trace)
[06:28:09]  source file: unsorted.dm,1379
runtime error: 
[06:28:09]Simple animal being instantiated in nullspace
[06:28:09]proc name: stack trace (/proc/stack_trace)
[06:28:09]  source file: unsorted.dm,1379
runtime error: 
[06:28:09]Simple animal being instantiated in nullspace
[06:28:09]proc name: stack trace (/proc/stack_trace)
[06:28:09]  source file: unsorted.dm,1379
runtime error: 
[06:30:38]Simple animal being instantiated in nullspace
[06:30:38]proc name: stack trace (/proc/stack_trace)
[06:30:38]  source file: unsorted.dm,1379
runtime error: 
[06:30:45]Cannot read null.actions
[06:30:45]proc name: New (/datum/action/item_action/New)
[06:30:45]  source file: action.dm,112
[...]

PR #153, opened 2017-01-16
Possible runtime fix on Lighting list

Poojawa:
I haven't anywhere near the tools to test it solo, but basically  I copied what was done ages ago on TG by adding "" to the list() lines

might be a spectacular failure, but it's a quick fix idea

### DIFF
--- a/code/modules/lighting/lighting_source.dm
+++ b/code/modules/lighting/lighting_source.dm
@@ -33,13 +33,13 @@
 /datum/light_source/New(var/atom/owner, var/atom/top)
 	source_atom = owner // Set our new owner.
 	if (!source_atom.light_sources)
-		source_atom.light_sources = list()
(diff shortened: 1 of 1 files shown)


Poojawa:
@PJB3005 If you could sanity check this, since you've been doing a lot of the heavy lifting. It's to fix #152

PJB3005:
If this does anything it sounds like a genius way to break things at best.

That issue report is far from specific on what the issue is at all.

Poojawa:
It didn't throw any errors at me, but like I said it wasn't really a good test environment since I've no idea what to spawn/do to use a ton of light sources

Just one of those, 'fuck it might just work' simple things. Damnifino.

Poojawa:
Probably not the solution we're looking for.

PR #176, opened 2017-01-27
Mentor and Admin tweaks

Poojawa:
Should help some.

Mentors/Admins speaking on mentor channel should have their own colors now

Also moves Dangerous verbs into their own tab.

Drop bomb prompt is now exitable before choosing bomb size or custom.


Poojawa:
should be good to merge/test

PR #187, opened 2017-02-01
Racial refractor

Poojawa:
Putting all the mam and lizard/human bodyparts like tails, ears, and body_markings into generalized lists. each sprite accessory will then be able to be marked by general species (human, lizard, canid, felind, avian, aquatic, insect) to be used only by the marked species. This is to prevent incorrect body_markings from being applied.

It also lays a framework to use special species to special markings/parts that no-one else will be able to get.

- [x] Refractor all Citadel specific parts to a separate file to avoid losing stuff later
- [ ] Ensure preferences.dm is taking allowed_species listing stuff.
- [ ] Ensure species are actually correctly taking their special accessories
- [ ] Find some means of 'sub species' selection or better use of body markings to have consistent colorations


Poojawa:
gunna do this better, I think

PR #188, opened 2017-02-01
size scale fixes

### LINKED ISSUE #60: Slipping/Falling Over
Slipping can cause your pixel_y and rotation to flip out. You can replicate this by slipping while you're already downed or switch your run/walk intent while you're knocked down from a slip. I don't know if this happens from things like stunbatons.

It fixes itself when you change your run/walk intent after you've stood up. While silly, this is actually pretty confusing if you're in combat or what not.

Here's a few dead syndicate who were exposed to a slipping foam grenade. They ended up becoming upright and offset down a bit. The character on the right is alive after slipping and switching move intent before standing up.
pic

Poojawa:
fixes #60

base work for shrink and purge chems. need to add mixin' chems for chemistry.

Size gun available to be built, needs research.


Poojawa:
Chems are deliberately disabled btw. just until I more fully test 'em

PR #196, opened 2017-02-04
Tail wagging works

Poojawa:
Wew

Gunna add to this PR with taurs as well, so no merging quite yet


Poojawa:
Taurs implimented, but they'll need some pixel tweaking/positioning. 

Also their layers are SUPER WIP AAAAAAAAA

But hey! for once something I tried to add to the list isn't shit anymore. :D

PR #216, opened 2017-02-12
Adds a autoimplanter to research designs

Roland410:
Adds a autoimplanter to research designs, quite high tech and material costs come with it though, but you won't need another person to implant you.


Poojawa (approved):
I'm still hesitant on having such a powerful tool available for anyone to use. But this can be reverted if my hesitation proves correct.

TalkingCactus:
I feel the same way. A testmerge would probably be the best option here.

TalkingCactus (approved):

PR #224, opened 2017-02-13
Slime/Jelly people are immune when stepping on glass shards

LetterJay:
This adds a check to slime/jelly people when stepping on glass shards so they no longer trip/take damage.

I feel like this could be a balance issue, but not a huge one. It was suggested, so here it is. I dunno if it's worth being merged or not.


TalkingCactus:
should've added this as a species trait so it can be easily applied to new species

PR #235, opened 2017-02-18
Several Balance and Convenience Changes

TalkingCactus:
Drones slowed down a smidge so they're a bit more controllable.

Swarmers slowed down to normal speed.

Industrial welders are faster, huge welders are huger.

Cake hat force reduced to 6 from 15.


Poojawa (approved):
They all seem good to me. the Cakehat still irks me, but I'm probably still being irritated by past history of seeing powergamers abuse it.

PR #261, opened 2017-03-04
Fix DIR for Waste Loop O2 Gas Filter in Atmo

Penguaro:
Changed DIR of Gas Filter for O2 in Waste Loop from 1 to 4

:cl: Penguaro
fix: Changed DIR of Gas Filter for O2 in Waste Loop from 1 to 4
/:cl:

[]: # Current Default start for TgStation (AKA BoxStation) map has the O2 Gas filter pointing the wrong direction, causing the loop to start broken and inoperable past Nitrogen filtering. Newer players or smaller rounds without an engineer may miss this and a broken loop does not seem canon to the general story.

atmo

### DIFF
--- a/_maps/map_files/TgStation/tgstation.2.1.3.dmm
+++ b/_maps/map_files/TgStation/tgstation.2.1.3.dmm
@@ -53120,7 +53120,7 @@
 /area/atmos)
 "cmq" = (
 /obj/machinery/atmospherics/components/trinary/filter{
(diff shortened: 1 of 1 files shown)


Penguaro:
Odd, apparently the map has merge issues. Will correct and re-commit.

TalkingCactus:
Your name goes after the first CL tag.

PR #270, opened 2017-03-10
Improved Atmospherics Logging

TalkingCactus:
Improves the message sent to admins when atmospherics equipment is tampered with. Includes `FLW`, `JMP`, and `?` buttons for all of them.


TalkingCactus:
travis is having a fit, but it works fine

PR #271, opened 2017-03-11
Radiation Shutter Control Fix

Penguaro:
🆑 Penguaro
fix: Changed Access Level from **24** (_Atmospherics_) to **10** (_Engine_) on **Radiation Shutters Control**
/🆑

[why]: # The Radiation Shutters Control was set to Atmospherics. If a server was of normal population, a Station Engineer wouldn't have Atmo access and thus could not close the shutters.


TalkingCactus:
We don't have these shutters anymore since our insatiable lag won't let us have fun.

Penguaro:
Bummer. I was starting to like the supermatter engine and it's quirks.

Issue #269, opened 2017-03-10
Clones have white eyes and no genitals


Little-119:
As it says in the title, freshly cloned humans have completely white eyes and lack any genitals.

PR #277, opened 2017-03-14
Cloning-Genitalia-Fix

ktccd:
:cl: Cloned Genitals Fix
fix: Various methods of cloning now gives the organs the clone's DNA says they should have and updates genital appearance along with clone updating of appearance.
/:cl:


LetterJay:
Since this is @TalkingCactus 's thing, I'll let him review/merge.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -50,6 +50,7 @@
 
 	internal_organs += new dna.species.mutanteyes()
 	internal_organs += new /obj/item/organ/brain
+	give_genitals()
```
TalkingCactus, line 57: `if(!(NOGENITALS in dna.species.species_traits))`

TalkingCactus (changes requested):

code/modules/mob/living/carbon/human/human.dm
ktccd: I'm... uhm, not sure what this means, I'm not very used to github. I don't really understand what you want changed :S.

TalkingCactus:
I don't suppose you could bundle a fix for the other half of #269 ? Clones apparently have white eyes, but I haven't seen it myself.

code/modules/mob/living/carbon/human/human.dm
ktccd: Ok, so reading up a bit... this seems like, if anything, it should be in the give_genitals() proc, rather than hardcoded in this specific place. Nothing should ever get genitals if it has NOGENITALS trait in its dna and anything that has genitals needs to have dna to begin with, right? 'cause it's a carbon/human proc?

ktccd:
There. Added sanity, removed old and harmful sanity and so on.
And yes, now the eyes fix is in here as well. It's completely unrelated to why genitalia was broken, but since it shares the same issue ticket and it seemed like a simple fix, I just added it in here as well.
Turns out, handle_body code was checking if you were NOT husked AND no disabilities before they drew eyes or lips. Only one of those things should be relevant and it's not disabilities. Blind people also have eyes.

ktccd (commented):
IDK how the request changes thing works, but I implemented the suggestion in my latest commit...

ktccd:
Travis loves me again!

LetterJay:
need an update on this

ktccd:
What'd you mean? Update how? O.o

code/modules/mob/living/carbon/human/human.dm
TalkingCactus: It checks it as a failsafe, but doing one check here instead of potentially doing it several times later is better.

ktccd: I already added a check in give_genitals now though? It's not... like... better to check outside it. The proc responsible for handling genitals should be doing the checks, right?

TalkingCactus: That works, yes. (I was referring to each individual proc inside `give_genitals()`)

TalkingCactus (approved):

PR #283, opened 2017-03-15
Ghost-role arousal fix

ktccd:
[Changelogs]: # Fixes it so ghost role spawners respect arousal preferences of the ghost.

:cl: Ghost-role arousal
fix: Ghost role spawners should now respect your arousal preferences.
/:cl:

[Justification]: # Preferences should be used or not exist, no point in ignoring them.

### DIFF
--- a/code/modules/awaymissions/corpse.dm
+++ b/code/modules/awaymissions/corpse.dm
@@ -78,6 +78,7 @@
 				MM.objectives += new/datum/objective(objective)
 		special(M)
 		MM.name = M.real_name
+		M.canbearoused = M.client.prefs.arousable
 	if(uses > 0)
 		uses--
 	if(!permanent && !uses)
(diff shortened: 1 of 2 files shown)


TalkingCactus:
Set up a check inside of `on_species_[gain/loss]()` that toggles your `canbearoused` var based on if the species has the NOGENITALS flag as well.

LetterJay:
@ktccd

ktccd:
Ok, so I'm working on making this a bit less shitty. And to also use NOAROUSAL flag, instead of NOGENITALS flag, since we apparantly have that, and that's pretty much exactly the flag it should use ^^.
Currently, this means it works for most race-changes, but I'm having problems with getting it to stick to the mob_spawn items. If I run the equip() proc on them from their spawn item, it DOES fix the setting, so something must be fudging it up elsewhere since that proc should be run in their create()...

ktccd:
This isn't what I imagine my preferred fix would be, but I know honestly FAR too little to be able to articulate properly what a solid fix would be.
What I **imagine** I'd prefer using:
Any mob/living/carbon, when it's made, would have it's arousal set based on species_traits.
THEN anytime a client is attached to such a mob, their settings is also taken into account. My problem is I have no idea WHERE this happens in the code. Or even if there is such a proc? It feels so.... hack-y to just make a check to catch the cases when the mob_spawn didn't change race...

PR #285, opened 2017-03-16
Citadel.1.1.0

Leon-Leonardo:
- fixed inaccessible chemmaster in chemistry
- moved hydroponics, replaced it with a Casino
- added Ian and spare ID's/PDA's to HoP office
- fixed the security of the Civ armory
- added cameras to bridge airlocks
- fixed CE office APC
- added some redundant powerlines
- added airlock to R&D entrance
- fixed stacking machine in cargo and made it so you can put crates on the conveyor
- added the auxiliary base
- added escape pod 4
- added the "near station" space area where appropriate
- moved mining shuttle

citadel 1 1 0


Leon-Leonardo:
Still having an issue with the navbeacons if anyone can help me with them

Set them up as best I could following what I observed on Box, but they just go back and forth between two beacons. I deleted the ones I had for now.

PR #287, opened 2017-03-16
Another attempt at fixing air.dm

Verkister:
I found out the process_cell proc did exist, but it's for turfs only. The runtiming call for it had no proper exclusion. No idea if this works, but the test runs showed no problems so far.


LetterJay:
Remote merge. We can test it next round and revert if need be.

PR #296, opened 2017-03-19
'Give' verb

LetterJay:
:cl: 
add: Adds the 'Give' verb from /vg/ and others. Simply go near a player and click the 'Give' verb (or type it) in the IC tab. The player you are trying to give the item to must have nothing in their hands and a box will appear to confirm if they want the item or not.
/:cl:

Also comments out some unused vars. From #270


LetterJay:
I'm sure there's a way to remove the need for both hands to be empty, but I'm not sure what the check would be.

Roland410:
https://github.com/Baystation12/Baystation12/blob/74b6687fe7d92f35c2b95c143e16370b33298c29/code/modules/mob/living/carbon/give.dm

Roland410:
Althought I think that's the same thing and I might be just tired.

PR #297, opened 2017-03-19
Access changes for chaplain g.tech and engineers

katnep:
changes made
chaplain has no theatre access anymore
Engineers now have atmos access
g.techs have been stripped of science access

### DIFF
--- a/code/modules/jobs/job_types/assistant.dm
+++ b/code/modules/jobs/job_types/assistant.dm
@@ -16,7 +16,11 @@ Assistant
 
 
 /datum/job/assistant/get_access()
-	return ..()
+	if((config.jobs_have_maint_access & ASSISTANTS_HAVE_MAINT_ACCESS) || !config.jobs_have_minimal_access) //Config has assistant maint access set
+		. = ..()
+		. |= list()
+	else
+		return ..()
 
 /datum/job/assistant/config_check()
 	if(config && !(config.assistant_cap == 0))
--- a/code/modules/jobs/job_types/assistant.dm
+++ b/code/modules/jobs/job_types/assistant.dm
@@ -18,7 +18,7 @@ Assistant
 /datum/job/assistant/get_access()
 	if((config.jobs_have_maint_access & ASSISTANTS_HAVE_MAINT_ACCESS) || !config.jobs_have_minimal_access) //Config has assistant maint access set
 		. = ..()
-		. |= list()
+		. |= list(access_maint_tunnels)
 	else
 		return ..()
 
--- a/code/modules/jobs/job_types/assistant.dm
+++ b/code/modules/jobs/job_types/assistant.dm
@@ -16,11 +16,7 @@ Assistant
 
 
 /datum/job/assistant/get_access()
-	if((config.jobs_have_maint_access & ASSISTANTS_HAVE_MAINT_ACCESS) || !config.jobs_have_minimal_access) //Config has assistant maint access set
-		. = ..()
-		. |= list()
-	else
-		return ..()
+	return ..()
 
 /datum/job/assistant/config_check()
 	if(config && !(config.assistant_cap == 0))
--- a/code/modules/jobs/job_types/assistant.dm
+++ b/code/modules/jobs/job_types/assistant.dm
@@ -18,7 +18,7 @@ Assistant
 /datum/job/assistant/get_access()
 	if((config.jobs_have_maint_access & ASSISTANTS_HAVE_MAINT_ACCESS) || !config.jobs_have_minimal_access) //Config has assistant maint access set
 		. = ..()
-		. |= list(access_maint_tunnels)
+		. |= list()
 	else
 		return ..()
 
--- a/code/modules/jobs/job_types/cargo_service.dm
+++ b/code/modules/jobs/job_types/cargo_service.dm
@@ -267,8 +267,8 @@ Janitor
 
 	outfit = /datum/outfit/job/janitor
 
-	access = list(access_janitor)
-	minimal_access = list(access_janitor)
+	access = list(access_janitor, access_maint_tunnels)
+	minimal_access = list(access_janitor, access_maint_tunnels)
 
 /datum/outfit/job/janitor
 	name = "Janitor"
--- a/code/modules/jobs/job_types/cargo_service.dm
+++ b/code/modules/jobs/job_types/cargo_service.dm
@@ -267,8 +267,8 @@ Janitor
 
 	outfit = /datum/outfit/job/janitor
 
-	access = list(access_janitor, access_maint_tunnels)
-	minimal_access = list(access_janitor, access_maint_tunnels)
+	access = list(access_janitor)
+	minimal_access = list(access_janitor)
 
 /datum/outfit/job/janitor
 	name = "Janitor"
--- a/code/modules/jobs/job_types/engineering.dm
+++ b/code/modules/jobs/job_types/engineering.dm
@@ -77,7 +77,7 @@ Station Engineer
 	access = list(access_engine, access_engine_equip, access_tech_storage, access_maint_tunnels,
 									access_external_airlocks, access_construction, access_atmospherics, access_tcomsat)
 	minimal_access = list(access_engine, access_engine_equip, access_tech_storage, access_maint_tunnels,
-									access_external_airlocks, access_construction, access_tcomsat)
+									access_external_airlocks, access_construction, access_tcomsat, access_atmospherics)
 
 /datum/outfit/job/engineer
 	name = "Station Engineer"
--- a/code/modules/jobs/job_types/medical.dm
+++ b/code/modules/jobs/job_types/medical.dm
@@ -127,8 +127,8 @@ Geneticist
 
 	outfit = /datum/outfit/job/geneticist
 
-	access = list(access_medical, access_morgue, access_chemistry, access_genetics, access_cloning, access_research, access_xenobiology, access_robotics, access_mineral_storeroom, access_tech_storage)
-	minimal_access = list(access_medical, access_morgue, access_genetics, access_cloning, access_research)
+	access = list(access_medical, access_morgue, access_chemistry, access_genetics, access_cloning, access_research, access_mineral_storeroom)
+	minimal_access = list(access_medical, access_morgue, access_genetics, access_cloning,)
 
 /datum/outfit/job/geneticist
 	name = "Geneticist"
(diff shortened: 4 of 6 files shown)


Roland410:
Has my full support, this will make maintenance actually be maintenance and give everyone else a harder time getting in and doing spooky things.

Roland410:
Miners don't need maint access, chaplain doesn't need theatre access, might want to add engineers atmos access even if it's not skeleton seeing as people rarely play atmos, remove scientist tech storage access and reserve that for RD/heads, same with robo, add security access to detective on skeleton.

katnep:
seems like a reasonable change, i will get to adding it after i finish this piece

Roland410:
Oh and also not sure if the change to the assistant maint thing will cause runtimes or not, so probably best to just remove or comment that part completely? I mean it shouldn't, but yeah, Citadel code is spaghetti sometimes.

Roland410:
@katnep Make sure to do the changes I asked for when you have the time.

code/modules/jobs/job_types/assistant.dm
```diff
@@ -16,11 +16,7 @@ Assistant
 
 
 /datum/job/assistant/get_access()
-	if((config.jobs_have_maint_access & ASSISTANTS_HAVE_MAINT_ACCESS) || !config.jobs_have_minimal_access) //Config has assistant maint access set
-		. = ..()
-		. |= list(access_maint_tunnels)
-	else
-		return ..()
+	return ..()
```
TalkingCactus: This is not something you should be changing. Talk with the host if you want to change something in the configuration files.

code/modules/jobs/job_types/cargo_service.dm
```diff
@@ -267,8 +267,8 @@ Janitor
 
 	outfit = /datum/outfit/job/janitor
 
-	access = list(access_janitor, access_maint_tunnels)
-	minimal_access = list(access_janitor, access_maint_tunnels)
+	access = list(access_janitor)
+	minimal_access = list(access_janitor)
```
TalkingCactus: Janitors have maintenance access so they can clean up the oil, broken glass, and dried blood; as well as setting mouse traps.

TalkingCactus (changes requested):
These changes are either unnecessary or incorrectly executed.

Poojawa:
Assistants having maint access is there because literally otherwise there is 0 access anywhere for anything. Makes having interesting tunnels completely pointless.

Tweaking jobs like say, G-tech access from having basically all of medbay and science makes more sense than taking away the janitor's access from maint.

katnep:
@poojawa to clarify g-tech is geneticist right?

Roland410:
I actually forgot to put in the geneticist having pretty much a LOT of access, but can’t say what really because I’m not home yet.

code/modules/jobs/job_types/cargo_service.dm
Roland410: Mouse traps kind of make sense, but the point of maintenance in real life is the authorized personnel use it to maintain wiring/machinery etc.

Roland410:
Geneticist should not have access outside of medbay, so remove access_research, access_xenobiology, access_robotics, access_mineral_storeroom, access_tech_storage.

code/modules/jobs/job_types/assistant.dm
Roland410: While you are somewhat right, I don't think it should ever be enabled, but I might be biased. As I said in another comment, maintenance should be reserved for personnel who HAVE to do something there.

Poojawa:
Actually, G-tech is both medical and science, but ultimately under the authority of the CMO.

Genetiecs, Cloner, research, and maybe medbay storage/chemistry should be all they get.

Roland410:
Geneticist should be a medbay staff and NOT have research access because that implies they should do RnD.

Roland410 (commented):
Properly did review below. I sometimes hate Github.

code/modules/jobs/job_types/engineering.dm
```diff
@@ -77,7 +77,7 @@ Station Engineer
 	access = list(access_engine, access_engine_equip, access_tech_storage, access_maint_tunnels,
```
Roland410, line 77: Add atmos access here too.

code/modules/jobs/job_types/medical.dm
```diff
@@ -127,8 +127,8 @@ Geneticist
 
 	outfit = /datum/outfit/job/geneticist
 
-	access = list(access_medical, access_morgue, access_chemistry, access_genetics, access_cloning, access_research, access_xenobiology, access_robotics, access_mineral_storeroom, access_tech_storage)
-	minimal_access = list(access_medical, access_morgue, access_genetics, access_cloning, access_research)
+	access = list(access_medical, access_morgue, access_chemistry, access_genetics, access_cloning, access_research, access_mineral_storeroom)
```
Roland410, line 130: No need for ORM access.

Roland410 (changes requested):
I think some stuff managed to get out of the PR somehow? Maybe it's just me, but I think we are missing a few commits/changes.
Miners still don't need maint access, remove bartender and cook having morgue access, remove scientist tech storage access and reserve that for RD/heads, same with robo, add security access to detective on skeleton.

code/modules/jobs/job_types/medical.dm
TalkingCactus: ORM is for chemistry.

code/modules/jobs/job_types/engineering.dm
TalkingCactus: it's already there

Roland410: My bad, it is indeed there.

katnep: can i just say, some of these access changes seem to be a bit drastic, cook is meant to have morgue access to mince bodies, which if we are trying to justify this with how a space station should work, is what would happen since recycling bodies into food, sci would need tech access since they can use some of the boards in there and no one really uses it as is.

Roland410: In my opinion cook should not be using those bodies to make food, there are a lot of alternatives. Also scientist doesn't need tech access because that should mainly be for engineers who need to fix a bombed place etc.

katnep: we are on a server with vore, if someone can eat another person, they can grind up souless bodies and eat them

engineers repair the structures, scientists maintain the machinery

Roland410 (approved):
After finally managing to agree on these, and if Cactus/Pooj say the code is fine, this should be ready to merge.

LetterJay:
This still has changes requested

katnep:
I'm pretty sure all the changes have been disputed or they have been made

TalkingCactus (approved):

code/modules/jobs/job_types/security.dm
```diff
@@ -124,7 +124,7 @@ Detective
 
 	outfit = /datum/outfit/job/detective
 
-	access = list(access_sec_doors, access_forensics_lockers, access_morgue, access_maint_tunnels, access_court, access_brig, access_weapons)
+	access = list(access_sec_doors, access_security, access_forensics_lockers, access_morgue, access_maint_tunnels, access_court, access_brig, access_weapons)
```
TalkingCactus: Detective should not have security access. This gives them access to cells and security equipment (which is undoubtedly better and should be reserved for law enforcement), and other things unnecessary for their job.

TalkingCactus (changes requested):

code/modules/jobs/job_types/security.dm
katnep: i had added the detective skeleton crew access to security for the fact that security players make up a tiny fraction of cit playerbase, let alone regular ones, so when we get secret with only a detective in sec it makes it easier on the crew so they don't get steam rolled by some guy hacking into the armoury five seconds in and killing them all

Roland410: I myself have requested these additions for the reasons stated by Sota.

TalkingCactus: This isn't the solution we're looking for. Giving the detective security access will make them feel obligated to conduct arrests and do security's job. Furthermore, it turns the detective into nothing but a security officer with extra access to lucrative equipment. Can't let this stay.

Roland410: It's only on skeleton though, which shouldn't be a problem anyways?

TaxMaster: Honestly, if its skeleton crew and there is only a detective in the security force, they should be filling in for the normal officer role until either one joins or the round ends.

TalkingCactus: Thing is, then there's no incentive to play security over detective. They're two distinctive roles; merging their duties in such a clear cut way is just repeating the same shit tg ended up with in terms of the geneticist's access. I've no problem with the detective filling the gaps during skeleton crew, but their gear is definitely sufficient to do both jobs. A cell and a stun baton are not required to fulfill those duties.

Roland410: While I'd normally agree, leaving the detective the only option of gunning down someone who does something bad on lowpop or holding them in a chair or something makes it really unprofessional.

katnep: there is plenty of incentive to play security, if we make it clear that detective still has to take orders from sec and has no authority over them, additionally, in roles like medbay if you choose medical doctor on skeleton crew you still get chem access, you aren't told the medivends and the store room are enough, because on low pop, a stick and a non lethal six shooter is not gonna stop the tag team of rampaging changelings

katnep: in essence this encourages powergaming, because they aren't being given the tools they need, so they are gonna find a way to get them that will land them in trouble with server staff

TalkingCactus: I expect the player to know when it's acceptable to use their gun. They're not without their radio and police baton, either. If you think putting someone in a cell for 1 minute is better than taking them into interrogation and/or just giving them a stern warning, then I don't know what to tell you.

TalkingCactus: Take it to the forum and inform the Host(s)/Head Admin(s) if you really think this is the right thing to do.

TalkingCactus:
Update the title and description and this can be merged.

TalkingCactus (approved):

TalkingCactus:
@katnep See above comment.

katnep:
done i believe

PR #299, opened 2017-03-19
citadel.1.2.1

Leon-Leonardo:
replaced the base turf of all the outside asteroid sand with the right
turf, hopefully this fixes the hang on loading
replaced all the maint loot spawns with the vanilla one
fixed the area for the lavaland shuttle
fixed the wooden tables of the bar and maint bar
added vents and scrubbers to the port hallway
added firelocks where they were missing


TalkingCactus:
You shouldn't leave old map files lingering in the game files. If you need to access a previous revision of the map you can just go through GitHub's history or you can save them elsewhere. Map files can be >1MB, which piles up and makes for a longer load time for first-time joiners.

Leon-Leonardo:
That's the thing, I'm not. I delete them before committing or making a PR. Unless I'm missing the magic "remove this unneeded file" button I don't know what to do about the old version of the map remaining in the git repo.

TalkingCactus:
Well, you better find out.

Leon-Leonardo:
¯\_(ツ)_/¯ Not really much I can do about it from here.

Leon-Leonardo:
there

PR #307, opened 2017-03-24
One Liner

TalkingCactus:
May or may not fix `list index out of bounds` errors.

**Me IRL:**
me irl


TalkingCactus:
oh wait i accidentally pushed it with a direct commit

PR #310, opened 2017-03-24
Bleeding edgy

LetterJay:
Syncs with the most recent /tg/ build plus Citadel's own.


LetterJay:
If this doesn't fix it, I'm sure we can account for the overlays used in the species as far as the timer crashing goes.

PR #320, opened 2017-03-27
Slime emote

ktccd:
:cl: ktccd
add: Gave slimes the custom emote. Finally, slimes can express themselves non-verbally! Finally they can pull out the mime inside of them and compete in charades! ~~Finally they have no excuse not to be ordered to the dorms by the xenobiologist!~~
/:cl:

[why]:
Because slimes aren't mute and/or immobile. I imagine they'd be able to express themselves somewhat. And it's hard to do so much as "nudges X person" without smashing them with a glomp and hurting them. Also because lewds is a thing here.


TalkingCactus (approved):

TalkingCactus:
For future reference, your name goes at the beginning of the changelog.

ktccd:
oh, did not know that. Thanks!

PR #323, opened 2017-03-29
auto end tweak

LetterJay:
reduces the 3 hr mark to 2 hrs. Why? Because these 3 hour long rounds are getting a bit tedious.

### DIFF
--- a/code/controllers/subsystem/shuttles.dm
+++ b/code/controllers/subsystem/shuttles.dm
@@ -354,7 +354,6 @@ var/datum/controller/subsystem/shuttle/SSshuttle
 			transit_requesters += M
 
 /datum/controller/subsystem/shuttle/proc/autoEnd()
-	world << world.time
 	if(world.time > GameTime && EMERGENCY_IDLE_OR_RECALLED) //3 hours
 		SSshuttle.emergency.request(null, 1.5)
 		priority_announce("The shift has come to an end and the shuttle called.")
--- a/code/controllers/subsystem/shuttles.dm
+++ b/code/controllers/subsystem/shuttles.dm
@@ -47,7 +47,7 @@ var/datum/controller/subsystem/shuttle/SSshuttle
 
 	var/lockdown = FALSE	//disallow transit after nuke goes off
 
-	var/GameTime = 72000 //time before in deciseconds in which the shuttle is auto called. Default is 2 hours.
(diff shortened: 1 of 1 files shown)


silicons:
Ruins rp

Roland410:
It actually doesn’t. By the way mulligan was made for <60 minutes too, so even antags will have enough time.

lm40:
make it a variable so an admin can extend it if necessary, but that would be an edge case anyway and this change is overall good

LetterJay:
There, I changed it to be listed as a var under the shuttle subsystem.

It's known as GameTimer. Default is still 2 hours.

code/controllers/subsystem/shuttles.dm
```diff
@@ -47,6 +47,8 @@ var/datum/controller/subsystem/shuttle/SSshuttle
 
 	var/lockdown = FALSE	//disallow transit after nuke goes off
 
+	var/GameTime = 72000 //time before in deciseconds in which the shuttle is auto called. Default is 2 hours.
```
TalkingCactus: Consider changing this to something which identifies it clearly as the time limit.
(maybe without capitals, too)

code/controllers/subsystem/shuttles.dm
```diff
@@ -352,7 +354,7 @@ var/datum/controller/subsystem/shuttle/SSshuttle
 			transit_requesters += M
 
 /datum/controller/subsystem/shuttle/proc/autoEnd()
-	if(world.time > 108000 && EMERGENCY_IDLE_OR_RECALLED) //3 hours
+	if(world.time > GameTime && EMERGENCY_IDLE_OR_RECALLED) //3 hours
 		SSshuttle.emergency.request(null, 1.5)
```
TalkingCactus: A 15 minute shuttle call will usually result in someone upping it to blue alert and then calling a 10 minute shuttle. You can either send the shuttle 15 minutes early(so it arrives at the time limit), reduce the time to 10 minutes or less, and/or prevent the ability to recall the emergency shuttle after the time limit has passed.

LetterJay:
I'll update this some more in the future. Right now it's just in place. I may even up the timer. Besides, admins can edit it in the SSshuttle to be longer if need be for the time being. Eventually I'll add it as an actual config.

PR #324, opened 2017-03-30
Captain display fix

ktccd:
:cl: ktccd
tweak: Captain's display case is now opened with captain access, instead of centcomm spec ops.
/:cl:

[why]: # 
The file says there was a reason for the spec_ops access level, but doesn't say what reason. So instead, every captain has to smash his display case to get his own gun. The gun is even named after the captain, no reason it should be locked away from them.


ktccd:
WAIT WHAT. What the fuuuuck is wrong with my PR. Can I stop this? It's just supposed to have the last commit in it!

Poojawa:
the merge tracking branches are fine, 

best way to "fix" those is to reset your master branch before a new fork.

ktccd:
I got no idea how to fix this PR though. Should I just re-do it?

Poojawa:
it's fine, you only have one changed file and 2 changed lines

ktccd:
Well, that's because the other commits is me merging citadels fixes into my branch since y'all updated so much ^^.

TalkingCactus:
A merge will only be required if there are conflicts, which are displayed when you make the PR and at the bottom of this page.

ktccd:
So what do I need to do?

TalkingCactus:
Nothing, actually. If merging with the latest updates causes you problems then not updating will work just fine (unless there are conflicts). 

Worded that kinda funny, sorry.

ktccd:
Yeah, the only problems I had on merging was the download freezing my client. I got it working now, but it's what caused those merge commits to be on this PR branch instead of my master branch...

What I meant with needing to do, was what do I need to do with this PR? Like, does it work as-is?

TalkingCactus:
It does, but I give time for others to peer review code typically.

ktccd:
Alrighty, sounds good ^^.

TalkingCactus:
Are they able to put the gun back after it's taken out?

ktccd:
yes?
This literally just changes the access. means you can click to open it or lock it with your ID.
((You could before also, but you had to be a centcomm special ops for some darned reason))

TalkingCactus:
Right, just checked it myself too. It's no surprise that I've never seen anyone actually open one of these things without tripping the alarm.

Poojawa:
As a note, you can also fix them by using a welder.

PR #325, opened 2017-03-30
Ninja tech-theft fix

ktccd:
:cl: ktccd
fix: Made it so Ninjas can actually steal tech from the server and R&D console now. Major props to LeoZ and Jalleo for helping me with figuring out why the for loop was so busted!
/:cl:

[why]: # 
bugfix, necessary to complete objectives (You CAN still steal tech by hitting the suit in your inventory while holding a tech disk in your hand though).


TalkingCactus:
You should push this to /tg/ as well.

ktccd:
Already done, they got their own version of it, because IDK how to push the same PR to different places.
And also because this one had those merge things that shames me in the commit logs.

PR #326, opened 2017-03-31
Minor Edits to Box

Cw3040:
Added larger maint bar, converted smaller maint bar to a storage room.

### DIFF
--- a/_maps/map_files/TgStation/tgstation.2.1.3.dmm
+++ b/_maps/map_files/TgStation/tgstation.2.1.3.dmm
@@ -62717,6 +62717,11 @@
 /obj/machinery/vending/kink,
 /turf/open/floor/wood,
 /area/maintenance/bar)
+"eIp" = (
+/obj/structure/table/wood/poker,
+/obj/item/weapon/storage/pill_bottle/dice,
+/turf/open/floor/wood,
+/area/maintenance/bar)
 "gsu" = (
 /obj/structure/table/wood,
 /obj/item/device/flashlight/lamp/green,
@@ -62791,10 +62796,6 @@
 	d2 = 4;
 	icon_state = "1-4"
 	},
-/turf/open/floor/wood,
-/area/maintenance/bar)
-"lPa" = (
-/obj/structure/table/wood/poker,
 /turf/open/floor/wood,
 /area/maintenance/bar)
 "twx" = (
@@ -85255,7 +85256,7 @@ Pfx
 idC
 Qmz
 Dat
-lPa
+eIp
 yug
 Fsp
 Fsp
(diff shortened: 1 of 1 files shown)


Roland410:
There is no real need for another maint bar, and even IF there were, it doesn't need that many/any windows AT ALL.

Cw3040:
I can rectify the window situation. As for the lack of "need for another maint bar", it's the only one. I turned the old bar into a storage room, with various maintenance stuff.

Roland410:
Exactly, there IS already a maint bar, also if you were to look at ye olde maint bar we had, replicating or copying that would be much better.

Cw3040:
I was going for more the size and feel of MetaStation's maintenance bar. That seems to be what people want.

Cw3040:
I've done some stupid bullshit so expect like 5 commits showing a major github blunder (largely due to my lack of knowledge of github)

TalkingCactus:
You need to convert the map file to tgm with the dmm2tgm tool located in the `../tools` folder.

Edit: Usually when it's done right, there aren't several thousand changed lines. Looking into it, though you do have the header.

Edit Edit: Forgot map files are almost 100,000 lines nowadays. 2,000 lines isn't actually that much. Looks good.

Cw3040:
I did no edits except in the picture, but changes say something about removing /area/shuttle/syndicate for whatever reason.

TalkingCactus:
90% of the walls should not be windows, the entrance needs to not share an adjacent tile with another door and should be moved to a more accessible location, there's a cob web attached to the entrance airlock, remove the booze-o-mat and replace it with boxes of drinking glasses(perhaps in a closet/cabinet), the bar should take a corner position where the counter can wrap around it in an "L" so nobody sits across the room from the bartender, the entrance airlock needs to be changed from "Maint Bar Access" to "Maintenance Bar", and it needs to use the `../maintenance/bar` area with its own APC.

Cw3040:
Okay
Pic related to commits: http://i.imgur.com/KclYOJi.png

Cw3040:
Update: Bar has been tweaked to fulfill requests.

TalkingCactus:
Move the cable so it goes through the door or a window so it can be easily tracked and modified by players without removing a wall.

Cw3040:
Check

LetterJay:
Something I want to stress is to make sure this will work using the maploader.

Cw3040:
It loads on all tests I've run on this specific map. Please elaborate.

EDIT for below: it was a meme

Cw3040:
I have done additional testing with the map and have found no issues with it conflicting. I do not think it will conflict with the maploader, given it is a small change to an existing map.

TalkingCactus:
The errors don't show up until Jay tries to load it with the server. I've tried twice to insert my maintenance bar in the past, so hopefully you've done it right.

TalkingCactus (approved):

Cw3040:
I don't know if it's done proper, but a testmerge is very possible.

PR #328, opened 2017-04-01
Syndicate borer added

ktccd:
:cl: ktccd
add: Syndicate agents, ~both the nuclear variant and~ **only** traitors, can now purchase our Syndi(tm) brand cortical borers! Nigh-guaranteed to be loyal to you and aid you in your mission, these borers sport all the latest and greatest of borer technology. _Reproductive capabilities may be void in certain states. Seller not responsible for providing souls to be housed in the borer body, please contact your local Syndi(tm) retail outlet for further customer service, refunds are available_
/:cl:

[why]: # 
Because borers are fun and provide some interesting ways to use them as a traitor. The icon for the item itself is a place-holder, as I lack spriting abilities and it looks ok as it is at least.
The borers of this variant are incapable of reproducing and will greentext together with their owner, which is not necessarily the same person as their host. An owner COULD for example buy a borer for their teammates, or send them to take control of a high-value host.
This kind of borer is a subtype of the regular kind, so it allows the possibility for further modifications if that's something we want, such as adding other chemicals to their production, though I want this PR to work as a basic set before I try anything weird.

### DIFF
--- a/code/game/gamemodes/antag_spawner.dm
+++ b/code/game/gamemodes/antag_spawner.dm
@@ -298,6 +298,7 @@
 obj/item/weapon/antag_spawner/syndi_borer
 	name = "syndicate brain-slug container"
 	desc = "Releases a modified cortical borer to assist the user."
+	icon = 'icons/obj/device.dmi' //Temporary? Doesn't really look like a container for xenofauna... but IDK what else could work.
 	icon_state = "locator"
 	
 obj/item/weapon/antag_spawner/syndi_borer/spawn_antag(client/C, turf/T, mob/owner)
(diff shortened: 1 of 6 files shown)


ktccd:
I would like some feedback on costs. I put it at 5TC, because I honestly don't know how much things are worth in the syndicate toolset. It kind of looks like things with high-killing power costs more than utility and support items...

katnep:
At least 10tc and probably disable it on nuke ops

ktccd:
Don't nuke ops explode when they die anyway? How would a nukie abuse this? O.o

katnep:
Well nuke ops has a lot of options as is, and being able to get free adrenals/healing is pretty ridiculous

katnep:
Either make a nuke ops grade one that costs more or disable it on them and around d 10/12 on Traitor

ktccd:
There, added in and done. Also configured it for surplus shenanigans.
Notes for the future:
If it turns out that this item doesn't get ghosts to join up often enough, then it might be a good idea to disable surplus crate probability from it, as I don't think you can refund those items.

ktccd:
Changed the changelog to show this.

ktccd:
So is this fine for merging now?

PR #330, opened 2017-04-01
Check Mate

TalkingCactus:
# For too long, us normies have suffered under the oppressive regime of this server's staff. From this day forward, we hereby remove their ability to express themselves and cast them to the pits of human superiority. I have a few words for those who this affects: 

The furries's worst transgressions are systematically whitewashed by the press. To counteract that whitewash I will use the remainder of my space here to expose the furries for what it really is. None of what follows is my own original research. Rather, all of it is taken from wiser people than I, and it is these people who deserve the credit for first observing that the furries makes a lot of exaggerated claims. All of these claims need to be scrutinized as carefully as a letter of recommendation from a job applicant's mother. Consider, for example, the furries's claim that matters of racial justice should enter a period of “benign neglect”. The fact of the matter is that it says that it can succeed without trying. Such statements are not just wrong; they're worse than wrong. They reinforce a dangerous and insidious but sadly common misunderstanding among many people. They disguise the fact that it's guilty of selective moral outrage. By that, I mean that the furries picks and chooses what it's going to be outraged about, then turns around and does the exact same thing to someone else. If you think that that's gloomy then consider that the furries desperately wants us to believe that the best way to make a point is with foaming-at-the-mouth rhetoric and letters filled primarily with exclamation points. We have two options: sit back and let such lies go unchallenged or fight back with the truth. I have decided to fight back. I shall do so by spreading the truth about how poisoning the air, water, and soil is a mug's game. The only reason it does things like that is because we have a dilemma of leviathan proportions on our hands: Should we stop its encroachments on our heritage, or is it sufficient to build a sane and healthy society free of its destructive influences? The answer is rather depressing, but I'll tell you anyway. The answer begins with the observation that the furries has been trying for some time to sell the public on a fetishism-based government. Its sales pitch proceeds both pragmatically and emotionally. The pragmatic argument: Solecism is absolutely essential to the well-being of society. The emotional argument: Big emotions come from big words. As you can see, neither argument is valid, which should indicate to you that the furries will tear down everything that can possibly be regarded as a support of cultural elevation by the end of the decade. Alas, this is not a tinfoil-hat conspiracy theory. It is cold, hard fact. A related fact is that by refusing to act, by refusing to avoid the extremes of a pessimistic naturalism and an optimistic humanism by combining the truths of both, we are giving the furries the power to perpetuate myths that glorify diabolism.

That's not the most frightening thing about the furries. Have you heard that it has managed to establish a vast number of sleeper cells around the world that, on the furries's command, will let down ladders that the disdainful, self-righteous, and phlegmatic scramble to climb? I find information like that disturbing on so many levels that I can't help but want to derail the furries's beggarly little schemes.

In these days of political correctness and the changing of how history is taught in schools to fulfill a particular agenda, if I want to become the target of prejudice, ridicule, discrimination, and physical violence, that should be my prerogative. I sincerely don't need the furries forcing me to. The furries's snippy fans fundamentally believe that cannibalism, wife-swapping, and the murder of infants and the elderly are acceptable behavior. Alas, this deeply held belief is fiction from start to finish. Every piece of evidence I can find makes it abundantly clear that if the furries truly wanted to be helpful, it wouldn't replace the search for truth with a situationist relativism based on sanguinolent presentism.

The furries insists that only one or two members of its entire sodality of brutish, unambitious wackos are scummy evildoers. Only one or two members? This is, to put it charitably, an understatement of the facts. It would be far more accurate to say that there's a famous mathematical proof that pertains to the furries. Essentially, this proof asserts that given that the furries enjoys pondering new ways to pit the haves against the have-nots, then, loosely speaking, it must be the case that its generalizations are a mere cavil, a mere scarecrow, one of the last shifts of a desperate and dying cause. Interestingly, the furries has recently begun conceding that perhaps it's not the case that its debauches are the result of a high-minded urge to do sociological research. However, it still insists that anyone who resists it deserves to be crushed. As the ancient Romans used to say, “Falsus in uno, falsus in omnibus.” Translated, this means, roughly, that the furries wants to commit confrontational, in-your-face acts of violence, intimidation, and incivility. Personally, I don't want that. Personally, I prefer freedom. If you also prefer freedom then you should be working with me to expose some of its more dubious financial dealings while remaining true to those beliefs, ideals, and aspirations we hold most dear.

The furries is the picture of the insane person on the street, babbling to a tree, a wall, or a cloud, which cannot and does not respond to its escapades. I have just one word for the furries: plethysmographically. I believe that the best way to overcome misunderstanding, prejudice, and hate is by means of reason, common sense, clear thinking, and goodwill. The furries, in contrast, avows that the sky is falling. The conclusion to draw from this conflict of views should be obvious: You don't have to say anything specifically about the furries for it to start attacking you. All you have to do is dare to imply that we should urge lawmakers to pass a nonbinding resolution affirming that a record of its acts of hypocrisy would fill volumes. Finally, whatever your thoughts or feelings about the furries are, I urge you to help me remove the misunderstanding that the furries has created in the minds of myriad people throughout the world.


TalkingCactus:
# I'M BEING LABELED

Leon-Leonardo:
tl;dr

PR #339, opened 2017-04-04
Huge Sprite Rework

ktccd:
:cl: ktccd
fix: Fixed how Taur sprites are layered, finally making genitals show for Nagas as they should.
tweak: Vaginas and Breasts now follow the same naming convention for sprites that penises use, so all genital organs can be rendered using the same piece of code. This means that breast sizes have dropped to single-letter variants, since those are what we got sprites for (Because the cup-size is now the Size of the obj).
tweak: Tweaked how colours are chosen (Humans can enable/disable matching genitals to
their skintones, everyone else has to pick their colours).
tweak: Breasts now have shapes! So far we only got the "pair" shape, but we got code-support for other variants.
add: Vaginas can now also be chosen and will render properly, meaning we now have more of them to choose from (Some of them terrifying). They don't really use a size, but there's code support for it if sprites would be made for it.
tweak: Breast items now change their size and colour to the owner's when/if they somehow lose the organs (gibbing for example). Both Breasts and Vaginas change their description to match what the object is set to look like, sprites are lacking in that area however.
tweak: Minor variable error in the penis-description code fixed.
fix: Taur sprites renamed to use the correct layers. Shephard sprite bug fixed where there was a missing 2 pixels in the middle of the body.
/:cl:

[why]: # 
Our custom spriting was a bit of a mess and should be working now. There's still some stuff that can be done and improved upon, such as adding testicles onmob sprites and more variants, shapes, sizes etc. But now we got code support for it, so adding more shapes should be fairly simple, and adding more sizes even simpler.
I'll attach a changelog yml file to this in the next commit, since I don't know the number of my PR before making one.


ktccd:
There. I did it, I fucking did it!
Yay! What a freaking way to go about learning how icons and overlays work. Like learning to swim by partaking in the mid-winter sauna + swimming.

TalkingCactus:
This should've been reviewed first.

LetterJay:
Normally I do leave it to be reviewed. The only reason I merged it so quick is I was there with him while he was testing it.

LetterJay:
I can revert it if there's an issue.

PR #345, opened 2017-04-07
Adminhelp prompts will now be larger to allow you to proofread large responses and PMs

TalkingCactus:
# Makes it much easier to spot spelling/grammar mistakes so you look like less of a retard.
## Also adds a new `MARK READ` button to adminhelps to let the player know their adminhelp is being looked into.
Also added the large box to /me emotes.
proof
more proof

no, there aren't too many buttons attached to adminhelps


cebutris:
Yes please

PR #347, opened 2017-04-09
April sync for testing

Poojawa:
Compiles on my end, but untested. same as usual folks.


Poojawa:
Also travis has been made to hate generic lists. so that's fun.

PR #342, opened 2017-04-04
Huge Sprite Rework

ktccd:
Reverts Citadel-Station-13/Citadel-Station-13#340
:cl: ktccd
fix: Fixed how Taur sprites are layered, finally making genitals show for Nagas as they should.
tweak: Vaginas and Breasts now follow the same naming convention for sprites that penises use, so all genital organs can be rendered using the same piece of code. This means that breast sizes have dropped to single-letter variants, since those are what we got sprites for (Because the cup-size is now the Size of the obj).
tweak: Tweaked how colours are chosen (Humans can enable/disable matching genitals to
their skintones, everyone else has to pick their colours).
tweak: Breasts now have shapes! So far we only got the "pair" shape, but we got code-support for other variants.
add: Vaginas can now also be chosen and will render properly, meaning we now have more of them to choose from (Some of them terrifying). They don't really use a size, but there's code support for it if sprites would be made for it.
tweak: Breast items now change their size and colour to the owner's when/if they somehow lose the organs (gibbing for example). Both Breasts and Vaginas change their description to match what the object is set to look like, sprites are lacking in that area however.
tweak: Minor variable error in the penis-description code fixed.
fix: Taur sprites renamed to use the correct layers. Shephard sprite bug fixed where there was a missing 2 pixels in the middle of the body.
/:cl:

[why] # 
Our custom spriting was a bit of a mess and should be working now. There's still some stuff that can be done and improved upon, such as adding testicles onmob sprites and more variants, shapes, sizes etc. But now we got code support for it, so adding more shapes should be fairly simple, and adding more sizes even simpler.
I'll attach a changelog yml file to this in the next commit, since I don't know the number of my PR before making one.

PR #348, opened 2017-04-09
Genital overlays Fixes for PR#342

ktccd:
This PR consists of the requested changes to https://github.com/Citadel-Station-13/Citadel-Station-13/pull/342
Can't push them to THAT PR, since it is actually a revert of a reverted merge of the old PR.
Before anyone asks, YES, we tried a ton of ways to re-make the old PR. No, we could not. No, pushing to the old PR branch doesn't update it. Yes, this is the best work-around I could think of.
So if the revert is accepted, these fixes needs to be applied as well.

:cl: ktccd
Tweak: Genitals size calculated only once, no repeated code
Tweak: Vagina description grammar fixes implemented and remade to be easier to read in code.
Tweak: Preferences save file fix implemented also. Preferences capitalization fixed.
Fix: Genitalia now uses skin_tone when they should and messing about in the preferences no longer overwrites the previous colors selected, allowing toggling of the skin_tone setting.
Tweak: Vaginas are now drawn behind penises and rendered on the same layer.
Add: Genital arousal framework for alternate icons during arousal. Currently only vaginas use it, but it's there if we get, for example, someone to draw sheath icon_states.
Tweak: Preferences show correct cock shapes on creation and randomising a character also randomises the shape of all genitalia.
/:cl:

[why]: 
Because these changes requested can't really be added to the old PR, since the old PR got merged then reverted and the reviewed code is actually a reverted revert-merge, which is messy but it's the only way we can do this it seems.


ktccd:
Closing this because Pooj helped me figure out how to push to a revert-revert-merge-PR. Shit's bananas.

PR #349, opened 2017-04-10
Fixes digestion messages displaying to everyone nearby

lm40:
no idea why this was the way it was, but the arguments for visible_message were wrong

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -43,7 +43,7 @@
 
 				//Send messages
 				owner << "<span class='warning'>" + digest_alert_owner + "</span>"
-				M << "<span class='userdanger'>" + digest_alert_prey + "</span>"
+				M.visible_message("<span class='notice'>You watch as [owner]'s form lose its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
 
 				owner.nutrition += 400 // so eating dead mobs gives you *something*.
 				playsound(owner.loc, "death_gurgles", 50, 0, -5)
@@ -85,7 +85,7 @@
 
 				//Send messages
 				owner << "<span class='warning'>" + digest_alert_owner + "</span>"
-				M << "<span class='userdanger'>" + digest_alert_prey + "</span>"
+				M.visible_message("<span class='notice'>You watch as [owner]'s form lose its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
 				owner.nutrition += 400 // so eating dead mobs gives you *something*.
 				playsound(owner.loc, "death_gurgles", 50, 0, -5)
 				digestion_death(M)
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -42,7 +42,7 @@
 				digest_alert_prey = replacetext(digest_alert_prey,"%belly",lowertext(name))
 
 				//Send messages
-				to_chat(owner, "<span class='warning'>"digest_alert_owner"</span>")
+				to_chat(owner, "<span class='warning'>[digest_alert_owner]</span>")
 				M.visible_message("<span class='notice'>You watch as [owner]'s form loses its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
 
 				owner.nutrition += 400 // so eating dead mobs gives you *something*.
@@ -84,7 +84,7 @@
 				digest_alert_prey = replacetext(digest_alert_prey,"%belly",lowertext(name))
 
 				//Send messages
-				to_chat(owner, "<span class='warning'>"digest_alert_owner"</span>")
+				to_chat(owner, "<span class='warning'>[digest_alert_owner]</span>")
 				M.visible_message("<span class='notice'>You watch as [owner]'s form loses its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
 				owner.nutrition += 400 // so eating dead mobs gives you *something*.
 				playsound(owner.loc, "death_gurgles", 50, 0, -5)
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -42,8 +42,8 @@
 				digest_alert_prey = replacetext(digest_alert_prey,"%belly",lowertext(name))
 
 				//Send messages
-				owner << "<span class='warning'>" + digest_alert_owner + "</span>"
-				M.visible_message("<span class='notice'>You watch as [owner]'s form lose its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
+				to_chat(owner, "<span class='warning'>"digest_alert_owner"</span>")
+				M.visible_message("<span class='notice'>You watch as [owner]'s form loses its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
 
 				owner.nutrition += 400 // so eating dead mobs gives you *something*.
 				playsound(owner.loc, "death_gurgles", 50, 0, -5)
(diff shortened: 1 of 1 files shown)


LetterJay:
who all was seeing this happen? Just ghosts?

lm40:
anyone within visible range, but only for fast digestion. now the fast digestion messages are the same as the normal ones.

LetterJay:
that's what the visible_message proc does.

ktccd:
Seems like an intentional feature?

LetterJay:
I dunno. I'd have to look more into it. Or let Poojawa. I know Cactus won't touch this stuff.

lm40:
well yeah but it was showing everyone the giant red letters when it was (probably) supposed to just be showing them the "form loses its additions" part. it would be an easy fix by switching around the order of the arguments, but since the normal digestion doesn't show any of this at all i just axed the whole thing.

ktccd:
I meant that the fact the only purpose for that line of code is to display to people the message, it seems very intentional. I guess it's meant so you can't secretly fast-digest someone?
I take it this code is either from another era here, or from vorestation? Where I think you could use vore for antag purposes and the reason it's displayed this way was to give prey a chance to be saved or preds to be caught?
EDIT:
Basically, fast-digestion is a decent murder and body-disposal method and that line is probably there for some kind of balance reason?

lm40:
vorestation doesn't have fast digestion

ktccd:
Personally I don't mind people outside a person being able to tell when they suddenly digest their own bodymass in food, as sprites don't convey this.
But I think the usage of "userdanger" span class is abused here. IDK much about chat formatting, but the name userdanger seems to imply this is meant for _IMPORTANT SHIT KILLING **YOU** RIGHT NOW_ kind of text.

ktccd:
And this is obviously NOT a danger to the person receiving the chat message about it. Would lowering the span format to the danger class, which is meant for dangerous shit to OTHER people, work for a fix?

lm40:
it was never meant to display the danger message to the bystanders, that's why this is a bug

ktccd:
~~So don't remove it, fix it?~~
Scratch that, it sounded mean. Fuck I'm too tired to manage being polite :S.
But what about the point I already raised about it?

lm40:
i don't really see the point of having this message at all so i'm just removing it

ktccd:
Damn, conversations moving too fast for me to be able to edit in time ^^.
What about the point I raised earlier though?

lm40:
it's more of an rp mechanic than a game one so i don't think this is really necessary

ktccd:
But... I do? Like, even from an RP point of view.
The sprite system we have don't currently allow for being able to tell when people are full or anything, so there's no way to notice a person suddenly shrinking down half their bodymass without constantly spamming examine and suddenly the message about them being full is gone.
I kind of like that there's a prompt that does let you notice a noticeable change like that.

Now, obviously I don't like that the span class is incorrect, but that's an easy fix.

lm40:
this is (mechnically) the same thing that happens when you finish digesting someone at regular speed and that shows no message, it should be consistently one way or the other and this way was marginally easier

ktccd:
Well, I'd like that one to have something similar too honestly. So for consistency, adding a similar thing there would also work. This is a very visually noticeable thing which I think is kind of weird that you can SEE if you look at the person but it's odd if you can't notice it changing without spamming examine.
(And I don't really like taking the easy way out just for the sake of being easier :S)

lm40:
i'm too tired to argue about this so it's easier to just change it, either way i want to resolve the issue of random bystanders feeling people gushing through their system or w/e

ktccd:
I think you removed the messages that are sent to the prey. Those are the M << message things.

ktccd:
Also, wait, bystanders "feel people gushing through their system"? You mean the OWNER messages are sent to bystanders?

lm40:
yes, the order of the arguments were wrong, that's why it was a bug.

ktccd:
THAT is a different kind of bug ^^. I think that's because the visible_message contained those strings where it shouldn't have? Would removing the [digest_alert_owner] and [digest_alert_prey] pieces from the visible_message fix it?

ktccd:
because those things should have been sent to their respective players with the << operator.

lm40:
no, the error was in visible_message, the << bit worked perfectly fine and it's why this bug only existed when people fast digested

ktccd:
Yeah, so keep the part of visible message that shows they kind of digested someone, but don't send them the digest_alert strings? 'cause those are usually written from the perspective of the prey/pred.

ktccd:
... would be cool to have a 3rd person string for this purpose, but meh, not the scope of this PR.

lm40:
...that's what i did.

ktccd:
Didn't you say you removed everything?
EDIT: Which is not what I'm suggesting here. I'm suggesting just removing the danger_alert parts of the visible_message piece.

lm40:
i did but i fixed it the other way since you were complaining and i'm too tired to argue

ktccd:
Yes, but when you did THAT commit, you ALSO removed the prey receiving a message at ALL...

lm40:
see visible_message

ktccd:
Yes... but this would display a message from the prey's point of view to everyone, wouldn't it?

lm40:
// Show a message to all player mobs who sees this atom
// Show a message to the src mob (if the src is a mob)
// Use for atoms performing visible actions
// message is output to anyone who can see, e.g. "The [src] does something!"
// self_message (optional) is what the src mob sees e.g. "You do something!"
// blind_message (optional) is what blind people will hear e.g. "You hear something!"
// vision_distance (optional) define how many tiles away the message can be seen.
// ignored_mob (optional) doesn't show any message to a given mob if TRUE.

/atom/proc/visible_message(message, self_message, blind_message, vision_distance, ignored_mob)

ktccd:
FUCK, I missed the comma. I thought it was just appended at the end of a string.... FML, my bad.

ktccd:
Yeah, your fix was correct then. Damn I should be sleeping regularly, I make dumb mistakes like this...

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -43,7 +43,7 @@
 
 				//Send messages
 				owner << "<span class='warning'>" + digest_alert_owner + "</span>"
```
TalkingCactus: to_chat(owner, "\<span class='warning'\>"digest_alert_owner"\</span\>")

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -43,7 +43,7 @@
...
-				M << "<span class='userdanger'>" + digest_alert_prey + "</span>"
+				M.visible_message("<span class='notice'>You watch as [owner]'s form lose its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
```
TalkingCactus: loses

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -85,9 +85,7 @@
 
 				//Send messages
 				owner << "<span class='warning'>" + digest_alert_owner + "</span>"
-				M << "<span class='userdanger'>" + digest_alert_prey + "</span>"
-				M.visible_message("<span class='userdanger'>[digest_alert_owner]</span>", "<span class='warning'>[digest_alert_prey]</span>",
-									"<span class='notice'>You watch as [owner]'s form lose its additions.</span>")
+				M.visible_message("<span class='notice'>You watch as [owner]'s form lose its additions.</span>", "<span class='warning'>[digest_alert_prey]</span>")
```
TalkingCactus: loses

code/modules/vore/eating/bellymodes_vr.dm
```diff
@@ -85,9 +85,7 @@
 
 				//Send messages
 				owner << "<span class='warning'>" + digest_alert_owner + "</span>"
```
TalkingCactus: to_chat(owner, "\<span class='warning'\>"digest_alert_owner"\</span\>")

TalkingCactus (changes requested):

lm40:
>doesn't compile

w/e i'll fix this tomorrow

nvm this probably will work i'm blind and dumb

LetterJay:
This show as pending reviews requested, but I don't see anything?

TalkingCactus (approved):

TalkingCactus:
I had to approve them, also yeah sorry I forgot to replace the quotation marks with square brackets.

PR #350, opened 2017-04-10
Subtle emote fix

ktccd:
[Changelogs]: 
:cl: ktccd
fix: Remade the subtle proc to be an emote. you can now use *subtle or the verb in the IC tab to use this emote. It works like a normal hearable or visible emote, with a range of 1 tile radius.
/:cl:

[why]: # 
Because the previous proc had issues with simple_mobs using it (at least) and was very difficult to debug because of... code I don't understand. It was NAMED an emote, but wasn't **actually** an emote datum, which is now fixed.

### DIFF
--- a/code/modules/mob/say_vr.dm
+++ b/code/modules/mob/say_vr.dm
@@ -115,8 +115,10 @@ proc/get_top_level_mob(var/mob/S)
 	emote_type = EMOTE_VISIBLE
 
 ///////////////// VERB CODE
+/mob/living/verb/subtle()
 	set name = "Subtle"
 	set category = "IC"
 	if(say_disabled)	//This is here to try to identify lag problems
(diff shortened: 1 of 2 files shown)


ktccd:
How do I request a review from a maintainer?

Poojawa:
To be fair, I pulled this from Virgo, so it was one of their coders who made it. so if this works in game, then heck ye. all good.

Poojawa (approved):

ktccd:
Yeah, maybe theirs was coded before emote datums were a thing, but I couldn't understand their code. I think what made simple_mobs unable to use it might've been the use of hearers(), but I'm unsure.
EDIT: As a bonus, now we will be able to choose if the subtle emote we use is audible or visual.

PR #351, opened 2017-04-11
travis friendly sync updating

Poojawa:
Current noted issues:

Icon previews are screwy again, doesn't refresh properly. Icons are OK ingame.
Mentor system is on the fritz, Hippie's system isn't updated with the new wrappers.
ERP system hasn't been tested, but seems to be 'decent'
'Who' command counts every name instance instead of just the one. Also counts Admins under players.

missing some backend tweaks, like captain's display case, etc.


Poojawa:
@LetterJay feel free to debug, as well. so we can get this out that much faster. I'll be working on it tonight too

PR #353, opened 2017-04-11
Lewd viro

ktccd:
:cl: ktccd
add: Added new citadel-only symptoms to viruses.
/:cl:

[why]: # 
Because we got these citadel-only things which could be used by virology. Mostly came to think about it when I figured if there was some way to get airborne aphrodisiac, when the word "airborne" made me think of virology.
# Ty to Snugglesnek for giving the idea of a virus that messes with genitalia. Sadly, the way we handle genitalia is very snow-flakey and not generic for all genitals, so yeah...


ktccd:
Yeah, please **do not merge** yet, I still got one more commit coming up for this one, the PR is mostly open so I can get feedback in the meantime.

Poojawa:
Probably pre-package these into a virus outbreak too, make it pick which to grow based on gender and already established organs maybe.

ktccd:
So you mean make it so the viral outbreak can result in these?
Also, it doesn't care about gender, because genitals do not care about gender. It only checks if you have them, and if so it makes them grow a size.

TalkingCactus:
I always wanted virus outbreak to be a bit more dangerous. Have it be a pandemic scenario where the virus mutates rapidly 'n stuff. Anyways, most transformations of genitals will be placed in procs when the system is more complete. Probably a good idea to hold off on this until then.

code/citadel/cit_virology.dm
```diff
@@ -0,0 +1,244 @@
...
+	transmittable = -2
+	level = 2
+	severity = 1
+	var/big_thoughts = list("You feel a bit stiff.","Your loins itch.","You feel like you don't want to wear clothing.")
+
+
+/datum/symptom/macro_genitals/proc/grow_genitalia(mob/living/carbon/human/M)
+	if(M.has_penis())
+		var/obj/item/organ/genital/penis/O = M.getorganslot("penis")
+		if(O.length < COCK_SIZE_MAX) //Couldn't find a macro or anything for the max size of a dong, so hardcoded, sadly.
```
TalkingCactus: Useless comment, defines save on memory and are easy to maintain for constants that are referenced a lot.

code/citadel/cit_virology.dm
```diff
@@ -0,0 +1,244 @@
...
+			O.length = O.length + 1
```
TalkingCactus: `O.length = Clamp(O.length++, COCK_SIZE_MIN, COCK_SIZE_MAX)`
or
```
O.length++
if(O.length > COCK_SIZE_MAX) O.length = COCK_SIZE_MAX
```

code/citadel/cit_virology.dm
```diff
@@ -0,0 +1,244 @@
...
+			O.size = breasts_size_list(size + 1)
+			O.update()
+			to_chat(M,"<span class='warning'>[pick("Your boobs suddenly gain several cupsizes!","You breasts rapidly expand outward.","You feel your chest grow much larger.")]</span>")
+*/
+	return
+
+/datum/symptom/macro_genitals/Activate(datum/disease/advance/A)
+	..()
+	if(prob(SYMPTOM_ACTIVATION_PROB))
+		//I think genitalia code is implemented on human level only, so let's not mess up
```
TalkingCactus: `if(ishuman(A.affected_mob))`

code/citadel/cit_virology.dm
```diff
@@ -0,0 +1,244 @@
...
+		var/mob/living/carbon/human/M = A.affected_mob
+		if(M)
```
TalkingCactus: Move this below the `..()` and change it to 
```
if(!A.affected_mob) return
```
Though I don't think this proc would even call if there wasn't an affected mob.

code/citadel/cit_virology.dm
```diff
@@ -0,0 +1,244 @@
...
+	level = 2
+	severity = 1
+	var/small_thoughts = list("You feel a bit stiff.","Your loins itch.","You feel like moving is a bit easier.")
+
+
+/datum/symptom/micro_genitals/proc/shrink(mob/living/carbon/human/M)
+	if(M.has_penis())
+		var/obj/item/organ/genital/penis/O = M.getorganslot("penis")
+		if(O.length > COCK_SIZE_MIN) //Couldn't find a macro or anything for the max size of a dong, so hardcoded, sadly.
+			O.length = O.length - 1
```
TalkingCactus: See Extendong comment; Same thing but inverse.

TalkingCactus (changes requested):

code/citadel/cit_virology.dm
ktccd: yeah, I think I forgot to remove the comment after finding the constant. When I wrote the comment, I had just put "20" there XD.

code/citadel/cit_virology.dm
ktccd: so clamp forces it to stay between the boundaries?

code/citadel/cit_virology.dm
ktccd: oh, yeah. might be weird to recast non-humans to humans. heh...

code/citadel/cit_virology.dm
ktccd: I think my idea... was something like, M would be null if A wasn't human. But I was just bad at how recasting types worked. I think maybe the result of recasting A as human, if it wasn't human, would probably not be null after all.

code/citadel/cit_virology.dm
TalkingCactus: Yes

code/citadel/cit_virology.dm
TalkingCactus: Monkeys can carry viruses as well. Xenos might be able to as well, haven't tested that in a while.

ktccd: so there's no point in doing if(M) then.
I guess it makes sense, none of the other viruses seem to do that.

TalkingCactus: Mhmm

code/citadel/cit_virology.dm
ktccd: ✅

code/citadel/cit_virology.dm
ktccd: ✅

code/citadel/cit_virology.dm
ktccd: ✅

code/citadel/cit_virology.dm
ktccd: ✅

code/citadel/cit_virology.dm
ktccd: ✅

ktccd:
oh goodness...

ktccd:
Ok, so, comparing the breasts_onmob.dmi with my commited one, I can't really spot why there's a conflict...
I renamed the sizes to use numbers instead of cupsizes, that's pretty much it...

TalkingCactus:
https://github.com/Citadel-Station-13/Citadel-Station-13/commits/master/code/citadel/icons/breasts_onmob.dmi

ktccd:
Wait, what did the other PR change with lighting?
There... isn't lighting in that file? O.o

TalkingCactus:
I'm showing you where the changes occurred since you couldn't find it.

ktccd:
Yeah, but... the file looks the same? Doesn't it?
I tried comparing with the file I took from master...

EDIT: I guess to be extra-sure, I'll just download the master version and rename the icon states.

TalkingCactus:
They're changed, trust me. Look at the PR description for the most recent commit.

ktccd:
The PR doesn't say what the exact pixel fix is, so I can't manually compare the files.

EDIT:
As it is now, I have downloaded the .dmi file from the master-branch and renamed the icon_states so they work with the new size system. After doing this change, git does not see any change when comparing the binary file from my local files with my commit here, so I assume the binary is identical now (Meaning the current commited version should be the same as the master-branch version + my name change).

TalkingCactus:
Why do you need to compare them? Just take then newest one and rename them.

ktccd:
I got no idea what is wrong with the .dme file though... I wasn't sure I changed that? Except maybe add in my new file to the list of files to compile?

ktccd:
I did. But you know how git won't commit a file if it hasn't been changed? That's the issue I'm having.
Downloading the master-file and renaming the icon_state names results in a file that github can't see a difference in...

EDIT: i.e the file I've already committed to this PR is the master-file version with names changed, but it still conflicts because I made this PR a while ago and github can't compare binaries it seems.

TalkingCactus:
Your only salvation may be a new PR, though I'm fairly certain there is a way to circumvent Git hating .dmi changes. As for the .dme, you just need to merge with the latest changes from the master and compile.

ktccd:
Well, whoever handles merging can resolve conflicts I'm sure. I know there's a way to handle binary file conflicts, by simply saying which of the 2 to use over the other.
Also, the .dme... I mean... Won't merging master into this PR fuck up the list of commits? Like, won't it contain every commit merged to master for the last week?

ktccd:
I mean, it's really freaking messy... but I could maybe just remove my .dme and .dmi changes, commit with master-versions of them and send a new PR with only those changed? It feels really messy though >_<.

ktccd:
How can the .dme have conflicts when I took the file from master? What.....

code/citadel/organs/genitals.dm
```diff
@@ -32,6 +32,12 @@
 	update_appearance()
 	update_link()
 
+////Size-changing procs should return TRUE if size change happened, FALSE if not
+/obj/item/organ/genital/proc/grow_size()
+
+/obj/item/organ/genital/proc/shrink_size()
+
+
```
TalkingCactus: These both should be combined into `AdjustSize(amt)` so it can accept an argument for changing the size. Look at the arousal setting procs for clues for achieving this. This proc would do `organ.size = (organ.size + amt)` and clamp it within the bounds set through the defines in `cit_defines.dm`. Doing this allows you to do increase and decreases in the same proc.

After that, add another proc called `SetSize(size)` which can be used to force a size change to a specific point.

TalkingCactus (changes requested):

ktccd:
Trying out this conflict solution. Praying to God it works....
I just added my virology file to the master-branch version of the .dme

ktccd:
Since this compiled now, I'm just putting out a reminder that this is only up here for review.
*No merging* until... IDK. I need to make a 2nd PR with the breasts_onmob.dmi file to go with this, but I don't think I can MAKE the PR until this merges... and I don't want this to merge without it... since that'd make boobs not work without the file... :S

ktccd:
WHAT. WHY? WHEN? HOW? It passed already!

ktccd:
Resolved conflicts. I'm assuming whoever renamed breast shapes to "Pair" instead of "pair" is correct here and going with their changes.

ktccd:
Alright, all conflicts resolved and the other PR is open. pinging @TalkingCactus to review this agin?

ktccd:
Closing it 'cause even if it compiles, I can't actually test if it works after all the conflicts. That, and there's that thing with the .dmi file which kept giving me headaches.

PR #356, opened 2017-04-12
Removes Sizeplay

Poojawa:
Removes sizeplay entirely.


Cw3040:
Yes please.

ktccd:
Does this mean no more huge/small people or size-ray guns?

Poojawa:
That would be correct. You can blame Cactus for this removal.

ktccd:
Whelp... IMO, I never liked those things. Being a different size was enough of a combat changer to be a super-power.

PR #359, opened 2017-04-12
Eevee/Eeveetaur additions!

Squi-RL:
Add: Eevee tail, ears, and neck fur
Add: Eevee taur body
Edit: Adds code to allow players to select the aforementioned Eevee things

[why]: because I was commisioned to make and pull this
here's hoping it works this time...
eevee example a
eevee example b
eevee example c
veetaur example a
veetaur example b
veetaur example c


Squi-RL:
Whoo! no conflicts!

PR #360, opened 2017-04-14
April sync

Poojawa:
So since I guess the preview icons not working isn't a problem, here's the full update then.


LetterJay:
**CONFLICTION**

PR #361, opened 2017-04-14
adds vore back

CitadelStationBot:
:cl: 
add: Added the base vore features back. I'm not sure why this was removed....
/:cl:


LetterJay:
I was logged into the wrong account, but whatever.

PR #362, opened 2017-04-14
Fixes compile issues with vore code

Poojawa:
Some things were changed.


Poojawa:
think that was just about all of 'em.

PR #363, opened 2017-04-15
Fixes the maintbar

Cw3040:
Small fix to maint to fix the maint bar and turns the old maint bar back into storage


Cw3040:
well, i got too hasty and maybe issues

PR #364, opened 2017-04-15
Map Fixes

Cw3040:
PR to fix parts of BoxStation. Changes:

-New maint-bar area re-added, should work properly now
-Old maint-bar turned into storage room
-Small portions of engineering maint were set to the area "Fore-Starboard maint" and were changed to reflect their actual area
-Things should function as expected.


Cw3040:
Something has gone horribly wrong, closing.

PR #391, opened 2017-04-15
Fixes Taur Overlay issues

ktccd:
:cl: ktccd
fix: Committed the icon_state fix that got excluded from my overlay PR due to conflicts. Taurs should now be rendered on the correct layer, shepherd taurs shouldn't have a hole in their body and cow taurs should now no longer have weirdly duplicate icon_state names.
/:cl:

[why]: # 
Because this was supposed to be in my main overlay fix PR, but because github hates binary file conflicts, it had to be done outside it.


ktccd:
@TalkingCactus this is ready for review, btw. It was supposed to be in the overlay fix PR, but due to timing bullshit it needed to be in its own PR to avoid conflicts.

PR #421, opened 2017-04-16
character data slot fix

LetterJay:
:cl:
fix: fixes a runtime where character slots can't be changed
/:cl:


LetterJay:
Travis is being stupid. Need this pushed or people can't swap characters. It compiled.

PR #423, opened 2017-04-16
Species Content

LetterJay:
**Species Content Update**

This is the long awaited, by a good bit, of people who have been wanting the species that were never added back. I'll be adding as many back as I can with what little sprite skill I have. I'm sure there will be issues and a few misaligned pixels, but I'll double back and get help as needed.


Poojawa:
What species were/are missing? I got literally all the markings ported over.

Poojawa:
I ask this because I specifically broke the species down into the consolidated parts to avoid having a massive list of species to maintain, and to make mix-match easier

Poojawa:
work done is now in #429

PR #430, opened 2017-04-17
Ckey locked tails

Poojawa:
Will later also be used for taur bodies, markings, ears, etc.

Uploading to bleeding-edgy for any debugging ideas. 

Current issues:
Human species by default can select everything.
Switching to any other species gives 
`Runtime in preferences.dm, line 1073: Cannot read null.ckeys_allowed`

wew


Poojawa:
Oh, gg. bork'd by github.

PR #452, opened 2017-04-18
[MIRROR] tgstation-server 2.0

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/26271
--------------------
@ Cyberboss 

Note, this also includes some improvements and bug fixes not yet on the servers. (marked with a (new) tag)

Test merges are in
Test merges support no longer requires editing the git config by hand, I found a way to do it without having to edit the remote. `git fetch pull/PR_NUMBER/head:pr-PR_NUMBER, git merge pr-PR_NUMBER` (new)
Update operations no longer has to fetch details for every PR. (new)

Tools to recompile or just clear test merges without updating (new)

Command to update without clearing test merges (new)

Fixed bug preventing active PRs from getting displayed when updating (new)

Fixed bug that caused a server crash if the server restarted after a successful operation but while another operation was running. (new)

Remove some final traces of the old map rotation system. (new)

Automated Fix errors tool.

Support for launching the bot and server when the computer starts up.

Prompt when updating/test merging telling you when you can stop and start another operation (like a test merge) (new)

Update to PR.bat now uploads the compiled changelog (before the test merge happens) (new)

Better readme (new)

Fix for bug that caused the a/b folders to become undeletable files (the libmysql.dll file is now symlinked so that the target never gets deleted during staging operations) (new)

Removed unused code relating to the old version of the bot (new)


LetterJay:
I would have to take the server offline for a bit to do this one.

LetterJay (commented):
I'm gonna have to pick a day to do this. Similar to shutting down the server like I did. I'll have to save the data folders and recreate the entire system with a new config setup.

Poojawa:
Is this still a WIP or?

PR #453, opened 2017-04-18
disables headpatting *wag

Poojawa:
Because there's always something to complain about from the normies.


ktccd:
It was kind of weird for humans to wag their non-existant tail :P

Poojawa:
This more or less just disables the *wag entirely since dna things are dumb.

code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -215,7 +215,7 @@
 		else if(check_zone(M.zone_selected) == "head")
 			M.visible_message("<span class='notice'>[M] gives [src] a pat on the head to make [p_them()] feel better!</span>", \
 						"<span class='notice'>You give [src] a pat on the head to make [p_them()] feel better!</span>")
-			if(dna && dna.species && ((dna.features["tail_lizard"] != "None") || (dna.features["tail_human"] != "None") || ("mam_tail" in dna.species.mutant_bodyparts)))
+			if(dna && dna.species && ((src.dna.features["tail_lizard"] != "None") && (src.dna.features["tail_human"] != "None") && (src.dna.features["mam_tail"] != "None")))
```
TalkingCactus: You don't need `src.` on these, since it's already reading from `src`.

TalkingCactus (changes requested):

code/modules/mob/living/carbon/carbon_defense.dm
Poojawa: Already realized that, tbh. just changing from mam_tail from `in dna.species.mutant_bodyparts` to != "None" "fixed" the headpat

I was simply trying to overcome it before giving up because going home from work.

Poojawa: since, atm, the code is as good as commented out, it doesn't read any bodypart for the *wag

cebutris:
I support this. Even in mammals, contact with the head does not spontaneously induce tail wagging. *wag should only be voluntary.

Poojawa:
Terrible Awoo.

LetterJay:
I feel like this is less of a complaint and more of a bug in regards to it not being specific to any actual species and causing even mobs w/o a tail to wag.

LetterJay:
Honestly, there's no reason to make PRs to cater to the 'normies'. Especially something like this. If anything I'd suggest finding a way to add a check if the mob has a tail or not to initiate the wag as opposed to removing it outright.

TalkingCactus:
The title is misleading, the PR is doing just that.

Poojawa:
In its current state, it isn't actually misleading. the checks aren't correct.

PR #455, opened 2017-04-18
Taur icon fix

ktccd:
:cl: ktccd
fix: Human legs no longer render on top of taur legs.
/:cl:

[why]: # 
It was kind of a temporary fix so taur legs look less obviously weird (for example for driders).
A better fix would require remaking how taurs work entirely to use a new type of leg of some kind... which is a bit beyond my ability right now :S.


ktccd:
uhm....
wait a second, this got an extra commit I am unsure if I planned to send in... the .dmi file shouldn't have been changed... I think?

ktccd:
Closing and remaking it properly....

PR #456, opened 2017-04-18
Fixes Human Leg issue

ktccd:
:cl: ktccd
fix: Taurs no longer render human looking legs on top of their taur legs.
/:cl:

[why]: # 
This is more of a quick-fix for taurs. To do a proper fix, we'd need to make taurs use proper legcode, but that'd need a complete refactor of taurs, which is beyond me at this point.


code/modules/mob/living/carbon/update_icons.dm
```diff
@@ -244,7 +238,12 @@
 	var/list/new_limbs = list()
 	for(var/X in bodyparts)
 		var/obj/item/bodypart/BP = X
-		new_limbs += BP.get_limb_icon()
+
+		if(istaur && (istype(BP, /obj/item/bodypart/r_leg) || istype(BP, /obj/item/bodypart/l_leg)))
+			continue
+		BP.update_limb()
+		new_limbs += BP.get_limb_icon()
```
Poojawa, line 245: this used to be
 `var/image/temp = BP.get_limb_icon()`
`  		if(temp)`
 ` 			new_limbs += temp`

PR #470, opened 2017-04-19
Sprite Accessory Ckey locks

Poojawa:
Enables use of ckey-specific tagging for donater and special use sprite accessories such as tails/body markings/etc. Species are unable to be locked, due to the way roundstart_species is stored in the code. It'll have to be an upstream change.

WIP: now to fix it showing the full path, fweeeee


Poojawa:
https://i.gyazo.com/ee42ba94ac91393dcce79dda70f2e6a7.png what looks wrong in this picture. :thinking:

improvedname:
It being neon pink
Unless you mean the path of course

Poojawa:
Oh right, this is being done better anyway.

PR #499, opened 2017-04-21
Fixes Initializations

### LINKED ISSUE #494: Cannot click dildos
The title really explains it. You cannot left-click on dildos to do stuff like change shape, etc or just to pick up. This also is an issue if they are in your bag, as you have to dump out your whole bag just to get at it.

Fix pls :V

TalkingCactus:
Fixes #494 
Dildos and genitals should now initialize correctly and be able to be interacted with normally.

### DIFF
--- a/code/citadel/toys/dildos.dm
+++ b/code/citadel/toys/dildos.dm
@@ -17,30 +17,7 @@ obj/item/weapon/dildo
 	var/random_color 	= TRUE
 	var/random_size 	= FALSE
 	var/random_shape 	= FALSE
-	//these lists are used to generate random icons, stats, and names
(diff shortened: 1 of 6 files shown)


code/citadel/toys/dildos.dm
```diff
@@ -17,30 +17,7 @@ obj/item/weapon/dildo
 	var/random_color 	= TRUE
 	var/random_size 	= FALSE
 	var/random_shape 	= FALSE
-	//these lists are used to generate random icons, stats, and names
-	var/list/possible_colors = list(//mostly neon colors
```
Poojawa, line 21: You could have simply put these as var/static/list as well. since they're only used here

PR #556, opened 2017-04-25
[MIRROR] Replaces /image with /mutable_appearance, where appropriate

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/26518
--------------------
mutable_appearance is a new 511 feature. In cases where you're creating an image to use as an overlay, it makes more sense to use a mutable_appearance instead if you can. The image will create a static appearance for not just the image but also each intermediate step if you change vars along the way. The mutable appearance avoids this unnecessary and expensive process. The only situation that requires an image instead of a mutable_appearance is if the overlay is supposed to be directional. MA's ignore dir while images don't. I dunno why, probably another BYOND-ism.

I added a convenience function, mutable_appearance(), designed to emulate image(). Also went ahead and set the default plane of /mutable_appearance to FLOAT_PLANE because it's fucking 0 by default.

Several overlays that were images were changed to just text strings when I could. add_overlay("string") has the same result as add_overlay(mutable_appearance(src.icon, "string")) and saves from having to instance the mutable_appearance.


LetterJay:
This needs to be merged before #567

Poojawa:
Christ, finally.

PR #570, opened 2017-04-26
[MIRROR] The blood loss change everyone's been raving about

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/26539
--------------------
Blood loss now scales linearly on a body part once you take 20 brute damage for that part.

Oxygen damage from blood loss now starts at 85% instead of 90%. 

**This brings some much needed sanity to blood loss. Where getting shot in the face/chest/leg would currently mean ZERO blood loss, but 3 shots to the chest would mean you're taking blood loss damage (pale/dizzy/oxyloss/etc.) after 30 seconds.** 

You'll see a little more light bleeding with this change but light bleeding is just that, LIGHT, you can safely ignore it for many minutes before it begins affecting you. To Kor: Its still 2 Goliath hits to the same body part before Miners start bleeding.


TalkingCactus:
Bandages might be used now.

PR #604, opened 2017-04-27
Fix for incorrect/missing extra layers

LetterJay:
:cl:
fix: Fixed layers for taurs/nagas not showing correctly.
/:cl:

It uses the old method. While more expensive to render, it actually works. either I'm overlooking something or MA just desn't like to use more than one extra layer.

Taurs

Nagas


LetterJay:
Speeeeeeeeeeeedmerge

PR #644, opened 2017-04-29
[MIRROR] Replaces Double Agents with Internal Affairs

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/26651
--------------------
:cl: Kor
del: Removed double agents mode.
add: Replaced it with Internal Affairs mode.
/:cl:

Double Agents have been rebranded as Nanotrasen Internal Affairs. Will be easiest just to paste their intro text to explain.

You are the Nanotrasen Internal Affairs Agent!
Your target is suspected of [crime], and you have been tasked with eliminating them by any means necessary to avoid a costly and embarrassing public trial. While you have a license to kill, unneeded property damage or loss of employee life will lead to your contract being terminated.
For the sake of plausible deniability, you have been equipped with an array of captured Syndicate weaponry. PDA blah blah blah uplink code
Finally, watch your back. [target] has friends in high places, and intel suggests someone may have taken out a contract of their own to protect them. 

tl;dr You are a double agent but meant to focus on your target and not kill wantonly. All agents know its double agents. This PR is obviously dead in the water unless headmins want to make them an exception to rule 4.

Why: DA is massively unpopular because it just turns into everyone killing everyone. It doesn't really function as anything other than traitor with more traitors right now. Since there isn't much worth salvaging from the mode currently I figured it'd be a good place to test if solo antags with restrictions are at all fun in our ruleset without risking a mode people actually care about.


Kevak:
I quite like this idea.

PR #648, opened 2017-04-29
Gives help intent groin functionality to wicker dolls

silicons:
Honk!


silicons:
>I just broke line endings

PR #780, opened 2017-05-05
[MIRROR] Moves languages to a holder datum, gives mind its own language holder

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/26776
--------------------
Fixes #26682.
Contains changes from #26650.

:cl: coiax
fix: If you know more than the usual number of languages, you'll now remember them after you get cloned.
del: Humans turning into monkeys will not suddenly be able to speak the monkey language, and will continue to understand and speak human.
add: Ghosts can now modify their own understood languages with the
language menu.
add: Any mob can also open their language menu with the IC->Open Language Menu verb.
add: Silicons can now understand Draconic as part of their internal databases.
/:cl:

- All atoms now have a single var for storing language details, `language_holder`.
- All language procs get this holder when manipulating languages.
- Living mobs, if they possess a mind, return use mind's `language_holder` instead for all language operations.
- When a mind is transfered into a mob, if it does not have any existing `language_holder`, it copies the information from the mob's existing holder. This will mean that fresh minds in swarmer bodies will inherit the swarmer's language details.
- New proc `could_speak_in_language` which determines if the mob COULD speak in a language, ignoring checks like OMNITONGUE or them not knowing the language in the first place. This makes things a lot simpler.
- Language holders have a `shadow_languages` var, which are a list of languages that the mob grants by occupying, but do not transfer with the mind.

### Bugs

- [x] Silicons no longer have machine language.
- [x] Ghost roles, and spawned syndicate cyborgs are unable to understand any languages
- [x] The language menu could not be opened for a syndicate cyborg
- [x] AIs sometimes randomly forget all languages

- [x] Sometimes the language holder disappears, and I don't know why


LetterJay:
@kevinz000 I have no idea how to work with tgui.

PR #888, opened 2017-05-10
[DNM] [MIRROR] Probably fixes the ending cinematic to Gang War

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/27084
--------------------
So it turns out that calling the nuke cinematic with nothing more than MISSED = 1 doesn't give Gang War and appropriate ending. Override doesn't appear to default to mode properly. 

Also might stop it from playing twice or more.


silicons:
this broke master upstream..

silicons:
fix is up

PR #929, opened 2017-05-13
[WIP] Custom Items!

fludd12:
Gives the ability to give select players specific, custom items, as well as allowing them to be restricted to certain job roles. Also adds some for my own character, but that can be atomized if anyone objects.


code/controllers/subsystem/job.dm
```diff
@@ -416,6 +416,23 @@ SUBSYSTEM_DEF(job)
 	if(config.minimal_access_threshold)
 		to_chat(M, "<FONT color='blue'><B>As this station was initially staffed with a [config.jobs_have_minimal_access ? "full crew, only your job's necessities" : "skeleton crew, additional access may"] have been added to your ID card.</B></font>")
 
+	if(M.mind.key in GLOB.custom_item_ckeys)
```
fludd12, line 419: Yes, I know this isn't where it should go. No, I don't know where it should go. Once I find a home for it, though, it should work.

fludd12:
#965 does this a lot better than I did!

PR #965, opened 2017-05-14
Custom Roundstart Items

silicons:
tl;dr loads items by ckey from config, spawns it when that ckey player joins the game as a living person, stuffing it into their backpack, and if that doesn't work, hands and drops on floor if hands are full.
Only currently works for humans, other mobs will have it just drop on the floor.

config should be fomatted like this:
`//File should be in the format of ckey|exact job name/exact job name/or put ALL instead of any job names|/path/to/item=amount;/path/to/item=amount`

`//Each ckey should be in a different line`

`//if there's multiple entries of a single ckey the later ones will add to the earlier definitions.`

`//is obviously a comment.`

`//Recommend defining one job per line, but do what you want.`

Job specifics will support antag roles in addition to regular ones :D

Everything's good, just needs some minor bugtesting!


fludd12:
Well, this completely outdid my system, excellent work!

fludd12:
Might I suggest instead of hands, trying to place it in the backpack, then failing that, hands, and then floor?

silicons:
@fludd12 It forces it into the backpack unless you don't have one or you're not human

silicons:
but yeah, I'm going to add job specifc stuff, hold on.

Poojawa:
If you can slip this into custom bodyparts as well that would be amazing. For things like markings/tails/etc.

silicons:
right resuming work..

silicons:
Alright boys, everything wooorks EXCEPT for roundstart spawning
Aka I'm trying to look for the best place to hook it into mob spawn code..
@LetterJay you want this ONLY roundstart, roundstart + latejoin, or whenever the specified player is created in a station role that matches the configs (which would include adminspawning them into a job)?

Poojawa:
It'll be best for by character basis. since people might like to have multiple special items for various characters/jobs.

silicons:
Kk

silicons:
i
really
need to finish this

Tacolizard:
BRING THIS TO TG YOU DOOFUS

silicons:
one sec.

PR #975, opened 2017-05-15
Cryo icon state and Wizard tweaks

Poojawa:
:cl: Poojawa
fix:  Fixed Cryopod icon states
tweak:  Adjusted Wizard spell power costs
/:cl:

Because no antag deserves that much power on a station this small.

### DIFF
--- a/code/game/gamemodes/wizard/spellbook.dm
+++ b/code/game/gamemodes/wizard/spellbook.dm
@@ -109,7 +109,7 @@
 /datum/spellbook_entry/rod_form
 	name = "Rod Form"
 	spell_type = /obj/effect/proc_holder/spell/targeted/rod_form
-	cost = 3
+	cost = 40
 
 /datum/spellbook_entry/magicm
 	name = "Magic Missile"
@@ -119,6 +119,7 @@
 /datum/spellbook_entry/disintegrate
 	name = "Disintegrate"
 	spell_type = /obj/effect/proc_holder/spell/targeted/touch/disintegrate
+	cost = 6
 
 /datum/spellbook_entry/disabletech
 	name = "Disable Tech"
@@ -140,6 +141,7 @@
 	name = "Time Stop"
 	spell_type = /obj/effect/proc_holder/spell/aoe_turf/conjure/timestop
 	category = "Defensive"
+	cost = 4
 
 /datum/spellbook_entry/smoke
 	name = "Smoke"
@@ -202,6 +204,7 @@
 	name = "Bind Soul"
 	spell_type = /obj/effect/proc_holder/spell/targeted/lichdom
 	category = "Defensive"
+	cost = 4
 
 /datum/spellbook_entry/teslablast
 	name = "Tesla Blast"
@@ -277,7 +280,7 @@
 	if(surplus>=0)
 		dat += "[surplus] left.<br>"
 	return dat
-
+/* // these have never been fun, like, ever.
 /datum/spellbook_entry/item/staffchange
 	name = "Staff of Change"
 	desc = "An artefact that spits bolts of coruscating energy which cause the target's very form to reshape itself."
@@ -287,12 +290,13 @@
 	name = "Staff of Animation"
 	desc = "An arcane staff capable of shooting bolts of eldritch energy which cause inanimate objects to come to life. This magic doesn't affect machines."
 	item_path = /obj/item/weapon/gun/magic/staff/animate
-	category = "Assistance"
+	category = "Assistance" */
 
 /datum/spellbook_entry/item/staffchaos
 	name = "Staff of Chaos"
 	desc = "A caprious tool that can fire all sorts of magic without any rhyme or reason. Using it on people you care about is not recommended."
 	item_path = /obj/item/weapon/gun/magic/staff/chaos
+	cost = 4
 
 /datum/spellbook_entry/item/spellblade
 	name = "Spellblade"
--- a/code/game/gamemodes/wizard/spellbook.dm
+++ b/code/game/gamemodes/wizard/spellbook.dm
@@ -106,10 +106,10 @@
 	name = "Fireball"
 	spell_type = /obj/effect/proc_holder/spell/aimed/fireball
 
+/*
 /datum/spellbook_entry/rod_form
 	name = "Rod Form"
-	spell_type = /obj/effect/proc_holder/spell/targeted/rod_form
-	cost = 40
+	spell_type = /obj/effect/proc_holder/spell/targeted/rod_form */
 
 /datum/spellbook_entry/magicm
 	name = "Magic Missile"
--- a/code/modules/spells/spell_types/rod_form.dm
+++ b/code/modules/spells/spell_types/rod_form.dm
@@ -12,7 +12,6 @@
 	action_icon_state = "immrod"
 
 /obj/effect/proc_holder/spell/targeted/rod_form/cast(list/targets,mob/user = usr)
-	return
 	for(var/mob/living/M in targets)
 		var/turf/start = get_turf(M)
 		var/obj/effect/immovablerod/wizard/W = new(start, get_ranged_target_turf(M, M.dir, (15 + spell_level * 3)))
--- a/code/modules/spells/spell_types/rod_form.dm
+++ b/code/modules/spells/spell_types/rod_form.dm
@@ -4,7 +4,7 @@
 	clothes_req = 1
 	human_req = 0
 	charge_max = 250
-	cooldown_min = 100
+	cooldown_min = 200
 	range = -1
 	include_user = 1
 	invocation = "CLANG!"
@@ -12,6 +12,7 @@
 	action_icon_state = "immrod"
 
 /obj/effect/proc_holder/spell/targeted/rod_form/cast(list/targets,mob/user = usr)
+`	return
 	for(var/mob/living/M in targets)
 		var/turf/start = get_turf(M)
 		var/obj/effect/immovablerod/wizard/W = new(start, get_ranged_target_turf(M, M.dir, (15 + spell_level * 3)))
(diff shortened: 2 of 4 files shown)


code/game/gamemodes/wizard/spellbook.dm
```diff
@@ -109,7 +109,7 @@
 /datum/spellbook_entry/rod_form
 	name = "Rod Form"
 	spell_type = /obj/effect/proc_holder/spell/targeted/rod_form
-	cost = 3
+	cost = 40
```
Kevak: This looks like an 'ided pls nerf' PR. The only tweak to wizard spellcosts is Rodform, which raises cost from 3 to 40.

code/modules/spells/spell_types/rod_form.dm
```diff
@@ -4,14 +4,15 @@
 	clothes_req = 1
 	human_req = 0
 	charge_max = 250
-	cooldown_min = 100
+	cooldown_min = 200
```
Kevak, line 7: Now this bit right here makes sense. Tweaking the charge time is sensible. Raising cost from 3 to inaccessible 40 is not.

code/modules/spells/spell_types/shapeshift.dm
```diff
@@ -14,10 +14,10 @@
...
 	var/list/current_shapes = list()
 	var/list/current_casters = list()
-	var/list/possible_shapes = list(/mob/living/simple_animal/mouse,\
-		/mob/living/simple_animal/pet/dog/corgi,\
-		/mob/living/simple_animal/hostile/carp/ranged/chaos,\
-		/mob/living/simple_animal/bot/ed209,\
+	var/list/possible_shapes = list(/mob/living/simple_animal/mouse,
+		/mob/living/simple_animal/pet/dog/corgi,
+		/mob/living/simple_animal/hostile/carp/ranged/chaos,
+//		/mob/living/simple_animal/bot/ed209,
```
Kevak, line 20: This bit I am soso on, On one hand, removes wizard ED's, on the other hand. Removes wizard ED's. Although to be honest, shapeshift is more or less a memespell. No real big loss.

code/game/gamemodes/wizard/spellbook.dm
```diff
@@ -106,10 +106,10 @@
 	name = "Fireball"
 	spell_type = /obj/effect/proc_holder/spell/aimed/fireball
 
+/*
 /datum/spellbook_entry/rod_form
 	name = "Rod Form"
-	spell_type = /obj/effect/proc_holder/spell/targeted/rod_form
-	cost = 40
+	spell_type = /obj/effect/proc_holder/spell/targeted/rod_form */
```
Kevak: From what I can tell with this, you're cutting it out entirely now. That's even less logical. The rechargetime tweak you made further down already worked. No need to remove a feature.

code/modules/spells/spell_types/shapeshift.dm
Poojawa: infinite taser + infinite cuffs = wew powergamingly bad.

Kevak: Keep in mind ED's are susceptible to ions. As well as lasers. Tasers do not go through windows, which lasers do. ED's have far less durability than a carbonmob on top of that.

Also keep in mind that wizard is balanced around Single Person Vs Entire Station and not Bullshit Ragin. Ragin mages was always meant to be an adminbus gamemode and was never meant to be in regular rotation. The shittery we see daily is because we not only have it in rotation, but also have it weighted very high for some reason.

code/game/gamemodes/wizard/spellbook.dm
Poojawa: It can be admin-given, rather than purchased in spellbooks entirely. 

For real though, it's not even a good idea to begin with. Yes let's become an object that instantly kills everyone it touches and punches through R-wall without issue.

code/game/gamemodes/wizard/spellbook.dm
```diff
@@ -277,7 +280,7 @@
 	if(surplus>=0)
 		dat += "[surplus] left.<br>"
 	return dat
-
+/* // these have never been fun, like, ever.
```
Kevak, line 283: Yes they have. Staff Of Change is one of those gimmick weapons. 50/50 chance you'll kill yourself with it when you Ashdrake an angry assistant.

code/game/gamemodes/wizard/spellbook.dm
Kevak: It instacrits. It does not instakill. If someone is fast enough they can and will snag and pull the critted person away to medbay. Wizards are all about mobility. Of course they're gonna go through Rwall. They have like. Three or four other spells and items that let them bypass secured areas. Maybe even five if we count wands and staffs as separate. Six if we are counting Staff of Chaos's Fireball/Door chance.

Currently rodform is the only reliable way for a wizard to down people (and leave corpses, which is essential for artifacer/necrostone builds) barring Summon Lesser Guns/Arcane Blast. Fireball has a tendency to explode the wiz, lightning bolt is useless, has a charge up time and needs to be tweaked, tesla blast I haven't tried. And we aren't covering Shapeshift cause we're discussing that elsewhere or E-Nath cause that doesn't leave corpses.

code/modules/spells/spell_types/shapeshift.dm
Poojawa: I'm well aware of how Wizard is balanced. and it's made to have an active population of 60+ with a full security team who are able to use their radios and stay together. And also a station that's actually wanting to fight rather than just do the job. 

And the best way to stop shittery, is to nullify the parts that are really shitty.

Kevak: Just dropping wiz as a gamemode entirely would be my suggestion when it's phrased like that. Make it adminbus only until we get a sufficient population. Or raise the population required for it. Either works.

Poojawa: > tweaks base level cost

> UHG JUST DROP IT ENTIRELY

tweaking a few game breaking spells ain't gunna hurt wizards.

Kevak: You seem to be misinterpreting my mental state as well as my point. You are phrasing your counterpoint in a manner that implies that you believe wizard as a gamemode should not occur on low pop. Your stated issue with the spell is it requires large numbers of players to make it balanced. The easiest way to handle that is to raise the population requirement from its current.. I think 10? to 45-50 similar to nukie/blob. It's a gamemode with a certain number of players in mind. Why not simply ensure that it only happens with that number of players?

code/game/gamemodes/wizard/spellbook.dm
Poojawa: Magic missile - tracking stun x4, unblockable
Disintegrate - instant gib, unblockable
Fireball - AoE Fire/explosion, targeted, unblockable. Use in a 3x3 area to win
Lighting bolt - Electrical stun, targeted, unblockable, Use in an ambush situation for instant winning.
Staff of chaos - Instant death, Fireball, or polymorph has options. All unblockable
Wands - Firebolt/death/change
Hulk mode - Laser eyes and immunity to stuns, punches send people flying and stun them
The Traps - multiple damaging traps + a stun trap, easily dropped if running.
Time stop - complete projectile protection/avoidance. 

If your entire argument is about how much you enjoy the rodform as the only reliable way to be robust, I dunno what to tell you fam. anything that instantly puts you in a helpless state is literal death in this game.

Kevak: You again are misinterpreting my point and mindset. Please do not put words into my mouth. It would be appreciated if you toned down the hostility as I am not being hostile towards you. Regardless. I will go over your points.
Magic Missile-Nonlethal, better as a panic stun for bailing out. Does not down long enough to crit without looting stationside gear for use. Not Reliable.
Disintegrate-Does not leave corpses, melee only. Reliable but Unsuited for Artificer and Necromancer Builds
Fireball-Risk of hitting self, among the highest causes of wizard fatalities. Destroys Rwalls after a few shots. Semireliable.
Lightning bolt-needs to be charged before each use unless that has changed. Ambush situations are rarely a thing with wiz. Lightning Bolt/Tesla Blast are however Reliable assuming they are charged in advance. Point ceded for them.
Staff of Chaos-Instant Rejuv, Door Bolt No Effect, Teleport, are also all options. I do believe there are a couple more in it. Keep in mind that about 50% of the Polymorph bolts will make the target more dangerous. Not Reliable. The point being discussed is about Reliable ways to down.
Wands-Reliable but interferes with Artificer Build. Wands run out, Charge is an extra spell required for reliable wandbuild. Overall reliable, ceding this one.
Hulk Mode-Reliable, Melee Only. Ceding.
The Traps-Insufficient experience with spell, unable to provide counterpoints, seems useful for chases based on observation, less so for offense. 
Timestop-Does not drop target. Hilarious when coupled with Summon Lesser Guns or Arcane Blast. Not Reliable for point of discussion but still a fun spell.

silicons:
@Kevak if you think rod form is the only way to reliably down someone I don't know what to tell you.
Unless you're going against simple mobs (Why are you going against simple mobs unless you used gimmicks?!) rod form is NOT the only way to stop a human, and infact it's not that good compared to things like fireball and lightning bolt.
Now, lightning bolt, useless? No, it's one of the most overpowered spells, it does a shitload of damage if you aim it right and it stuns. It's honestly more powerful than I intended it to be when I coded this and had it put in upstream. Fireball is one of the most powerful weapons if you have a way to stunbreak so you can keep your distance from your targets.
Arcane barrage and lesser summon guns are jokes. They're 20 damage. That's nothing compared to fireball which is a hard 50/50 GGNORE on unarmored (And armored too before I nerfed explosions, once again, upstream). Fireball is way more spammable than rodform, and while slower, can demolish almost any obstacle in your path within a reasonable amount of time.

So yeah, rodform isn't the only powerful spell around, and if you can only win wizard mode by using it, it's a you problem.

Poojawa:
So unless there's any other huge issues with Rod form removal, this'll be merged by the end of the day.

PR #986, opened 2017-05-15
[MIRROR] Adds Mass Insert proc to sql

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/27139
--------------------
Takes a table name and list of rows to insert, and inserts them

Rows are in the format of `row[column] = value`

Supports auto generated `ON DUPLICATE KEY UPDATE` querys as well as manually specifying that part of the query.

Generated queries are pretty formatted to make debugging sql errors easy.

Blackbox converted to use the system.

@ Shadowlight213 
@ optimumtact 
@ Jordie0608


Poojawa:
Since this is database stuff.

PR #1051, opened 2017-05-18
No More Free Parasites

cebutris:
Traitors can no longer get free parasites by refunding the box after taking out the injector. Holoparasite injectors (and guardian card decks and holocarp fish sticks) now delete themselves when successfully used. The price for traitor parasites has been increased from 10 to 15


cebutris:
A side effect of this is that you can get infinite cardboard and paper, but it's better than the alternative.

PR #1063, opened 2017-05-20
Adds a staff of the knotfather/knotted staff/dildo staff into the game.

regtaf:
Adds a 'knotted staff' into the game as an admin spawnable item.

**WHAT IT DOES:**
- fires knotted dildos at long ranges similar to the honkmother staff (in fact most of the stats are the same as the honkmother staff), except without all the stunning and honking
- people hit by the magical dildos will get more arousal (5 arousal) every time they're hit by a dildo, and a message that tells them that they're 'slightly more horny'
- basically allows you to hold a dildo on a stick that shoots more dildos to make people horny through **magic**

**WHAT IT DOESN'T DO**
_The dildo was lodged into Assistant's chest!_
**[Common]** **Assistant** says, "**OH GOD THERE'S A DILDO IN MY CHEST HELP!!**"

`i hope you enjoy this idea`


cebutris:
Can you make sure that this doesn't get spawned when a wizard uses "Summon Magic"? We already have the wand of nothing screwing unlucky people out of fun magiks, we don't need this staff also doing that

code/modules/projectiles/projectile/bullets.dm
```diff
@@ -164,6 +164,28 @@
...
+	icon = 'code/citadel/icons/dildo.dmi'
+	icon_state = "dildo_knotted_3"
+	range = 200
+
+/obj/item/projectile/bullet/dildo/on_hit(atom/target, blocked = 0)
+	. = ..()
+	if(iscarbon(target))
+		var/mob/living/carbon/M = target
+		M.arousalloss += 5
+		M.update_arousal_hud()
```
deathride58: M.adjustArousalLoss(5)

Might also wanna add a check around here to see if arousal's even enabled, so people don't get the "You feel slightly hornier..." message if they've got arousal disabled.

regtaf:
the summon magic event has the list of wands/staves hard coded, it doesn't take from the staff/wand object tree (see code\modules\spells\spell_types\rightandwrong.dm line 130 and beyond)

this is why you don't actually see the honkmother staff whenever the staves are summoned! it's just bad luck if you get the 'wand of nothing'

silicons:
Is this clear to merge :^)

regtaf:
if you want 'w'

silicons:
if you want to make people horny use a syringe gun of crocin and bicardine...

PR #1092, opened 2017-05-22
[MIRROR] Gives gangs new surplus rifle, mounted machine gun, and more

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/27195
--------------------
We're pretty overdue for a gun that isn't design to kill you in under 2 seconds. The new surplus rifle (8 points) requires two hands to fire, is very bulky, and it take a few seconds between each shot. Its a cheap way for people to defend a gang's turf without giving them a murderbone tool. Uses 10mm rounds (30 damage a shot). 

The mounted machine gun (50 points) is something I've had sitting around. I wanted to make gang gunfights a little more tactical which is why I added stuff like the sniper rifle. The mounted machine gun is something you buckle into and makes both hands unusable while using. It fires a volley of 10 rounds of 20 damage bullets with moderate spread over 2 seconds, then requires 2.5 seconds to "cool off" before another volley can be fired. The machine gun never runs out of ammo and extends sight range by 3, so its an excellent suppression weapon, but its spread and low damage per bullet means you aren't going to be blowing people away either. Its stationary status mean that the gun is almost useless in open spaces (if you can flank the gunner after their initial volley you have almost 5 seconds to put them down before they can fire again) or any position where someone can fire on you from cover.

I'm also giving gangs access to a sawn-off improvised shotgun (6 points) that fires only one shell at a time.

This PR also adds my previously added gear to the item list I overlooked, which should allow them to be purchased now. 

🆑 Robustin
add: Gang influence is now decentralized, each gangster has their own influence that they can increase by spraying (and protecting) their tags and new influence-enhancing bling. 
add: Gang uniforms are now created based on your gang's color and can be purchased by any gang member. They will increase the wearer's influence and provide protection, but will make it fairly obvious which gang you belong to. 
add: Gangs have access to a new surplus rifle; it is a semi-automatic rifle that is very bulky and has a very low rate of fire. Gang members can buy this gun for just 8 influence.
add: Gangs have access to the new machine gun turret; it unleashes a volley of bullets with an extended view range. It does not run out of ammo, but a significant delay between volleys and its stationary nature leaves the gunner vulnerable to flanking and return fire. Holding down the trigger will allow you to aim the gun while firing. It will cost gangs 50 influence. 
add: The sawn-off improvised shotgun is now available to gangs for 6 influence. With buckshot shells being easy to produce or purchase, this gun gives the most "bang for your buck" but its single-shell capacity will leave you vulnerable during a pitched gunfight. 
add: The Wetwork boots give gangs access to a lightly armored noslip variant. 
tweak: The armored gang outfits are now slightly more resistant to ballistic and bomb damage
/🆑


LetterJay:
Moved to #1130

PR #1080, opened 2017-05-22
The Return of the Flightsuits

silicons:
Performance with flightsuits have been massively increase from some optimizations done before, I believe the server can support them now, can always pull them back out if it can't.
:cl: Skylar Lineman
experimental: Central Command has re-approved flightsuits designs for testing by the local research department!
/:cl:

PR #1236, opened 2017-05-27
Re-enables flightsuit processing subsystem

silicons:
#1080


Poojawa:
technically already merged, so

PR #1298, opened 2017-05-30
[MIRROR] [READY]Replaces R&D accelerator laser with beam rifle, adds MouseMove and mouse parameter tracking to client, adds blast cannons!

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/27618
--------------------
Beam rifles - Hitscan sniper rifles that have a firing delay based on aiming angle!
You hold your mouse down to aim, and when the beam is green you're able to fire, but not before.
You will be penalized if you move your aim based on how much angular change there is, so it will be extremely ineffective at close ranges while at long ranges it will be harder and harder to dodge!
2 seconds to aim, 0.2 second increase per angle changed, 0.6 second increase per tile moved while aiming.
 Pierces 2 structures, does 90 impact structure damage and 30 damage to structures around that structure, has a 50% chance per tile to set tiles on fire within a 5x5 centered on the impact location.
Does 60 damage to mobs, with a 25 damage AOE around the impacted atom.
4 second fire delay.
Can only be fired twohanded.
10 shots per charge, gun is weight class large, and can only be attached to armor and back otherwise.
Can not be fired while unscoped.
All of these stats are negotiable.
And ofcourse, this is more expensive than accelerators.
https://puu.sh/vZO97/dfc349fd56.gif
:cl:
rscadd: Research and Development have recieved designs for new prototype Beam Marksman Rifles. These rifles require a short aiming cycle to fire, however, have extreme velocity over other weapons.
experimental: Aiming time is 2 seconds, hold down mouse to aim, aiming time increases if you change your aim based on angle changed, or if you move while aiming. The weapon can not be fired while unscoped.
rscdel: However, someone tripped and pulled out the power cord while your servers were being updated with the latest revision of accelerator laser cannons. All data have been lost...
/:cl:

This is going to end so well :^)
In comparison after 10 tiles accelerator lasers will have 10 more damage than these so they're arguably not as powerful in some situations <:^)

Ofcourse, everything's codersprited.

Blast cannons - Basically concentrates a bomb's effect to a projectile's approximately 1 wide line of destruction. In almost all cases, crappier than just maxcapping.
The upside?
They're uncapped.
currently adminspawn unless someone has an idea for them, I might just remove them as they're much less thought out than beam rifles which is the primary purpose for the PR.


LetterJay:
>1,691 files changed.

okay.jpg

PR #1328, opened 2017-06-01
[MIRROR] Moves tgstation-server

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/27904
--------------------
See #26534 

The plan is to put these two in their own repository and have it as a submodule on the main repo


Cyberboss:
This should be fine to merge as is if you guys don't care about forking the server tools

PR #1403, opened 2017-06-04
[MIRROR] Removes ragin mages

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/27872
--------------------
Requested by design lead


Poojawa:
As dumb as this mode is, as long as we don't have it in our rotation, it'll be fine.

PR #1411, opened 2017-06-05
TGS3 config changes

LetterJay:
Original:
Merging these early because they are non-functional changes and without them, it makes cloning the main repo with TGS3 an utter pain as @Iamgoofball found out.

    Adds server_to_tool_bridge_port.txt which will be used for DreamDaemon -> Server communication
    Ensures all descriptive comments are prefixed with two #'s so that the server GUI can build it's controls correctly
    Adds a disabled field to map configs for disabling a map without removing it
    Added runtimestation as a disabled map
    Added formatting to admin_ranks.txt


LetterJay:
Closes #1226

PR #1418, opened 2017-06-05
[MIRROR] Removes Revenant scriptures entirely

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/28061
--------------------
:cl: Joan
rscdel: Removes Revenant scriptures entirely.
/:cl:

These are the win more scriptures. You use them only while winning, and they cause you to win more than you already were.
That sucks. Clockcult doesn't need to win more if it's already winning.


Poojawa:
apparantly already handled at some point

PR #1732, opened 2017-06-22
removes 'oof'

LetterJay:
:cl:
sounddel: Removed oof
/:cl:


cebutris:
Truly it is an end of an era

PR #1780, opened 2017-06-26
[MIRROR] Removes database specifiers from sql queries.

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/28839
--------------------
Seriously, these default to the active database, and its not at all likely somebody will use the default name. /tg/ sure has hell doesn't. Doesn't it make more sense to make it easier for server operators to apply these by not making them have to change the database name on each query?


silicons:
@LetterJay server side stuff

PR #1827, opened 2017-06-29
TAUR SWEATERS and keyhole

katnep:
:cl: Fashion and Facism are one and the same
add: BBsweaters now available in clothesmates along with a keyhole sweater as a premium item
/:cl:

[why]: #people want it


cebutris:
@fludd12

fludd12:
everything is wonderful

PR #1828, opened 2017-06-29
STIMPACKS POWERCREEP EDITION

katnep:
:cl: powercreep
add: added stimpacks and stimpack kits to the vendor
as well as my patented powercreep kit of survival pens (fairly priced so as to avoid actual powercreep)

/:cl:

[why]: # because i add this back everytime it gets erased and no one has complained yet, if anything mining players seem to appreciate it


katnep:
wait fuck i broke it, fixing it now

PR #1845, opened 2017-07-01
[MIRROR] Adds the blood-drunk miner as a potential mining "mini" boss

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/28829
--------------------
It's very unique, very aggressive... VERY aggressive.
It dashes at you, fires a KA projectile, and attacks almost as rapidly as a player does in melee.

It does not currently appear since I want to use the tiles from #28760 in the ruin it shows up in.

- [x] Actual effect for crusher trophy, as if you can actually get the trophy off of it without SUPER DYING


Iamgoofball:
why

@LetterJay

PR #1863, opened 2017-07-03
Changes the mining reward katanas to use the regular path

ChayseRamsay:
item/weapon/katana/cursed was intended to be used for the wizard cursed items event, not for a reward.

:cl: RealDonaldTrump
fix: Changes the mining reward katanas to use the main path, as opposed to the katana intended for the wizard cursed items event
/:cl:


cebutris:
Pretty sure the katana is cursed intentionally

ChayseRamsay:
The cursed variant does literally nothing except remove the ability to place it on your belt and back. It was designed to fix a bug involving the no_drop flag during wizard events. It seems pretty pointless using it as the mining loot drop.

PR #1882, opened 2017-07-05
July 5th sync Bleeding-edgy Test Merge

Poojawa:
Just so people can review it for anything I missed.

Basically, test if you can still be a furry, eat people, and if dogborgs work. VG clothing will be fixed eventually, don't much care about it right now.


Poojawa:
Oh.

PR #1883, opened 2017-07-05
July 5th TG sync

Poojawa:
Invalidates all current open pull requests older than this post.

This is to be merged only if the review branch passes inspection.


silicons:
owo

PR #1944, opened 2017-07-08
Fixes underwear not being counted in exposure checks

silicons:


LetterJay:
@kevinz000 fix your shit

TalkingCactus:
I think we should add a var to humans/carbons that just stores the cover flags of all the items you're wearing so we don't need to calculate it so much.

PR #1980, opened 2017-07-13
Holoparasites no longer can heal from within.

king5327:
:cl: Support Holoparasite Fix
add: Healing no longer functions from within the host.
/:cl:

[why]:
Because holoparasite needs to have some risk in healing.

EDIT: I legit have not tested it, I just added one line in Github.

### DIFF
--- a/code/modules/mob/living/simple_animal/guardian/types/support.dm
+++ b/code/modules/mob/living/simple_animal/guardian/types/support.dm
@@ -29,14 +29,15 @@
 /mob/living/simple_animal/hostile/guardian/healer/AttackingTarget()
 	. = ..()
 	if(toggle && iscarbon(target))
-		var/mob/living/carbon/C = target
-		C.adjustBruteLoss(-5)
-		C.adjustFireLoss(-5)
-		C.adjustOxyLoss(-5)
-		C.adjustToxLoss(-5)
-		var/obj/effect/temp_visual/heal/H = new /obj/effect/temp_visual/heal(get_turf(C))
-		if(namedatum)
-			H.color = namedatum.colour
+		if(!(src.loc == summoner))
+			var/mob/living/carbon/C = target
+			C.adjustBruteLoss(-5)
+			C.adjustFireLoss(-5)
+			C.adjustOxyLoss(-5)
+			C.adjustToxLoss(-5)
+			var/obj/effect/temp_visual/heal/H = new /obj/effect/temp_visual/heal(get_turf(C))
+			if(namedatum)
+				H.color = namedatum.colour
 		if(C == summoner)
 			update_health_hud()
 			med_hud_set_health()
--- a/code/modules/mob/living/simple_animal/guardian/types/support.dm
+++ b/code/modules/mob/living/simple_animal/guardian/types/support.dm
@@ -29,15 +29,16 @@
 /mob/living/simple_animal/hostile/guardian/healer/AttackingTarget()
 	. = ..()
 	if(toggle && iscarbon(target))
-		if(!(src.loc == summoner))
-			var/mob/living/carbon/C = target
-			C.adjustBruteLoss(-5)
-			C.adjustFireLoss(-5)
-			C.adjustOxyLoss(-5)
-			C.adjustToxLoss(-5)
-			var/obj/effect/temp_visual/heal/H = new /obj/effect/temp_visual/heal(get_turf(C))
-			if(namedatum)
-				H.color = namedatum.colour
+		if(src.loc == summoner)
+			return
+		var/mob/living/carbon/C = target
+		C.adjustBruteLoss(-5)
+		C.adjustFireLoss(-5)
+		C.adjustOxyLoss(-5)
+		C.adjustToxLoss(-5)
+		var/obj/effect/temp_visual/heal/H = new /obj/effect/temp_visual/heal(get_turf(C))
+		if(namedatum)
+			H.color = namedatum.colour
 		if(C == summoner)
 			update_health_hud()
 			med_hud_set_health()


ktccd:
Wouldn't this make them useless? They're pretty crap in actual combat and if you're out of combat to heal, then there's cheaper ways to do it.... 15TC for an out-of-combat heal, when a medkit or a few pills would do just as well...
That means you're mostly paying for the teleport ability, which while cool is also an out-of-combat ability of some utility use...

king5327:
Hardsuit costs 8 tc.
Parasite costs 15 tc.
Both make you spaceproof, but one also lets you survive three carp attacking you at once (edit: while you're in space too), and can heal itself (by proxy). I figure at the very minimum the Guardian should have to reveal itself to make use of this.

Edit: also, teleport ability is probably worth 3 or 4 TC on its own. Getting somewhere secluded for 5 seconds and suddenly being at your safespot is... wow. An extra second for each other person that's getting sent to their doom/safety.

ktccd:
You aren't factoring in how unsubtle the teleport is, screaming antag to anyone who sees, or the unclonable effect.
Also yes, paying 15TC is a lot. It's NOT a hardsuit, a hardsuit does more than just spaceproofing, it armors you and shit.
Portable spaceproofing (like the syndi-suit): 4TC
out-of-combat Teleport: generously 4TC, though we have no other gear-equivalent.
Combat helper value varies based on robustness of helper, but technically also leads to death by turning to ashes since this model isn't much good in combat.
Healing outside combat: 4TC
Total value of this all is about 12TC, but here's the kicker: I don't think people would buy this even if it was 12TC. for 12TC I can get so much better stuff! Sure, the values are correct, space-suits are 4tc, and so is medical kits, but those aren't considered good value items.


EDIT:
Basically, my problem is this:
15TC is a LOT. Something worth 15TC is going to be THE core of your plans that round. You invest that in a single item to be your MAIN weapon and tool for antag stuff. It's not unreasonable to heal without being OBVIOUSLY seen (you still have a unique particle effect that I think is noticable by others?) and it's not so much that you can't be killed through the heal (unless they use only fists, but 15TC vs fists should not be a contest). 
Compare other such expensive items and we got a gun that two-shot destroys you, the revolver, and it costs LESS than this.

code/modules/mob/living/simple_animal/guardian/types/support.dm
```diff
@@ -29,14 +29,15 @@
...
 	if(toggle && iscarbon(target))
-		var/mob/living/carbon/C = target
-		C.adjustBruteLoss(-5)
-		C.adjustFireLoss(-5)
-		C.adjustOxyLoss(-5)
-		C.adjustToxLoss(-5)
-		var/obj/effect/temp_visual/heal/H = new /obj/effect/temp_visual/heal(get_turf(C))
-		if(namedatum)
-			H.color = namedatum.colour
+		if(!(src.loc == summoner))
```
silicons: remove src.

code/modules/mob/living/simple_animal/guardian/types/support.dm
```diff
@@ -29,14 +29,15 @@
 /mob/living/simple_animal/hostile/guardian/healer/AttackingTarget()
 	. = ..()
 	if(toggle && iscarbon(target))
```
silicons, line 31: infact replace your specific checks with a
if(src.loc == summoner)
     return

silicons:
:-1: imo unnecessary

king5327:
@ktccd 
Keep in mind however that, having the hardsuit slot free, it is very much possible for the user to get heavier armor. The case in point I'm referencing is that a few games ago a traitor in this setup raided the armory, got sec armor, survived a bunch of fights with both crew and carp alike, and managed to survive like an hour of constant blood loss too.

The only thing I can't tell is whether or not they had been using internals, but that's the last of my problems.

ktccd:
That doesn't sound unfeasible though? It sounds pretty anectdotal.
I've seen rounds where the revolver was used to kill almost the entire crew by just stocking their backpack with some cheap speedloaders and then two-shotting everyone. I've seen double-esword users empty a server.
The round you're referencing is anectdotal evidence that I can't really make use of, it's hard to separate all the vague information there: How many armed people did he fight, how well armed were they, how many times did he have to fight, how often? Was his success due to personal skill, lack of security skill or luck? did raiding the armory help a lot or how much was the armory gear responsible for winning?

The fact is, the parasite does nothing to stop stuns, slips or stamina damage (as I've myself experienced my the user being simply knocked down with the secborg laser knocking you out), There's definately counterplay as these are the most common ways to lose a fight. Even if it freed up a hardsuit slot, I can't really see this being bought if it costs you 75% of your resources AND doesn't help significantly in combat. I'd MUCH rather buy ANY of the other models which are all rather useful to me in combat.

king5327:
The entire success was due to the guardian being able to heal damage faster than it was dealt, as well as drag the guy off. It was 4v1, they downed and started beating the person. Parasite pops out, drags them off while healing, and from there it just looked like a scene from a slapstick. Cuffs were on, people were flying around, throwing things to try to get to the target, meanwhile the guardian dragged the guy around until they finally got 5 seconds of peace and poofed to the safe spot.

ktccd:
That sounds like none of you were used to fighting holoparasites in general. The best way to kill one is for YOU to be the one dragging the person while shooting them, they have trouble keeping up and have a much harder time protecting their owner while also fighting you off.
Not saying the parasite isn't strong, but it's not as unbeatable as you're making it out to be...

silicons:
as someone who plays security for hours on end on servers much worse than citadel in terms of murderbone and destruction by antags i can tell you that when holoparas were enabled there i never had trouble killing them.

ChayseRamsay:
This is a case of "I died plis nerf" in my opinion.

silicons:
@Citadel-Station-13/group-a reopen and merge if you want this but otherwise no

PR #2027, opened 2017-07-15
Enable borer random event

ktccd:

:cl: ktccd
tweak: Enabled random borer event, increased minimum timer before borer event can happen, decreased possibility of borer event, down to default value, increased minimum living, non-AFK players required for event to happen.
/:cl:

Because borers are fun, but lowpop borers are less fun. Also our rounds are expected to last longer than TG rounds, so increased timer and lower weight is required if we don't want borers to happen on every extended round. I don't even know why the weight was HIGHER than the default before....

Final note: I kind of want J to give me his thoughts on this, because I have a vague memory of him changing this at some point so IDK if this was removed because of TG synching or not.
(I know TG disabled it randomly happening before they removed it completely for example, so it might've been a result of a TG-synch)

### DIFF
--- a/code/game/gamemodes/miniantags/borer/borer_event.dm
+++ b/code/game/gamemodes/miniantags/borer/borer_event.dm
@@ -3,6 +3,7 @@
 	typepath = /datum/round_event/borer
 	weight = 10 //Default weight
 	max_occurrences = 1
+	min_players = 20 //10 is MINIMUM needed, but this is not a gamemode that does well in lowpop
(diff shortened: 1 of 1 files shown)


ktccd:
@LetterJay I have a vague memory of you being involved in the last change, or at least of having some discussion with you about it, so I'd like to hear from you about this? :S

ktccd:
Alright, talked to Jay about it on the coderbus channel, no need for talking here also then I guess :D.

silicons:
@LetterJay

PR #2039, opened 2017-07-17
[MIRROR] Beam Rifle Zooming Rework + Click catcher memes + bunch of other random stuff that players won't use

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/28551
--------------------
- Click catcher now supports mousemove/mousedrag. Basically, mouse location can now be captured even if the user isn't mousing over a visible location
- Added procs to allow for projectiles to be fired with just an angle for pixel projectiles, instead of requiring a target turf and pixel x/y
- Added procs to get angle of user's mouse from their viewpoint (Time to rework gang machine guns again!)
- Beam rifles now have different zoom modes
- Free directional zooms out and tracks the angle of your mouse from the center of the screen. However, you can't target with very good accuracy on this (Shots can't be properly aimed on non dense objects/lying down mobs.)
- Locked directional zooms like free directional but doesn't automatically turn if your aim changes.
- Center view, just increases your view in all directions (2x weaker)
- No zoom mode, in which you just retain your normal view.
- You can select beam rifle zooming rates to be instant or stepped. 
- Stepped zooming rates zoom out 5 tiles per second. This will likely help with people not being able to use it without lagging because their computers aren't as beefy!
- Beam rifles no longer require zoom to be fired
- Beam rifle aiming beams now instantly update instead of on process
- Beam rifle aiming beams are now one object instead of 150. This'll help with the lag caused by it during gameplay that I've observed.
- Angular penalty reduced by 0.1 for a nice even number.
- Instances of client.view = <view> have been replaced with client.change_view() as that'll properly update the click catcher
- Hopefully shooting yourself in the face when you hit a blob tile or whatnot is fixed with the new and improved code..
Props to @ RemieRichards and @ MrPerson!

closes #2055
closes #2060
closes #2109


silicons:
there's a truckload of issues this caused on upstream so don't merge until my fix is mirrored...

PR #2140, opened 2017-07-26
Delayed objective [Latejoiners can now be selected as traitor targets]

ktccd:

:cl: ktccd
add: A new objective is now available for traitors, where their target to assassinate has not yet arrived at the station.
fix: Random objective target that used to not be random now is.
/:cl:

Dodging antag targets is "unfun", encourages lowpop gamemodes and mulligans and midround antags don't help it much.
This might only apply to traitors, but it's something and **I want feedback** on this. Should it be more likely than normal kill objectives? Less? Equal? What about the timer, is 1min to 10min randomly a good size to avoid people stalking near arrivals and such for their possible targets?

As for the target thing, it's something I noticed when I looked at their code for reference, the NORMAL find_target_by_role does NOTHING to randomise the list of possible targets, so ANYONE using it will target the same, unfortunate person in a non-random way. I think this is why team-lings always had such an easy time with their kill objective, since team-lings is one of the few antags that re-use this proc that I know of (to avoid targetting each other).

### DIFF
--- a/code/_globalvars/lists/mobs.dm
+++ b/code/_globalvars/lists/mobs.dm
@@ -23,3 +23,5 @@ GLOBAL_LIST_EMPTY(simple_animals)
 
 GLOBAL_LIST_EMPTY(language_datum_instances)
 GLOBAL_LIST_EMPTY(all_languages)
+
+GLOBAL_LIST_EMPTY(latejoiners) //All latejoining people, for traitor-target purposes.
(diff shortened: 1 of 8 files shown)


LetterJay:
This ready to merge? @ktccd

ktccd:
Well, yes, it works and all, I was just kind of wanting to see if anyone's got anything to say about it, any input or something :S.

deathride58:
+support, antags really need more objectives, and this seems like a pretty unique one

katnep:
I like it, currently people are late joining so they don't have to worry about being a target, this will hopefully either discourage that behavior or right the problem this causes

PR #2170, opened 2017-07-28
[MIRROR] Adds 3D Sound System

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/29455
--------------------
Adds in a 3D sound system, courtesy of Paradise and Baystation.

The gist of this system is that sound carry further, but falloff in a manner relative to distance.

There's also a performance boost to playing sounds, as well. Previously, when a sound was played, it'd loop through everyone who could hear that sound and create a new sound object for them, then delete it.

This changes that so the sound object is only created *once* then modified to every single person that hears it. This optimization can be put into place should this generalized system fail to go through.

I'm sure the explosion sound handling could be changed to be a bit different with this, but I really don't want to touch that.

:cl: Fox McCloud
add: Sound should carry further, but should get quieter and quieter the further you are from it
/:cl:


Poojawa:
Why did this get merged with .rej

PR #2197, opened 2017-07-31
[MIRROR] Adds Surgical toolarms to R&D, makes them less robust

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/29365
--------------------
Also removes the W_CLASS override from arm augments, it didn't appear to be doing anything at any rate.

:cl:
add: Surgical toolarm to protolathe and exofab
tweak: Toolarm tools are less robust but more efficient at surgery
tweak: Arm augments are no longer a huge item
/:cl:

[why]: Same as before. It doesn't feel good for the surgical toolarm to be left to languish, and keeps the tools no more effective than a kitchen knife in terms of damage.

Toolarm tools do no more than 10 force damage and do not exceed W_CLASS SMALL. This will reduce their chances of knockdown and their actual effectiveness in combat. I hope everyone feels this version is less powercreep-y.


ktccd:
@kevinz000 You mentioned on the original PR it was broken, is... is things still broken?

PR #2242, opened 2017-08-03
Fixes cloning and DNA change losing genitals.

ktccd:

:cl: ktccd
fix: Cloning and podding people no longer forgets to put the genitals back in you.
fix: Changeling and other DNA transformations now apply the appropriate genitalia.
/:cl:

Because miners kept losing their penis in lavaland and can't even find it to transplant it back. Also, things in their DNA should really follow with the clone.
And comparing the genitals to find who is a changeling got old really fast. Changelings are powerful enough to fake a dong if they need to...
(Also, this applies to transformation stings! So have fun stinging someone's dick away)


sirelgate:
yey

PR #2329, opened 2017-08-10
Adds Taur cocks

ktccd:

:cl: ktccd
add: Adds flared, knotted or tapered dicks for Taur bodies.
tweak: Resized the cow-taur's over-extended belly to match the profile picture, no longer completely blocking the taur penises.
Fix: Barbed, knotted penises and other such long-named cocks are now visible on the ground if dismembered.
/:cl:

Because Taurs wants to have cocks situated between their hind-legs.
And the cow magically got fatter if you looked at it from the front than from the side, this was obviously an error.


ktccd:
WAIT, don't merge yet! I forgot to make them have an icon if dismembered!
EDIT: Still working on it, turns out we had a flawed design in how we drew them, if I fix this then barbed-knotted cocks will stop being invisible also.

ktccd:
OK done now I hope!
Fixed a lot of stuff but it's mostly a refactoring so the bugfix is ALMOST not visible to players.
Only players who had barbed, knotted penises AND who was gibbed and tried to find their now-invisible penis would've noticed that bug. No wonder it was unknown.

PR #2374, opened 2017-08-14
xenomorph round start update

sirelgate:
A long overdue overhall and readdition of xenos

This updates the long dead xenomorph round start race with one of my own creation. Abit weaker to fire, but tougher in bodily chitin than their previous player counterparts. They have a long head that restricts helmets and are digigrade. They can wear masks and have all the functioning human organs with no hold over's from xenos.
They are effectively humans that have gained the external appearance and slightly weaknesses and toughnesses of the unrelenting xenomorph hive. 

They have some small level of armor standing in as their chitin, and to make up for the no-helmet and digigrade tag. But much like xenomorphs they are weaker to fire and heat.

Currently the race comes with both genders and head and dorsal variations.


sirelgate:
fixed a few over sights, it's ready to rock and roll

sirelgate:
added pierce immune to fix the issue with digigrade legs, since shoes give pierce immune

PR #2469, opened 2017-08-23
[MIRROR] Refactors forensics to a datum

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/29542
--------------------
there is now a forensics datum that contains like, blood, fibers, adminprints, and normal prints. everything uses it now.

No changelog because there is no visible change except in the code

Please note: /atoms still have print/fiber/blood related procs which just points to the forensics datum's procs


silicons:
TG is fucked hard.

PR #2509, opened 2017-08-26
[MIRROR] Pulls the plug on Cerestation (for real this time)

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/30196
--------------------
:cl: MMMiracles
del: Cerestation has been decommissioned. Nanotrasen apologizes for any spikes of suicidal tendencies, sporadic outbursts of primitive anger, and other issues that may of been caused during the station's run.
/:cl:

[why]: # The more I work on remapping the departments to resolve complaints the more I wonder if I'm just better off scrapping it entirely and starting from scratch.

Yeah, probably shouldn't of held up the initial removal PR, but I had some hopes of trying to salvage the time I spent but I've come to the point where I might as well start over with a better base with what I know now instead of several months back. 

The shuttle stays in though because I've only heard positives about it and because having the space titanic as a shuttle is funny.

:pray:


kevinz000:
would like player input.

silicons:
you know what?
not my problem :^)

ktccd:
Needs to be merged, since none of the problems that plagues Cerestation will be fixed since the maker of the map has decided to start off from scratch.
And the map is obviously also intended for huge population sizes, the likes of which we rarely see here, on top of somehow rolling it as lowpop making the entire round an unbalanced mess...

PR #2511, opened 2017-08-26
Alright alright instead of syncing icons to TG lets just fix the code instead!

silicons:
I'm HOPING this does it properly, but I'll be going through and seeing if any are cit-specific.


ktccd:
I'm not seeing anything directly wrong, but I *did* just wake up and I'm only half-aware of what this is all about :S.

PR #2513, opened 2017-08-27
Lets humans use MUTCOLORs [Maximum Heresy]

ktccd:
:cl: ktccd
tweak: Heresy has come to the station, humans must now select custom skin colours! Specific RGB values can be enterred in order to get just the right tone of skin your special snowflake needs, just like you can for other humanoids!
/:cl:

Exactly what it sounds like. Enables humans to use mutant colours and disables the preset skintones.
Preset skintones are still doable, since mutant colours can be accurately adjusted by putting in RGB values for the colours, but it'd be more manual work. On the other hand, you can now get a more accurate
skin tones. Prepare for the incoming orcs, elves and other special people, the heresy has just begun.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/humans.dm
+++ b/code/modules/mob/living/carbon/human/species_types/humans.dm
@@ -2,10 +2,10 @@
 	name = "Human"
 	id = "human"
 	default_color = "FFFFFF"
-	species_traits = list(MUTCOLORS_PARTSONLY,EYECOLOR,HAIR,FACEHAIR,LIPS)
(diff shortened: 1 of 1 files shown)


ktccd:
Let's wait with merging this stuff until we get some feedback from people about this.
Do people prefer having a short list of human skintones to pick from, or do they want the freedom of more colours, but needing to pick them manually rather than just from a named list of a few?

ktccd:
Well, it's a test-merge in practice at least. Let's see how people like it.

silicons:
h-haha

PR #2582, opened 2017-09-02
Fixes headpatting *wag and Vore spaceworthiness

Poojawa:
Because dying to exposure in someone/a borg is not intended.


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -613,7 +613,7 @@ INITIALIZE_IMMEDIATE(/mob/living/carbon/human/dummy)
 
 	//Check for dresscode violations
 	if(istype(head, /obj/item/clothing/head/wizard) || istype(head, /obj/item/clothing/head/helmet/space/hardsuit/wizard) || istype(head, /obj/item/clothing/head/helmet/space/hardsuit/syndi) || istype(head, /obj/item/clothing/head/helmet/space/hardsuit/shielded/syndi))
-		threatcount += 3
```
silicons, line 616: WHY

Poojawa: Because people running around in syndie hardsuits for the memes. :v I've pretty much always had it this threatening

Poojawa: It doesn't make sense that Beepsky will arrest monkeys, but not punish contraband.

PR #2583, opened 2017-09-03
Extra Screenshake and Screenshake accessories.

deathride58:
["When someone shows you a game and asks you how to improve it, usually the answer is 'Hey, add some screenshake' because nobody does it and it's super easy and it makes your game better." - Jan Willem Nijman](https://youtu.be/AJdEqssNZ-U?t=1000)

ss13eswordscreenshake
ss13pistolscreenshake
ss13shotgunscreenshake

PR does exactly as it says on the tin. It's modular.

:cl: deathride58
add: Firing most guns will now cause your screen to shake subtly.
add: Being hit by melee weapons with a force over 10 will now cause subtle screenshake
add: Hitting someone with a melee weapon with a force over 10 will now cause subtle screenshake
add: Hitting an object with a melee weapon with a force over 20 will now cause subtle screenshake
add: Added two options to the Game Settings menu: Screen Shake and Damage Screen Shake. Screen Shake controls (most) screenshake in the game, and allows you to set the percentage of screenshake from most sources, or simply turn it off entirely. Damage Screen Shake controls the screenshake when you get hit by weapons, and has three settings, on, off, and only when down. The defaults are 100%, and only when down, respectively.
/:cl:


Poojawa:
While I can appreciate the 'force feedback' on item use, Make this an optional thing with it defaulted to off. Some people might get some serious motion sickness.

This also makes it insanely hard to click people in some instances.

deathride58:
Alright, added a setting to enable/disable/reduce screenshake in it's entirety. The "reduce" setting for screenshake will reduce screenshake to 25% of it's original strength. Also added a setting to disable only the screenshake caused by item damage, including the option to only show it when knocked down.
By default, screenshake is enabled, and item damage screenshake is set to only show when down. Quick testing with macros shows that all screenshake caused by your own actions will end well before click delay ends, with the only exception being items var-edited to have 300+ force. It's small enough to not have a negative effect on gameplay, but big enough to add noticeable impact where you'd expect it.
With the defaults, combat will feel a little bit more satisfying, yet will remain completely unaffected, even if your assailant is able to land their hits first.

Since these settings are able to be changed ingame via the game preferences menu, it shouldn't be too hard to toggle them off on the fly.

katnep:
ayy just saying, isn't the sniper rifle is sorta kinda balanced by the screenshake, since it makes it impossible to dodge and see after firing

deathride58:
Since sniper rifle screenshake lasts for a mere fraction of a second, not really. It ends long before you can fire again, attack, or move three tiles. Most, if not all, screenshake in TG exists purely for flavor, albeit with some rare cases having unintentional negative effects on gameplay. This PR adds an option to either reduce or entirely remove said screenshake, which gives the players a choice between fancy effects, more accuracy in rare cases, or somewhere in between.

LetterJay:
This is getting stale. @deathride58

PR #2592, opened 2017-09-04
Disables the swarmer events for now

LetterJay:
I feel like it's more proper to disable them via var settings than commenting out the entire file. I also feel it's better to go back and look at these before flatout removing them from the event table.

:cl: Jay
del: Removes swarmers from the event table for the time being.
/:cl:

reverts #2584

### DIFF
--- a/code/game/gamemodes/miniantags/bot_swarm/swarmer_event.dm
+++ b/code/game/gamemodes/miniantags/bot_swarm/swarmer_event.dm
@@ -1,9 +1,8 @@
-/* 
 /datum/round_event_control/spawn_swarmer
 	name = "Spawn Swarmer Shell"
 	typepath = /datum/round_event/spawn_swarmer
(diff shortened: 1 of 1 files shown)


Poojawa:
Yee probably.

I still feel swarmers don't add anything to the game, and you can't ban people without just blanket banning them from all ghost roles.

Been seeing Swarmers attempt to eat the SME doors, actively hunt people, get on the shuttle to try and teleport people into space. Just general station grief.

katnep:
im just throwing this out there, but wouldn't alot of these problems be solved if we added a prompt on the swarmer much like the drones one, explaining what your job is, like big red text and everything
and for counterplay ion carbines are cheap AF

Poojawa:
You can't make ion carbines if the swarmers eat R&D, and they have the lawset already told to them.

PR #2611, opened 2017-09-05
[MIRROR] Fix cargo shuttle not docking correctly

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/30422
--------------------
#30356 made returning null in `dock` not okay

The other dock overloads probably need fixing too.

@ AnturK


Poojawa:
Already implmented. also terrible mirror

PR #2622, opened 2017-09-06
Cryopods for logging off

Poojawa:
:cl: Poojawa
add: Ports Paradises' Cryopod system, allowing players to exit a round without clogging up important job slots.
/:cl:

WIP atm, and will likely be hand-in-hand with a whitelist permitted respawn mechanic so people can rejoin a round. This is intented for heads of staff players who didn't feel up to playing whatever role but got random selected into it by the game.  We've all been there, after all.

Wew.


cebutris:
This will probably need guidelines for use when added, but that's a rules issue. 

Also, the CL says "Cyropod" instead of "Cryopod"

Poojawa:
wew. gunna keep working on this.

PR #2625, opened 2017-09-06
Adds space ruin oldstation.dmm to blacklist

Raeschen:
:cl: Raeschen
tweak: oldstation.dmm blacklisted
/:cl:

Oldstation.dmm creates like 130 hivebots which create 250-300ms npc pool MC usage. I tried editing it previously but then a tg mirror reverted it, so this is to try disabling it instead.


LetterJay:
The config folder isn't affected by PRs.

Raeschen:
@LetterJay Oh, apologies. would you disable it your side then? Creates so much lag

PR #2643, opened 2017-09-07
[MIRROR] Ammo cleanup

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/30361
--------------------
Creates a one-to-one mapping between casing and bullet types to make future changes easier. No longer will we have to deal with `weakbullet3` being shared across multiple ammunitions.

Cleans up ammo definitions to make them easier to read and maintain. It's far from perfect, but its a start.

Removes the gang, haem, and toxic bullets for being unnecessary. Some of you might be thinking: "Those sound cool!!! Keep them!!!" But in reality they were just renamed regular bullets that dealt less damage.
 
Combines meteorshot bullets with breaching bullet. These were practically the same anyways.

Removes overloaded improvised shells because people complain that they are broken. Someone can readd these when they figure out how to make them not broken. The regular improvised shells remain.


LetterJay:
This PR assumes gangs were removed.

PR #2694, opened 2017-09-11
Applies WRITE_FILE to Citadel saves + Belly clone fix

Poojawa:
Applies the WRITE_FILE status to our snowflake save file stuff, same as TG's is.

Also should transfer over saved belly preferences to newly cloned mobs.


Poojawa:
Need to test that this even works first, can't do testing on my laptop anymore because fuck knows.

Poojawa:
the cloner meme is annoying as fuck. uhg.

Basically, all it needs to do is perform the same copy_to code the vore/living.dm does upon mob creation. Sooo.

PR #2704, opened 2017-09-11
[MIRROR] Makes the station Z level into a list instead of a single define

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/30297
--------------------
For multi z

I left a define for the main/first floor of multi z maps because I don't want to rewrite weather right now, and we might want some things to only impact the main floor.

I'll either add mapping helpers that add more zs to the list or beg for help to add/read lists from the map jsons


Poojawa:
Shit is broken as fuck

PR #2723, opened 2017-09-12
[MIRROR] Fixes adjust config

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/30654
--------------------
Closes #30651
Changes config format because config depending on order is insane.
Also bit more meaningful name.


Poojawa:
We need probably to revert #2654 to the .sav file again and risk not having workable pets again.

PR #2747, opened 2017-09-14
Makes APC cell_type actually pick a cell

Poojawa:
Ported from https://github.com/tgstation/tgstation/pull/30327

Setting a APC's cell_type just sets the max_charge var for the standard cell inside, resulting in a ton of var-editted increased capacity standard cells being spread out around the station.

This makes it so those cells are actually the proper types instead of just being a mass of varedited ones.

regex;
/obj/machinery/power/apc{\r\n cell_type = 10000;
to
/obj/machinery/power/apc/highcap/ten_k{

.

/obj/machinery/power/apc{\r\n cell_type = 15000;
to
/obj/machinery/power/apc/highcap/fifteen_k{

.

/obj/machinery/power/apc{\r\n cell_type = 5000;
to
/obj/machinery/power/apc/highcap/five_k{

.

/obj/machinery/power/apc{\r\n cell_type = 2500;
to
/obj/machinery/power/apc{
dirty edit.

.

cell_type = 0
to
start_charge = 0
^ (this was bad since it means that those cells would NEVER hold a charge and shouldn't have been done in the first place.)

Then just clean up any leftover cell_type = (there should only be a few if any)


Poojawa:
Delta station wasn't converted to .tgm or something.

PR #2867, opened 2017-09-20
Maybe removes eggsacs from xeno based meaty ores

Poojawa:
Because people don't understand why its a problem.


ktccd:
thank you, so very much. It's like, either this or hand out mass-banning for people touching the things...
It happens EVERY damn shift and people KNOW they don't get in trouble for doing it. I fucking feel it, people do it on purpose to end "boring" extended shifts....

PR #2883, opened 2017-09-21
Adds Spinfusors

Michiyamenotehifunana:
:cl: Toriate
add: Adds Spinfusors
add: Spinfusors and their disks can be made with the protolathe in science
/:cl:

It's a neat little idea stolen straight from Tribes. The idea is that this thing would shoot flightsuits easily out of the air, cutting down on shitlerism with a hard counter. The explosion has been tested to not punch holes everywhere, so collateral damage shouldn't be too intense when you're trying to give blue disk specials to flying dickwads.


Michiyamenotehifunana:
This thing requires some slight reworking, such as not letting a bunch of eggheads spawn infinite rocket launchers.

Issue #2880, opened 2017-09-20
Examining breaks when a ` is present in your flavortext


cebutris:
Basically, if a ` (backtick, NOT an apostrophe) is present in the flavortext, your character will be entirely unable to be examined

Poojawa:
Since they touched it last, and moved it into features. 

Probably just needs a sanitize check rewrite. it was never supposed to be a DNA text feature.

ktccd:
Yeah, I'll get to work on this tomorrow (Don't have more time right now before uni). Sanitizing weird character does seem like a good idea >_<.

PR #2926, opened 2017-09-24
Adds sanitation to flavor_text

ktccd:
No changelog needed really. Just adding a fix I really should've put in when I last touched flavor_text, but I just assumed all input was already sanitized.
Which was kind of silly, so now it is.
Tested inputting various symbols, including the ` (backtick) and was able to examine my character properly.


ktccd:
Resolves https://github.com/Citadel-Station-13/Citadel-Station-13/issues/2880 I hope.

PR #3016, opened 2017-09-30
[MIRROR] Stops putting RCD, bluespace bags, and staves into ORM

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/31093
--------------------
Closes #30764


WewLads:
This doesn't work with Bluespace Mining Satchels as you forgot to add the keyword to bluespace mining satchels.

PR #3093, opened 2017-10-03
Shock Collars Are Now A Thing.

Kevak:
Exactly what it says on the tin.

dreamseeker_2017-10-03_07-34-54

Shockcollars are now a thing.

They function more or less identically to electropacks+a bit more fluff to the messages it gives the wearer and minus the capacity to be turned off on account of electropacks having rather fucky definers with how item pathing works that I'll clean up later in a different pr. They cannot be removed by the wearer. They can be removed by other people. They obviously go in the neck slot. To use, acquire remote signaler and tune to frequency, like with electropacks.

Shock collars have been added to the hacked item list in autolathes. Check the security tab. They cost a tad less than electropacks on account of being collars and not fuckoff backpacks.

Did testing, shouldn't be any bugs. The ones that I found have been removed.

:cl: Kevak/Neri
add: Added Shock Collars to hacked autolathes.
imageadd: On character sprites, credit to Toriate
imageadd: Object sprite, credit to Toriate
/:cl:

-Why. Does one really need to ask 'why' here? Tis kinky as fuck, we don't have enough collar types, and it is neat.

~Edit. Made a small tweak. You can now name the collars of your victims.


pantherpusher:
Would you consider having it become actually dangerous when emagged?

Kevak:
I could probably do that. It would not take very much time/effort.  The issue is with balance. Considering how spammable it is (once ever 7ish seconds, 7-8ish second stun+however much damage the theoretical emagging would do. Adds up fast, cannot be removed or negated in any way outside of wearing an exosuit that covers the head fully to prevent putting on in the first place/wearing something in neckslot.) and how easy it is to put on someone(Unlike electropacks and back slots, very few people wear things on their neck, resulting in a 10 second application of collar rather than the 30+ with an electropack to remove backpack and then attach the EP. Which is not removable by the wearer.). It's the reason electropacks don't have a damage via emag function from what I understand, despite being notably more time consuming to put on someone.

I'd say that's up to the maintainers to decide if I should do it or not. Prolly in a later PR if at all. I personally don't think it's a good idea for balance reasons however.

pantherpusher:
Understandable

Kevak:
In a later PR I might come back to it and do something with adjusting the stuntime/recharge time and maybe make it burn out after a few uses if emagged. But not for this PR. This PR is just making it a thing in the first place.

Kevak:
Made a small tweak. Normal collar functionality is now enabled. (Aka ye can name em.)

code/game/objects/items/devices/radio/electropack.dm
```diff
@@ -143,4 +143,82 @@ Code:
...
+
+/obj/item/device/electropack/shockcollar/receive_signal(datum/signal/signal)
+	if(!signal || signal.encryption != code)
+		return
+
+	if(isliving(loc) && on)
+		if(shock_cooldown != 0)
+			return
+		shock_cooldown = 1
+		spawn(100)
```
deathride58, line 177: either use addtimer here, or make shock_cooldown use world.time and check it against world.time instead.

Kevak: Alright. Will note this is what the original electropack does. So will tweak for it too. Electropack code is ooooold.

PR #3094, opened 2017-10-03
[MIRROR] Changes the broken floor sprite

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/31234
--------------------
Makes it that black and purple checkerboard pattern from other games. Now we can have broken things be a reference in and of themselves. 

(Real reason: People can identify it's a broken easier as it blatantly stands out being not correct)


Poojawa:
"meh"

PR #3097, opened 2017-10-03
RED BULL GIVES YOU WINGS

deathride58:
image

:cl: deathride58
tweak: RED BULL GIVES YOU WINGS
/:cl:


Kevak:
I would really rather this not be a thing. Maybe make an adminspawnable variant, but keep flight potions as is. Stale corporate branding is stale.

PR #3143, opened 2017-10-05
Readding cult hardsuits

Kevak:
Readds cult hardsuits to cultforge options. Preeetty straightforward. I have tested it. Report bugs on this git.

:cl: Kevak/Neri
add: Readds Cult Hardsuits. or more specifically, makes cult forges be able to make hardsuits. They never got removed from the code.
tweak: Cultforge uses Option instead of Choice. (Aka ye get a menu instead of three buttons)
balance: Cultforge can make hardsuits again.
fix: Fixed Cultforge procs, Slightly cleaner code.
/:cl:

Why: If it ain't broke don't remove it. Removing a feature is frankly never a good thing. It's understandable upstream on TG when they have to balance around wanton powergaming. However in this case it's pretty silly even for them. I am fairly sure the main reason they axed culthardsuits upstream is because the Choice proc can only handle 3 options and they could not be assed to turn it into an Option. Laziness at its best.


Kevak:
I will be totally honest. I don't know where that first commit came from. It seems to do nothing and touches no files so.. I guess it's just there?

Kevak:
Hold on. Github is being stupid. Lemme actually put the files in.

Kevak:
Thar we go. As a side note. This makes that one tweak to the timerproc obsolete. Cause I just folded that in.

PR #3173, opened 2017-10-06
Fixed Drying Rack

WewLads:
Fixed the drying rack. The drying rack can now properly dry anything labeled as a consumable plant.

:cl: Fixed Drying Rack
fix: Fixed the Drying rack refusing plants.
/:cl:

[why]: We need to smoke damn it!

### DIFF
--- a/code/modules/food_and_drinks/kitchen_machinery/smartfridge.dm
+++ b/code/modules/food_and_drinks/kitchen_machinery/smartfridge.dm
@@ -278,8 +278,6 @@
 			update_icon()
 
 /obj/machinery/smartfridge/drying_rack/accept_check(obj/item/O)
-	if(istype(O, /obj/item/reagent_containers/food/snacks/grown))
-		return TRUE
 	if(istype(O, /obj/item/reagent_containers/food/snacks/))
 		var/obj/item/reagent_containers/food/snacks/S = O
 		if(S.dried_type)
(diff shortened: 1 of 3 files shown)


code/modules/food_and_drinks/kitchen_machinery/smartfridge.dm
```diff
@@ -278,6 +278,8 @@
 			update_icon()
 
 /obj/machinery/smartfridge/drying_rack/accept_check(obj/item/O)
+	if(istype(O, /obj/item/reagent_containers/food/snacks/grown))
+		return TRUE
```
deathride58: This will cause runtimes when putting things that normally don't go inside the drying rack in, since this completely bypasses the check for an item's dried_type.

WewLads: Well what sort of things? I've tried it with all sorts of plants and it works fine.
I mean sure you wouldn't normally smoke a cabbage but there's no harm in it either.

WewLads: I'm looking at the object tree and the only questionable thing that would be included is carpet in plant form.

silicons:
Why isn't this done upstream?

WewLads:
/tg/station drying rack works even with the same weird logic code, so I honestly have no clue what the real problem is. However, this is a solution that has the same effect that makes everything work fine without runtime errors.

WewLads:
@kevinz000 Alright we figured out the actual real problem. The drying rack spawns wrenched without connected power despite being in a room with power, but if you unwrench it and wrench it works again. This is an issue having to do with the power system which I have no experience in dealing with. I'm just going to close this and make an issue thread for someone else to deal with.

PR #3359, opened 2017-10-13
Fixes the Wish Granter

cebutris:
Because, to be quite honest, there is literally zero reason that this needs to exist in it's current state. It's only purpose is to let lavaland powergamers selfantag and go on a murderspree.

Is this an i ded? 
Yes, but I still believe it's a valid change.

There exists two wishgranters in the code, one that just makes you an avatar, found in the cube, and one that gives you a choice. I removed the creation of the avatar from the first one, though it still gives hulk/xray/tk/heatresistence so it's still pretty good. The second one dusts the user when they choose "To kill".

### DIFF
--- a/code/game/machinery/wishgranter.dm
+++ b/code/game/machinery/wishgranter.dm
@@ -29,7 +29,7 @@
 
 	else
 		to_chat(user, "You speak.  [pick("I want the station to disappear","Humanity is corrupt, mankind must be destroyed","I want to be rich", "I want to rule the world","I want immortality.")].  The Wish Granter answers.")
-		to_chat(user, "Your head pounds for a moment, before your vision clears.  You are the avatar of the Wish Granter, and your power is LIMITLESS!  And it's all yours.  You need to make sure no one can take it from you.  No one can know, first.")
+		to_chat(user, "Your head pounds for a moment, before your vision clears. ")
 
 		charges--
 		insisting = 0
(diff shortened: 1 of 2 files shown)


deathride58:
You probably could just give a *lot* of extra debuffs alongside the normal effects, instead of making the option kill you outright without any warning.

Alternatively, twisting the wish by spawning a horde of megafauna in the player's area instead of giving them antag status would work pretty well. The wish is only "to kill", after all, it doesn't specify *what* the user wants to kill.

cebutris:
I don't actually even know if the "wildwest" wishgranter (the one that lets you choose "to kill") actually spawns. I was thinking of making it twist the wish to force the user to commit suicide after a delay. Technically, they are killing, only it's themselves. 

Powergamers would just use the megafauna to get extra loot, which is still better than using the wishgranter to kill everyone else for the lulz

Kevak:
We are not Goon. Instadeath is not a valid reward for reaching something that takes effort. Put some effort into the PR.

cebutris:
The only one that gives you instadeath gives you four choices. Choosing one kills you, the others give rewards. 

Also, the Die of Fate can give instadeath, and that takes a whole lot of effort to get.

Kevak:
Die of Fate is thematically appropriate to do that. This is not. Die of fate is a roll of chance _After_ multiple uses of a chance based object that hurts you. Thematically it fits. 

Wish granter takes a bunch of effort to get in the first place and instadeath is not thematically appropriate.

silicons (changes requested):
if you're going to remove the option just remove the option this isn't fun for anyone ever.

Kevak:
Not to mention this is not a ‘fix.’ This is a reversion and removal of a feature. _Fix_ that title.

cebutris:
I just shouldn't touch code ever

Kevak:
Touching code is fine. It’s how people learn. The issue is the reasons and bringing round specific salt into what is intended to be a permanent feature. Ideds are rarely good code. Just look at that one time revolvers accidentally shot gibbullets.

PR #3462, opened 2017-10-17
Medihound Fixes

Poojawa:
:cl: Poojawa
fix: fixed medihound having to respect devourable settings.
fix: fixed other borg sleepers to not inject any regents.
add: Tgui menus for borg sleepers. Is pretty.
fix: Fixed analyzer nose to work as intended.
fix: infinite energy loop is gone.
tweak: Removed borg hypo from Medihound. The sleeper works now.
balance: Medihound sleeper will swap from anti-tox to pure toxin if the patient is a slime.
/:cl:

So while I have most of Tgui working, the only thing that actually functions is the eject button. @kevinz000  if you'd be a dear and proof over it, I honestly can't figure out what I'm missing between the pair.


Poojawa:
C H R I S T

code/citadel/dogborgstuff.dm
```diff
@@ -389,18 +414,42 @@
 		if(patient) return //If you try to eat two people at once, you can only eat one.
 
 		else //If you don't have someone in you, proceed.
+			if(!isslimeperson(target) && ("toxin" in injection_chems))
```
deathride58, line 417: isslimeperson() does not take into account jellypeople, a species that also takes inverted toxin damage. It'd be more efficient to check for the TOXINLOVER flag in the target species' traits instead.

Poojawa: Yes it does.  Or rather, no-one plays jelly people, so w/e

PR #3510, opened 2017-10-19
Removes powergaming cancer from sec

Roland410:
[Changelogs]:
🆑 Roland410
del: Removed powergaming cancer from sec aka rwalls in normal cells and chem dispensers.
/🆑

[why]: Security does not need either of them.


Roland410:
Well, I have no idea why it conflicts, yay..

deathride58:
This seems like more of a kneejerk reaction PR than anything

Also you forgot to use the mapmerger

EDIT: and my other PR updating security in response to feedback also includes removing the chem dispensers, so that change isn't really neessary

deathride58:
image
Map doesn't even load proper, meaning your changes cannot be confirmed.

Poojawa:
Why, exactly, does Sec not need R-walls?

also map merge

Roland410:
I swear I used map merger, but I guess I'll have to redo it. 
Also sec does not need rwalls at cells because this is not a prison station, electrified grilles are more than enough to contain a prisoner. If someone wants to break out a prisoner and they even manage to do so, it's security's fault for not keeping attention.
Chem dispensers are also way overkill for security and will only enable powergamers. The other map changes I like myself, but I thought these were stupid as hell.

deathride58:
R-walls in place of cell windows means that if someone breaks into the cells, the warden can't simply just close the cell blast doors and not have to deal with it further. R-walls can also be thermite'd a lot easier than blast doors and reinforced windows can, which buffs chemists when it comes to breaking prisoners out. R-walls in place of windows also means a prisoner can't just throw their shoe repeatedly at the window and underlying grille to escape. R-walls in place of windows also helps deter riots from occurring simply because the clown ended up with a 2 minute timer in the brig.
On top of that, the changes to Boxstation introduced alongside the cell window > rwall replacement actually make it much easier to break prisoners out, since all that's really stopping someone with a welding tool and wrench are reinforced tables and windoors. This means the security lobby is actually a pretty important choke-point during riots, one that can no longer be completely controlled by simply slamming the cell shutters down and tossing a couple of barrier grenades. The vulnerability is being introduced to Meta and Delta alongside other tweaks (like removing the chem dispenser) in the follow-up PR.

Taking a change like this at pure face value without analyzing the full effects of the change is only bound to deter further changes and tweaks from taking place, as contributors will get the impression that players will always get up in arms over balance changes.

tl;dr: Cell windows > Rwalls makes sec easier to break into, even moreso with the changes that were brought along with it.

Roland410:
Sec was never meant to be a stronghold or anything like that, and rwalls are a LOT harder to disassemble without being caught in the process that say windows or even normal walls. Sure, you CAN thermite through, but that means you have to get thermite first somehow, and unless you're a chemist, you can get access, or chemistry is broken into so you can just walk in to make some, it's not going to happen.

deathride58:
With the changes on Boxstation, you can just break a windoor and hop a table to avoid having to touch the R-walls at all when breaking prisoners out of the cells, no thermite required. I went over this in my wall of text. The follow-up PR I made replicates this vulnerability across Delta and Meta.
With cell windows in place of R-walls, wardens can simply slam the cell shutters down to make cells near-impossible to break into or out of, since you simply cannot break blast doors without being an antag.

Just because it looks secure doesn't mean it is secure.

Poojawa:
Sec is literally supposed to be a stronghold. It's where the armory is located. It's the final line of defense against rev and cult.

Shame sec gear sucks against groups.

Roland410:
I never said I readded the shutters, deathride, just replaced the walls like they used to be.
About the breaking out, I meant that alone you have no chance to break out with rwalls, but you do with electrified grilles.

cebutris:
You're not supposed to be able to get out of the brig alone unless you prepared ahead of time with specific gear

silicons:
Please fix your code/maps and then remake the PR with a non joke/meme title and description and include why you think this would improve the game and server.

Poojawa:
and for the record, I should re-add murderous tesla to grilles in security, then you can actually argue they keep people contained. lol

PR #3524, opened 2017-10-20
[MIRROR] Allows species to be blocked from ever being roundstart, can now choose your Halloween species

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/31877
--------------------
:cl: Kor
add: You can now select your Halloween race, rather than having it assigned randomly via event.
/:cl:

Why: So we can stop having terrible slippery slope arguments about roundstart races that are meant for antagonist/event/holiday/admin use (and so we can get epic "blocks lizards from roundstart" PRs from goof for the next five years)

As an example/fun use of this, the holiday event no longer randomly turns crew into zombies/skeletons,  but those races will be selectable during Halloween.

To make the above work properly with inheritance, nightmares and infectious zombies are blocked from roundstart (why would you ever want those anyway?)

I'm not sure this is the right solution but I will do very silly things in an attempt to compromise


Poojawa:
This will make it easier to select donation reward species to specific people

PR #3529, opened 2017-10-21
Nerfs Autolathed Foam Guns

Michiyamenotehifunana:
:cl: Toriate
balance: OP foam stuff now cost slightly more to build from autolathes
balance: OP foam stuff now turned into total memes instead of practical weapons, all burst firing foam guns have at least 60 spread. Stealth pistol also has 60 spread.
balance: Foam X9 and Mag Rifle now requires two hands to fire
fix: Foam Force X9 and Mag Rifle now has its correct weight class of NOT FITTING IN FUCKING BACKPACKS
/:cl:

This will turn very dangerous and OP weapons (when loaded with riot darts) into total memes, just like they should have always been. Now you can't hide a fucking assault rifle in your backpack and unload 5 riot darts into some bloke to instantly drain all their stamina!

### DIFF
--- a/code/citadel/cit_guns.dm
+++ b/code/citadel/cit_guns.dm
@@ -190,13 +190,15 @@
 	needs_permit = 0
 	mag_type = /obj/item/ammo_box/magazine/toy/x9
 	casing_ejector = 0
-	spread = 45		//MAXIMUM XCOM MEMES (actually that'd be 90 spread)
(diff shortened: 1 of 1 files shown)


deathride58:
ＩＴＳ　ＮＥＲＦ　ＯＲ　ＮＯＴＨＩＮ’

PR #3556, opened 2017-10-22
Hyposprays mk ii

Poojawa:
:cl: Poojawa
add: Added hypospray mk iis! Swap chemicals and applications for better triage!
/:cl:

Mostly because they're neat. Initial idea/code pulled from Virgo. Has been something I've always wanted to add in since I played CMO ages ago. Spites by Cameron~ they're pretty snazzy honestly.

2a94948aa93acb40031b7e431620ea17

Hyposprays IIs are able to be 'hot swapped' with different chemicals with different Vials. normal Vials are only 30u capacity, but the CMO's fancy set up (which will be a preloaded box in their locker) has 120u capacity, and the ability to dump all of it into someone in one go. Thus, making sure it's still a very prized possession for the CMO to forget about and never use as a Theft objective. I'll probably swap it to 60u once I consider how OP dumping 120u of morphine into someone is. oops.

the new and improved hypos are also able to swap between 'Inject' and 'patch' methods of application, so hypos can apply reagents that are only able to be used via patch as well. Hypos are not instant, however, and will require time to apply the medicine - they are better than syringes only due to their ability to swap being inject and spray modes. 

Hypos will be refillable via syringe, and will be able to be produced outright from the Chemmaster. The large hypo vials are of limited quantity and will not be available via cargo or science.


Poojawa:
Current issues for my record keeping are:
Ensure the damn hypos accept their modified capacity and that only the vials are certain to work.
Ensure that one can actually swap carts. Likely just needs to be a right click action verb instead of trying to hijack the attackhand code. Fuckin' Phil.
Update Chem masters to produce hypo vials. 
Ensure you cannot drink or spill the fucking hypo vials.

wew.

code/citadel/hypovial.dm
```diff
@@ -0,0 +1,100 @@
...
+
+/obj/item/reagent_containers/glass/bottle/vial/preloaded/tricordrazine
+	name = "vial (tricordrazine)"
+	icon_state = "hypovial"
+	comes_with = list("tricordrazine" = 30)
+
+/obj/item/reagent_containers/glass/bottle/vial/large/preloaded/CMO
+	name = "large vial (CMO Special)"
+	icon_state = "hypoviallarge-cmos"
+	comes_with = list("epinephrine" = 30, "kelotane" = 30, "antitoxin" = 30, "bicaridine" = 30)
```
Poojawa: Ya I know this is dumb, but my brain defaulted to putting shiny together.

PR #3568, opened 2017-10-23
removes riot foam darts from autolathe

cebutris:
With the new accessibility of dart guns introduced by Camile's memeguns, this has become a bit of a problem, especially with public autholathes existing on most maps. This PR removes them from the autolathe. I was thinking of adding them to a contraband crate in cargo (I.E the stuff you get from multitooling the board, not emagging the console), but I'll wait for more feedback about this PR

🆑 Cebutris
del: riot foam darts removed from autolathe
/ 🆑

### DIFF
--- a/code/modules/research/designs/autolathe_designs.dm
+++ b/code/modules/research/designs/autolathe_designs.dm
@@ -655,7 +655,7 @@
 	materials = list(MAT_METAL = 4000)
 	build_path = /obj/item/ammo_casing/shotgun/incendiary
 	category = list("hacked", "Security")
-
+/*
 /datum/design/riot_dart
 	name = "Foam riot dart"
 	id = "riot_dart"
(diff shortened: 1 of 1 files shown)


pantherpusher:
This is probably for the best, people died over these darts

Michiyamenotehifunana:
A better way to ensure the memes remain memes instead of abusable weapons would probably be to prevent the autolathed foam weapons from loading riot darts, though that would require either more code for the guns, or replacement of the foam guns with energy equivalents.

I'm not entirely sure how to prevent riot darts from being loaded into foam guns since they use the same calliber and all the code I could find that limits what guns can fire depend entirely on the callibers being different.

cebutris:
@Michiyamenotehifunana 's solution is so much more preferable but requires more coding knowledge than I possess.

Michiyamenotehifunana:
I coded the stuff in myself to begin with and I lack the knowledge to actually limit their ammo types without messing with callibers. Swapping them out with energy equivalents would really be the only solution if the foam gun abuse keeps up  or an actual coder swoops in to save us.

deathride58:
I'll see what I can do to prevent riot darts from entering autolathe toys when I finish up map syncs

Poojawa (approved):

ZeroNetAlpha:
This was dead and stale for months. There's also another PR that does the same exact thing already. Why was this re-opened?

Poojawa:
Reee.

PR #3593, opened 2017-10-24
Revert "Revert "[MIRROR] Late join antagonists will try to target late join players""

LetterJay:
Reverts Citadel-Station-13/Citadel-Station-13#3584

Also includes the fix from @deathride58's PR upstream


code/modules/mob/dead/new_player/new_player.dm
```diff
@@ -488,6 +490,7 @@
 	. = H
 	new_character = .
 	if(transfer_after)
+		mind.late_joiner = TRUE
```
deathride58: Make sure ya get rid of this line right here, it's what breaks everything

PR #3621, opened 2017-10-26
[MIRROR] [512] The great \ref purge

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/31824
--------------------
This change is a result of the default \ref behaviour changing in 512. \ref no longer uses the object's tag var if available. REF() replicates that behavior and this also adds a new datum var for determining whether something should use its tag which defaults to FALSE.

Replaced:
`\\ref\[([^\[]+?)\]`
With:
`[REF($1)]`

Uses of \ref like `text("...\ref[]...", src)` Were manually fixed as there weren't many.


deathride58:
PR was merged to upstream's 512-experimental branch.

ninjanomnom:
You can feel free to just close this, this is for upcoming 512 testing 
image

PR #2211, opened 2017-07-31
Adds a HUD for borers

ktccd:
[Changelogs]: # (Please make a changelog if you're adding, removing or changing content that'll affect players. This includes, but is not limited to, new features, sprites, sounds; balance changes; map edits and important fixes)

:cl: ktccd
add: Borers of all kinds now have a HUD element to see other sentient borers and their hosts. (Hosts cannot see this HUD)
/:cl:

[why]: # (Please add a short description [on the next line] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding:) 
It's slightly annoying as a borer to not know who is infested with a borer, especially seeing as the borers are supposed to have a hive-mind of a sorts.
This avoids the annoyance of trying to infect someone's host, and also lets borers quickly see if a borer has awoken or not.
Hosts cannot see this HUD, because they're not a telepathic hive-mind. They can always ask their borer to point out people for them if they want to.

PR #3635, opened 2017-10-26
Undoes Borer HUDs

ktccd:
:cl:ktccd
del: Removed borer-HUDs
/:cl:

Critical flaw in antag-HUD system prevented proper use, since automatic reverting of original PR is no longer possible, (Probably due to another antag-hud being added afterwards), this is a manual one.
Basically, antag-HUDs can only ever have ONE icon, and the latest one applied undoes any previous ones.
This is kind of a flaw in how antag-HUDs work in general, so this bug exists in many places, but seeing as I'm maintaining borers, I'm removing it for them, it was only a minor quality-of-life thing anyway, not very critical. Refactoring antag-HUDs to work with more than 1 antag is waaaaay outside what I can reasonably do...

### DIFF
--- a/code/__DEFINES/atom_hud.dm
+++ b/code/__DEFINES/atom_hud.dm
@@ -40,8 +40,7 @@
 #define ANTAG_HUD_SINTOUCHED	16
 #define ANTAG_HUD_SOULLESS		17
 #define ANTAG_HUD_CLOCKWORK		18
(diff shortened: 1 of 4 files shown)


ktccd:
This basically undoes https://github.com/Citadel-Station-13/Citadel-Station-13/pull/2211 except since it can't be reverted, I'm doing it this way instead.

PR #3639, opened 2017-10-27
[READY] Redtext bug fix + Endround objective check cleanup

deathride58:
I HAVE NO IDEA WHAT THE FUCK IM DOING BUT HERES TO HOPING THIS WORKS

:cl: deathride58
fix: Antags are able to greentext again
/:cl:

this basically removes a lot of the copypasta in the crew objective check in the gamemodes by moving the (Optional) text to a ternary and moving the crew objective check right over the bit that makes antags fail.


deathride58:
Got around to testing it locally, seems like it resolves the bug.

PR #3687, opened 2017-10-30
CRITICAL BALANCE FIXES TO CITADEL'S GUNS

Michiyamenotehifunana:
:cl: Toriate
del: No more ballistics for the captain
balance: Magweapons are no longer superior to everything
balance: instastuns removed from all cit exclusive projectiles
fix: 10mm Soporifics now work
balance: 10mm soporifics require 2 shots to take effect, down from 3. Also has sleep duration increased up to par with sniper soporifics
add: HoS loadout now starts with the Model D Energy Gun, replacing the default egun
/:cl:

The entire playerbase has been ignoring massive balance issues up until an hour ago. This should solve most of it.

### DIFF
--- a/code/citadel/cit_guns.dm
+++ b/code/citadel/cit_guns.dm
@@ -719,10 +719,10 @@ obj/item/projectile/bullet/c10mm/soporific
 	if((blocked != 100) && isliving(target))
 		var/mob/living/L = target
 		L.blur_eyes(6)
-		if(L.getStaminaLoss() >= 50)
-			L.Sleeping(400)
+		if(L.getStaminaLoss() >= 60)
+			L.Sleeping(300)
 		else
-			L.adjustStaminaLoss(60)
+			L.adjustStaminaLoss(25)
 	return 1
 
 /obj/item/ammo_casing/c10mm/soporific
--- a/code/citadel/cit_guns.dm
+++ b/code/citadel/cit_guns.dm
@@ -111,7 +111,7 @@
 	w_class = WEIGHT_CLASS_BULKY
 	can_suppress = 0
 	burst_size = 1
-	fire_delay = 20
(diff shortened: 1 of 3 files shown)


Poojawa:
150 hours of testing, what could have possibly gone wrong.

code/citadel/cit_guns.dm
```diff
@@ -719,10 +719,10 @@ obj/item/projectile/bullet/c10mm/soporific
 	if((blocked != 100) && isliving(target))
 		var/mob/living/L = target
 		L.blur_eyes(6)
-		if(L.staminaloss >= 60)
-			L.Sleeping(250)
+		if(L.getStaminaLoss() >= 50)
```
silicons: ........................

Michiyamenotehifunana: Oh god I fucked it up

Michiyamenotehifunana: ...wait, did I? It didn't seem to work at all with just `staminaloss`, at least that's what I was told when a bunch of people on the live server tried it.

Poojawa: Why doesn't it just inject soporific to the mob?

Michiyamenotehifunana: Because it uses the same code as sniper rifle soporifics rather than being a syringe thingy.

Michiyamenotehifunana: ...unless this is about the fact that it was nerfed to needing three shots before it was buffed again to two shots here because it's so useless and antags would rather kill their victim instead of trying to put them to sleep.

Michiyamenotehifunana:
@kevinz000 There, soporifics are back to the old 3 shot to sleep. _Just pretend the 2 shot numbers were a mistake._

PR #3714, opened 2017-10-31
Further rebalancing of Cit_guns projectiles

Michiyamenotehifunana:
:cl: Toriate
balance: Mag weapon projectiles now have slower projectile speeds and reduced range. Non-lethal shots brought up to 3 shots to knockdown, lethal shots now weaker than ever.
balance: Non-lethal 9mms now have drastically lower range and slower projectile speeds
/:cl:

The changes to the magweapon projectiles should make them less broken, require more skill to land hits, and not excessively weak.

The 9mms are still unimplemented after their hasty removal, and probably won't be coming back anytime soon despite me touching them in this PR. The reduced range of rubber shots would bring them closer to tasers, except slower still, and with no instastun. The frangible bullets now make more sense for reducing collateral damage due to their 20 tile range.

### DIFF
--- a/code/citadel/cit_guns.dm
+++ b/code/citadel/cit_guns.dm
@@ -213,6 +213,8 @@
 	damage = 15
 	armour_penetration = 10
 	light_range = 2
+	speed = 1.2
+	range = 25
 	light_color = LIGHT_COLOR_RED
 
 /obj/item/projectile/bullet/nlmags //non-lethal boolets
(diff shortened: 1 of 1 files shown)


Michiyamenotehifunana:
I think I should just remove frangible 9mm bullets entirely. They were originally created because I didn't plan on making the Mk.37 use 9mm and wanted to limit their lethal capabilities a lot.

PR #3877, opened 2017-11-10
[MIRROR] Adds icons for items with randomization

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/32523
--------------------
Adds icons for items that have randomized features (ie change colors, subtypes, ect) when they spawn so they are more easily distinguished from normal versions in DM.

These are not visible ingame.

Also fixes invisible engineering scanners and headphones in DM.

image to confirm nothing was messed up ingame.
image

DM example;
image


deathride58 (changes requested):
Missing uniforms.dmi and items_and_weapons.dmi from the original PR

PR #3882, opened 2017-11-10
[MIRROR] SSoverlay improvements and tweaks

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/32371
--------------------
Changed up how overlays was done to account for the fact its a queue and not a processor. (it was using almost none of the processing subsystem framework)

Made the overlay loop faster by making it not cut the list until the end.

Made build_appearance_list process queued overlays in atoms before converting to appearances to fix some edge case issues with atoms being used as overlays

Added a simple generic benchmark stat tracking system.

Overlays will now log how long each type took to process.

I don't know how much overhead tracking this adds to overlays. I may put it behind testing or something, but I do want to test this on the serbers to get some stats.


deathride58:
Relies on the \ref purge, which has already been merged.

PR #3936, opened 2017-11-14
[MIRROR] adds 1 gbp to armhulen's account (grammar fix)

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/32698
--------------------
in every other case it uses our instead of your, so it's weird that it doesn't do that here. I kept it for the latter part of the removal thing, because it makes sense. You aren't a ling anymore.


deathride58 (changes requested):
PR is supposed to modify line 1039. It's supposed to change "Your powers are awoken" to "Our powers have awoken"
This change is not present in the PR.

PR #3951, opened 2017-11-15
No fuzz insects, renamed instead of added.

SpiderPsycho:
[Changelogs]: # (Please make a changelog if you're adding, removing or changing content that'll affect players. This includes, but is not limited to, new features, sprites, sounds; balance changes; map edits and important fixes. An example changelog has been provided below for you to edit or remove. If you need help, read https://github.com/tgstation/tgstation/wiki/Changelogs)

:cl:
add: Added insects without fuzz
del: Removed fuzz
tweak: changed the name of the fuzzless sprites to insects so the character creation can call them properly.
/:cl:


LetterJay:
It'd probably be better to add a variant instead of outright removing it.

SpiderPsycho:
Pooj and Not-A-Min told me it'd probably be better if I do it this way. Anyways, the moth fuzz can be made into a beard if you want it. I'll get started on that if need be.

deathride58:
Moth fuzz is already covered by the moth marking.

PR #3984, opened 2017-11-17
[MIRROR] JSON feedback

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/32188
--------------------
Replaces the string and int based feedback system with one that records data entirely in JSONs (LONGTEXT in Maria). For how this works see: https://github.com/Jordie0608/tgstation/blob/6ef229ddc4aa227b2f911def66b716bfa572fca2/code/controllers/subsystem/blackbox.dm#L129-L182
The blackbox subsystem no longer stores every radio message sent just to count how many there were and is updated to identify the current radio channels. The subsystem now stores the highest tech levels attained since everywhere else this is stored can be lost during the round, deleting the data.

Some feedback keys have been changed or removed, a spreadsheet detailing all of these changes and notes pertaining to them can be found at: https://file.house/q7nI.ods

DM side functionality tested to work. A couple of stat related fixes are scattered in, most being incorrect arguments being used.

Conversion of old data is done by a python script. As my first time in python it's not very pretty but I've tested that it handles all our feedback data as of 2017-10-23 01:34:06. Details on how it operates are in the script itself

@ nfreader 
@ MrStonedOne

closes #4011 
closes #4068


deathride58:
@LetterJay You'll probably wanna run feedback_conversion_2017-11-12.py on the server when ya get around to merging this.

PR #4054, opened 2017-11-21
[MIRROR] Removes material exports

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/32926
--------------------
:cl: Kor
del: Following complaints that cargo has been selling all the materials they receive rather than distributing them, mineral exporting has been removed.
/:cl:

I'm not 100% sold on this but it is probably the most common request for techwebs so far, to address the fact that cargo just tells the rest of the station to get fucked and sells all the materials instead

Long term I think something like what Robustin suggested with departments able to generate credits by doing their work, and trading those credits to cargo for materials might work, but for now it is just too tempting to sell all the stuff rather than share.


deathride58 (approved):

silicons:
dnm until techwebs

PR #4082, opened 2017-11-22
Donator item modularization + manual sync of sprites with upstream + manual mirrors + donor item fixes

### LINKED ISSUE #4288: Vermillion clothing icon turning into SUIT when worn
[Directions]:  Round ID 10683, I was hacking into atmospherics that round and came upon a vermillion clothing, This goes onto your exosuit slot, I decided to put it on because why not, as i did, its icon changed to SUIT..Thats really it. https://gyazo.com/08d72a3c3edf1108383092bafbf7d7d4

deathride58:
This is going to need review to make sure the sprite syncs didn't remove any cit-specific sprites.

Closes #3877 (Manually mirroring it)
Fixes #4288 
Manually mirrors https://github.com/tgstation/tgstation/pull/32678


LetterJay:
Most sprites are either in custom.dmi or in their own folder, so they shouldn't be overwritten.

deathride58:
Currently, most of the donator sprites are spread throughout the base files, meaning they will always get overwritten when the files are synced with upstream manually or via mirror.
This PR fixes that by moving most of the donator sprites to custom.dmi, and properly makes use of icon_override to ensure the items render properly when worn on characters.

PR #4097, opened 2017-11-24
[READY] Makes walking on tiles slowly make them dirty, similar to Baystation, and adds footstep sounds (Featuring sounds from CEV Eris!)

deathride58:
Title. They're neat, subtle features.

:cl: deathride58
add: Walking on tiles will now slowly make them dirty.
add: Objects moving around will also slowly make tiles dirty.
add: Walking around will now emit footstep sounds. These sounds are from CEV Eris.
/:cl:

The footstep sounds are from CEV Eris.


cebutris:
Can we add Paradise's jackboot sounds that make it sound like you're walking through wet dirt or mud and it's so loud you can hear it from across the station (please dont actually do this)

PR #4171, opened 2017-11-30
[MIRROR] Fix turbine computer showing broken

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/33144
--------------------
:cl: psykzz
fix: Fixing the broken turbine computer
/:cl:

I previously broke the turbine but no one noticed, fixing it before someone complains.


deathride58 (changes requested):
Needs TGUI recompile

deathride58 (approved):

PR #4195, opened 2017-12-03
Changes some export prices around lavaland items and adds some few new ones

jughu:
This balances some prices to not be super high. yet be rewarding for cooperation with cargo

:cl: Improvedname
add: Adds the crusher trophies to be exported for high prices
balance: Balances some export prices to not be insanely high
/:cl:

### DIFF
--- a/code/modules/cargo/exports/weapons.dm
+++ b/code/modules/cargo/exports/weapons.dm
@@ -73,55 +73,79 @@
 	
 // relics of lavaland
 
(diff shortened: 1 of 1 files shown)


deathride58 (approved):
Could probably modularize everything here so it won't get overwritten by upstream changes, but it looks pretty good

jughu:
Imma try to push this to upstrram first then since this included my final changes and if that doesnt go ill moderize

PR #4198, opened 2017-12-03
[MIRROR] Adds emergency lights, built into light fixtures

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/33213
--------------------
:cl: Xhuis
add: Light fixtures now turn an ominous, dim red color when they lose power, and draw from an internal power cell to maintain it until either the cell dies (usually after 10 minutes) or power is restored.
add: You can override emergency light functionality from an APC. You can also click on individual lights as a cyborg or AI to override them individually. Traitor AIs also have a new ability that disables emergency lights across the entire station.
/:cl:

Light fixtures now spawn with a small 120 W power cell inside of them. If the area experiences a power outage or lighting power is otherwise cut, the lights will automatically fall back on these cells to go into a very low-power "emergency mode." In this state, the lights will emit a very dim red glow that becomes darker as the cell runs out of power, and the lights can be removed without any gloves.

You can insert any power cell into a light fixture, but if it has a charge above 300 W, the fixture will burn out. This is because light fixtures are designed to handle a tiny trickle of power (one second uses 0.2 W in emergency mode) and not a 10 kW monster of a power cell designed for use in industrial machinery. 

In emergency mode, lights consume 0.2 W per tick from their internal cell, meaning they should last for about ten minutes on paper. When *not* in emergency mode, they recharge their internal cells at the same rate, though I may remove that function. You can produce the tiny, miniature cells at an autolathe for a very small amount of glass if you want to include emergency lights in construction projects.

You can override an area's emergency light fallbacks from an APC, preventing them from turning on. As an AI or cyborg, you can also click on lights themselves to override them individually. Traitor AIs can also now purchase a one-use Disable Emergency Lights ability for 10 processing power that disables them all across the station, though they can be reactivated through APCs individually.

I felt like this was a more interesting alternative to the lights just shutting off if power is cut. The glow is dark enough that the station is by no means illuminated, but it's much more atmospheric - rather than walking along a black screen, you might see the shadowy figures of crew or other things moving through the glows, or a blood trail cast into dark relief by the waning light.

=========
Closes #4208 (Merged into this PR.)


deathride58 (changes requested):
Needs TGUI recompile, missing lighting.dmi from original PR

deathride58 (approved):

PR #4212, opened 2017-12-05
[MIRROR] In-code key handling system, Take 2

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/32751
--------------------
:cl: MrPerson & ninjanomnom
refactor: Completely changed how keyboard input is read.
add: Holding two directions at the same time will now move you diagonally. This works with the arrow keys, wasd, and the numpad.
tweak: Moving diagonally takes twice as long as moving a cardinal direction.
fix: You can use control to turn using wasd and the numpad instead of just the arrow keys.
del: Some old non-hotkey mode behaviors, especially in relation to chatbox interaction, can't be kept with the new system. Of key note: You can't type while walking. This is due to limitations of byond and the work necessary to overcome it is better done as another overhaul allowing custom controls.
/:cl:

An updated version of #18863 by @ MrPerson 

Ready for testmerging and I'll be working out any other issues we have with it.

Video of it working https://puu.sh/ylF1Q/1285840bab.mp4


deathride58:
Do not merge until upstream fixes this. Apparently it breaks default/non-hotkey mode.

PR #4232, opened 2017-12-05
Removes borer event from rotation

cebutris:
Following a discussion in admin chat I have made this. Doesn't effect Syndicate Borers, and admins can still trigger the event, but it won't trigger on it's own

### DIFF
--- a/code/game/gamemodes/miniantags/borer/borer_event.dm
+++ b/code/game/gamemodes/miniantags/borer/borer_event.dm
@@ -1,8 +1,8 @@
 /datum/round_event_control/borer
 	name = "Borer"
 	typepath = /datum/round_event/borer
-	weight = 0 //Default weight
-	max_occurrences = 1
+	weight = 0
+	max_occurrences = 0
 	min_players = 20 //10 is MINIMUM needed, but this is not a gamemode that does well in lowpop
 	earliest_start = 24000 //40 min, double default timer
 
--- a/code/game/gamemodes/miniantags/borer/borer_event.dm
+++ b/code/game/gamemodes/miniantags/borer/borer_event.dm
@@ -1,7 +1,7 @@
 /datum/round_event_control/borer
 	name = "Borer"
 	typepath = /datum/round_event/borer
-	weight = 10 //Default weight
+	weight = 0 //Default weight
 	max_occurrences = 1
 	min_players = 20 //10 is MINIMUM needed, but this is not a gamemode that does well in lowpop
 	earliest_start = 24000 //40 min, double default timer


code/game/gamemodes/miniantags/borer/borer_event.dm
```diff
@@ -1,7 +1,7 @@
 /datum/round_event_control/borer
 	name = "Borer"
 	typepath = /datum/round_event/borer
-	weight = 10 //Default weight
+	weight = 0 //Default weight
 	max_occurrences = 1
```
deathride58: Could probably change this to 0, as well, so the event system doesn't fuck up during a long round and spawn borers due to all other events hitting their max occurrences

deathride58 (approved):

Roland410:
Why though? Borers are small creatures that get valided by 90% of the people and the remaining 10% usually loves them. It enables players to get back into the game without it being majorly distruptive to the round.
I highly disapprove of this change.

deathride58:
They were removed completely upstream for a reason. This PR does not remove them completely like upstream did.

This PR is a thing due to how overpowered borers can be in the right hands. Having a free meth generator with almost no downside or counter is a preeeeeeeeeeeeeeeeeetty big thing. On top of that, a borer outside of a body is capable of instantly stunning targets for quite a long time.
This PR is also a thing due to how many folks have to be bwoinked when borers get triggered. People do not seem to understand what "You need to reproduce, protect your host from harm" means, which results in borers showing a gross disregard for their hosts' health. If this PR gets merged, it'd save a *lot* of work for staff members, and make things a lot less stressful for players.

Roland410:
Borers can be countered with SUGAR, literally a base chem which you can mass produce. Reproducing takes quite a while, and you can't use SSDs at all since ages ago. I'm not a staff member here (or really anywhere currently) but when I was, borers were NEVER a problem, at most a mild annoyance when they misbehaved.
Borers are as I said quite balanced themselves, and they are a RANDOM event which happens not too commonly.

deathride58:
Borers are still obtainable via syndicate uplink even with this PR.

Borers being overpowered or not, having a random event that results in a lot of major server rules (nonconsensual erp, dont actively go against your objectives, etc) being broken by salty ghosts, to the point where it almost always results in people getting notes, is pretty bad for server health. Staff members shouldn't have to micromanage players every round the event subsystem decides to roll the borer event.

And you can indeed use SSDs to reproduce as a borer, but it almost always results in problems for everyone involved.

Borer stuns can only be countered via meth or equivalent antistun. Sugar does not stop borers from entering a body or stunning it. 50u sugar metabolizes out of the body faster than it takes for the borer to stun and enter the body. Even then, sugar only disables the borer for as long as sugar is in the body, it does not instantly kill borers. On top of sugar not being lethal to borers, there's no way to tell if someone even has a borer, which makes powergaming with meth insanely easy for people lucky enough to get a robust borer in them. Since borers can inject meth with cooldowns and the host's OD threshold being the only two restricting factors, that essentially means robust borers are free, uncounterable, invisible antistun generators. If you're familiar with TG's stunbased combat at all, you'd know exactly why this is a major problem. This PR restricts borers to be a traitor item, meaning they won't be a problem during extended, nukeops, changeling, wizard, etc, when this gets merged. Also bear in mind that swarmers were disabled here due to their balance issues.

Borers are also capable of taking control of a host with no downsides or limits, for a bare minimum of 40 seconds if the host isn't SSD. 40 seconds is more than enough for a troublesome player to put someone in crit as a borer with no downsides, which almost always results in staff members having to repeat the entire "You have to protect your host as a borer, you have to reproduce" spiel to the borer in question. 40 seconds is also more than enough for a troublesome player to push someone into an ERP scene without their consent. If these things were insanely rare, and only happened once every two hundred rounds or so, it wouldn't be much of a problem. But these are things that happen every single time borers are rolled as a random event. Staff members should not have to warn and ban people for the same exact things over and over and over and over and over.

It'd be much more efficient to get rid of the problem at it's core instead of choosing to deal only with it's symptoms. Guess what this PR does.

Roland410:
While you have perfectly solid points, I never saw myself borers going WAY out to grief people, might have been at times when I'm not online, or maybe I have missed it, but when I told people in OOC that this is going to go through, noone said that "it was finally time", and most of them were surprised and didn't want them to go. Sure, you could always get a borer from TC, but it's rarely bought for its insane cost where you have to gamble if you get a robust player or a noob who just clicked on the YES button.
I don't wish to dispute that admins have a bit more work when borers misbehave as I said already, but I'd like to ask the admins/devs to reconsider this change.

PR #4281, opened 2017-12-09
[READY] Adds night shifts

deathride58:
:cl: deathride58
add: When it's night time in the server's timezone, all lights aboard the station will dim to simulate a graveyard shift. It's possible for lights to dim in the middle of a shift.
/:cl:

Inspired by Aurora's implementation, as seen here: https://github.com/Aurorastation/Aurora.3/pull/1049

Normal/daytime lighting
image

Night shift lighting
image

## TODO:

- [x] Make red alert disable the automatic light dimming
- [x] Add an automatic announcement when the lights switch between night and day mode mid-round. Something something circadian rhythms.


deathride58:
image
Compile failed due to travis breaking.

It's more or less ready for merge whenever.

MrsTonedOne:
So, one thing on my list to look at doing, that might help you here, is to add a bloom/oversaturation plane by just adding a second lighting planemaster (replicating that one lighting bug*) with a 0 alpha.

You can then add the ability to have eyes "get used to" low lighting by giving the bloom planemaster a slight alpha (that could grow over time), this would brighten up darker tiles akin to the first level of ghost darkness toggle but make brighter tiles seem overexposed or too bright.

*I haven't tested it out so i don't know if replicating that lighting bug will be that easy, but i know it involvings having a doubled up set of planemasters.

PR #4346, opened 2017-12-13
Restores the PDA sprites by cecily and toriate (And modularizes it)

deathride58:
Regression caused by a hard sync of sprites with upstream's

:cl: deathride58
fix: The standard-issue PDAs are now holographic again.
/:cl:


Michiyamenotehifunana:
There are two `pda-clear`s in the .dmi. One is the old one, and the other is the new one. Would this cause any issues?

deathride58:
The extra pda-clear state won't really cause any issues ingame, as it didn't really cause any issues before the regression. I'm not exactly sure how byond handles multiple icon states with the same name in a file, but I do know that one iconstate has priority over the other.

PR #4434, opened 2017-12-18
TG sync

silicons:


deathride58 (changes requested):
You removed pretty much every additional file from the .dme in 3ba42d0
Might be best to start again from scratch, and remember what I nagged you about in the previous sync PR. Remember to make sure admin logging for pumps also remains in the code.
Also, map files again. Boxstation, Metastation, Deltastation, Omegastation, and Pubbystation are all already synced with upstream, you can revert the changes to them.

PR #4466, opened 2017-12-20
Removes pierce immunity from roundstart xenos

Michiyamenotehifunana:
:cl: Toriate
balance: Roundstart xenos are no longer pierce immune, also no more extra damage from heat.
/:cl:

Because giving outright immunities to roundstart races is _totally_ balanced. Also removes the heat damage multiplier.


Kersakof:
Thank you.

PR #4469, opened 2017-12-20
[MIRROR] Tidy telecomms radio code, make PDA server real telecomms machinery

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/33647
--------------------
:cl:
refactor: Telecomms and radio code has been cleaned up.
tweak: The PDA message server is now a real piece of telecomms machinery.
tweak: PDA logs now include the job as well as name of the participants.
fix: The Syndicate frequency no longer hears its own messages twice.
fix: Talking directly into a radio which is also hot-miked no longer double-talks.
/:cl:

Follow-up to #33381 and last big change in this series. Radio broadcasting code is redone. The PDA server is made a real piece of telecomms machinery (PDA messages previously already faked subspace transmissions to check if telecomms was up, now it's real). The PDA message monitoring computer has also received a couple of tweaks (like fixing the broken `<A>` tags).

This has been tested in several real rounds privately with no noticeable ill effects, but it definitely should be testmerged before being merged.

I considered splitting this into a PR for the radio code and a PR for the PDA server, but it would involve hacking old PDAs to work on the new broadcasting and then immediately deleting that code, so I decided against it for now.

Fixes #32148.


LetterJay:
This relies on this https://github.com/tgstation/tgstation/pull/33236 which was never mirrored.

PR #4471, opened 2017-12-21
[dnm] new tgsync branch

silicons:
starting from scratch and seeing if this works better. ugh.


deathride58:
I'll give you headpats if this is the PR that finally gets us up to sync with upstream

PR #4493, opened 2017-12-22
[MIRROR] Job landmarks now show the mob with the proper job outfit. 

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/33714
--------------------
Also adds a proc to generate these icons automatically.

The proc(which is acessible by enabling debug verbs -> mapping tab) is quite costly so it really shouldn't be used anywhere besides updating the icon when a job is changed. The generation also includes cyborg ,AI and X that are used on landmarks. 

Closes #33631


deathride58:
Travis was broken when this got mirrored

PR #4494, opened 2017-12-22
[MIRROR] [READY]Refactors projectile trajectories, fixes piercing projectile inaccuracy and beam rifle visual deflections

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/32433
--------------------
Fixes #32343
Fixes #29315
Projectiles now use a new /datum/point thingy, which measures in pixels from the edge of the map.
projectiles themselves use /datum/point/vector, designed for pixel-accurate angular trajectories.
Changes some loc = to forceMove()s.
:cl:
fix: Projectiles that pierce objects will no longer be thrown off course when doing so!
code: Projectiles now use /datum/point/vectors for trajectory calculations.
fix: Beam rifle reflections will no longer glitch out.
/:cl:


deathride58 (approved):
Mirrored when travis was broken

PR #4520, opened 2017-12-24
[MIRROR] Prefixed all disability defines with DISABILITY_ to fix a name collision

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/33783
--------------------
The BLIND disability define used to have a value of 1, which was the same as the BYOND built in value of the BLIND bitflag used in the mob.sight var. When disability defines became strings in #33055, any operation that used BLIND as a bitflag (ventcrawling and shuttle docking computers) would runtime.

Fixes #33762


deathride58 (dismissed):
Relies on whatever added add_disability and all the other disability procs. Unable to find what added them.

LetterJay:
This also syncs a lot of the monkey related shit from the overhaul

PR #4523, opened 2017-12-25
readds shock collars

silicons:
woops the rnd sync nuked it


deathride58 (changes requested):
Make it compile, table cat

PR #4598, opened 2017-12-30
adds functional cebusoap

cebutris:
Adds in a functional version of my soap, purely for admin purposes.

I'll still spawn with the crap version but the real ones are avalible


fludd12:
cebusoap

PR #4818, opened 2018-01-14
Fixes character slot changing erroring out due to attempt_vr runtimes

deathride58:
REE

this is why you actually test the code instead of assuming something specific is happening

:cl: deathride58
fix: You can now change character slots again
/:cl:


deathride58:
Something in the changelog is causing the tool compile to fail.

PR #4826, opened 2018-01-14
Hopefully fixes human tails/ears not saving, and fixes new/random characters having underwear

### LINKED ISSUE #4092: Ears and Tails on humans do not save properly
Possibly related to #3182, ears and tails on humans do not save when set, leaving you as a base human the next round unless you manually reset it every single round. This is an issue because catgirls need to be catgirls.

deathride58:
These changes are a little hacky, but they're only intended to be temporary until the underlying issues are resolved.

The underwear fix can be removed whenever someone figures out how to get underwear layering properly with genitals.

The human tails/ears save fix can be removed when the actual issue causing its inability to actually save is discovered, though it can be kept and moved to modular_citadel to allow human mutant bodyparts to stay in sync with cit's mutant bodyparts.

:cl: deathride58
fix: New and random characters no longer get generated with underwear
fix: Human tails and ears now save properly again! As a side effect, tails/ears that were previously exclusive to mammals are now available to humans.
/:cl:

Fixes #4092

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -20,32 +20,35 @@
 		else
 			return "000"
 
-/proc/random_underwear(gender)
+/proc/random_underwear(gender)//Cit change - makes random underwear always return nude
 	if(!GLOB.underwear_list.len)
 		init_sprite_accessory_subtypes(/datum/sprite_accessory/underwear, GLOB.underwear_list, GLOB.underwear_m, GLOB.underwear_f)
-	switch(gender)
+	return "Nude"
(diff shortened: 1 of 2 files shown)


Poojawa:
Probably also get this to default tits/giant cocks to 'none' in the preview as well, that'd be greeeeat.

The underwear stuff needs to be re-enabled at some point.

deathride58:
Character previews retaining the first genitals rendered is a completely unrelated issue with how the character previews handle organs. It's an upstream issue, since the same bug makes dullahans break head rendering for all future character previews.

PR #4839, opened 2018-01-15
Shock collars

LetterJay:
:cl: Jay
add: Adds shock collars.... again
/:cl:

Somehow these were removed.


Poojawa:
You should move this out of the vanilla file, because it's just a datum.

deathride58:
Eh, we can move this to modular_citadel some time later

PR #4856, opened 2018-01-15
[MIRROR] Xenobiology Species Rework

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/34166
--------------------
:cl: XDTM
add: Reworked the species side of xenobiology.
add: Instead of an universal unstable mutation toxin, green slimes can now produce slime mutation toxin (plasma), human mutation toxin (blood), and lizard mutation toxin (radium).
balance: Slime mutation toxin requires several units and some time before acting, unless you're already a slimeperson (in which case it will randomize your subspecies)
add: Two new jellypeople subspecies have been added to make up for the lost species.
add: Stargazers have telepathic abilities; they can Send Thoughts, privately and anonimously, to any target they can see, and they can Link Minds with any aggressively grabbed mob to form a hivemind with them. Hiveminds have no limit but expire when the Stargazer dies or changes species.
add: Luminescents glow faintly and can absorb slime cores to gain special abilities. Each core has a minor and a major ability, generally based on what they normally do. Using these abilities does not use up the cores, but it has a cooldown depending on the ability's power.
balance: Slimepeople can now switch bodies on death or when unconscious, as long as they have another one.
/:cl:

I've always found grating that xenobiology had access to species they had no business working with, and slime magic can only explain so much. My idea is to separate those species into their respective departments/antags, instead of having them all into a centralized, cheap source. 
* Androids should be made by robotics through augmentation
* Podpeople are a botany feature
* Skeletons should be limited to liches or other magic antags
* Zombies should stay limited to romerol
* Golems are still available in adamantine slimes (and luminescents can morph into golems with them)
* Shadowpeople still lack an official place, but luminescents can become shadowpeople with black cores

The basic races are mostly human and lizard, and there are mutation toxins for those in case someone wants their species back after bodyswapping/ling mutation sting/etc.

Closes #34145 since it's non-compatible


deathride58 (changes requested):
missing changes in human_parts_greyscale.dmi

Poojawa:
No longer needed

PR #4892, opened 2018-01-16
Banana spider

theTiji0:
:cl: tiji0
add: Added banana spiders as a ghost controllable mob. Obtained through banana mutations
/:cl:

[why]: # 
Gives ghost players a small, responsabilities and impact free option to mess around with.


tgstation.dme
```diff
@@ -1925,6 +1925,7 @@
 #include "code\modules\mob\living\simple_animal\bot\medbot.dm"
 #include "code\modules\mob\living\simple_animal\bot\mulebot.dm"
 #include "code\modules\mob\living\simple_animal\bot\secbot.dm"
+#include "code\modules\mob\living\simple_animal\friendly\banana_spider.dm"
```
deathride58: The path in the .dme points to the wrong location. You placed banana_spider.dm in modular_citadel/code/modules/mob/living/, but didn't update the .dme

deathride58 (dismissed):
Ya just need to fix the .dme, then it should be good.

tgstation.dme
theTiji0: alright, It should be good now

deathride58 (approved):

PR #4894, opened 2018-01-17
[MIRROR] Revert Defines all subsystem priority values (#33537)

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/34528
--------------------
Reverts Defines all subsystem priority values (#33537)

Prioritys are a weight that controls how much of the byond tick the subsystem will be allowed to have, 50 was the default, so a subsystem with a priority of 100 would be allocated twice the amount of runtime as normal, and 25 would get half the amount.

Changing them to be consecutive numbers was improper and broke some of the things done to keep everything fair, and caused more overhead because costly but important subsystems were having to spread their workload out over more ticks.

@ tgstation/commit-access high priority please.

@ ninjanomnom Please redo this with just the define changes without changing the values.


deathride58:
Already handled by some other PR.

PR #4904, opened 2018-01-17
[WIP] IPCs as a species

GuppyLaxx:
A not-so-furry race!
Mechanically they'll work the same as humans, just kinda cosmetic, they can be changed very easily into something else, but we'll have to change them up for le balance
Also don't touch this PR for now because this shit barely works and was half assed while I was exhausted

:cl: GuppyLaxx
add: Adds IPCs as a species, totally not furry!
/:cl:


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1033,3 +1033,6 @@
 
 /mob/living/carbon/human/species/zombie/krokodil_addict
 	race = /datum/species/krokodil_addict
+
+/mob/living/carbon/human/species/ipc
```
deathride58: This can go into modular_citadel

config/game_options.txt
```diff
@@ -1,519 +1,520 @@
...
+## You probably want humans on your space station, but technically speaking you can turn them off without any ill effect
+ROUNDSTART_RACES human
+ROUNDSTART_RACES mammal
+ROUNDSTART_RACES avian
+ROUNDSTART_RACES aquatic
+ROUNDSTART_RACES insect
+ROUNDSTART_RACES xeno
+ROUNDSTART_RACES datashark
+ROUNDSTART_RACES guilmon
+ROUNDSTART_RACES ipc
```
deathride58: Revert changes to this file, try using notepad++ instead of whatever you used to edit this file, overly large diff is large

modular_citadel/code/modules/mob/living/carbon/human/species_types/species.dm
```diff
@@ -0,0 +1,12 @@
+//shamelessly stolen from hippie
+/obj/item/bodypart/var/should_draw_hippie = FALSE
```
deathride58, line 2: You can replace all references to hippie with cit.

Also, would ya be aight with porting over all of cit's species to use the should_draw_citadel vars/procs?

deathride58 (changes requested):
Beep!

config/game_options.txt
GuppyLaxx: turns out this staged the wrong file into the change, reeee

modular_citadel/code/modules/mob/living/carbon/human/species_types/species.dm
GuppyLaxx: i was gonna change it to citadel instead of draw_hippie, but i kept it cause meme

Poojawa:
Let's get this finalized, if we can. ~

GuppyLaxx:
Haven't been able to work on this, running on my phone data which means I cant put my 4k videos of **animal documentaries** in the back while coding

Poojawa:
Things to be updated since this has taken a very long time to be finished:

Mutant_parts.dmi needs to be checked to ensure body parts are all still there. Also the inclusion of hand sprites separate from arms.

IPC sprites should be added to one of the reference files for additional security in case they get overwritten.

tbh, all of the hippie refs need to be renamed to cit/citadel.

config/game_options.txt
```diff
@@ -1,521 +1,522 @@
-## HEALTH ###
```
silicons, line 1: fix your line endings

silicons (changes requested):

PR #4907, opened 2018-01-18
Craftable Intellicard Alt: Potat-OS

ghost:
- Adds the Potat-OS, a Portal inspired way to card AI's! Functionally identical to the intellicard, now with a %100 more starch.

- Crafted using a cable coil, a metal rod, an intellicard and a potato and found in the robot crafting menu.

pohtayto


deathride58 (changes requested):
You forgot to commit the addition to the .DME

deathride58 (approved):

PR #5026, opened 2018-01-22
[MIRROR] removes moth language

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/34745
--------------------
there was no need to give them a separate language, they already have a flight in low grav which is a pretty good feature. plus the only good race language is  draconic which are given to lizards and ash lizards because it makes sense lore-wise. (infact LIZARDS HAVE LORE AND MOTHS DON'T, MOTHS DON'T EVEN HAVE A WIKIPAGE)

infact i'm surprised no one has added anything related to moths on the wiki, it makes me think people only wanted this to spread their furry ideals.

also no this isn't obvious bait, i like to think moths are related to flypeople and that to become a moth, you were mutated into one and since flypeople don't get their own language, i'm making this pr

also i'm keeping the language icon there just in case.

@ KorPhaeron 

Fixes #34736 
Closes #34798


Poojawa:
Already handled

PR #5079, opened 2018-01-24
[MIRROR] Extends the disability refactor to include more traits, removing some snowflake code

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/34664
--------------------
This way you can add/remove traits without fear of other sources overriding them. Now you can add TRAIT_STUNIMMUNE to somebody without _what if hulk_

Notable changes: 
* Fakedeath now updates instantly, instead of waiting for the next life tick.
* Fakedeath now sets time of death when acquired.
* Removed extremely snowflake code in reagents that checked if you had morphine to remove slow immunity and so on.
* Hulk no longer overrides status_flag changes, in case there are any.


Poojawa:
Already handled, thank fuck

PR #5118, opened 2018-01-26
Restricts Supply Console to Quatermaster access

Poojawa:
:cl: 
balance: Supply Console is locked to Cargo Access only.
/:cl:

[why]: # Stops the clown from ordering corgis probably. also actually makes sense because the quartermaster is the one whose job it is to approve or deny shipments.

Papers, please.

### DIFF
--- a/code/modules/cargo/console.dm
+++ b/code/modules/cargo/console.dm
@@ -9,7 +9,7 @@
 		cannot transport live organisms, classified nuclear weaponry or \
 		homing beacons."
 	var/blockade_warning = "Bluespace instability detected. Shuttle movement impossible."
-	req_access = list(ACCESS_QM)
+	req_access = list(ACCESS_CARGO)
 
 	light_color = "#E2853D"//orange
 
--- a/code/modules/cargo/console.dm
+++ b/code/modules/cargo/console.dm
@@ -9,7 +9,6 @@
 		cannot transport live organisms, classified nuclear weaponry or \
 		homing beacons."
 	var/blockade_warning = "Bluespace instability detected. Shuttle movement impossible."
-	req_access = list(ACCESS_CARGO)
 
 	light_color = "#E2853D"//orange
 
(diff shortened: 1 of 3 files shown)


Kersakof:
I think it'd be better as QM and Cargo Tech on high pop, since you said Cargo tech and miner get QM access on low pop. Cargo would be invalidated without at least one person to do all the ordering, and cargo doesn't always have a QM. And when there's cultists on board, a job change is going to be the last thing people are worrying about.

Poojawa:
if we had some stats of rounds with QMs in 'em, that'd be good. if just to really judge if cargo tech access is needed.

also, hypothetical situations tend to get steadily more 'but what if' as they're discussed, so we should try to keep things to what we've seen is what usually happens. Which is more likely to be Cargo is either the first ones converted or horribly murdered in the ensuing clusterfuck of paper, swords, and lasers.

cebutris:
Should definitely allow cargo techs to access the console, and _maybe_ even miners. I think it should be Cargo Office access, or maybe even it's own seperate access (but I don't know how hard that would be to add)

Poojawa:
Seperate access is just adding a datum, but honestly that's unneeded. I'll shift it to just cargo techs.

Poojawa:
This also opens the flood gates for HoS to order stuff as well, since he has cargo access.

code/modules/cargo/console.dm
```diff
@@ -9,6 +9,7 @@
 		cannot transport live organisms, classified nuclear weaponry or \
 		homing beacons."
 	var/blockade_warning = "Bluespace instability detected. Shuttle movement impossible."
+	req_access = list(ACCESS_CARGO)
```
deathride58: you can slap this into modular_cit

deathride58:
did you test this to make sure it works? make absolutely sure you test it

Poojawa:
It's literally the same change I did to the mining shuttle console.

deathride58:
I remember someone else trying to do this same exact thing, only to find that access wasn't actually restricted at all.

Make sure you test this.

deathride58 (approved):

yourdoom9898:
My opinion isn't probably worth much, but the cargo techs and HOS having access is most likely a good idea. It stops the whole "we have no QM, need to go get a job change" issue, and it push comes to shove, nuke ops are invading the station or something, and Cargo is incapacitated, at least the HOS can still come in, and order much needed supplies for the fight.

PR #5120, opened 2018-01-26
[MIRROR] Holorecord fixes and additions

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/34882
--------------------
[Changelogs]: 

:cl: Dax Dupont
add: You can now rename and copy holodisks!
add: Holorecordings now can be looped.
tweak: Holodisks now have materials and the autolathe cost has been adjusted.
fix: Fixed holodisk speaker name not getting set when on non-prerecorded messages.
/:cl:

[why]: Fixes holorecords not setting speaker name. Checked for Unknown, since that's the default and it's required for the preprocessed disks. 
Removed dots from the buttons because it looks weird.


Poojawa:
Already handled

PR #5180, opened 2018-01-29
Interface color change

ghost:
- Makes the overall interface darker/less blinding in the information/status window and general border of the interface.

- Disables the loading of interface\skin.dmf and instead loads the modular_citadel\interface\skin.dmf

Got the idea from the lads over at [Oracle](https://github.com/OracleStation/OracleStation/), so the credit goes to them! Modified some portions of their .dmf to basically be a clone of our current information menu, minus a few things from the oracle skin. This should just be a simple color change, unless I goofed something up big time.


ghost:
Forgive the mess of a PR, don't know why those other things are showing.

PR #5186, opened 2018-01-30
[MIRROR] brief CC power patch - removes stargazers, adds a power cap of 50 kW, lowers tier unlock thresholds to compensate

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/34880
--------------------
:cl: Xhuis
balance: Stargazers have been removed.
balance: The clockwork cult now has a power cap of 50 kW.
balance: Script scripture is now unlocked at 25 kW, and Application at 40 kW, instead of 50 kW and 100 kW.
tweak: Important scriptures are now displayed with italicized names.
/:cl:

stargazers have been a balance hell since they were introduced, and it seems like there's no real way to balance them without either making them useless or not addressing the problem at large, so I removed them

integration cogs should be used for all power generation going forward; each cog produces 5 W of power per second, which is about a quarter of one stargazer; however, the massive amount of APCs on the station ensure that you'll still not run short on power if you ensure that they're cogged well

this also provides opportunities for ambushes through unexpected attacks on people busy removing integration cogs in maintenance, but makes it more feasible to disrupt the cult by attacking the heart of their operations, thus providing an ability to counterplay where none otherwise exists


Poojawa:
wew

PR #5235, opened 2018-02-01
Mentor System

Poojawa:
Oh boy, it's that time again, except this time it's actually working again

I did not do a SQL database test, but I'm pretty sure the old table is still there.

Brief test showed everything functioning as intended enough. there is a delay mentor command.


cfg/admin.txt
```diff
@@ -1,125 +1 @@
...
-cyberboss role=admin
-pjb3005 role=admin
-sweaterkittens role=admin
-feemjmeem role=admin
-jstheguy role=admin
-excessiveuseofcobby role=admin
-plizzard role=admin
-octareenroon91 role=admin
-serpentarium role=admin
-cebutris role=admin
```
deathride58: This entire file does basically nothing on a live server. There's no need to axe it.

deathride58 (dismissed):
Seems like it works

cfg/admin.txt
Poojawa: m e h

PR #5236, opened 2018-02-01
TG Sync 1/31

Poojawa:
Seems ready to merge. tales of warning about bucket lag though, but I'm sure we can hunt down and fix

🆑 Poojawa
fix: so many things.
add: Vore eject gun given to medical lockers, same as defib. Looks like a dragnet. will eject people out of all vore bellies. May need to fire more than once.
fix: Dogborg item blacklist is a thing again, scrubpups are able to process items into fuel.
tweak: Dogborgs are able to consume devourable simple mobs for fuel or for weird ERP.
tweak: Stargazer slimes are admin-only, way too much of a bad idea to hand out willy nilly.
tweak: Genital layering refractored to use mutable appearance. Also no longer always present on character screen.
fix: probly other things I forgot I tweaked, like cake hat nerfing
/ 🆑


Poojawa:
Ready enough to merge if there isn't anything else to fix, other than buckets

CitadelStationBot (dismissed):
Just revert the buckets PR and come back to it later, otherwise we'll be here forever. After that it'll be merge ready.

Poojawa:
We're gunna have to figure it out at some point. 

Literally, all that needs to be done as a hacky hotfix is to make it automatically clear buckets after initialization. That's all we're doing as admins and it fixes the issue.

PR #5327, opened 2018-02-05
[MIRROR] Return of ore stacking, various changes to lavaland bombs to facilitate reduced lag

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/35291
--------------------
First things first, @ kevinz000 had no part in the "bomb balance" aspects of this PR, he "only" made the original ore-stacking PR which I cherry-picked into this one. It just so happens to be the majority of this PR's contents so leaving his name out would have been unfair. 

:cl: Naksu and kevinz000
tweak: Ores are no longer completely impervious to explosions, but rather "proper" ores are destroyed by the strongest explosions and sand is blown away everything but the weakest ones. This shouldn't affect ore spawns from explosions
tweak: Lavaland bomb cap reduced to double the station bombcap.
refactor: Ores are now stackable. Ore stacks now have the appearance of loose ores and are not called "sheets" by machines that consume them.
code: Material container component now uses the singular names of stack objects.
refactor: /obj/item/ore/Crossed() is now removed in favor of the ore satchel utilizing a redirect component
/:cl:

stacks:
image

[why]: 

Every now and then someone lags the server to hell with bombs that never resolve without admin intervention or a server crash. Implementing a bomb cap didn't much help with this issue, as two bombs resolving at the same time is possibly worse than the megabombs you used to be able to make with fusion. This PR hits every part of the problem apart from explosions themselves, likely solving the lavaland special case problem with explosions.

Fixes #35336
Fixes #33712


Poojawa:
reeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

PR #5440, opened 2018-02-12
Ports prayer beads from Paradise

cebutris:
:cl: Cebutris
add: Added in prayer beads, code and sprites shamelessly stolen from Paradise. Chaplains can pray for people, to heal their wounds without the risk of braindamage, and cleanse their mind of unholy thoughts
/:cl:

[why]: Because Chaplains running around with swords is fun and all, but maybe we could have a few less murdery men of faith. Code and sprites from Paradise, modified to fit our codebase


deathride58 (dismissed):
ur coder!

Poojawa:
I like this idea, more friendly stuff is always appreciated.

PR #5452, opened 2018-02-13
Adds in special fluff items for TechnicalMagi

cebutris:
This adds in a worn sweater for TechnicalMagi (Naomi Reids), as well as a special collar for them, too. The collar induces arousal and hallucinations to those who wear it.

No, they didn't donate. No, there's no reason for this to be added other than the fact that they're special to me. This was mostly just a coding project for myself, so if it's closed on the grounds that they didn't donate,  that would be entirely understandable.


fludd12:
I'd gladly donate, even if this PR is closed! Been here for two years, definitely support you guys.

deathride58 (dismissed):

PR #5463, opened 2018-02-13
[READY FOR MERGE] The Giant Combat Reworks - Part 1: Goodbye stuns, hello staminacrit

deathride58:
This is the rework I've been pouring hours into recently.

:cl: deathride58
add: You can now sprint by holding shift.
add: Added combat mode. Press C or use the UI button to toggle it on and off. The code and the sound for toggling combat mode are from Interbay. The sound is mainly a placeholder until a more fitting sound can be found.
add: You can now use the right mouse button to perform actions while in combat mode. If something doesn't have a unique action for right clicking, it'll default to the normal left click action.
add: You can now move around and use items while resting. The code is from Interbay
add: You can now use the resist button to get up from resting. This will take time depending on your health and stamina, even if you try to use the rest button instead.
add: Reworked stamina, and introduced stamina crit. If you run out of stamina, you'll enter stamina softcrit, during which you'll be unable to attack, get up from resting, or perform various other actions. If you keep losing stamina after entering stamina softcrit, you'll enter full stamina crit, in which you'll be unable to move or interact with the environment until you're above 0% stamina again.
add: Added a stamina buffer. Performing actions that drain the user's stamina will now drain the stamina buffer before draining actual stamina. The stamina buffer will start to regenerate after going 5 seconds without performing a stamina-draining action. When the stamina buffer regenerates, it will use up stamina, at a rate of 0.5 points of stamina per 1 point of stamina buffer.
add: Added a rest button to the HUD
balance: 90% of all stuns have been removed in favor of stamina. Stuns will now simply knock the affected person down if they would have been 8 seconds or less, but will function for a "normal" stun with a length that's a tenth of the normal time if their normal time is longer than 8 seconds.
balance: Attacking with weapons will now cost stamina depending on the weapon's size. Some code from Interbay.
balance: Throwing things now costs stamina. Some code from Interbay.
balance: Default movement speed has been reduced by one tick to accommodate for sprinting. This does not require a config update.
balance: The slowdown you get when you're low on health has been reduced by one tick to accommodate for the shift from stun-based combat to stamina-based combat.
balance: Dogborg pouncing has been buffed from a knockdown of 4.5 seconds to a knockdown of 45 seconds. After the previously mentioned changes, this brings dogborg pouncing from 11.5 stamina + knockdown to 115 stamina + 4.5 second stun. **This is a temporary change to make dogborgs immune to the stun nerfs for now until a better solution can be found.**
balance: Crushers now regenerate stamina upon successfully detonating a mark.
add: Also laid down the groundwork for a psuedo z-height system. You can see a small glimpse of it by hopping on a table!
add: Added a couple of fancy buttons to the UI! Sprites are from Toriate.
add: Also added a stamina meter to the UI. Sprites are from Hippiestation
/:cl:

This is a huge, huge rework that will completely change how combat works. I'm hoping it'll make the combat on the server a little slower, and hopefully increase the skill ceiling of combat by quite a bit. I'm also hoping it'll remove the frustrations surrounding stunbased combat.

The primary inspiration for these reworks is Lifeweb. All similarities to Lifeweb and Interbay are intentional.

# TODO

- [x] Buff golden bolt and stunbaton
- [x] Make stamina have less of an effect on movement speed
- [x] Add buffers for actions that deal staminaloss to their user
- [x] Nerf resting item damage
- [x] Lay down groundwork for weapons having staminaloss penalties that aren't their arbitrary size class.
- [x] Add preference to make combat mode automatically disable itself upon reaching stamina softcrit
- [x] ~~Add preference to attempt to resist out of resting automatically upon getting out of staminacrit or reaching 0 staminaloss~~ Too clunky. Way, *way* too clunky. I really can't think of a way to implement this in a way that isn't so clunky.
- [ ] Add a couple more right-click interactions (OPTIONAL)

# KNOWN ISSUES, CAN BE FIXED IN SEPARATE PR

- Sleeping carp still has hard stuns unaffected by the stun nerfs (Resolved by #5617 )
- No martial arts have stamina penalties for their moves
- Stamina penalties for swinging weapons aren't always physically accurate
- Xenos still aren't vulnerable to stuns or staminaloss (Resolved by #5616 )
- Dogborg pounce


mattmatterson111:
:eyes:

BurbManager:
I think its a pretty fun addition, only problem is that i think stamina regenerates way too slowly.

deathride58:
Yeah, the main reason behind stamina regenerating slower than default TG speeds while standing is to stop stamina from being too meaningless. With TG's default speeds, it only takes 30 seconds to regenerate all your stamina from stamina softcrit, which makes staminaloss a complete joke in the heat of combat. I'll likely be buffing the standing stamina regeneration by a little bit to make things a tad bit less tedious, as it currently takes 1 minute, 40 seconds to regenerate all of your stamina while in stamina softcrit when standing. Stamina does regenerate at default TG speeds while resting, though, which does add a level of depth to combat, and makes the rest button somewhat useful in the heat of battle.

code/_onclick/item_attack.dm
```diff
@@ -68,15 +72,22 @@
 	add_logs(user, M, "attacked", src.name, "(INTENT: [uppertext(user.a_intent)]) (DAMTYPE: [uppertext(damtype)])")
 	add_fingerprint(user)
 
+	user.adjustStaminaLoss(w_class*1.25)//CIT CHANGE - makes attacking things cause stamina loss
```
silicons: items should have a 1-100 weight imo, instead of just w_class, w_class is used for balancing storage as much as weight, because i'd imagine a lightsabre, although "bulky", isn't actually that heavy.

silicons: long term stretch goal i guess i don't expect you to change everything's weights lol
or just make weights auto assigned if there's no specified weight and have it default to null =p

code/game/machinery/dance_machine.dm
```diff
@@ -400,15 +400,16 @@
 
 
 /obj/machinery/disco/proc/dance4(var/mob/living/M)
-	var/speed = rand(1,3)
+	//var/speed = rand(1,3) // CIT CHANGE
 	set waitfor = 0
-	var/time = 30
+	/*var/time = 30 CIT CHANGE -- replaces dance4 with rapid spinning so that disco balls dont make weird shit happen
```
silicons, line 405: >::::{

silicons: \>:{{{{{
just make it modify transform instead of resting because @Robustin  is a lazy ass

code/game/objects/items.dm
```diff
@@ -510,12 +510,18 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
...
+		to_chat(user, "<span class='danger'>You're too exhausted for that.</span>")//CIT CHANGE - ditto
+		return //CIT CHANGE - ditto
+
 	src.add_fingerprint(user)
 
 	playsound(loc, src.hitsound, 30, 1, -1)
 
 	user.do_attack_animation(M)
 
+	user.adjustStaminaLoss(10)//CIT CHANGE - makes eyestabbing cost stamina
```
silicons: :+1:

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1259,9 +1259,9 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 		var/hungry = (500 - H.nutrition) / 5 // So overeat would be 100 and default level would be 80
 		if(health_deficiency >= 40)
 			if(flight)
-				. += (health_deficiency / 75)
+				. += ((health_deficiency-39) / 75) // CIT CHANGE - adds -39 to health deficiency penalty to make the transition to low health movement a little less jarring
```
silicons: -39/75.....

CameronWoof:
I've delayed commenting on this PR until now, when Bhijn has stated that it's nearing completion and is almost ready for a full merge.

This is a massive set of changes, and while I appreciate the amount of work that goes into an update of this scope, I don't think that this is a good update for Citadel. Main points are:

**Stamina is extremely vague**, based entirely on the heart monitor HUD element that also tracks health. For a mechanic to be so integral to the game now, when you can't so much as break a window without going into stamina crit, it's extremely difficult to say for certain how much stamina you have and how much you need to perform an action without, essentially, killing yourself if a fight is going on around you.

**Heavy nerfs to melee weapons** in the form of high stamina costs and, in the case of stunbatons, actual efficacy jostle the balance of the game too much in favor of ranged weapons, which already had a distinct advantage.

**Lack of clarity** is a big one for me. "Spaceman is sideways" is a very clear, very definitive piece of visual feedback that a person is unable to act. With the addition of crawling and "fake stuns", it's now entirely possible for someone to be knocked onto the floor and still win a fight by firing their weapon or bashing someone with a high-damage melee weapon. Not to mention that they can still move.

**Manual standing** is an extremely finicky mechanic that really, really needs to be automated. Being in combat and being knocked into stamina crit prevents you from regenerating stamina or standing without manual input, which only serves to further punish players who aren't familiar with the unintuitive combat system this PR would instate. 

One of this PR's stated goals is to remove mass murder, and making it harder to compete in combat at all is *not* the way to do that. I've seen players stuck on the ground for minutes at a time because they're unfamiliar with the system, not understand (as it's not explained) that they can't regenerate stamina while in combat mode, and get killed by enemies that they'd managed to stun simply because they weren't disarmed.

**Usable Feedback**
- Stamina needs its own HUD element, with a concrete bar or real numbers and not a vague color system. 
- Stamina needs to regenerate in combat mode, OR combat mode needs to deactivate when a player is put into stamina crit.
- Players need to automatically stand up again when they have enough stamina. This should not be manual. 
- Anything that puts a player on the ground needs to disarm them. 
- Passive grabs on resting/crawling players need to prevent them from moving. The current Security meta is "keep stunbatonning them until the AI gets worried" and that's not elegant.

deathride58:
@CameronWoof thanks for the constructive feedback! Constructive feedback's pretty rare around these parts.

For stamina being vague, I'm planning on implementing a buffer system for both sprinting and the rest of the actions that drain stamina from the user. While those actions aren't being performed, the buffer will start to regenerate using the user's stamina so long as the user's outside of combat mode, and trying to perform those actions when the buffer is depleted will drain stamina completely. The buffer will feature its own unique HUD sprite that shows you how much you have left before you start getting tired. This should add a little more strategy and depth to gameplay while also resolving your, and various other folks' complaints of stamina being too vague.

For melee weapons having high stamina costs, I'm planning on resolving this down the line by making it so that weapons cost stamina depending on their actual physical weight instead of their arbitrary size. Weapons using the arbitrary size values for staminaloss calculations is currently temporary, and will be resolved in the future. Ranged weapons still need to be balanced for the new system, but they will be receiving nerfs when I get around to implementing stats/skills, inspired by the stats/skills implementation seen in Lifeweb and Interbay.

For the lack of clarity regarding spaceman being sideways, I'm planning on resolving the issue of a dude on the ground being able to simply whip out their esword to instantly win a confrontation against a spontaneous clown by making resting a lot less viable in terms of offense. Interbay handled it somewhat well by making resting cause weapons to lose most of their damage, which is what I'm planning on implementing soon. Being on the floor currently only has a negative impact during unarmed combat, but I'm planning on expanding it to affect melee and ranged combat. However, being on the floor shouldn't always be an instant ggnore, as that'll reintroduce the core problem of stunbased combat, being that stunbased combat is essentially rocket tag gameplay, something that makes combat effortless, which is something that should really be avoided now that we're shifting over towards a more roleplay-focused gameplay style. As for the complaint about spacemen being sideways meaning that the spaceman is unable to move or interact with their environment... You'll get used to it.

Manual standing is frankly a side-effect of the way stuns were reworked. Making standing automatic again would require quite a bit of work that would undoubtedly introduce a lot of oddities, namely with manual rest-to-crawl and numerous other features. Only reason getting up from stuns was previously automatic was because stuns were a hard status effect, you either had em or you didn't, meaning you got up the moment your stun was over. Now, stuns simply deal stamina damage and force the person to rest. Getting up from resting now depends entirely on your health and stamina. This implementation ensures future compatibility with upstream changes, while also giving much greater control over how being on the ground should be handled on the code side of things, and allows for more depth in getting up from the ground. I've tried my best to make manual standing a little more intuitive than it is in Lifeweb by making it plausible to use the resist button to get up from resting. I could add in a preference to automatically try to stand upon leaving stamina softcrit and/or successfully falling back to 0 staminaloss to make things a little more new-guy friendly, though this might lead new players missing the fact that it's now possible to move while on the ground.

For the smaller bits of feedback...

- See above walls of text
- Stamina regenerating in combat mode would be removing the main drawback of combat mode. I will, however, add a preference that defaults to on that allows players to automatically exit combat mode upon reaching stamina softcrit.
- See above walls of text. A preference will be added to allow players to automatically resist out of resting.
- Stunbatons, the golden bolt, and various other stuns already do force players to drop items in 99% of circumstances. I could make it so that getting knocked down calls the accident proc just like tabletripping and slippery tiles do, though, which will make it so that getting knocked down with the weaker stuns will make folks outside of combat mode drop their items on knockdown.
- I'm still very iffy on having aggressive grabs stop resting targets from moving, as is the case right now. I really don't want slip-grabbing to be the new meta to replace slip-cuffing. The security meta of stunbatonning over and over again can be resolved by making stunbatons do more stamina damage, though. Perhaps making it so only 2 stunbaton swings are required to push someone into full staminacrit can work? Same buff can be applied to the golden bolt.

code/modules/mob/living/carbon/human/species.dm
deathride58: Only values above 40 will have the -39 applied to them. its impossible to have the -39 applied with 0 health deficiency

code/game/machinery/dance_machine.dm
deathride58: im laaaaaaaaaaaaaaaaaaaaaaaaaaaaaazy

code/_onclick/item_attack.dm
deathride58: :+1:

mattmatterson111:
Being on the ground on InterBay also meant you couldn't hit above the groin, and you also had a higher miss chance (if I remember correctly), not just reduced weapon damage. At the end of InterBay's lifecycle being  knocked down meant you could fight a little but not much.

deathride58:
Did some codediving to refresh my memory, and it seems you are correct, though the higher miss chance seems to have only applied to unarmed combat. I might just be blind or searching for the wrong var, though.

Not sure if I'm overlooking a big chunk of code somewhere, but it seems like the only factor that made fighting on the ground non-viable in Interbay was the inability to aim higher than the groin. Due to how baymed works, that is what resulted in a much, much lower damage output potential, as legs aren't nearly as vital as a stomach or a heart. So, "Interbay reduced weapon damage while resting" was technically wrong, but also technically right. Either way, it did restrict damage output while on the ground by quite a bit.

Due to the vast differences between Baycode and TG, it's pretty hard to replicate all of the negative effects that come with being unable to aim above the groin, as a toolbox to the leg more or less has the same impact as a toolbox to the head with TG's medical system. However, I am planning on working on a medical rework down the line, which should let the inability to aim above the groin to be a lot more than just fluff.

I'll be seeing how reduced item damage while resting will work out for now. Players had quite a bit of fun with the ability to swing toolboxes around at full force while imitating slugs during the last two testmerges.

Poojawa:
Reminder that setting literally everything into preference will ensure that only a certain type is ever used (being default)

also preference creep is something we need to try to avoid, since .sav files are fucky and SQL would be annoying to have to database edit to add or remove to.

king5327:
My thoughts on this: it's approaching a combat overhaul the right way, but at the same time it's trying to change a very simple feature so entrenched in SS13 balance that almost every single mechanic needs a special case.

For example, the PK Crusher is a very heavy item that'll knock you out from the stamina damage before you even half-kill a drake (supposedly fixed before post). The esword should be a very light item but because its size changes when you extend it it does too much stamina damage to you for a light sword (status unknown). Dart guns can knock you out for a good few minutes with a single aimed barrage (not fixed by time of posting). No, not getting hit by riot darts, but firing regular ones.

Each and every one of these would have to be coded as an explicit exception to otherwise normal combat rules.

The main problem isn't the new mechanics, it's the fact that the new mechanics are directly based on the object's size class, as well as the fact you're not changing the damage slowdown to suit the new system. If you want this to fly, you must at the very least make each and every item have its own defined staminaloss and staminaloss on throw. No shortcuts.

Edit: Another thing that slipped my mind is resting. Moving _*should absolutely not*_ interrupt standing up from rest; furthermore, standing up should be automatic with no pref option to disable. Only successful disarms and punches should prevent it.

Lastly, combat and sprint should increase speed and damage on hit at the cost of no stamina regen. As it stands the game is balanced toward letting combat play out until somebody drops. The new meta under these changes is simply finding the weapon that does the most damage per time with the least stamina cost per hit. Which, because of the Crusher needing to be viable for mining, is undoubtedly going to be it.

Edit2: Also, RNG based aim sucks. The guy running away now has to contend with bullets that are randomized instead of predicting the other guy's aim. The shooter now has to contend with half the shots hitting the wall at a 70 degree angle from the original shot. (This is confirmed. A shot at one point, fired point blank diagonally up right hit a thing diagonally up three tiles and one left, with the dart dropping on the floor diagonally two up one left. - nobody has aim that shitty).

deathride58:
Items taking their stamloss from their arbitrary weight class is temporary, yes. There already is groundwork for making weapons use stamloss values that aren't their arbitrary weight class. The groundwork will be seeing a lot of use throughout the next two parts of the reworks. Dart guns being able to push someone into staminacrit from a single barrage is a pretty huge exaggeration, but if you can point me to which dart guns are supposedly dealing massive amounts of staminaloss, I can fix that with a single value change. I'm hoping you're aware that you still have 100 stamina to safely lose beneath that 20 points worth of buffer for staminaloss, as that seems to be a common misconception I've seen ever since I've implemented the stamina buffer. Stamina itself will be shown a lot more clearly when the server recompiles.

As soon as this gets merged, anyone will be able to make weapons cost different amounts of staminaloss just by changing a single value. Right now, anything that doesn't have that value defined will just use the arbitrary weight class for stamloss calculations, as it saves a *lot* of time of having to assign a physically accurate value for every single item in the game.

Moving cancelling getting up from resting is a code limitation. It'd be impossible to make getting up from resting ignore moving without also making it also ignore stuns, neckgrabs, etc, unless the code for handling timer-dependent actions were completely rewritten from the ground up to allow that one specific edge case of allowing folks to stand up while moving around, something that'd kill this PR's compatibility with upstream changes to timer code. Timer code gets changed pretty often upstream in the pursuit of optimization and decreasing memory usage.

Making standing automatic again and completely removing manual standing will require tossing out a solid 75% of the code in this rework. Removing manual standing will be decreasing gameplay depth by a lot, and would just bring things right back to where the motivation for these reworks began. The entire concept of manual standing is exactly what the reworks are being balanced around, as it's a massive improvement over the old 1-click-to-stun-and-have-your-opponent-be-unable-to-do-literally-anything-about-it combat system. Making automatic standing a preference really wouldn't be too optimal since there really isn't a way to implement an automatic standing system for these reworks that wouldn't be overly clunky.

Making combat more drawn-out is indeed one of the goals of the reworks. Making people think about what weapon they use when approaching an opponent is intentional. Combat, something that can leave someone sitting out of the round for up to three hours, should absolutely not be over in one single click under *any* circumstance.

As for the crusher, it's probably gonna be seeing a nerf as soon as I figure out how to get the stamina regen to only work in depressurized areas like how the PKA works.

And ya, the accuracy penalty for not aiming has already been nerfed. The previous value will be used for crew members that only have basic firearms skills in the next part of the rework

Poojawa:
all we mean by automatic standing as a toggle is let us have a toggle that manually stands or auto stand attempts without a message for spamming.

king5327:
As it is now, any gunfight can be over in one click with this one simple trick: lie down so that whoever's shooting you has to click you directly. This combined with the ridiculous RNG aim means that guns are completely and utterly useless against a crawling opponent with strong melee.

As for the crusher, my suggestion is that instead of fumbling with stamina regen code, it should recharge stamina fully on any hit against a marked target. This is an umbrella fix that allows crusher combat to be more or less identical to pre-update combat (complete with sprinting as long as you want) whilst being completely ineffectual in the rest of the game.

That having been said, Stamina has too big an impact right now while suffering from too little regen. I would suggest that you blanket halve all staminaloss values, and remove the 5-second stamina buffer cooldown so that it's always regenerating. Then, combat mode should simply prevent healthloss stamina from regenerating, while the stamina buffer runs continually.

As for the other things you noted: Yes, I'm aware that stamina is one of the hidden health stats (with rad and clone). However, right now you're just using stamina as some general nerfing statistic - stamina gives you no bonuses (compared to pre-update). Here's my suggestion: add the stamina buffer to the health calculation. Having a full stamina buffer, therefore, should fully offset having 20% healthloss, and the total stamina amount should act as a multiplier to reduce any flat stun times (like slips, flashes and lube). If you want dynamic combat, you have to provide people a way to use the mechanic, not just work around it.

deathride58:
The RNG added to guns can be removed by simply right-clicking in combat mode to aim. Aiming can be done much faster than a crawling opponent can move. The exact value for inaccuracy from not aiming has been recently been tweaked to not be so severe, from the range at which guns can reliably land a shot being 3 tiles, to 5 tiles. Guns aren't suddenly 100% useless just because you have to click one single time to aim. The requirement for aiming is so that running and gunning is no longer a viable strat, because without that requirement, there's really not much penalty to yakkety-saxing around a random assistant while firing a disabler at them. As a matter of fact, without the addition of aiming down sights, guns would be flat-out overpowered due to their minimal stamina costs and spammable nature.

Crushers already directly restore stamina upon mark detonation, if you've looked at the code, you'd know that it's just a simple AdjustStaminaLoss(-13) that gets triggered after mark detonation, nothing fancy. Making them restore *all* stamina would just be straight-up overpowered, moreso than the current value. Crusher combat already is more or less unaffected, only difference is that playing aggressively is now a much more viable option.

Stamina values have already been tweaked heavily from their original state in response to player feedback, playtesting, etc. Stamina being the biggest factor during combat is intentional. The regeneration is low to discourage people from trying to treat combat like TG and hippie and simply going in toolboxes-swinging without a single thought. This, along with the stun reworks, also means stunbatons, disablers, and other stamina-loss dealing items are now a lot more powerful than the bottles of water you can snag from almost any given corner of any given map.

The low stamina regeneration rates directly encourage folks to pace themselves during combat, it helps slow things down and make people think about their actions, one of the primary goals of these changes.

The stamina buffer having a delay on regeneration is to directly encourage people to pace themselves, as one of the major complaints was that trying to attack people was essentially killing yourself due to the shift towards stamina-based combat along with attacking costing stamina. If stamina buffer regenerated naturally 100% of the time with 0 delay, then there'd really not be much point to having the stamina buffer, as you'd then be able to constantly swing toolboxes around with little to no penalty nonstop, which would just be powercreep with the way combat works now. The stamina buffer works by pulling 0.5 stamina from your health to regenerate 1 point, and the rate of stamina buffer regeneration increases exponentially with the less you have in your staminabuffer. The delay for the regeneration to start is a little long, though, I'll be decreasing it to 3 seconds.

Stamina already is taken into account when getting up from resting. It has more of an influence on getting up from resting than it does for movement. "Stamina's just being used as a nerfing statistic" is because the entire combat system is being shifted away from stun-based combat to stamina-based combat. Most flat stuns have already been removed, there's only a handful of flat stuns now. Stamina buffer is a reward for players pacing themselves, the stamina buffer allows players to perform actions without affecting their actual stamina much. Making the stamina buffer have an influence in movespeed is something I'm a little hesistant to add, but I'll throw that in and see how it works.

king5327:
To your first counterpoint, I'd just like to remind you that no amount of right-click aiming will make shots hit somebody lying down without clicking them directly.

Lavaland has been entirely balanced around a stamina-free combat experience. What you're doing here is saying "no don't take a few seconds to dodge Bubblegum because otherwise you'll lose more stamina than your next hit regenerates and die miserably". If you weren't aware, staying too close to lavaland mobs makes them prime their melee so that you are guaranteed hit the next moment you step in range.

As for your next three points, I'd like to make the case that the staminaloss from hitting something should not ever be able to cause a knockdown. A decent stamina regen should allow them to prolong the fight or run away if it starts dragging on and reinforcements arrive. The buffer getting filled faster the lower it is should also contribute in this way, allowing you to keep attacking, but also losing speed bonuses.

I didn't know the stamina buffer pulled 0.5 stamina for 1 buffer point, but that actually adds to the balancing factor, since by making the buffer (but not stamina health) refill even in combat the fight can last twice as long - unless it is, as you put it, a blitzy fight with toolboxes swinging faster than the speed of sound and both parties sprinting.

Don't tell me "stamina helps buff standing back up" when pre-stamina getting up from resting was instant. I'm not comparing your PR's head to a previous commit, I'm comparing it to the normal combat system. Stamina does not offer a reward for people pacing themselves, it simply punishes those who don't. Offer someone more for having to deal with stamina management, not just a stay of execution. If flat stuns are directly affected by stamina, then you don't have to remove stuns, but instead just tune the multiplier to suit the patch. If you make standing up a zero-time action at 100% stamina health (that is to say, instant with no timer call, just whatever the call is to put them back on their feet) then you've added an actual reward to stamina management.

As for the stamina buffer influencing movespeed, as I said in my last post, it should be able to offset 20% physical damage's worth of slowdowns and other negative effects.

silicons:
im going to take a nap and then look at this i'm too sleepy for this.

ZeroNetAlpha:
Pulling triggers on burst weapons (Foam force guns, autorifles, etc) depletes multiple chunks of stamina instead of one per trigger pull.

king5327:
New and amusing bug has come in. I wasn't the one reporting this so I did my best to collect all the details.

From what I understand, running out of stamina while trying to stand up (somehow) causes a permanent state where rest will put you down, but you will be unable to stand back up on your own power, ever. No status messages are output by the attempt, according to the guy.

king5327:
More fun: Highlander swords should have no staminaloss.

deathride58:
@king5327 Aye, the bug where attempting to stand up while in staminacrit causes you to be permanently stuck resting has already been fixed! I'll also look into buffing the highlander sword.

@Zna12 Just fixed that! Burst-firing weapons should now deal staminaloss equivalent to one shot.

LetterJay:
Ping when this is MERGE READY and not just TEST MERGE READY.

deathride58:
@LetterJay It is merge ready, yes

Elllsiy:
I have comments on how this addition works with certain events but this doesn't look like the place to discuss it. Am I wrong? And if I am, Where do I go to discuss it?

Poojawa (approved):

deathride58:
@Elllsiy You can discuss it either here or on the #Development-Discussion channel of the Discord, aye.

code/modules/mob/living/simple_animal/bot/ed209bot.dm
```diff
@@ -274,7 +274,7 @@ Auto Patrol[]"},
 		if(BOT_PREP_ARREST)		// preparing to arrest target
 
 			// see if he got away. If he's no no longer adjacent or inside a closet or about to get up, we hunt again.
-			if(!Adjacent(target) || !isturf(target.loc) ||  target.AmountKnockdown() < 40)
+			if(!Adjacent(target) || !isturf(target.loc) || !target.recoveringstam || target.staminaloss <= 120) // CIT CHANGE - replaces amountknockdown with recoveringstam and staminaloss checks
```
silicons, line 277: staminaloss define?

code/modules/mob/living/simple_animal/bot/ed209bot.dm
```diff
@@ -301,7 +301,7 @@ Auto Patrol[]"},
 				back_to_idle()
 				return
 
-			if(!Adjacent(target) || !isturf(target.loc) || (target.loc != target_lastloc && target.AmountKnockdown() < 40)) //if he's changed loc and about to get up or not adjacent or got into a closet, we prep arrest again.
+			if(!Adjacent(target) || !isturf(target.loc) || (target.loc != target_lastloc && !target.recoveringstam && target.staminaloss <= 120)) //if he's changed loc and about to get up or not adjacent or got into a closet, we prep arrest again. CIT CHANGE - replaces amountknockdown with recoveringstam and staminaloss checks
```
silicons, line 304: define?

code/modules/mob/living/simple_animal/bot/secbot.dm
```diff
@@ -305,7 +305,7 @@ Auto Patrol: []"},
 		if(BOT_PREP_ARREST)		// preparing to arrest target
 
 			// see if he got away. If he's no no longer adjacent or inside a closet or about to get up, we hunt again.
-			if( !Adjacent(target) || !isturf(target.loc) ||  target.AmountKnockdown() < 40)
+			if( !Adjacent(target) || !isturf(target.loc) || target.staminaloss <= 120 || !target.recoveringstam) //CIT CHANGE - replaces amountknockdown with checks for stamina so secbots dont run into an infinite loop
```
silicons, line 308: define?

code/modules/mob/living/simple_animal/bot/secbot.dm
```diff
@@ -332,7 +332,7 @@ Auto Patrol: []"},
 				back_to_idle()
 				return
 
-			if(!Adjacent(target) || !isturf(target.loc) || (target.loc != target_lastloc && target.AmountKnockdown() < 40)) //if he's changed loc and about to get up or not adjacent or got into a closet, we prep arrest again.
+			if(!Adjacent(target) || !isturf(target.loc) || (target.loc != target_lastloc && !target.recoveringstam && target.staminaloss <= 120)) //if he's changed loc and about to get up or not adjacent or got into a closet, we prep arrest again. CIT CHANGE - replaces amountknockdown with recoveringstam and staminaloss check
```
silicons, line 335: define?

silicons (approved):
good good

PR #5478, opened 2018-02-14
Gives mineral storage access to those who need it

cebutris:
# DO NOT MERGE IF #5479 IS MERGED

:cl: Cebutris
add: Everyone who gets new mineral announcements should now have access to actually get those materials
/:cl:

[why]: Because why get announcements if you can't get them? I was originally gonna give virologists and chemists this as well but I learned they already have it **BUT ENGINEERS DON'T. WHAT IS THIS.**

Closes #5479


cebutris:
Compiled but untested, as is tradition

Poojawa:
Why do engineers have access to something the CHIEF Engineer is losing?

cebutris:
If this PR goes through, the other one should be closed. They're alternatives to eachother, two different solutions to one problem

deathride58:
@Poojawa Chief Engineer's access is untouched in this PR. This just adds mineral access for normal engineers, chief engineers keep their original access.

Poojawa:
Well, barkeep doesn't really need it, but station engineers and atmosia techs could use it. so

cebutris:
If the barkeep doesn't need it, why does he get mineral announcements?

deathride58:
Bartender can use uranium to make nukacola, which is basically discount meth that lacks any lasting side effects.

cebutris:
Not to mention all the other things they can do with uranium, plus some stuff with plasma and iron

LetterJay:
Close one of these PRs. There's no reason to have two open like this.

deathride58 (changes requested):
engineering.dm needs to be added to the .dme

deathride58 (approved):

PR #5479, opened 2018-02-14
Removes mineral storage access from those who don't need it

cebutris:
# DO NOT MERGE IF #5478 IS MERGED

:cl: Cebutris
del: Non-cargo crew no longer have access to the ORM
/:cl:

[why]: Because it's a cargo machine, cargo should be able to control what goes in and out of it

Closes #5478


Poojawa:
Cheng should also have access, due to being the primary constructor and needing materials.

cebutris:
Nope, this is JUST cargo, captain, and HoP. Nobody else.

Poojawa:
Soooo, Engineering gets futher shafted on materials, ok.

cebutris:
If this one goes through, it just means cargo needs to be more involved and give materials to the people that ask for it

LetterJay:
Close one of these PRs. There's no reason to have two open like this.

Poojawa:
Won't be needed

PR #5484, opened 2018-02-14
Emag limits

Poojawa:
:cl: Poojawa
balance: emags now have a 10 use endurance. Emag carefully, operatives.
/:cl:

[why]: # Instant gratification and a tool that is infinitely powerful is something better suited for servers who have no expectations of a round lasting more than 30m to an hour. We are obviously not shooting for that minimal time frame.

Emags will work for 10 uses before fizzling out and dropping at your feet. something something electrical current feedback. Really not hard to manage. This is to stem people away from just emagging all over the place willy nilly, with how pathetically simple it is to bypass doors anyway, there's no reason it should remain infinite use.

### DIFF
--- a/modular_citadel/code/game/objects/ids.dm
+++ b/modular_citadel/code/game/objects/ids.dm
@@ -44,3 +44,32 @@
 /obj/item/card/id/knight/examine(mob/user)
 	..()
 	to_chat(user, "<span class='notice'>Alt-click to recolor it.</span>")
+
+//=================================================
+
+/obj/item/card/emag_broken
+	desc = "It's a card with a melted magnetic strip, useless!"
+	name = "broken cryptographic sequencer"
+	icon_state = "emag"
+	item_state = "card-id"
+	lefthand_file = 'icons/mob/inhands/equipment/idcards_lefthand.dmi'
+	righthand_file = 'icons/mob/inhands/equipment/idcards_righthand.dmi'
+	flags_1 = NOBLUDGEON_1
+	flags_2 = NO_MAT_REDEMPTION_2
+	color = rgb(35, 20, 11)
+
+/obj/item/card/emag
+	var/uses = 10
+
+/obj/item/card/emag/afterattack(atom/target, mob/user, proximity)
+	. = ..()
+	uses--
+
+	if(uses<1)
+		user.visible_message("[src] fizzles and sparks. it is now burned out.")
+		user.dropItemToGround(src)
+		var/obj/item/card/emag_broken/junk = new(user.loc)
+		junk.add_fingerprint(user)
+		qdel(src)
+		return
+	..()
--- a/modular_citadel/code/game/objects/ids.dm
+++ b/modular_citadel/code/game/objects/ids.dm
@@ -66,10 +66,9 @@
 	uses--
 
 	if(uses<1)
-		user.visible_message("[src] fizzles and sparks. it is now burned out.")
+		user.visible_message("[src] fizzles and sparks. It's burned out!")
 		user.dropItemToGround(src)
 		var/obj/item/card/emag_broken/junk = new(user.loc)
 		junk.add_fingerprint(user)
 		qdel(src)
 		return
-	..()


cebutris:
How will this effect the ninjas katana, which has an emag function?

Poojawa:
Unaffected, this is just the card itself.

cebutris:
On the one hand, the emag can definitley be overused. On the other, limiting it like this seems a little... I dunno, bad? Perhaps there are other directions we could take it, like giving it "charges" that recharge over time, so that they can be used when you need it but you can't storm through an entire department at once, or maybe make it cheaper or something.

All in all I have no strong feelings one way or the other.

deathride58 (approved):
Optimal route for rushing the spare roundstart via emag on boxstation is 5 emag swipes. Optimal route for rushing the armory on box is 3 emag swipes.
This change won't be preventing the rushing strats made plausible by emags, but it definitely gives traitors something to think about if they plan to use the emag for more than just rushing.

+support, will jog some noggins

Iamgoofball:
just make this only decrease on emagging doors

king5327:
No.

silicons:
just give it a cd and charges

Poojawa:
Having a cooldown and charges continues allowing it to be infinite. Which is, kinda the thing I'm attempting to curb. 

Why does it need to remain infinite?

Also, how will those recharging points work? will it be a callback timer for every emag to check itself?

Iamgoofball:
@Poojawa Because by removing it being infinite you make it a shitty purchase.

Your issue is with it going through doors. Just make it have limited uses on doors and only doors.

deathride58:
Doors aren't the main focus.

Cargo consoles, mules, buttons, shuttle consoles, the gulag item reclaimer, sleepers, recyclers, and much, much more can all be emagged. Those emag functions almost always result in chaos. Being able to emag anything for as long as you want while only spending 6 TC is a little absurd the more you think about it, considering how crippling some of the emag functions tend to be. And if you run out of charges on your emag and wanna sow more chaos upon the station? Just spend another 6 TC on a new emag.

10 uses is double the required amount for the current emag meta. You can still use an emag to rush the spare roundstart just fine with limited uses, which is enough of a bargain on its own for only 6 TC.

Making uses limited means that all of the other emag functions require thought and strategy put into it instead of just clicking every single thing with an electrical current to wreak havoc upon the station.

deathride58:
As for recharging emags, making a one-use emag recharge item that lets you add 5 emag uses onto your emag that can be purchased for 2 TC would probably work out pretty well.

improvedname:
To be fair just remove emagging cargo and shit and make it a special traitor cargo purchase etc

improvedname:
Theres no reasons for emagging robots it your slave

Emaggs are supposed to fuck electronics up not being benifitfull 200% of the time
Slave robots can be achieved by the surgery bag. With the syndicate mmi.
Special cargo purchases could be put unded a cargo only traitor item(like a special disk)((this also removes xenobio meta of syndiborgs emagging cargo for traitor stuff)

fludd12:
It was once limited, which was removed after literally everyone complained for like a year straight.

Poojawa:
It was actually reversed arbitrarily because a coder didn't like it and I hadn't bothered to re-change it till now.

Same coder who disabled suicide because reasons.

fludd12:
I can't find any related PR for that, but I'll defer to your likely correct story.

I just know that on para, it once was limited, and by popular opinion became unlimited.

Poojawa:
@fludd12 https://github.com/Poojawa/Citadel-Station-13-5th-Port/blame/28621a9c1e0e31402a42471e6cd43c51c443d0b5/code/game/objects/items/weapons/cards_ids.dm#L73

fludd12:
I'm reasonably certain that wasn't "arbitrarily," but instead to maintain parity with TG before the pull bot was established, but I won't argue further, if you think this is needed, go for it, but it's going to make the emag taken about as often as the syndi-cigs.

Poojawa:
The PR bot is a relatively new development, and if you notice that was back in 5th port. I think I had it in 4th as well. 

So 2+ years ago. It curtailed people buying it and just emagging the station, but was still used as well due to how useful it can be in a pinch. You just had to plan it out better.

modular_citadel/code/game/objects/ids.dm
```diff
@@ -43,4 +43,32 @@
...
+	color = rgb(35, 20, 11)
+
+/obj/item/card/emag
+	var/uses = 10
+
+/obj/item/card/emag/afterattack(atom/target, mob/user, proximity)
+	. = ..()
+	uses--
+
+	if(uses<1)
```
Fox-McCloud, line 68: I know this was already merged, but uh, I'm going to nitpick anyway:

```
uses = max(0, uses - 1)

if(!uses)
```

would have been better logic for this.

PR #5490, opened 2018-02-15
Adds stunsword modification kits to the security vendor

cebutris:
:cl: Cebutris
add: Security vendors now have a stunsword modification kit available for purchase with a coin. There might be another somewhere in there, but that would require tampering with it, and you're a good little redshirt, aren't you?
/:cl:

[why]:   Makes my amazing swords available to players

### DIFF
--- a/modular_citadel/code/game/machinery/vending.dm
+++ b/modular_citadel/code/game/machinery/vending.dm
@@ -1,0 +1,3 @@
+/obj/machinery/vending/security
+	contraband = list(/obj/item/clothing/glasses/sunglasses = 2, /obj/item/storage/fancy/donut_box = 2, /obj/item/device/ssword_kit = 1)
+	premium = list(/obj/item/coin/antagtoken = 1, /obj/item/device/ssword_kit = 1)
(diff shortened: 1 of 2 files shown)


modular_citadel/code/game/machinery/vending.dm
```diff
@@ -0,0 +1,3 @@
+/obj/machinery/vending/security
+	contraband = list(/obj/item/clothing/glasses/sunglasses = 2, /obj/item/storage/fancy/donut_box = 2, /obj/item/device/ssword_kit = 1)
```
deathride58, line 2: Pretty sure you can do
`contraband[/obj/item/device/ssword_kit] = 1`
to avoid redefining the entire list

Same with premium

cebutris: Doing that exactly I get this error:

`modular_citadel\code\game\machinery\vending.dm:2:error: contraband: left-hand side must be an object variable`

`modular_citadel\code\game\machinery\vending.dm:3:error: premium: left-hand side must be an object variable`

PR #5551, opened 2018-02-17
Citadel'd Maps from #5523

Poojawa:
Mirroring #5523 
Original PR Text:
Adds new helpers to handle cyclelinking & locking airlocks.

The reasoning for this is to
1: Fix a number of broken icons on maps related to locked airlocks
2: Make it clear to mappers what it happening on the map without having to manually review an airlock's vars
3: Make it very obvious to mappers when there are issues with cyclelinking
4: Reduce the number of unique instances of vareditted items on maps

With these changes, I identified a very large number of incorrectly cyclelinked airlocks, and resolved all of said issue. Also went through and linked all unlinked airlock pairs where was it was necessary.


Poojawa:
wew. updated all of the supplemental maps as well, so yaaaay

Poojawa:
reeeeeeeeeeeeee travis

PR #5558, opened 2018-02-18
Reduces base rnd point generation rate to 1,200 points per minute

deathride58:
Title. Does exactly as it says on the tin.

:cl: deathride58
balance: RnD's base research point generation rate has been decreased to 1,200 points per minute.
/:cl:

Prior value was 3,258 points per minute. This meant RnD got completed around the 40 minute mark. This change lets techwebs actually act like the timegate it's supposed to be.

### DIFF
--- a/modular_citadel/code/controllers/subsystem/research.dm
+++ b/modular_citadel/code/controllers/subsystem/research.dm
@@ -1,0 +1,2 @@
+/datum/controller/subsystem/research
+	single_server_income = 20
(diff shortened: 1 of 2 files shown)


fludd12:
If you're doing this, could you enable building more servers generating more points per minute? It's currently disabled in the code, at least, last I checked.

deathride58:
Building more servers to generate more points is disabled because it's way too easy to generate metric fuckloads of points just by building a bunch of extra servers roundstart.

It's plausible to get RnD done faster by setting off bombs, drinking alcohol as a scientist, or having robust engineers set up the engine and use the research point generation mode of rad collectors.

fludd12:
I mean, there's also an option to make more servers scale point generation logarithmically, which will make it not do that.

Poojawa:
I'd still restrict it to more like 500, but we'll see how 1200 goes.

PR #5578, opened 2018-02-19
Code owners citadel style

Poojawa:
Because it's neat and should be a thing on our code to prevent ghost nerfs.


.github/CODEOWNERS
```diff
@@ -4,67 +4,15 @@
...
-/_maps/ @ShizCalev
-/sound/ @ShizCalev
-
-# Multiple Owners
-
-/code/controllers/subsystem/air.dm @duncathan @MrStonedOne
-/icons/ @ShizCalev @WJohn
+#deathride58
+/modular_citadel/ @deathride58
+/code/citadel/ @deathride58
```
deathride58: You can probably change this to just /code/ so that I can scream at anyone who thinks of making non-modular changes

Poojawa: which will end up being every single mirror'd PR.

deathride58: Yes

deathride58 (dismissed):
Could probably add Jay to __donator.dm in the loadout gear stuff and custom_items.dm so that it'll automatically request a review from Jay for any donor fluff that gets added

deathride58 (approved):

PR #5594, opened 2018-02-20
fixes possible missing sprites

Poojawa:
Xenomorphs might've been invisible randomly, they shouldn't be any more


deathride58:
Travis broke

PR #5616, opened 2018-02-20
Removes xeno stun and stamina immunity

deathride58:
Title. Intended to go alongside #5463 

:cl: deathride58
balance: Xenos no longer have stun or stamina immunity
/:cl:


cebutris:
While they shouldn't be immune, they should probably be massively resistant or at least have some sort of faster regeneration. They're living biological weapons designed for murder, and should be able to murder people more efficently than an actual person

PR #5617, opened 2018-02-20
Makes martial arts use Knockdown() instead of Stun() for their stuns

deathride58:
Title. This makes it so sleeping carp and psychotic brawl obey the new stun nerfs in #5463 without affecting existing functionality.

:cl: deathride58
balance: Sleeping Carp and Psychotic Brawl will now both use Knockdown() instead of Stun() for their stuns, making things a lot more consistent with the recent stun nerfs.
/:cl:

### DIFF
--- a/code/datums/martial/psychotic_brawl.dm
+++ b/code/datums/martial/psychotic_brawl.dm
@@ -47,7 +47,7 @@
 			if(!istype(D.head,/obj/item/clothing/head/helmet/) && !istype(D.head,/obj/item/clothing/head/hardhat))
 				D.adjustBrainLoss(5)
 			A.Stun(rand(10,45))
-			D.Stun(rand(5,30))
+			D.Knockdown(rand(5,30))//CIT CHANGE - makes stuns from martial arts always use Knockdown instead of Stun for the sake of consistency
 		if(5,6)
 			A.do_attack_animation(D, ATTACK_EFFECT_PUNCH)
 			atk_verb = pick("punches", "kicks", "hits", "slams into")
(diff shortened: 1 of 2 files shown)


jakeramsay007:
Martial Arts are supposed to be the most deadly form of hand-to-hand combat in SS13. Sleeping Carp, off the top of my head, currently costs 19tc. Part of SC is that the combos utilise the stuns that you can do to do extra damage while the person still has a chance to break the stun and fight back, which is why I believe that it should be unaffected, otherwise you can chainstun and staminalock a person with sleeping carp instead of it being a hard stun that clears immediately upon it timing out after two seconds or so.

Poojawa:
No-one ever buys sleeping carp unless it's on sale anyway.

Also we still need getting up not to be interrupted by movement, or make it a hard toggle like run/walk is. Combat and movespeed is still too fast to have as long of a delay as it does.

deathride58:
Actually, even if this PR doesn't go through, it'll still be entirely plausible to push someone into staminacrit and keep them in staminacrit with sleepingcarp due to the already existing usage of the Knockdown() proc in sleepingcarp. This "stunlock" applies the same to every single stamina-dealing and stun-dealing weapon in the game, that's how the combat reworks are supposed to work. The difference between the previous stunlocks and staminacrit is that staminacrit requires more than one single click to initiate. There is no reason that the stuns in sleepingcarp should be unaffected by the stun reworks in the combat reworks, as keeping things inconsistent will do nothing but confuse players.

Without this PR, sleepingcarp will have the issue of making it way, way too easy to stunlock someone. The three stuns affected in this PR are plausible to chain together to initiate a stunlock from the very first hit on the opponent, with or without combat reworks. The changes in this PR will make the stun reworks properly apply to the new stuns.

There's no reason to keep sleepingcarp inconsistent from the rest of the stuns affected by the rework, as the sleepingcarp stuns served the same exact purpose as the rest of the stuns affected by the rework.

Mechanical consistency shouldn't have to be an uphill battle like every other minute change.

PR #5623, opened 2018-02-21
Vore preference fix + Sounds respect wall barriers

Poojawa:
:cl: Poojawa
fix: Fixed preferences from overridding vore bellies on different characters with the last saved. Maybe. Worked on local.
soundadd: added a button for prey to restart their sound loop, if it cut out and they want it back on.
soundadd: Vore sounds respect walls, people can stop bitching about dorm room vore RP.
/:cl:

yaaaay.


Poojawa:
8^) do I get to self merge this one?

deathride58 (approved):

PR #5624, opened 2018-02-21
Blob update

EtheoBoxxman:
Ups the minimum players from 20 to 40 for a blob event to happen.


deathride58 (changes requested):
Modularize the new value

Poojawa:
I'd rather the minimum start time upped rather than player population., you only need a few crew with meme guns to clear a blob out.

PR #5654, opened 2018-02-22
fixes arousalloss not being a valid damage type for humans

silicons:
>:[


Poojawa:
Oh, the must've been missed. oops

PR #5726, opened 2018-02-27
[MIRROR] Tesla Corona Analyzers

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/35965
--------------------
:cl: Poojawa
add: Tesla Corona Analyzers! Study the seemingly magic Edison's Bane for supplemental research points!
/:cl:

Pubby does not come with an Supermatter engine, thus cannot produce tritium for fusion warheads for science. This would also permit a return of Tesla Engines for other uses as well, if/when a point generation nerf is implemented.

The maximum point cap and resistance values are just randomly assigned, and should probably be tweaked as needed. Using the research setting induces a 20 (up from 2) Power_loss value, effectively making voltages from 120,000W strikes to 6,000W, but there is no cap to power generation in this setting. Points per Tesla_act are hard capped at 200, but like I said I pulled it out of my ass assuming that having multiple energy balls doesn't just call tesla_act multiple times at once, rather than just adding to the power generation as a whole. 

Power setting is like normal, with the added small bonus of points to reflect the focus of hard power transfer.  You can swap between modes as you would the gas freezer/heater, as I found that the best alternative to trying to hack in a mod kit or multitool fuckery. 

Currently none of the maps are edited for specific machines, so they'll all default to power mode unless deconstructed and swapped over. 

Proof of working:
`ADMIN LOG: Poojawa/(Flick Kelley) called Tesla Corona Analyzer's tesla_act() with the arguments 200.
Available points: 10 (+0 / minute)`

`ADMIN LOG: Poojawa/(Flick Kelley) called Tesla Corona Analyzer's tesla_act() with the arguments 547387.
Available points: 210 (+0 / minute)`

Sprites! 

zap fixed

daea49e50ab37cfff413aaaa3056cf24


Harakoniwarhawk:
I can see this being very useful on Delta and Pubby, even if the point generation isn't that high, the basic coils upgraded produce more than enough power to add plenty of these.

Poojawa:
speed merge my self merge of my own pr.

totally how it works, right?

cebutris:
Is it still self merging if it's a mirror of your own PR on another codebase? 🤔

PR #5727, opened 2018-02-27
Adjusting Prayer Beads to be more useful

jakeramsay007:
:cl: RealDonaldTrump
tweak: Removed the probability check from prayer beads for their low amount of healing, as well as lowering the time needed to heal from 15 seconds to 10 seconds. Slimepeople won't get harmed by prayer beads either.
/:cl:

Since you sacrifice the ability to deal damage with your various null rod weapons to be the true embodiment of a healslut, the beads now actually are reliable healing tools.


Poojawa:
these should heal arousal as well. just because. like 10-15 points worth.

cebutris:
Pray the gay away

deathride58 (approved):

pantherpusher:
Can you wear them around your neck?

PR #5739, opened 2018-03-01
Triphase Laser Rifles

jakeramsay007:
:cl: RealDonaldTrump
add: Added the Triphase Laser Rifle; a weapon designed to fire a true laser at the speed of light. Medium damage with a long cooldown and low ammo capacity. The Triphase Laser Rifle uses five round energy magazines that can be reloaded in a recharger. Both the Triphase Laser Rifle and Energy Packs can be produced from the security protolathe once beam weapons are researched.
/:cl:

image


silicons (dismissed):
code is good enough if it's already tested and all
balance wise i don't care but think long and hard about the can of worms this will open

jakeramsay007:
Let it be known that I have no idea how the shit I correct .dmi conflicts.

PR #5755, opened 2018-03-02
[MIRROR] Modernizes SQL admin loading

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/35264
--------------------
Updates sql based admin loading to support excluding flags from ranks and to allow for tempminning. Since changes are largely piggybacking off the existing systems for loading admins, legacy loading is unchanged and still supported. Functionality for adding and removing individual verbs from ranks is removed. Any database errors when loading admins forces a revert to legacy loading.

When sql admins is enabled actions on the permission panel are prompted to be either written to the DB or just for the current round, where applicable. Additionally when sql is admins enabled the list of admins and ranks is reloaded from DB and recorded to `admins.txt` and `admin_ranks.txt` as a backup. This is also where updating `lastadminrank` is handled so only permanent changes are kept.

The interface for changing rank flags now uses @ vuonojenmustaturska bitfield switcher, which has been modified to allow changing window height, width and slider color. The exclusion flag picker uses red sliders for clarity. For reference, the bitfield switcher looks like: https://github.com/tgstation/tgstation/pull/34078#issuecomment-359104142

`rights2text()` updated for current admin flags and to output both included and excluded flags.
`code\modules\admin\permissionverbs\permissionedit.dm` moved to `code\modules\admin\permissionedit.dm` as it was the only file there.

All three admin tables are modified to remove unused columns and improve the remaining. `admin_log` now records operation type per log for easier searching, however note there's currently no interface for reading the log.

A python script is included to handle importing admins and ranks from `admins.txt` and `admin_ranks.txt`

General functionality tested to work, pending more testing to root out possible edge cases.
@ MrStonedOne 
@ nfreader


Poojawa:
Merge whenever this is good, I guess

PR #5771, opened 2018-03-03
What blob

ZeroNetAlpha:
[Changelogs]: Blobs now wont spawn until at least an hour into the shift and with no less than 50 people.
:cl: ZeroNetAlpha
tweak: Blobs now start at 60 minutes and only if 50 players are online.
/:cl:

[why]: Long drawn out discussion in Discord led to this.


yourdoom9898:
Can we mute Cameron for like a week so development-discussion stops exploding every day?

LetterJay:
This is redundant since the playerlimit is a config option.

ZeroNetAlpha:
I'm going off what the blob has on it's event 'blob.dm.' the 40 player is what was on there to begin with, just copied the format and set it to 50

deathride58:
Actually, the config option playerlimit is just for the gamemode. This is for the random event.

Poojawa:
60m is a good setting, yes yes.

PR #5774, opened 2018-03-03
Bomb points scaling

ZeroNetAlpha:
[Changelogs]: #Citadel Station is not /TG/Station13. We have a different crowd than TG. We also have our passive point generation halved. Which is an issue. Reverts #5684 .

:cl: ZeroNetAlpha
tweak: Restores Doppler arrays to the way they were before PR #5684 came along
/:cl:

[why]: We're not TG. We won't be TG. If having enough skill to gain research points quickly though gas mixing is a bad thing, then why are we here?


Poojawa:
okay it's obvious you didn't read the reasons behind why it was nerfed 

without the curve, it's possible to only need one explosion and *completely finish R&D* 

That is not acceptable. There are going to be supplemental ways to get points.

modular_citadel/code/game/machinery/doppler_array.dm
```diff
@@ -0,0 +1,132 @@
+GLOBAL_LIST_EMPTY(doppler_arrays)
```
silicons, line 1: why are you copying the entire file whne you have to override a single proc?

DaedalusGame:
hi i made the pr you're reverting and it's balanced for complete removal of passive gen in the future

deathride58:
Yo, as I've stated in the Discord multiple times, bomb sizes are exponential. Point gains are square rooted now.

When you do the math, the point gain nerfs actually just make the point gains linear instead of exponential.

Trying to revert proper calculations for bomb size research point generation just shows a complete lack of understanding of how bombs and explosion sizes currently work.

PR #5789, opened 2018-03-04
Vore code to json and more

Poojawa:
:cl: Poojawa
add: Added release sound effects for choosing. If having a non-belly, belly.
add: Added a 'silent' modifier to bellies. 
add: Added a vore subsystem to handle belly processing, should be less clunky
balance: rebalanced heal mode and dragon digestion.
soundadd: added client preference based sound toggles to both eating/release/struggle and digestion noises seperately.
refactor: vore prefs save to JSON now instead of .sav. much easier really.
refractor: dragon vore is working, though with this game who knows.
server: your prefs should automatically transfer, but it's likely you'll lose pre-loaded bellies, BACK UP EVERYTHING. (you should be doing this anyway too)
server: Creating new characters will import the previous' bellies. but once you click 'save' they'll make the new JSON file will all the info. This is per slot, so if you change your character, you'll have to change belly stuff.
/:cl:

wew lad.

90% of this code was nabbed from Virgo. because heck it's good. Just to lend some credit <3


cebutris:
Now, I know that after I say this I'm going to get flak about being anti-vore or whatever...

But this seems like a pretty substantial buff for the heal mode. Why is vore able to heal people? How does that work? How is that balanced?

Poojawa:
No-one uses it anyway. it'll also drain nutrition like mad. 

also ~magic~

Poojawa:
Good to go!

PR #5817, opened 2018-03-05
Filet-O-Sharkpeople

ZeroNetAlpha:
[Changelogs]:Aquatic Species are now properly giving fish fillets when they're run through the gibber instead of regular meat steaks.

:cl: ZeroNetAlpha
tweak: Tweaked Aquatic Species to give fillets when run through the chef's gibber.
/:cl:

[why]: # If you're going to gib a fishperson, it makes sense that they drop fish fillets.


modular_citadel/code/modules/mob/living/carbon/human/species_types/furrypeople.dm
```diff
@@ -0,0 +1,2 @@
+/datum/species/aquatic
```
Poojawa: You don't need to edit this, because furrypeople.dm is our snowflake file.

deathride58: Could probably move the entire furrypeople.dm to modular_citadel for the sake of consistency

ZeroNetAlpha: So move furrypeople.dm to modular_citadel?

deathride58 (dismissed):

deathride58 (approved):

deathride58:
Time to bully those walking, talking pillars of sushi

PR #5848, opened 2018-03-07
Tesla adjustment

Poojawa:
40m into the round, R&D is finished and drowning in hundreds of thousands of points.

I did mention the original values were not even tested.


Poojawa:
what the fuck

PR #5849, opened 2018-03-07
tweaks Tesla Point income

Poojawa:
40m into the round, R&D is finished and drowning in hundreds of thousands of points.

I did mention the original values were not even tested nor were they being anything other than blatantly generous. 

github did a dumb on the last one.

### DIFF
--- a/code/modules/power/tesla/coil.dm
+++ b/code/modules/power/tesla/coil.dm
@@ -85,7 +85,7 @@
 		playsound(src.loc, 'sound/magic/lightningshock.ogg', 100, 1, extrarange = 5)
 		tesla_zap(src, 5, power_produced)
 		if(istype(linked_techweb))
-			linked_techweb.research_points += min(power_produced, 10)
+			linked_techweb.research_points += min(power_produced, 2)
 		addtimer(CALLBACK(src, .proc/reset_shocked), 10)
 	else
 		..()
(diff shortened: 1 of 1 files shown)


deathride58 (approved):
Be sure to send this upstream

Poojawa:
already have.

kiwilover8:
Provided it isn’t incredibly difficult to add additional generators to the Tesla array, I see no problem in bringing the point income down to a reasonable 720~ if the 4 generators to points still allows enough to power the station without additional added

JogBoba:
That kiwilover8 comment was me, old account I forgot I had.

Poojawa:
That's part of the intention is to make it so engineering needs to choose, power or research contribution. 

once I get an estimate on power income for analyzers alone, I'll end up tweaking their penalty moreso. the basic setup should be profitable but not anything to speak of. all of them to analyzing should be a pretty hefty deficit in power, it's only 1/20th, but might need to be 1/50th

JogBoba:
My intention with that comment is that RnD should be able to be finished in a bit over 2 hours with just passive gain, and I feel the engine being able to power the station and provide that at the same time would be a better alternative than increasing the base server gain, which is why I asked if it could still power the station while producing the needed 700+.

Poojawa:
oh no, self merge.

PR #5907, opened 2018-03-12
Hypospray kits and reworks!

Poojawa:
:cl:
add: Pre-packaged Hypospray MK II kits are now being supplied to station Nanomed Plus units!
add: The CMO's mkII kit is added to their locker as well.
add: Medical Doctors and Chemists will spawn with a hypospray kit in their backpacks.
tweak: Trek Chems are given new colors other than pink to better differentiate in the new vials
tweak: Hypospray functionality refractored thanks to Kfive.
/:cl:

Hooray!

This also adds a combat hypo kit, which is basically the CMO kit + lots of goodies. Try not to dump everything into your buddies tho


Poojawa:
Not modularizing the color change, I'm gunna be replacing all the reagent code soon enough.

modular_citadel/code/game/objects/items/storage/firstaid.dm
```diff
@@ -0,0 +1,140 @@
...
+/obj/item/storage/firstaid/o2
+	icon_state = "oxy"
+
+/obj/item/storage/firstaid/tactical
+	icon_state = "tactical"
+
+//hijacking the minature first aids for hypospray boxes. <3
+/obj/item/storage/hypospraykit
+	name = "Hypospray kit"
+	desc = "It's an kit containing a hypospray and specific treatment chemical-filled vials."
```
GrayRachnid: A kit.

modular_citadel/code/game/objects/items/storage/firstaid.dm
```diff
@@ -0,0 +1,140 @@
...
+/obj/item/storage/hypospraykit/tactical/PopulateContents()
+	if(empty)
+		return
+	new /obj/item/defibrillator/compact/combat/loaded(src)
+	new /obj/item/reagent_containers/hypospray/mkii/CMO/combat(src)
+	new /obj/item/reagent_containers/glass/bottle/vial/large/preloaded/combat(src)
+	new /obj/item/reagent_containers/glass/bottle/vial/large/preloaded/combat(src)
+
+/obj/item/storage/hypospraykit/cmo
+	name = "CMO Hypospray medkit"
```
GrayRachnid: Would probably be better to break down CMO here as most other head of staff personal items follow that rule.

Poojawa: "Chief Medical Officer Hypospray medkit" would be cutoff name length wise

GrayRachnid: True that

modular_citadel/code/game/objects/items/storage/firstaid.dm
```diff
@@ -0,0 +1,140 @@
...
+	icon = 'modular_citadel/icons/firstaid.dmi'
+
+/obj/item/storage/firstaid/Initialize(mapload)
+	. = ..()
+	icon_state = pick("[initial(icon_state)]","[initial(icon_state)]2","[initial(icon_state)]3","[initial(icon_state)]4")
+
+/obj/item/storage/firstaid/fire
+	icon_state = "burn"
+
+/obj/item/storage/firstaid/fire/Initialize(mapload)
```
Poojawa: @deathride58 I can't help but feel these aren't needed,

king5327:
Hold the update, there are a handful of fixes on the way.

Poojawa:
It's... fixed already though. you can use hyposprays out of the box without needing to eject/reload first.

king5327:
The changes currently working in my tests:

- The hypospray is no longer a reagent_container (because it shouldn't hold reagents standalone).
- The vials can now be built in the autolathe, they are available in single units from the med-vendor.
- The transfer amount now depends on the vials (before you could stick small vials in the CMO spray and try and spray 60 units).
- Bluespace vials with double capacity exist, but are unobtainable.
- Some naming cleanups (Example: CMO hypospray kit becomes deluxe hypospray kit)
- Medical belt now holds the hypospray naturally instead of relying on its inheritance of the regular CMO hypospray
- Assorted minor improvements, possibly might even be able to replace the old Hypo in upstream /tg/
- No, the CMO's can't pierce covered body parts anymore (because being able to penetrate biosuit would be OP)
- Instead of the CMO's pack getting the same chems in bigger amounts, have it have exotic chems instead.
- Combat Hypo mk.II can be tacreloaded, dropping the old vial on the floor when the new one's inserted.
- Emagging a hypospray removes its delay and makes it tacreloadable. No capacity increase. (Maybe emagging it shouldn't make it tacreloadable)
- Icon change for the combat spray.
- Alt and attack_self have been swapped. Alt-click now changes the injection amount. attack_self pops the vial out.
- Mode change has been streamlined and automatically swaps modes now, instead of having a popup after already having to right click it.
- User can instantly spray themselves.

Suggestions
- Having varied delays depending on spray versus inject mode.

Also, obligatory complaint about unexpected reagent colour change.

One last thing: The only edit made outside of modular_citadel was the one about the med-belt, but it shouldn't cause any conflicts ever in the future.

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -931,7 +931,7 @@
 	id = "bicaridine"
 	description = "Restores bruising. Overdose causes it instead."
 	reagent_state = LIQUID
-	color = "#C8A5DC"
+	color = "#BF0000"
```
king5327, line 934: So, why are the colour changes necessary? The outline on the vial should cover the easy IDing.

Poojawa:
I don't agree with giving this any emag functions. 

the 'tac-reload' efforts sound fine for the combat version, but part of the chore and self imposed balance intended is for people to have to inventory juggle a bit so you can't just tag yourself with a super heal mix and not have any problem in dosing up with 200+u of chems. 

There's a reason patches don't act instantly anymore. which is why I didn't want them to instantly apply to the user either.

Hyposprays have the potential of being *very* powerful, which is why I wanted to get ahead of the curve and make sure they didn't have crazy amounts of runaway powergamer function. delays delays delays. 

CMO equipment was given a bonus because you can upgrade syringes to ignore thickmaterial flags, plus it's more of a convenience factor. The combat hypo also ignores this flag because nukeops and ERT aren't gunna bother shrugging out of armor to get meds. That and heads of staff gear should always be markedly better than anyone else's rather than just a very slight upgrade, s'why they're of use for theft objectives. 

The chemical adjustment in the CMO's vials is okay as well. and will serve a use for when I overhaul how chemicals work. <3

I am quite fine with the mode switch streamline, I'll admit I was more focused on getting the thing working than the little things like that. it being swapped to its own delivery system instead of a container in itself is also good. 

However, people will need to eject vials and adjust them in their hands now, which satisfies my want to have people careful with their dosage amounts.

This is actually closer to the idea I was intended, but had assumed I needed to work within the system instead of changing the way I thought about the situation. Well, more so than I had already. :P

king5327:
I disagree with the CMO having thickmaterial bypass, but I overlooked that the combat one should have. The CMO does not operate in life-or-death scenarios.

The spray self delay makes sense, since the hypo's main function is to inject into the bloodstream through the skin. Having it act as a patch would mean having to pass it over an area of the body rather than in just one tiny spot.

As for the chemicals in the CMO's kit, I also wanted to include synthflesh (making the CMO have to swap between inject and spray) but that would mean losing dexalin. Not a major issue since the deluxe mix handles it, but tricord and saline-glucose already heal those damage types anyway.

As for the dosage changing needing an unload - that's not true. Alt-click will let you change dosage without popping it out, but the effect is still disruptive on the keyboard.

Lastly, I don't believe the chem dispenser should be able to make vials.

edit: With penetration back into the question, I'm removing the emag's quickload enable and replacing it with penetration.

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
CameronWoof: Trekchems are baby chems and could use some visual training wheels to help with learning.

Poojawa:
The inject self should still have a delay as well, as it's one of those things that would end up abused. pull from belt, click click yourself, healing an absurd amount almost right off. 

I'd like to hear further feedback as to why vials shouldn't be in the chem master, when it can already produce bottles, patches, and pills.

I also would very much not persist in having emag functions on hyposprays. Consider the reasoning also being that emags are a limited resource, and a function like that is only allowed for people who can actually buy one at all, which is solely traitors.

CameronWoof:
Controls could use some work.

Click with empty hand - eject vial
Attack_self - set injection amount

... and then have it select spray or inject automatically. Which is less fun, but better than annoying controls.

Poojawa:
Yeah there's tweaks in terms of how it's used, the attack self is a hardcoded response thanks to /tg/ coding

king5327:
My changes bring it in line with the same paradigm as most unloadables. The HmkII is not a heavy gun, you shouldn't need both hands to unload it.

Vials should be in the autolathe, not the chemmaster, because of their use. Bottles are useless outside of syringe, splash or drink use - which most reagent_containers already have. However, bottles don't splash on throw. Vials, on the other hand, can be used to very quickly inject large amounts of preloaded chems in the same time it would take for 5 units of syringe.

@CameronWoof Spray and inject are equivalent to patches and syringes respectively. So, for example, spraying synth will heal brute/burn, while injecting salglu will raise the blood level.

Poojawa:
to be honest, their only use is in the hypospray, so it would make sense to put them where chemistry can make more at will. :P

silicons (dismissed):
you don't need \the if you're doing [atom] btw

king5327:
That's not an unreasonable perspective, mine is simply that they're lower capacity, more functional beakers.
That's why there are bluespace variants in my changes, ready to be integrated with the nerds at any moment.

modular_citadel/code/modules/reagents/reagent container/hypospraymkii.dm
```diff
@@ -112,23 +112,23 @@
...
 		if(!quickload)
-			to_chat(user, "<span class='warning'>\The [src] can not hold more than one vial!</span>")
+			to_chat(user, "<span class='warning'>[src] can not hold more than one vial!</span>")
 			return FALSE
 		unload_hypo(vial, user)
 	if((istype(I, /obj/item/reagent_containers/glass/bottle/vial)))
 		var/obj/item/reagent_containers/glass/bottle/vial/V = I
 		if(!is_type_in_list(V, allowed_containers))
-			to_chat(user, "<span class='notice'>\The [src] doesn't accept this [vial].</span>")
+			to_chat(user, "<span class='notice'>[src] doesn't accept this [vial].</span>")
```
king5327: Since I can't find it anywhere, how does \the work. From what I understand, it's ignored if the word following it is a proper noun (that is, capitalized and not initiated with \improper) and allowed to go through otherwise. If improper nouns work as the most recent commit suggests, the line I'm marking would possibly read "The hypospray doesn't accept this the hypovial."

king5327: How does improper noun parsing work anyway? From what I've seen, \the and \The are used when an improper noun needs to be preceded with "the", and is only applied when the name of the object isn't capitalized (or if it must be capitalized, it isn't preceded with \improper).

Would this line possibly read as "The hypospray doesn't accept this the hypovial." ?

Poojawa: this is also why I didn't want [vial] to be named. :v

king5327:
I still disagree with adding vials to the chem dispenser, on principle. But it's not like it's that big a deal.

Poojawa:
what do you mean 'on principle' 

How many autolathes do you have access to in medbay?

What if you need a hypospray but the only one sitting out is without a vial, are you just going to go 'WELP' and die because you can't make more vials?

I literally just ran into that exact issue last night, and being able to make a new vial was fantastic.

code/modules/reagents/chemistry/machinery/chem_master.dm
```diff
@@ -307,6 +307,34 @@
...
+				V.name = trim("[name] hypovial")
+				adjust_item_drop_location(V)
+				reagents.trans_to(V, 30)
+
+			if(vol_part)
+				V = new/obj/item/reagent_containers/glass/bottle/vial/small(drop_location())
+				V.name = trim("[name] hypovial")
+				adjust_item_drop_location(V)
+				reagents.trans_to(V, vol_part)
+			. = TRUE
```
deathride58, line 336: needs CIT_CHANGE comments

code/game/objects/items/storage/belt.dm
```diff
@@ -156,11 +156,12 @@
...
 		/obj/item/reagent_containers/blood,
 		/obj/item/tank/internals/emergency_oxygen,
-		/obj/item/pinpointer/crew, 
+		/obj/item/pinpointer/crew,
 		/obj/item/gun/syringe/syndicate,
 		/obj/item/implantcase,
 		/obj/item/implant,
-		/obj/item/implanter
+		/obj/item/implanter,
+		/obj/item/hypospray
```
deathride58, line 164: // CIT CHANGE

deathride58 (changes requested):

code/modules/reagents/chemistry/machinery/chem_master.dm
Poojawa: this'll be getting a second pass upon #6045 don't worry

code/game/objects/items/storage/belt.dm
Poojawa: done. ~

deathride58 (approved):

PR #5929, opened 2018-03-12
makes traits more paranoid, also removes the double application

Poojawa:
I'm not sure I'm able to really test this all that well, but it seemed fix locally.

It shouldn't runtime anymore either.


deathride58 (approved):
Send to upstream

Poojawa:
Not a problem up there for the doubles, the runtimes i dunno about since I don't have access to that runtimer

@kevinz000 if you've seen the traits run timing lemmie know

PR #5937, opened 2018-03-13
Small combat balancing: Small crawling nerf, Esword buff

deathride58:
Title.

:cl: deathride58
balance: Crawling underneath standing spacemen now requires a 3 second delay. A notice will be displayed to everyone around you during this delay. Added due to overwhelming demand from everywhere.
balance: Eswords now only take 3.4 staminaloss per swing while activated, but now cost a lot less while disabled!
code: Laid out a little bit of groundwork for the total_mass var. This will come in handy when I get around to implementing stats and skills.
/:cl:


Useroth:
I don't believe this fixes d-eswords.

deathride58:
Double eswords costing more stamina than regular eswords is intentional, yes.

Double eswords are considerably strong compared to normal eswords due to their bullet deflection and high damage.

silicons (approved):

PR #5939, opened 2018-03-13
Removes syndicate uplink items from the autolathe

CameronWoof:
Did you know that a box of 40 riot foam darts would run a traitor two telecrystals, if they couldn't print them out en masse from an autolathe for free? 

Because they do.

:cl: CameronWoof
del: Riot foam darts from autolathe
/:cl:

PR #5947, opened 2018-03-14
Degrading Riot Darts

ZeroNetAlpha:
[Changelogs]: Each time a riot dart is shot, it will do five less stamina damage with the next shot. This is meant as an alternative to #5939 #3568 

:cl: ZeroNetAlpha
balance: Foam Riot Darts will now loose 5 stamina damage every time they are re-used until they turn into a regular foam dart.
/:cl:

[why]: I don't want riot darts to be essentially removed completely from the game, so I propose an alternative. There'll be another PR coming soon that moves ammo to the sec protolathe.


CameronWoof:
Does not solve the problem. They will just waste more station resources printing more grief darts.

ZeroNetAlpha:
> There'll be another PR coming soon that moves ammo to the sec protolathe.

Like I said in the PR post. This is a two part special.

silicons (approved):
Very copy-paste and bad implementation but whatever.

silicons:
Actually you know what I think this needs a rethink since the combat update.

PR #5894, opened 2018-03-10
Actually remove xenobio & update maps

Poojawa:
This has #5893 's changes, in addition to removing the xenobio consoles and kill chambers from roundstart. It is possible convert the rooms back if you're clever about it.

This is based upon the tentatively positive feedback I received regarding the brief time that the console removal change was actually in effect. Leading me to believe that despite certain detractor's vitriol, it wasn't actually the end of the department at all. People were having more fun interacting with slimes and being better at monkey wrangling. 

This also is a better organized one, and doesn't have missing power wires in Metastation's xenobio. 

I'd much rather this one be merged, but to appease *certain* individuals, it's open to discussion.

PR #5953, opened 2018-03-15
Makes xenobio consoles upgradeable via disks obtainable from bluespace research, makes most xenobio console abilities exclusive to disk upgrades

deathride58:
Alternative to #5894 

:cl: deathride58
add: Xenobiology consoles can now be upgraded via upgrade disks, which can be obtained via the bluespace research nodes in RnD. Once researched, these disks can be produced in the electronics section of the Science protolathe.
balance: Most xenobio console abilities now require upgrades to use. Xenobio consoles will start out with their only ability being slime scanning. Basic bluespace research will grant the ability to remotely place and recycle monkeys. Practical bluespace research will grant the ability to remotely grab and drop 1 single slime at a time. Advanced bluespace research will grant the ability to remotely feed potions to slimes, and will increase the maximum amount of slimes the console can hold to 10.
/:cl:

People seem to be more receptive to this idea than they are to Xenobio consoles as a whole being made RnD-exclusive.

Main issue with xenobio consoles is that they reduce the amount of effort, skill, and patience required to breed endgame-tier slimes, while also making the surprisingly in-depth slime relationship mechanics completely unknown to the vast majority of players. This PR offers a solution to this, letting the xenobio consoles stay as roundstart machinery, while also making their effort-reducing abilities tied to research.


JogBoba:
As someone who’s been a major proponent of the console modification, I feel that 25 slimes at once is too many, as well as ranged potions. However, making it able to have its current functionality at max tier parts would be nice, perhaps increasing to at most 10.

deathride58:
Ranged potions are a part of the vanilla TG xenobio console.

TG's limit on slimes able to be picked up via xenobio console is 5 slimes max. Main reason I'm increasing this for a max-upgraded xenobio console is so that these changes aren't pure nerfs. 25 slimes max is a little extreme, though, I'll admit. Probably gonna be nerfing it down to just 10.

ShadowDeath6:
so how much earlier are these unlocked compared to current xenobio consoles

I'm also standing by my statement that slime relationships are not useful unless you plan on doing antagonistic things with them considering the job revolves around killing slimes

deathride58:
Applied bluespace research, where current xenobio consoles are in RnD, can be rushed in less than 15 minutes with just the base passive research point generation. With this PR, the majority of the xenobio console functions are locked behind a design unlocked with the applied bluespace research node.

Advanced bluespace research can be rushed in 30 mintues with just the base passive research point generation.

ShadowDeath6:
if they're locked behind the same node that the current xenobio console is in R&D, does that not make this pr's changes functionally identical to the ones mentioned in the referenced pr?

deathride58:
Yes.

The difference between this PR and the other PR is that this PR lets the xenobio console gradually gain its abilities as nodes get researched, while the other PR makes the xenobio console something you either have or you don't. The other PR also causes a lot of confusion, since those that don't pay attention to the changelog will think the xenobio console has been completely removed from the codebase with the other PR.

Remote monkey placement/recycling, for example, is locked behind the basic bluespace research node, which can be easily grabbed near roundstart. Monkeys in particular are something that a lot of folks have gripes about since they're able to hold a fight on their own. Having monkey placement/recycling easily accessible via a cheap roundstart rush-able research node means that those who want to use the console to avoid the effort of wrestling monkeys into the pens can do so near the start of the round, while keeping slime wrangling something that has to be done manually until research has progressed to a certain point.

JogBoba:
I must have missed ranged potions being added then, as I’ve never used it nor seen it used. It seems overly power creepy. I do think it’s also important to note that mining will normally be gotten before the bluespace techs. Nerfing it to 10 will make it ok with me, provided the kill room isn’t mapped out.

Poojawa:
This is an excellent alternative solution. 

By generating community awareness, we were able to compromise on this issue in a way that serves what we should be going for and tweaks an issue for the better.

Good work on this.

My only concern is at this point is the potion application, especially when crossbreeding becomes a thing again.

JogBoba:
I do think potion application should be removed or restricted further. I also believe that the parts should each be moved up one node and a fourth/fifth node added for current/buffed functionality

Poojawa (approved):

Knightfall5:
Eh, I've changed my mind. The research required to get xenobiology going isn't terrible and the console starts mapped in, so it's not nearly as bad as the other system.

silicons (approved):

PR #5993, opened 2018-03-18
Chemistry application rework

Poojawa:
:cl: Poojawa
tweak: medicines are no longer omni-applicable excepting in certain cases. Be careful of your patients! 
tweak: pills and patches reduced to 5u, but bottles are buffed to carry considerably more
tweak: New method of chemical application, INHALE. For when you're breathing and don't have internals on.
/:cl:

Shakin' up the medical tree some. Making chemicals actually take some skill to apply, rather than just drilling 700 units of chems into your teeth.

with the new hyposprays as well, it shouldn't be too difficult to adapt to a new meta ~


GuppyLaxx:
image
????

Poojawa:
Moved the reagent folder to modular citadel so I can not have to dick about with overrides

s'why it says 52 files changed instead of like, 3

Nightspark43:
I don't like the whole, pills and patches are 5 units only, think, maybe 15-25 at least, most people don't even use bottles, at all, so nerfing pills and patches so hard likely won't go over too well, even with the new hypospray stuff.

ShadowDeath6:
I feel like the intention of this pr is just to make people use your hyposprays more, honestly

make pills and patches hold more, 5 at a time is completely worthless outside of making meth patches or other chems that are ok in low doses

I also think "considerably more" when the change is 20% is a bit misleading, but whatever

MimicFaux:
Will brute/burn kits now contain bottles of their contents to even out their cost:medicine ratio?

PR #5995, opened 2018-03-18
sabre and sheath

LetterJay:
Adds another donator item(s).

@deathride58 you may need to take a look at this since I don't know how to module left and right hand sprites.


deathride58 (changes requested):
You can modularize inhands using the lefthand_file and righthand_file vars.

Also, modularize the belt sprites

LetterJay:
The hell do you mean modularize the belt?

Plus I have no idea how the left and right hand system works aside from pulling it from the respective file, so I don't see how that'd play out either.

deathride58:
The sheathe sprites aren't making use of icon_override or alternate_worn_icon, they're just pulling from stock TG files as it is right now. It should be fairly easy to modularize them.

Inhands work exactly as the vars say. When an item's held in the left hand, it'll pull icon sprites from the lefthand_file var, and if it's in the right hand, it'll pull icon sprites from the righthand_file var.

LetterJay:
@deathride58

deathride58 (approved):

PR #6085, opened 2018-03-22
Makes blood cult combat rework compatible.

GrayRachnid:
[Changelogs]: # Changes the amount of stuns blood cult has each time they are carved onto the body, each use is a 1.6 second knockdown and gives out 40 points of stamina damage. One or two uses will knockdown and slow but your target would still be able to crawl away. Three uses will knock them down longer and staminacrit them briefly. Four uses is a guaranteed stun for a while, note that each use will deal 3 points of brute damage to the arm you are using the stun on.

:cl:
balance: Blood cult stun now has 4 stun charges instead of one. Three uses will give you 120 stamina damage and a 4.8 second knockdown and 9 point of damage. Four uses will give you 160 stamina damage and a 6.4 second knockdown and 12 points of damage. 

/:cl:

[why]: # Because they can't stun anyone right now.

### DIFF
--- a/code/modules/antagonists/cult/blood_magic.dm
+++ b/code/modules/antagonists/cult/blood_magic.dm
@@ -139,9 +139,8 @@
 	name = "Stun"
 	desc = "A potent spell that will stun and mute victims upon contact."
 	button_icon_state = "hand"
-	charges = 4
 	magic_path = "/obj/item/melee/blood_magic/stun"
-	health_cost = 3
+	health_cost = 10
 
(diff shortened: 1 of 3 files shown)


deathride58 (changes requested):
Modularize by overriding the datum's values in a file in modular_cit

deathride58 (changes requested):
Just needs to be added to the DME now. Also you accidentally reverted the health_cost back to 10

deathride58 (approved):

PR #6088, opened 2018-03-23
Makes sure the server can never ever crash ever again

LetterJay:
:cl: Jay
tweak: Reduced autocall to two hours.
/:cl:

### DIFF
--- a/code/controllers/subsystem/shuttle.dm
+++ b/code/controllers/subsystem/shuttle.dm
@@ -56,7 +56,7 @@ SUBSYSTEM_DEF(shuttle)
 
 	var/lockdown = FALSE	//disallow transit after nuke goes off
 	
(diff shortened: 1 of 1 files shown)


deathride58 (dismissed):
Speedmerge this,

deathride58 (approved):

JogBoba:
Please just test merge this

silicons:
techwebs is currently balanced for 3 hours.

king5327:
I mean, it's not like the server sometimes OOMs 1h into a round after a full restart.

MimicFaux:
This fixes the symptom, not the cause

king5327:
I don't know anything about the server configuration, but my first instinct would be to increase swap space (since memory leaks can be swapped out and stay swapped out). If it persists then I'd resort to such drastic measures.

ariever:
Even a fix that's supposed to make sure the server never breaks again is no match for **DELTA**.

PR #6115, opened 2018-03-26
Metabreaks event reports

Poojawa:
Several events now have a 50% chance of being a 'classified' message. This is to break people from being able to perfectly guess what events are and to stop people from gearing up in response. 

Check those papers!

### DIFF
--- a/code/modules/events/anomaly_bluespace.dm
+++ b/code/modules/events/anomaly_bluespace.dm
@@ -10,7 +10,7 @@
 
 
 /datum/round_event/anomaly/anomaly_bluespace/announce(fake)
-	if(prob(50))
+	if(prob(90))
 		priority_announce("Unstable bluespace anomaly detected on long range scanners. Expected location: [impact_area.name].", "Anomaly Alert")
 	else
 		priority_announce("A report has been downloaded and printed out at all communications consoles.", "Incoming Classified Message", 'sound/ai/commandreport.ogg') // CITADEL EDIT metabreak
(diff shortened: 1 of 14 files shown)


Nightspark43:
I'll say here what I said in the Discord, this assumes that there are either active heads, heads who give a shit, or a bridge that people can look into, so if the round ends up a pubby map with no heads, it becomes a 50% chance for any event to be completely unknowable without breaking in.

At the very least the anomalies should not be in this, due to how common and destructive they can be if not found.

JogBoba:
I’ll also repeat what I said, cargo events being logged as reports doesn’t make sense, as it’s centcom contacting cargo to tell them to send off the shuttle so it’s a non-issue. I’m fine with the “it’s getting hijacked by so and so” being printed instead, but for most of them it doesn’t make sense, why would pizza be a classified report for example?

Poojawa:
we have rounds of 20-70 players.

There is always at least one staff head, or barring that, one silicon unit. You can also walk by the bridge and have a looksee yourself. 

Cargo events add additional non-hostile report chances. it's to further clog people's assuming what a report ends up being

cebutris:
I have my doubts about this, most of which have already been expressed. But it's not going to be the end of the world if this is put in place, I'd be willing to see it in play and suggest changes/removals from that

MimicFaux:
Metabreaking is (usually) always a good thing, and rolling the dice for a classified report seems like a good way to do it.

ghost:
dang c*ders

Brimcon:
I'd say remove the dangerous ones, like Bluespace, Hyperflux and the Pyro one at the very least from this 50% chance.

deathride58 (approved):

PR #6132, opened 2018-03-27
Adds Particle Blasters to armories

Michiyamenotehifunana:
:cl: Toriate
add: Centcom has just approved a procurement order for new pump action particle blasters. Armories will be stocked with them.
/:cl:

This PR maps in pump action disablers because they are straight out adminspawn only right now. Two are placed where DRAGnets spawn, and if the map doesn't have them, one is placed on the ion rifle/temp gun/reflective vest rack.


deathride58 (approved):

improvedname:
I love how nobody can argue with this being an thing because the dragnet does the same thing but better, i would love if this replaced the dragnets

PR #6078, opened 2018-03-22
MORE LOADOUT STUFF

ghost:
[why]: # muh customization
also i did all this on the browser so i never tested if i accidentally broke everything, will properly do stuff later when i get home
:cl:
add: Added more loadout options
add: Added a cosmetic version of the goliath cloak
/:cl:

PR #6174, opened 2018-03-29
(Partially) Globalizes Polychromic Clothing Code

Michiyamenotehifunana:
:cl: Toriate
add: Added Polychromic Shorts and kilts to loadout
code: changed the way polychromic clothes works
refactor: refactored polychromic things such that the vars and (some) procs are globalized
/:cl:

This PR finishes the half-assed job that #6078 couldn't bother with, while at the same time globalizing polychromic vars and procs to be applicable to all clothing items in general, however a workaround is required due to how upstream handles individual categories of clothing, i.e. having snowflake worn_overlays() procs that override everything else.


deathride58 (approved):
You could probably make the roundstart polychromic clothing a donor reward. Discuss it with Jay

Code's good though :+1:

PR #6228, opened 2018-04-01
Nerfs beam rifle effectiveness vs blob

silicons:
yawn yawn this was supposed to be a "we got adv tech lets powergame because we're lazy" not a "we can't fight a blob without it" and if it's the latter people need to change up their tactics.
:cl:
experimental: beam rifle damage vs blobs nerfed by 80%.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/misc/beam_rifle.dm
+++ b/code/modules/projectiles/guns/misc/beam_rifle.dm
@@ -509,6 +509,8 @@
 		return 0.4
 	if(istype(target, /obj/structure/window))
 		return 0.5
+	if(istype(target, /obj/structure/blob))
+		return 0.2
 	return 1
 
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/misc/beam_rifle.dm
```diff
@@ -509,6 +509,8 @@
 		return 0.4
 	if(istype(target, /obj/structure/window))
 		return 0.5
+	if(istype(target, /obj/structure/blob))
```
deathride58: if you aren't sending this to upstream, add
`//CIT CHANGE - removes an easy ggnore method of dealing with blobs`
here

deathride58 (changes requested):

deathride58 (approved):

PR #6252, opened 2018-04-03
looc now obeys ooc bans

silicons:


deathride58 (approved):
Deserving of a speedmerge.

PR #6254, opened 2018-04-03
testmerge only, stability not guaranteed nor tested - robustin overlay patch manual mirror

silicons:


jakeramsay007:
Has bugs related to overlay handling in other parts of the code, would not recommend merging nor testmerging till robustin does some more work on this.

KasparoVy:
Test-merging this on a local branch reproduces the issue on the live server where character setup preview icons break (don't render) after the subsystems initialize.
I figured it's best to document what the culprit is
Robustin's doing some voodoo

PR #6273, opened 2018-04-04
[MIRROR] Shuttles have additional baseturfs, and other minor baseturf changes

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/36388
--------------------
:cl: ninjanomnom
fix: Shuttles have proper baseturfs now.
fix: Mineral walls properly use their baseturfs when destroyed/drilled.
add: A new engineering goggle mode allows you to see the shuttle area you're standing in.
admin: Buildmode works a bit better with baseturfs now and can properly only remove the top layer of turfs when editing. Note that as a result the order you place turfs is important and a wall placed on space means when the wall is removed there will be space underneath.
/:cl:

This adds an area hook for turfs getting the PlaceOnTop proc called so shuttle areas are able to intercept it and add a special turf used for determining where the bottom of the shuttle is. I'm not too certain on this being the best way to handle it but the alternatives restrict the kind of turfs that are allowed to be on shuttles which I didn't want to do as it was part of the whole intent of the baseturf change to allow any turf type on shuttles.

The alternative to the area hook would require a tool which attaches turfs to a shuttle which is why there's a shuttle area viewing tool in this pr. I considered doing it this way first but decided it was too clunky.

**In addition to other changes:**

Tiled/Reinforced floors have proper baseturfs now.

A large amount of the remaining ChangeTurf calls were swapped over to various baseturf helpers so, for example, mining doesn't create a lavaland turf if somehow the mineral wall is on plating.

Escape pods, the aux base, and syndicate drop pod have been converted over to shuttle templates as well.

The error when the timid var is set incorrectly has been changed to a stack_trace.

The change turf file has been made all lowercase as per standard.

fixes #33984


ninjanomnom:
Any and all shuttles you have will have to be converted into templates for this to work. If you don't they will not move.

PR #6309, opened 2018-04-08
Turians are here!

DravenPeace1:
:cl: DravenPeace
add: Add a new specie! The Turians!
add: Add more ear, tail and body markings! There is now jellyfish bell, orca tail and body mark, axolotl tail and ears!
/:cl:


DravenPeace1:
Oh god
Why

PR #6330, opened 2018-04-09
They are here!

DravenPeace1:
:cl: optional name here
add: Added a new specie! The Turians!
add: add new body parts, jellyfish, axolotl and orca ears, tail and body marking!
/:cl:


jakeramsay007 (changes requested):
Don't make a whole new folder titled Turian for these, use modular_citadel

DravenPeace1:
I did moved them

PR #6334, opened 2018-04-10
[MIRROR] Use a rust DLL for logging

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/36858
--------------------
By moving our logging to a DLL we see a drop in CPU/real time of 2-3 orders of magnitude. This is due to BYOND opening and closing file handles on every write, causing incredible amounts of unneeded overhead. The logging library also handles timestamps for us, further increasing performance gains.

This library will also allow for further offloading in the future, such as completely replacing file2text() and friends.

A pre-compiled DLL is bundled, but Linux users will have to compile manually. Directions can be found at the [rust-g repo](https://github.com/tgstation/rust-g).

Log output is enhanced with millisecond time stamps:

```
[2018-04-01 15:56:23.522] blah blah blah
```

This includes runtimes as well, which benefit from the same timestamp improvements and no longer have hacky splitting code to add their own timestamps.

Log shutdown is handled in a dedicated proc called as late as possible, as rust-g integration expands this will be factored out into a generic native code shutdown proc.

Finally, here are some benchmarks:

image

:cl: neersighted
code: Reduced lag by speeding up logs 10000%
code: Logs now include millisecond-level timestamps
/:cl:


deathride58 (approved):
This is a major, ***MAJOR*** performance improvement. Since I don't have remote access, this will have to be set up and compiled by Jay.

PR #6355, opened 2018-04-11
[DNM] Reverts randomized airlock wires

LetterJay:
One of the topics people keep bringing up. Probably one of the very very few items I disagree with as far as changes go, but I'll leave this for further discussion in the discord.


Poojawa:
You don't even play, why would you disagree with this anyway.

PR #6376, opened 2018-04-11
Replaces random airlock wires with departmental airlock wires

deathride58:
Title. This idea is much more well-received than the current system of making airlock wires 100% random.

:cl: deathride58
tweak: Airlocks are no longer 100% random. Instead, their wires are randomized depending on their department.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -99,7 +99,7 @@
 
 /obj/machinery/door/airlock/Initialize()
 	. = ..()
-	wires = new /datum/wires/airlock(src)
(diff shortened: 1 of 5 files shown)


LetterJay:
This retains the security feature while not making it a pain in the ass for engineers to quickly resolve airlock issues, so I approve of this simply for that.

LetterJay (approved):

RogueSteampunker:
Definitely, DEFINITELY a good compromise. It makes it so a multitool isn't a skeleton key to all the doors, but it also doesn't make insuls a prerequisite to actually get through a single door.

silicons (approved):

PR #6630, opened 2018-05-03
Reduces power gain from Tesla coils

Poojawa:
Let's make more elaborate coils the norm, rather than being able to really profit from the bare minimum.

### DIFF
--- a/code/game/objects/structures/grille.dm
+++ b/code/game/objects/structures/grille.dm
@@ -267,7 +267,8 @@
 			var/obj/structure/cable/C = T.get_cable_node()
 			if(C)
 				playsound(src, 'sound/magic/lightningshock.ogg', 100, 1, extrarange = 5)
-				tesla_zap(src, 3, C.powernet.avail * 0.01) //Zap for 1/100 of the amount of power. At a million watts in the grid, it will be as powerful as a tesla revolver shot.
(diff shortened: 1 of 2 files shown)


FalcoGer:
I think the station should be able to power all of it's system with the engine it is given, as it had been able to with any other engine so far.
so with such a change, clearly extensive testing is required: with all APCs empty, the primary SMES should still charge while supplying power to them all.
At which point infinite power doesn't add all that much more convenience to it all.
On the other hand, not having all stations be able to be powered will leave the crew in a bind with low/non engineer count crews.
The problem with the main engine is and has always been not it's power output, but it's danger to the station if containment fails. Changing that aspect of the game seems a little over the top.
Conclusion:
If powering all systems is desired, the required testing is a waste of time and it's easier to just leave it as it is instead of introducing potential problems.
If not powering all systems is desired by nerfing the engine, then this is a show stopper for citadel in unbalanced rounds or people hosting citadel code with low people count.
I made a case.

Is it not customary to REVIEW changes by another party and discuss them before they are merged?

deathride58:
This PR was closed, not merged.

We have our github set up to require that PRs be reviewed before they're able to be merged at all.

A proper fix for the corona analyzer cheese will be coming soon.

PR #6682, opened 2018-05-09
[MIRROR] Update DMAPI to version 3.2.1.0

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/37429
--------------------
This is primarily a compatibility layer that enables both forward and backward compatibility for all past and future APIs based on the detected running TGS version. It also bundles all it's includes to make future upgrades not have to modify the dme, heck, could even use a submodule if it wanted (not happening). No other changes necessary.

There's an upcoming event system and new chat management functions. Check them out here: https://github.com/tgstation/tgstation-server/blob/303448457e1529c900fa183a8b3c7f2e5459acf1/DMAPI/tgs.dm

Also added /datum/proc/CanProcCall()


LetterJay:
Wew. This is gonna need to be mirrored to the RP build.

deathride58:
@LetterJay Two steps ahead of you, the TGS3 port on the RP server was ported directly from that branch.

deathride58 (approved):

PR #6753, opened 2018-05-15
Makes the Panic Bunker message a bit more helpful

LetterJay:
This should make shit less confusing for new joins


deathride58 (dismissed):
Might wanna add a `//CIT CHANGE - makes the panic bunker disconnect message point to the discord` to the end of the modified disconnect reason

deathride58 (approved):

PR #6755, opened 2018-05-15
[!EXPERIMENTAL!] [NON-MODULAR] Makes atmos process 2.5x faster, and makes it a lot more violent

deathride58:
Title. This change isn't modular yet. It's purely experimental. For testmerge only right now, to collect feedback and make sure the server won't explode from it.

:cl: deathride58
add: Atmos now has the ability to throw people and objects around (ported from para)
tweak: Humans now have a pressure resistance of 8, reduced from the original value of 25.
tweak: Atmos now processes once every 2 ticks instead of once per 5 ticks.
tweak: The minimum amount of pressure for objects to move as a result of atmos has been reduced. The original threshold is now used to determine if an object can be thrown.
/:cl:

### DIFF
--- a/code/controllers/subsystem/air.dm
+++ b/code/controllers/subsystem/air.dm
@@ -10,7 +10,7 @@ SUBSYSTEM_DEF(air)
 	name = "Atmospherics"
 	init_order = INIT_ORDER_AIR
 	priority = FIRE_PRIORITY_AIR
-	wait = 5
+	wait = 2
 	flags = SS_BACKGROUND
 	runlevels = RUNLEVEL_GAME | RUNLEVEL_POSTGAME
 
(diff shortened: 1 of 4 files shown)


lightcat:
I am snowdragon01 and I approve of this change

deathride58:
This has been testmerged for long enough.

The server surprisingly handled the faster atmos process rate insanely well, though the server did end up slowing the atmos process rate down automatically during stressful situations.

The pressure resistance changes and pressure throwing port made hull breaches dangerous for the first time in forever, though my later nerfs to pressure movement ended up being a little too strong. If this PR is ever revisited, it'd be best to adopt values that are somewhere in between the initial pressure interaction values.

I'll absolutely be revisiting this PR in the future, but for now, I've gotta get other stuff off my plate.

PR #6808, opened 2018-05-20
[MIRROR] [ready] cargo bounties

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/37833
--------------------
:cl: Pubby
add: Cargo bounties from Nanotrasen. Earn cargo points by completing fetch quests.
remove: Several cargo exports have been reduced or merged into cargo bounties.  
/:cl:

**what:**

Each round cargo receives a randomly-generated list of bounties, which are little fetch quests that involve shipping items to CentCom. When they complete a bounty, they get cargo points.

These run alongside exports, but due to the added cargo points bounties provide, several exports have had their prices reduced or turned into bounties. Exports are still profitable, but you'll want to do bounties as well.

image

**why:**

Exports have problems. They're too hard to balance, and they're not comprehensive. The only way to learn them is to read the source, and those that do go on to abuse the system and earn way more credits than intended. It's not surprising how many times exports have been nerfed or removed.

Cargo itself has problems. The job doesn't have much depth, and feels very samey round to round. The job is supposed to involve buying crates for others, but really it's all about buying crates for yourself because other parties don't have any negotiating power. 

That brings us to bounties:
- easier to balance because they're finite
- can be applied to things like food without having to define comprehensive prices for every food type
- randomly generated nature means each round is a little different and you can't use the same strategy every round
- they have a UI which means you don't have to source dive to find prices
-  if other departments can help complete a bounty, other departments have negotiating power and cargo has more interaction


Cyprex:
While the bounties are a great feature, Several unreasonable nerfs come along with this PR(for example adamantine for 7500->500, most other material prices halved, plasma 300->200). On extended rounds you also tend to run out of (doable) bounties.

PR #6860, opened 2018-05-23
fixes flash icons

Hatterhat:
because people keep whining about it and because kev wanted me to do it for him with a webeditor pr

[Changelogs]:
:cl: hatterhat after being yelled at by kevinz
fix: deletes an entire line that might fix flash icons idk blame him if it gets worse
/:cl:

[why]: kev told me to do it
image


jakeramsay007 (approved):
Looks good, should work.

silicons (approved):

PR #6943, opened 2018-05-28
Metastation Cleanup

Cw3040:
Making it so that MetaStation doesn't have null objects (the Stock
Exchange computer previously located at
/obj/machinery/computer/stockexchange) and is usuable/editable with FastDMM.


Cw3040:
I have been informed that the directory is NOT cit_map_files and that i was just using the wrong dmm. Closing PR now.

PR #6947, opened 2018-05-29
obsolete

silicons:
As much as I love abusing this to be nearly unstoppable this has gone on way too far.


silicons:
travis error is a map error, unrelated to pr.

PR #7018, opened 2018-06-05
High-Tier Map Memes

Cw3040:
Making the map not the same as /tg/'s while retaining functionality of /tg/'s shit like VR.


Cw3040:
maintbar
F on the wall stands for "False"

deathride58 (changes requested):
Remove the drone dispenser you added to meta. Drones are an absolute pain in the ass to deal with from an administrative standpoint 90% of the time; they're admin-spawn only now for good reason.

Also, try remaking your changes to Pubby. Travis is being weird about it, might be due to the finite symbol bug.

Cw3040:
"Finite symbol bug" explain pls

Cw3040:
I'm legit thinking that Travis just fails on the last commit of a series

Cw3040:
Travis has failed only when BUILD_TESTING=true and it seemingly fails on a random map tile.

Cw3040:
alright i guess it's just fucked now
time to un-fuck it

PR #7023, opened 2018-06-06
Telebatons/policebatons buffed to be a viable stun weapon (but not as good as stunbatons/prods), secways nerfed to non-sprint movespeed with sprint buffer (but this time they're actually responsive instead of being laggy and the buffer is weaker)

silicons:


deathride58 (changes requested):
Put the files in the modular_citadel/code/ folder or so help me god

deathride58 (dismissed):
The stamina damage on batons can be toned back juuuust a smidge

deathride58 (approved):
*headpats*

PR #7036, opened 2018-06-08
Cw's dumb Map Edits v.4

Cw3040:
Let's hope Travis doesn't get mad at this one, might be a problem with FastDMM. As usual, commits have their changelogs.


deathride58 (approved):
After 54432340698 PRs, hopefully it would have been worth the wait

PR #7083, opened 2018-06-12
nerfs techwebs passive generation, pushes back full max time by 15 minutes

silicons:
would rather not but lmao 6 minute autocloning


deathride58 (changes requested):
The old value, before the modular override broke, was 20. Please use the old value. Also, add `//CITADEL EDIT - reduces the rnd point gain rate` and remove the modular override

Wilchenx:
>6 minute autocloning
>4 minute bohbombs

why not just go back to the old reliable science?

deathride58 (approved):

PR #7086, opened 2018-06-12
Corrects climax cooldown delay message, makes relieve message pronoun…

ShizCalev:
…sensitive

The cooldown message was dividing the delay by 20ds, meaning it was would tell the user to wait 4 seconds instead of the proper 8 seconds. Also switched it to use DisplayTimeText() so it will no longer say `you must wait 1 seconds`


deathride58 (approved):

GuppyLaxx:
why are you here get out

silicons (approved):

PR #7182, opened 2018-06-19
removes unnecessary <br> on README.md

LetterN:
[why]: <br> not need, this isn't html ree


LetterN:
actually, i'l just do this on the tg repo

PR #7332, opened 2018-07-04
[MIRROR] Refactors component signals to properly register for specific signals

CitadelStationBot:
Original PR: https://github.com/tgstation/tgstation/pull/38798
--------------------
This changes how signals work so that datums keep track of what components registered for what signals. This means it's easy to tell when a signal would have no one listening and to skip the whole process.

Additionally this means components can register for signals on not just their own parent, making the signal redirect component obsolete for use inside other components. It is still useful for non components which want to listen to a signal. Uses in existing components have been converted.

Global signals are now just signals registered on `SSdcs`.

The amount of `_SendSignal()` calls during init have shot way down as a result of all this.
Before: 180046
After: 27

This whole refactor is basically a cost trade of memory for improved performance and functionality. We do get slightly better memory cost for components which would potentially use many redirect components but we didn't have too many of those.


ninjanomnom:
If you've added or modified any component this won't necessarily conflict with that directly but will break nonetheless. `RegisterSignal()` needs a new arg now for it to work properly.

PR #7364, opened 2018-07-12
flightsuit speed rebalancing

silicons:
:cl:
experimental: flightsuit no longer gives innate jetpack speed in nograv. it already has its own movement processing.
experimenta: flightsuit braking mode is no longer as crappy and might actually be useful now
/:cl:
hopefully cuts down on the unlimited space saxxing.


Cw3040 (changes requested):
Make the stabilizers not crash you into walls on 30 output and we'll be kosher

deathride58 (approved):

PR #7379, opened 2018-07-17
Removes Voodoo Doll ability to see target

Cw3040:
-why would voodoo dolls allow the user to see their victim


silicons:
go nerf it some other way

PR #7380, opened 2018-07-17
Ports Oraclestation's genpop, adds it to Boxstation

deathride58:
Title. Oracle's initial genpop PR can be found at https://github.com/OracleStation/OracleStation/pull/419 for reference. Genpop in this PR is more or less functionally exactly the same as it currently is in Oraclestation.

For those that don't know what genpop is: Genpop basically shifts focus away from sec's dunk tanks, and towards the prison. Instead of standard procedure being to put prisoners into the dunk tanks, genpop makes shoving prisoners together into the prison wing standard procedure. It also reduces the work security has to do, as prisoners will be able to leave the brig on their own the moment their sentence finishes up, without any action required on security's part. Just toss the prisoner into genpop, and go.

Quick guide to placing someone into genpop, for those too lazy to read oracle's PR:
Step 1: Drag your prisoner onto an open prisoner locker
Step 2: Fullstrip your prisoner. Don't leave a single thing on their body aside from handcuffs.
Step 3: Drag your prisoner off of the locker and click it to enter details into the system, which will spawn an ID that is now actively counting down
Step 4: Put the prisoner uniform that just spawned onto the prisoner
Step 5: Give the prisoner the ID
Step 6: Take your cuffs
Step 7: Push the prisoner through the entrance turnstile
~~Step 8: Walk away and completely forget about the prisoner, and then arrest them after their sentence is up because you thought they escaped before their sentence is up~~

:cl: deathride58
add: Added genpop! It's currently only available on Boxstation, but genpop support can be expanded to all the other maps pretty easily!
/:cl:


CameronWoof:
And... what do you do when the prisoners get into fights? The reason the dunk tanks are so good is that they avoid security having to deal with violent prisoners mauling each other. What happens if you need to add time? To flash the prisoners? This looks completely horrible for Security.

EDIT: Just saw the flasher in the middle of the room. Still not a fan of the fact that it doesn't look like you can tase/disable through what used to be nice windows. Are turnstiles airtight?

deathride58:
If you need to add time, you can do so by clicking the prisoner's assigned locker. The prisoner's ID will update remotely, allowing you to reduce or increase their sentence. You can also use this to change a prisoner's status from temporary prisoner to permanent prisoner if you so desire, which is a good way to nullify prisoner IDs that might've been stolen without having to risk entering the prison, which should hopefully deter prisoners from acting out of order. There's really nothing stopping prisoners from mugging each other for their IDs in an attempt to get out faster, but security does have the tools necessary to keep rambunctious prisoners inside. The turnstiles can be shot through just fine, and items can be thrown through them, so crowd control is still entirely plausible from a distance. Turnstiles are not airtight, but if that proves to be a problem, it should be pretty easy to add firelocks underneath turnstiles on the map.

The dunk tanks still exist for violent prisoners that really need to be kept in a cage, and are renamed from cells to solitary dunk tanks to reflect their new status.

Dankseed:
Let's try it out before judging too much.

deathride58:
This is made obsolete by #9072

PR #7384, opened 2018-07-20
Ports Construction Hardsuits back from 5th port. Not mapped yet.

RoyalKurwa:
Construction Hardsuits are bright orange hardsuits. The difference being from engineering ones is that they're supposed to be used for quick fixes or general station work, could also be surplus incase more than 2 basic engineers sign up. Quicker movement speed than the regular Engisuit but not gamebreakingly so (hopefully).

Suits aren't mapped in yet because I'm unsure of how this works but there would be two in the secure tech storage as one of the main issues of them on 5th was the ease of access into the construction area.

:cl: RoyalKurwa
add: Added Construction Hardsuits
/:cl:


code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -202,6 +202,25 @@
 	max_heat_protection_temperature = FIRE_IMMUNITY_SUIT_MAX_TEMP_PROTECT
 	helmettype = /obj/item/clothing/head/helmet/space/hardsuit/engine/atmos
 
+	//Construction hardsuit
+/obj/item/clothing/head/helmet/space/hardsuit/engine/construct
+	name = "construction hardsuit helmet"
+	desc = "A special helmet designed for work in a hazardous, low-pressure environment. Much of the protective qualities have been given up for superior movement."
+	icon_state = "hardsuit0-construction"
+	item_state = "con_helm"
+	armor = list(melee = 5, bullet = 5, laser = 5, energy = 5, bomb = 5, bio = 100, rad = 0)
```
deathride58: wrap all of these var names in quotes. not doing so will result in the armor not applying at all

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -202,6 +202,25 @@
...
+	item_color = "construction"
+	resistance_flags = FIRE_PROOF
+
+/obj/item/clothing/suit/space/hardsuit/engine/construct
+	name = "construction hardsuit"
+	desc = "A special hardsuit designed for work in a hazardous, low-pressure environment. Much of the protective qualities have been given up for superior movement."
+	icon_state = "hardsuit-construction"
+	item_state = "con_hardsuit"
+	armor = list(melee = 5, bullet = 5, laser = 5, energy = 5, bomb = 5, bio = 100, rad = 0)
```
deathride58: wrap all of these var names in quotes. not doing so will result in the armor not applying at all

deathride58 (changes requested):

improvedname:
If theres no way to get this why add them sprites look meh
But without aan purpose code bloat

RoyalKurwa:
Still trying to figure out on how to add things, I'll ask someone for some help doing it.

Poojawa:
this was basically a way for assistants to get more hardsuits. since engineering almost never needed them.

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -202,6 +202,25 @@
...
+	resistance_flags = FIRE_PROOF
+
+/obj/item/clothing/suit/space/hardsuit/engine/construct
+	name = "construction hardsuit"
+	desc = "A special hardsuit designed for work in a hazardous, low-pressure environment. Much of the protective qualities have been given up for superior movement."
+	icon_state = "hardsuit-construction"
+	item_state = "con_hardsuit"
+	armor = list("melee" = 5, "bullet" = 5, "laser" = 5, "energy" = 5, "bomb" = 5, "bio" = 100, "rad" = 0)
+	helmettype = /obj/item/clothing/head/helmet/space/hardsuit/engine/construct
+	slowdown = 0.75
```
deathride58, line 222: Outside of very specific edge cases, this only differs from slowdown=1 when the server's tickrate is set to 40 ticks per second. The live server's tickrate is 20.

deathride58 (approved):
I can see these being pretty alright for adminbus, or available as surplus hardsuits that only spawn when more than a certain amount of engineers are active, similar to how extra sec lockers spawn in when there's enough sec officers latejoining.

deathride58:
Has conflicts with hard sync

nik707:
If this is fixed and re-opened, I'd probably approve it, Until then...

PR #7410, opened 2018-07-26
fixes the reason I and a lot of you lock up for more than half a minute when trying to join the server

silicons:
:clap:
see "asset cache css duplication" on tg git
credits to oranges and anturke and whoever else found it


deathride58 (approved):
Testmerge didn't break the server/10

PR #7429, opened 2018-08-02
family heirloom toolboxes are now bulky items but take less stamina to swing, have 2 extra damage, as well as 20 extra damage to structures/objects

silicons:
greytide worldwide


Cw3040:
but
how will they fit into backpack then?
they're supposed to be heirloom box for a reason

deathride58 (approved):
Pretty strong for what's supposed to be a negative trait, but I'm interested in seeing how it'll work in practice. If people start breaking airlocks and shit with heirloom toolboxes, you're fired.

deathride58 (approved):

PR #7437, opened 2018-08-06
changes to pubby secpod/bridge

nik707:
[Changelogs]: Changes a few doors area types on command to make fixing them easier without an AI during greytide events.
Also adds an escape pod to security.

:cl: nicc
tweak: area alterations
add: escape pod
/:cl:

[why]: Bridge greytide opens a few airlocks that can't actually be worked by the command door remotes, which seems like a bit of an oversight. Changing the areas of said doors, which are the AI sat access doors, to AI sat area, means bridge won't get vented, and command staff would be struggling to hack doors closed when the command remote refuses to work. The alternative was changing the doors access levels, or the remotes levels, which seemed more game changing. So I opted for this change instead.
Pubby seems to be the only map lacking alternate escape, aside from the iffy monastery pod which can be in one of two places on opposite ends of the station. This should add an alternate escape for crafty antags, or good sec/command players.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -202,6 +202,3 @@ tools/MapAtmosFixer/MapAtmosFixer/bin/*
 
 #Jukebox audio files
 /config/jukebox_music/**/*
-*.dmm
-_maps/nanostation.dm
-_maps/nanotstation.json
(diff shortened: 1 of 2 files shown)


silicons:
your shit broke, pubby monastary has an escape pod instead of a shuttle now, wtf

deathride58 (approved):
yOuR ReMovAl oF oNe sIngLe pLAnT tHrOws oFF thE enTIre bAlAncE of tHe gAmE

PR #7440, opened 2018-08-08
fixes my inability to make a singularity cannon

silicons:


deathride58:
are you absolutely, 100% certain this won't cause weird shit? might be wise to add checks to make sure it's adminbus before returning true

PR #7441, opened 2018-08-08
Blurry eyes just got blurrier

FlattestGuitar:
First PR!

very blurry

obraz
The HUD remains untouched when it's blurred.

I also refactored the AO thingy to use the filter helpers, which should really be the only way anyone interacts with them because of how retarded the basic BYOND implementation is. I've also been told that introducing very strong opinions in my first PR makes me seem like I know my shit, even when I don't. Thanks.

:cl: Flatty
tweak: The blurry eye overlay has been replaced with an actual blur.
/:cl:


deathride58:
Could probably make it even stronger so that the only thing you can make out are basic shapes

code/__DEFINES/misc.dm
```diff
@@ -453,9 +453,6 @@ GLOBAL_LIST_INIT(pda_styles, list(MONO, VT, ORBITRON, SHARE))
 // Used by PDA and cartridge code to reduce repetitiveness of spritesheets
 #define PDAIMG(what) {"<span class="pda16x16 [#what]"></span>"}
 
-//Filters
-#define AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-2, size=4, border=4, color="#04080FAA")
```
deathride58: Why remove the define? The define makes it easier to find and keep track of a specific filter's properties.

FlattestGuitar: Because a filter with these specific properties is only used once, and with the `add_filter` proc you don't even use the in-built `filter()` proc, so it'd just be a list of properties. 

I can use defines if you really want, but I don't think there's much point.

deathride58 (dismissed):

silicons (dismissed):

deathride58 (approved):
Putting some thought into it, it'd be *really* neat if the eye blur effect were stronger the higher the eye_blur value is, and weaker the smaller the value is. That's likely out of this PR's scope, but LGTM

PR #7462, opened 2018-08-15
fixes telekenetic snatching from toilets where some captains hide the spare id

silicons:


deathride58 (approved):
your're removeing emergent gameplay!!!!!!

PR #7466, opened 2018-08-16
nya

Wilchenx:
🐈


modular_citadel/code/modules/mob/cit_emotes.dm
```diff
@@ -100,4 +100,12 @@
...
 		playsound(user, 'modular_citadel/sound/voice/snap3.ogg', 50, 1, -1)
-	. = ..()
\ No newline at end of file
+	. = ..()
+
+/datum/emote/living/nya
+	key = "nya"
+	key_third_person = "nya's audibly"
+	message = "nya's"
+	emote_type = EMOTA_AUDIBLE
```
deathride58: >emota

modular_citadel/code/modules/mob/cit_emotes.dm
```diff
@@ -100,4 +100,12 @@
...
+	sound = 'sound/effects/nya.ogg'
```
deathride58: your emote has to be a subtype of /datum/emote/sound to make use of this var

deathride58 (changes requested):

PR #7472, opened 2018-08-18
The Open Galaxy

DravenPeace1:
:cl: Ah yes "Reapers"
add: Finally adds turians, but as a body marking, to have a better look, select the mammal, remove the snout and you won't have any weird looks, you can still add ears and tail, even taur body if anyone is "unique" enough to use it.

:cl: Scolipede, i choose you!!
add: Adds a new whole specie! Scolipede have snout, body marking, taur, tail and ears, for a more good look, try to find dark red colors to fit with it.

add: Added a bit news things, such as bee body marking, insect tail, fish tail and gills which is found in ears, horn for anyone that want to make a demoness or goat, added a tentacle tail, which i can see it will be mostly be used for slimes, the tentacle tail itself was an old sprite of citadel, i only brought back since i found a place to fit it, the real creator of the sprite will have the credit he/she deserves.

fix: Fixed the sprite bug of the octopus taur, it also had it name changed to tentacle
/:cl:


DravenPeace1:
https://prnt.sc/kjzsvn

https://prnt.sc/kjzt1z

https://prnt.sc/kjztfv

https://prnt.sc/kjztoy

https://prnt.sc/kjzu2o

https://prnt.sc/kjzukd

These are simple images of how they look like (Note" The purple part of Scolipede was made bright purple to show the place it would be colored the dark red would almost hide it)

PR #7488, opened 2018-08-22
DMcode IRV poll viewer

silicons:
It has come to my attention that the player representative poll is unable to be viewed because statbus does not support IRV viewing and TGDB is not able to be set up at this time.
This fixes that.
Or will fix that when I finish it in a few hours.


config/dbconfig.txt
```diff
@@ -3,7 +3,7 @@
 ## administration, and the in game library.
 
 ## Should SQL be enabled? Uncomment to enable
-#SQL_ENABLED
+SQL_ENABLED
```
silicons, line 6: note to self to remove this when testing is done

deathride58:
Ping me when this is ready for testmerge

config/dbconfig.txt
deathride58: testings done, remove this when you want this merged because this is nice to have

PR #7499, opened 2018-08-25
Xenomicrobes now randomly transform you into one of three castes.

Tupinambis:
:cl: Tupinambis
tweak: Xenomicrobes will now have a random chance of transforming you into either a drone, hunter, or sentinel.
tweak: Flavor text for the xenomicrobe infection has been modified.
note: In order for this to work xenomicrobes had to be removed from the transforming diseases file.
/:cl:

[why]: Since xenomicrobes are currently only obtainable via a lavaland ruin, it is worthwhile to update them to be more robust, allowing the user to potentially become a drone and even a queen.

### DIFF
--- a/code/datums/diseases/transformation.dm
+++ b/code/datums/diseases/transformation.dm
@@ -167,36 +167,6 @@
 			if (prob(20))
 				affected_mob.say(pick("beep, beep!", "Boop bop boop beep.", "kkkiiiill mmme", "I wwwaaannntt tttoo dddiiieeee..."))
 
-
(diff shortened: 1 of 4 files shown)


nik707:
Wasn't the entire point of changing this ruin to begin with to remove a traitor item? Now you've essentially replaced it with a self-antag item.

deathride58:
There's good reason why xeno microbes transform you into a hunter specifically.

Tupinambis:
Considering the strong concerns for potential self antagging issues. I am closing this.

PR #7502, opened 2018-08-26
Makes the code not go "no you" when admins try to note people (NOTE: NOTES ARE STILL BUGGED, this just fixes the worst issue)

silicons:


deathride58 (approved):
fix the bug where noting is still broken, grmbl grmbl grmbl grmbl

PR #7505, opened 2018-08-27
Makes the captain's spare easy for acting captains to access.

Tupinambis:
:cl: Tupinambis
tweak: Removes the captain's spare from the locker and places it on a table within the captain's rooms. Applies to all stations.
tweak: Unlock air alarm and door bolts for Toxins. Adds scrubber and bomb hardsuit. Just updating it to be in line with the other stations.
/:cl:

[why]: As an acting captain, going through the process of grabbing an axe or ion rifle to access the captain's spare is frustrating. This change should allow heads of staff  to more easily take over as acting captain, should no HoP or Captain be available round start.

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -22131,7 +22131,13 @@
 "bfB" = (
 /obj/structure/table/wood,
 /obj/machinery/atmospherics/pipe/simple/supply/hidden,
-/obj/item/card/id/captains_spare,
+/obj/item/storage/box/matches{
+	pixel_x = -3;
+	pixel_y = 3
+	},
+/obj/item/clothing/mask/cigarette/cigar{
+	pixel_y = 6
+	},
 /turf/open/floor/wood,
 /area/crew_quarters/heads/captain)
 "bfC" = (
@@ -26100,13 +26106,12 @@
 /area/crew_quarters/heads/captain)
 "bpl" = (
 /obj/structure/table/wood,
-/obj/item/storage/box/matches,
 /obj/item/razor{
 	pixel_x = -4;
 	pixel_y = 2
 	},
-/obj/item/clothing/mask/cigarette/cigar,
 /obj/item/reagent_containers/food/drinks/flask/gold,
+/obj/item/card/id/captains_spare,
 /turf/open/floor/carpet,
 /area/crew_quarters/heads/captain)
 "bpm" = (
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -20756,6 +20756,8 @@
 /area/bridge/meeting_room)
 "bbZ" = (
 /obj/structure/table/wood,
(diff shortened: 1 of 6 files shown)


deathride58:
I really don't want to see a repeat of the great grey spare rush of late 2017.

Tupinambis:
If greytiders try to take the spare. Just bwoink them. If an antag rushes the spare roundstart, bwoink them for powergaming. It is silly to not put in a quality of life feature for heads (who already have a difficult job) just because some people will potentially abuse it.

silicons:
this is needed the original locker change was hugbox garbage tbh

deathride58:
At least move the spares to the back rooms so greyshirts have *at least* one airlock to go through if they want the spare.

Poojawa:
Greytide has 3 doors to go through. unless you've got a powergamer using the secret monkey technique to get in.

Putting it in a very shatter proof, head ID locked case would work to keep this mythic other-server greytide Bhjin's PTSDing over out while making sure heads or headstaff looted antags can aquire the spare.

deathride58:
Greytide only has disposals to go through. Literally the only thing a tider has to do to get into the Captain's office is turn one single disposals pipe around and hop in a disposals chute. Omega is the only map immune to this.

Also, putting the ID in a damage-proof case would just be straight-up hugbox. However, simply moving the spare to the back room would be sufficient enough to deter the return of the great grey spare rush, as that'll increase the minimum time to speedrun rushing the spare from less than one minute (!!!) to well over two minutes, since the tider will have to learn the door wires and all that other shit to actually get to the spare. Increasing the minimum time to speedrun spare rushes gives admins an actual chance to notice a tider rushing the spare.

When something is 100% unlogged, happens in less than a minute from the start of the round, and is insanely easy to do, that thing is guaranteed to go completely unnoticed, and therefore usually unpunished, most of the time it happens, unless, in the specific case of spare rushing, there happens to be an actual Captain present roundstart. But guess what this server rarely ever has.

tl;dr: Literally all you have to do to deter spare rushing and make the jobs of admins much easier is to simply put the ID in the captain's quarters.

Poojawa:
So. y'know. log it or something.

make a check that message_admins when someone who isn't a staff head enters the Captain's office before the round start summery check happens. 

Unless the entire meta of the server has instantly changed overnight, this has never actually been a problem on Citadel. You really need to slow down on using other server's issues as justification for nerfing core mechanics. 

Also, I didn't mean damage proof, though I did misspeak and meant to put 'resistant'. Something stronger than a fireaxe case (which almost no-one knows how to cleanly open these, wtf) and could be hooked up to trigger the same anti-theft lockdown as the gun case might. If you absolutely need some security. 

by the by, it helps to not spoil advanced tactics.

MimicFaux:
The captain's spare was originally put in the locker to avoid rushing it down and there it was left alone for a long time. Then, door wires were individually randomized.
Outside of mucking about with pipes, it takes significantly longer to break unlawfully break in since the wire change.

deathride58:
Did mapdiffbot break

deathride58:
Mapdiffbot heyo

PR #7513, opened 2018-08-29
adds windoor to atmos flaps

nik707:
[Changelogs]: suggest change to make atmos just that little bit harder to break into.

:cl: nicc
add: adds windoor to the plastic flaps in the atmos lobby
/:cl:

[why]: currently, in order to break into atmos to steal a hardsuit, you need a welder and a wrench. Crawl under flaps, jump into and out of the disposals unit, rip the wall open. GG 30 seconds or less. This makes that less of an easy thing to do.

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -36955,6 +36955,12 @@
 	location = "Atmospherics"
 	},
 /obj/effect/turf_decal/delivery,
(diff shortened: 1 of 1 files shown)


MimicFaux:
The end of an era.

Poojawa:
It's a bit more brute-force than I'd want, but with the advent of crawling it's needed. It used to be you needed to throw a bucket of water to skid under it.

silicons (approved):

PR #7517, opened 2018-08-31
Forces Vore prefs to use rust_g

Poojawa:
This probably fixes vore specific prefs breaking.

:cl: 
fix: Makes vore specific prefs call Rust_g else uses normal stuff.
/:cl: 

next, Cit_prefs


deathride58 (approved):
My snake's three stomachs still aren't saving at all even with this PR but this is a pretty significant performance improvement for the few people that vorepref saving does work for

PR #7525, opened 2018-08-31
removes honkstaff temporarily

nik707:
[Changelogs]: temporarily removes the honkmother staff from the lavaland ruin

:cl: nicc

del: honkstaff

/:cl:

[why]: The staff currently has a bug where it will get stuck between two turfs and spam the fuck out of anyone nearby with several hundred bullet strikes, and has been known to cause crashes. Temporary removal until fix is found.


MimicFaux:
Honk

PR #7534, opened 2018-09-04
[s] nerfs smuggler satchels cus wtf

nik707:
[Changelogs]: Reduces carry capacity of smuggler satchels

:cl: nicc

balance: reduces weight cap of smuggler satchel

/:cl:

[why]: at the moment you can put like 5 guns in them and then throw them in your backpack, so unless one of the better coders wants to just make it so you can't put guns in or something, this is the easy way to solve that massive powergame.


Poojawa:
I approve of this. they shouldn't be boxes that can carry normal items.

deathride58 (approved):

PR #7552, opened 2018-09-07
combine with mob buff PR - buffs kinetic crushers

silicons:
lets be honest it'll require quite a lot more kiting and you'll run out of stamina real damn quick without this.
lavaland is not just something that plays well with stamina update.

### DIFF
--- a/code/modules/mining/equipment/kinetic_crusher.dm
+++ b/code/modules/mining/equipment/kinetic_crusher.dm
@@ -15,9 +15,8 @@
 	force_wielded = 20
 	throwforce = 5
 	throw_speed = 4
-	light_range = 5
-	light_power = 1
-	item_weight = 2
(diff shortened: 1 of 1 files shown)


MimicFaux:
#bless

code/modules/mining/equipment/kinetic_crusher.dm
```diff
@@ -15,8 +15,8 @@
 	force_wielded = 20
 	throwforce = 5
 	throw_speed = 4
-	light_range = 5
-	light_power = 1
+	light_range = 7
+	light_power = 2
```
deathride58, line 19: buffing light_power beyond `1` makes the resulting light look pretty weird. this specific buff can be toned back

deathride58 (approved):

PR #7571, opened 2018-09-13
Syndicate Basic Encryption Key's: Teamwork makes the dream work.

Tupinambis:
LISTEN UP AGENTS: We have noticed that success rates among our agents increases when there is a line of communications between them. Sadly, this is a rarity when every TC counts. In order to ensure that our agents can more effectively operate in the field, we have added an encryption key in your uplink for the low low price of FREE. Get yours today agent and be badass as a TEAM!

:cl: Tupinambis
add: Adds an all new syndicate basic encryption key, which allows you to use the syndicate channel on your headset. This can be purchased for free under "Standard Issue Gear" from the uplink, but only once.
balance: Reduces the cost of the normal syndicate encryption key to 1 TC from 2 TC.
/:cl:

[why]: Currently, security can very easily face a traitor in a one versus one scenario in most situations. Part of this, is that traitors often work alone. Private radio communications will encourage traitors to work as a team and coordinate to complete their objectives (which unlike in lings, usually won't result in betrayal) will not only encourage teamplay and RP, but also give sec more of a challenge, and give traitors who have been caught by sec (or validhunters) a chance to escape with the help of their allies.

### DIFF
--- a/code/game/objects/items/devices/radio/encryptionkey.dm
+++ b/code/game/objects/items/devices/radio/encryptionkey.dm
@@ -17,7 +17,7 @@
 	syndie = 1//Signifies that it de-crypts Syndicate transmissions
 
 /obj/item/encryptionkey/syndicate/basic
-	name = "syndicate encryption key"
+	name = "syndicate basic encryption key"
 	desc = "An encryption key for a radio headset. To access the syndicate channel, use :t."
 	icon_state = "syn_cypherkey"
 	channels = list("Syndicate" = 1)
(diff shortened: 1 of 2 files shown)


Poojawa:
This is not something to buff.

Strip out the departmental eavesdropping as a compromise for becoming free.

Poojawa:
This is not something to buff.

Strip out the departmental eavesdropping as a compromise for becoming free.

Tupinambis:
That is in fact, what I did.

code/game/objects/items/devices/radio/encryptionkey.dm
```diff
@@ -16,6 +16,12 @@
 	channels = list("Syndicate" = 1)
 	syndie = 1//Signifies that it de-crypts Syndicate transmissions
```
Poojawa, line 17: This is inherited to /basic

This doesn't remove all-channel listening privileges as a result

Poojawa (changes requested):

Tupinambis:
Rethinking it this may not be the best idea balance wise regardless. Closing for now.

PR #7580, opened 2018-09-16
Adds simple harness item to game and loadout

izzyinbox:
:cl:
add: Added simple harness jumpsuit item
/:cl:

Since a lot of people seem to be using the tennis ball glitch as a stand-in for appearing nude, I've made a lower-body harness jumpsuit item to fill the need for appearing mostly nude but keeping sensors/item slots. Accessible via the loadout.


Poojawa:
Should always be in the habit of uploading an example picture of it in game, also this would be more spriteadd

PR #7587, opened 2018-09-21
Vore based mood events

Poojawa:
:cl: Poojawa
add: Added Vore based moodlets, primitive and ain't gunna adapt for not belly things but w/e
/:cl:

[why]: # Nom ~


deathride58 (approved):
The `fedpred` mood event's description isn't "Tastes like trash, just like me!" so this pr is irredeemable shit

10/10

PR #7599, opened 2018-09-27
Digitigrade Uniforms from Oracle 

Poojawa:
ports https://github.com/OracleStation/OracleStation/pull/920 <3

🆑 Quantum_m & Ziiro & Poojawa
add: All uniforms now have digi leg alternates!
/🆑

Override Icon was fully replaced in this PR, I'm sure Shoes will cause a conflict when they're merged, but ez fix.


deathride58 (approved):
Can be merged as soon as that "Added currently testmerged label" notification hits an age of 12 hours

PR #7603, opened 2018-09-28
[DNM][WIP] JSON Character saving

Poojawa:
this is more for review atm.

So far, I've converted, to my know-how, .sav to save to JSON.

I'm really not confident about feature list being properly updated, not that it does atm.

Compiles fine, it reads the .sav, but features are bork. Changing slots blanks out the viewing and prints a blank character.json file in the correct positioning. 

I know I probably need to serialize custom_names and features but I haven't the foggiest idea how to.


code/__HELPERS/persistence.dm
```diff
@@ -28,7 +28,7 @@
 
 // This is so specific atoms can override these, and ignore certain ones
 /atom/proc/vars_to_save()
- 	return list("color","dir","icon","icon_state","name","pixel_x","pixel_y")
+ 	return list("slot")
```
deathride58, line 31: ?????????????????????????????????????

PR #7614, opened 2018-10-03
Fixes Black Tango dress for digitigrade

Poojawa:
I'll have to look into double checking they're in the sprite folder later but, ~


deathride58 (commented):
This won't fix their broken icon

deathride58 (approved):

PR #7627, opened 2018-10-05
APC have Damage Resistance, Health, and armor buffs

Poojawa:
:cl: Poojawa
balance: APCs are harder to destroy.
/:cl:

only compile tested, but it's the same code as door armor so w/e


deathride58:
Why tho

Poojawa:
People keep wanting to make APCs even easier to rebuild, and they're super easy to break as is. Kinda silly tbh.

deathride58:
Let's see the reception of this after a day of testmerging

deathride58 (approved):

PR #7630, opened 2018-10-07
The cooler nuke op base (featuring more air)

Tupinambis:
:cl: Tupinambis
tweak: Changes large parts of the nuclear operative base, including the addition of a new barracks where the ops spawn, a personal bedroom for the leader, the rearrangement of numerous rooms, some added fluff, a slightly more roomy bar, a larger hanger, a kitchen, and various other small changes to make the base feel just a bit more alive. This was achieved by completely redoing the base from scratch so there was room to work with, as a result potential bugs could crop, please do report them.
fix: The drop pod chamber was airless. This seems like an oversight, and has been filled with air.
/:cl:

[why]: Should make the base a bit more immersive, makes it feel like the operatives could actually live there. Hanger has more room for EORG memes should the emergency shuttle be redirected. Also allows for larger potential to bus.

image


deathride58:
Looks good but mapdiffbot is dead so we're gonna testmerge this for a couple of rounds before fully merging

deathride58 (approved):

ComradeNecron:
ah yes, now you can get the METAL BOX without dying horribly

PR #7640, opened 2018-10-09
Adds minor protection to emag clicks

Poojawa:
:cl:
balance: Added sanity checks to emag usage
/:cl:

### DIFF
--- a/code/game/objects/items/cards_ids.dm
+++ b/code/game/objects/items/cards_ids.dm
@@ -95,7 +95,16 @@
 	var/atom/A = target
 	if(!proximity && prox_check)
 		return
-	A.emag_act(user)
+	if(isturf(A))
+		return
+	if(istype(A,/obj/item/storage/lockbox))
+		A.emag_act(user)
+	if(istype(A,/obj/item/storage))
+		return
+	if(!(isobj(A) || issilicon(A) || isbot(A) || isdrone(A)))
+		return
+	else
+		A.emag_act(user)
 
 /obj/item/card/emagfake
 	desc = "It's a card with a magnetic strip attached to some circuitry. Closer inspection shows that this card is a poorly made replica, with a \"DonkCo\" logo stamped on the back."
--- a/code/game/objects/items/cards_ids.dm
+++ b/code/game/objects/items/cards_ids.dm
@@ -95,16 +95,33 @@
 	var/atom/A = target
 	if(!proximity && prox_check)
 		return
+	//Citadel changes: modular code misfiring, so we're bypassing into main code.
+	if(!uses)
+		user.visible_message("<span class='warning'>[src] emits a weak spark. It's burnt out!</span>")
+		playsound(src, 'sound/effects/light_flicker.ogg', 100, 1)
+		return
+	else if(uses <= 3)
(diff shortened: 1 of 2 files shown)


deathride58:
This actually doesn't do anything but restrict potential emag effects to only specific hardcoded types. The code that actually reduces emag charges is in modular_citadel, which this PR doesn't touch

Poojawa:
The only things this hardcodes against is mob/living that isn't a simple animal bot (for emagging beepsky, etc), drones, or silicons. You can't emag humans, after all. 

Turfs and storage containers that aren't lock boxes are also assured to not to. everything else fires off with emag_act, including charge use processing.

deathride58:
https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/modular_citadel/code/game/objects/ids.dm#L85 This is where charge usage is handled. Your PR doesnt touch it

Poojawa:
That's.. correct, yes. It's not intended to be changing that. This solves the reported problem of Emags incorrectly using charges when inserted into backpacks or used on turfs.

deathride58:
Because this PR doesn't touch that code, this PR does not stop emags from consuming charges when the proc that the modular code overrides returns early.

Poojawa:
```
 That's a cryptographic sequencer
It's a card with a magnetic strip attached to some circuitry.
It is a tiny item.
It has 10 charges left.
You put the cryptographic sequencer into the backpack.
 That's a cryptographic sequencer
It's a card with a magnetic strip attached to some circuitry.
It is a tiny item.
It has 10 charges left.
```

works just fine?

Poojawa:
fixed.

code/game/objects/items/cards_ids.dm
```diff
@@ -95,7 +95,33 @@
...
+		playsound(src, 'sound/effects/light_flicker.ogg', 30, 1)	//Tiiiiiiny warning sound to let ya know your emag's almost dead
+
+	if(isturf(A))
+		return
+	if(istype(A,/obj/item/storage/lockbox))
+		A.emag_act(user)
+		uses = max(uses - 1, 0)
+		if(!uses)
+			user.visible_message("<span class='warning'>[src] fizzles and sparks. It seems like it's out of charges.</span>")
+			playsound(src, 'sound/effects/light_flicker.ogg', 100, 1)
```
deathride58, line 113: You can reduce the copypasta here by making use of `goto`

code/game/objects/items/cards_ids.dm
```diff
@@ -95,7 +95,33 @@
...
+	if(istype(A,/obj/item/storage))
+		return
+	if(!(isobj(A) || issilicon(A) || isbot(A) || isdrone(A)))
+		return
+	else
```
deathride58, line 118: This `else` here isn't necessary, as the check right above is an `if(); return`, and all of the other checks have a return at the end of them

deathride58 (approved):
It's spaghetti, but it'll work just fine, my own gripes about hardcoded typechecks aside

code/game/objects/items/cards_ids.dm
Poojawa: I don't know how to use that.

deathride58:
Remind me later and I'll do a small cleanup pass, my own personal gripes with this PR can be fixed later

PR #7641, opened 2018-10-09
fixes mining suits, wintercoats

Poojawa:
Mislabel on the winter coat, fixed

Seva/exo have unlazy finishings just to hold out till they're redone

Also fixes shoe blood and ears because w/e


deathride58 (approved):
Can be speedmerged

PR #7643, opened 2018-10-10
Adds Rollerbeds to MediHounds

CameronWoof:
:cl: CameronWoof
add: Medihounds now have rollerbeds for non-vore patient transport
/:cl:

[why]: Because it's better than nothing and I think that's hard to disagree with. Quick win buff to a famously struggling module.


Poojawa:
"Because it's better than nothing"

They already have patient transport.

CameronWoof:
They have patient transport that is very robust but only usable on less than half of the people playing. This gives them an alternative to dragging vore-disabled patients bleeding through the hallways, which I think we can all agree is an upgrade.

Poojawa:
blimey, I wonder why that is.

CameronWoof:
Please keep feedback related to this PR, not a settled issue from half a year ago. Denying the addition of a rollerbed because of your personal feelings on consent is petty.

deathride58 (approved):

deathride58:
Roller beds will be vastly inferior to the sleeper once the sleeper is made normie-compatible.

CameronWoof:
We can probably safely remove them at that point. Or leave them for **dual patient transport weewoo**

silicons (approved):

PR #7650, opened 2018-10-13
(HOPEFULLY) Fixes autostand making item dropping wait for stand attempts to finish

deathride58:
Reason number 3189235981239 why game devs hate unconstructive feedback with a passion. Shit like this never gets reported. At all.

:cl: deathride58
fix: Autostand no longer makes you invulnerable to dropping items when being knocked down with a force greater than 80
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1020,7 +1020,7 @@
 			client.move_delay = world.time + movement_delay()
 	lying_prev = lying
 	if(canmove && !intentionalresting && iscarbon(src) && client && client.prefs && client.prefs.autostand)//CIT CHANGE - adds autostanding as a preference
-		resist_a_rest(TRUE)//CIT CHANGE - ditto
+		addtimer(CALLBACK(src, .proc/resist_a_rest, TRUE), 0) //CIT CHANGE - ditto
 	return canmove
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/living.dm
```diff
@@ -1020,7 +1020,7 @@
 			client.move_delay = world.time + movement_delay()
 	lying_prev = lying
 	if(canmove && !intentionalresting && iscarbon(src) && client && client.prefs && client.prefs.autostand)//CIT CHANGE - adds autostanding as a preference
-		resist_a_rest(TRUE)//CIT CHANGE - ditto
+		addtimer(CALLBACK(src, .proc/resist_a_rest, TRUE), 0) //CIT CHANGE - ditto
```
silicons, line 1023: this is literally abuse of timers

silicons (approved):

code/modules/mob/living/living.dm
Poojawa, line 1023: 0 duration timers don't work

deathride58:
Byond doesn't have a built-in way to make a proc call asynchronous. 0-length timers are a hacky way to call a proc asynchronously, so long as the game is currently running and timers are currently processing. Judging from ingame testing, it actually works just fine.

Poojawa:
you just call spawn(0)

that's it. It does the half stutter pause.

PR #7660, opened 2018-10-14
Makes the O2 crate sprites consistent

CameronWoof:
:cl: CameronWoof
fix: The closed O2 crate now uses the same color scheme as the open one
/:cl:

[why]: Because it needed to be done, it took five seconds, and juicy, juicy GBP. Thanks to Toriate for the original sprites, of course.


deathride58 (approved):
Eligible for speedmerge

PR #7671, opened 2018-10-22
Makes human tails and human ears save again

deathride58:
Grmbl grmbl

:cl: deathride58
fix: Human tails and human ears now save again. Humans now use mam_tail and mam_ears for their snowflake bits instead of tg's equivalent, just as they did prior to human tail and ear saving breaking
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/humans.dm
+++ b/code/modules/mob/living/carbon/human/species_types/humans.dm
@@ -3,8 +3,8 @@
 	id = "human"
 	default_color = "FFFFFF"
 	species_traits = list(EYECOLOR,HAIR,FACEHAIR,LIPS,MUTCOLORS_PARTSONLY)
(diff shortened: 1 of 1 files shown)


Poojawa:
This... literally shouldn't do anything. 

if people are swapping between characters, shit gets reset/transferred between 'em, that's the actual bug

deathride58:
The bug is that human tails/ears reset every single round. This fixes that by just making the selection just use mammal tails/ears, which do save properly. This comes with the positive side effect of humans having a much more diverse selection of tails/ears

Poojawa:
err, they already have the same list as mammal ears/tails

https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/modular_citadel/code/modules/mob/dead/new_player/sprite_accessories.dm#L33

fix the actual save/load bug, not band-aid it

silicons (approved):

PR #7680, opened 2018-10-28
Removes vampires from being able to ventcrawl with items while in bat form, makes the bat named after the original mob

silicons:


deathride58:
Doesnt even compile
-rep

GuppyLaxx:
i cant even play here but it's a +1 from me because i know how garbage this is

deathride58 (approved):

PR #7688, opened 2018-10-30
makes it impossible to resist/move out of grabs while resting (though disarming your way out of the grab still works)

deathride58:
Title. For the laymen, this means sec will no longer have to push perps into hard stamcrit to actually arrest someone. For those that know what they're doing, this means that combat with non-sec will be a lot more lethal. This partially emulates the upcoming grab rework's restraining/pinning with the current grab system, which will make it a lot easier for those who know what they're doing to end a fight.

:cl: deathride58
balance: You can no longer resist/move out of grabs while you're resting. Disarming your way out still works, though.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1422,6 +1422,9 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 		if(target.pulling == user)
 			target.visible_message("<span class='warning'>[user] wrestles out of [target]'s grip!</span>")
 			target.stop_pulling()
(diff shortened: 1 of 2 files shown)


Poojawa:
why wouldn't you be able to fight out of a grab?

have you literally never seen wrestling?

Tupinambis:
This makes slips and knockdowns more viable tactics, and allows antags to be more effective. I strongly approve

deathride58:
@Poojawa You can still fight out of a grab by disarming the person grabbing you

silicons (approved):

PR #7700, opened 2018-11-06
Probably fixes rogue icon overrides

Poojawa:
Taurs should use normal legs to avoid bugs.

:cl:
fix: Taur suits should fix themselves properly when worn by other people.
/:cl:


deathride58 (approved):
Can be speedmerged if kev notices

Poojawa:
ok, so a week from now

📯

PR #7714, opened 2018-11-11
Lone Ops but they are an actual threat (hopefully)

Tupinambis:
:cl: Tupinambis
balance: Lone operatives no longer spawn with a syndicate hardsuit and a bulldog. They instead spawn with a black and red EVA suit and an extra 15 TC (40 in total)
image

/:cl:

[why]: Currently, lone ops have a very small chance (read: 0% chance) of succeeding unless either A: the disk was never secured or B: there is no security or greytide threat. This is because they are very limited in how they can attack the station. They start with a syndie hardsuit and if they dare be seen, people will immediately scream "NUKE OPS!" over radio. Ops are also forced to use a bulldog, which not everyone prefers, and which puts you right in the range that security wants you (taser, stunbatons, bolas, and more!). They don't currently have enough TC to change things up or get creative in their approach. They also don't get any of the free gear that comes with the syndicate infiltrator and base. The extra TC (effectively 7 extra, as syndie hardsuits are free by default) should hopefully allow lone ops to attack the station with more confidence and with greater odds of success.

### DIFF
--- a/code/game/gamemodes/nuclear/nuclear.dm
+++ b/code/game/gamemodes/nuclear/nuclear.dm
@@ -175,3 +175,20 @@
 		/obj/item/tank/jetpack/oxygen/harness=1,\
 		/obj/item/gun/ballistic/automatic/pistol=1,\
 		/obj/item/kitchen/knife/combat/survival)
+
+/datum/outfit/syndicate/lone
+	name = "Syndicate Operative - Lone"
+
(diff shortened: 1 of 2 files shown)


MimicFaux:
Gives some love to an already very rare, and statistically unsuccessful side-antag: Check

silicons (approved):

PR #7717, opened 2018-11-12
Pubby Engineering has been redesigned with a SM engine.

Tupinambis:
:cl: Tupinambis
add: The tesla engine has been replaced with a supermatter engine, and pubby engineering has been redesigned to accommodate this change.
del: The tesla engine has been completely removed along with any related components
fix: RD and R&D vents are now actually connected to the air distribution.
/:cl:

[why]: The current Pubby tesla engine is so easy to fuck up for how much sheer destruction it can release upon the station (and the server oh god the lag), that it desperately needed replacement. Often times people will unintentionally loose or anger the tesla, causing an early call, and this is not an uncommon event on Pubby. The supermatter is a much more robust engine than the tesla, is very much recoverable should something go wrong, and more can be done to get creative with it as an engineer.

image

### DIFF
--- a/_maps/map_files/PubbyStation/PubbyStation.dmm
+++ b/_maps/map_files/PubbyStation/PubbyStation.dmm
@@ -15275,6 +15275,13 @@
 	},
 /turf/open/floor/plating,
 /area/maintenance/department/cargo)
+"aKm" = (
+/obj/effect/spawner/structure/window/plasma/reinforced,
+/obj/machinery/atmospherics/pipe/simple/orange/visible{
+	dir = 4
+	},
+/turf/open/floor/plating,
+/area/engine/engineering)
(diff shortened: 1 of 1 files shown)


lolman360:
pubby is a special snowflake because of its engine
don't take it from pubby

Tupinambis:
> pubby is a special snowflake because of its engine
> don't take it from pubby

Being a special snowflake, or "unique" does not excuse what essentially comes down to poor gameplay, where incompetent engineers, antags, or some weird quirk that nobody anticipated, can completely ruin a round. This doesn't mean you can't build a tesla or singularity, it just means you won't have it roundstart.

MimicFaux:
Why not leave the components in storage?

deathride58 (approved):

PR #7719, opened 2018-11-13
Reduces shadow species light threshold leniency to accommodate for cit's lighting changes

deathride58:
Title. The unchanged threshold is insanely lenient, allowing spicy memes like standing right next to a light source without taking damage, but stepping onto the same turf as the light source results in you instantly starting to take damage. This fixes that by reducing the light threshold leniency to a rounded-up value of what get_lumcount() returns on a turf adjacent to an APC in maint

:cl: deathride58
balance: Nightmares and shadowpeople have had their light threshold for being considered in darkness reduced. Nightmares can no longer stand unharmed next to a light fixture during a nightshift 100% unharmed. The threshold is high enough so that APC lights on their own won't harm a nightmare, but being on the edge of a light source will.
/:cl:


Poojawa:
i ded tag. :^)

silicons (approved):

PR #7724, opened 2018-11-16
you are too exhausted to keep fucking

silicons:
don't actually merge


deathride58 (commented):
dont forget to add a stamloss check that stops the emote from going through if the user's in soft stamcrit or lower

MimicFaux:
WHEN WILL THE MIME NERFS STOP!?

PR #7726, opened 2018-11-16
Makes it plausible to join as ghost roles from the latejoin menu

deathride58:
Title. Here's how it looks
image

This should hopefully encourage players to play as ghost roles such as ashwalkers, as a lot of ghost roles are far more effective when they have multiple players.

:cl: deathride58
add: You can now access ghost roles from the latejoin menu.
/:cl:


BlackMajor:
Honestly this is actually a fantastic idea.

A lot of players hit observe hoping for a specific ghost role, only to find it not there, etc.

You'd best try to fit things like Positronic brains and such into this menu.

silicons (approved):

PR #7739, opened 2018-11-18
[PRIORITY] Pubby Station Air Injector Hotfix

Tupinambis:
:cl: Tupinambis
fix: the air injectors for the SM engine waste and for toxins waste should now receive power
/:cl:


deathride58 (approved):
okay for speedmerge if another maintainer sees

PR #7741, opened 2018-11-19
Makes auto fit viewport enabled by default for new players

deathride58:
"Hey why is my screen so small" is such a common question from new players, and it's got a simple answer of "Use the fit viewport button in the ooc tab"
But the pref that automatically presses that button is disabled by default thanks to TG. So let's make that pref press that button by default.

:cl: deathride58
tweak: Auto fit viewport is now enabled by default for new players. One less question off the FAQ
/:cl:


nik707 (approved):
this is a really big code change, i hope you tested it

PR #7743, opened 2018-11-19
Disables TG's biased preference-based map voting system in favor of an actual end-of-round map voting system

deathride58:
Title. This PR makes it so that, when the shuttle's leaving for centcom, the map system will no longer use heavily biased voting methods that effectively leave the next map up to RNG. Instead, the server will start up a vote that will actually let players have a say in what map will be played on the server next.
This will allow contributors, current and future, to have the freedom of being able to experiment with new map design ideas without having to worry about if their map will receive heavy backlash. This will allow comfybunker and other experimental maps like it to coexist with the existing maps while letting the players themselves decide "hey, we want to play this map" instead of having the game force everyone into it regardless of overall reception.
The issue with TG's preference-based voting system is that there's really not much player choice at all. Since players rarely ever tinker with their game preferences, TG's built-in map rotation is heavily, *heavily* skewed towards favoring whatever map is set as the default. And on top of that, maps that have 0 players who have the map set as their map preference still have an RNG chance to be picked, as a result of TG heavily trying to push a "hey, let's make sure every single map gets played at least once a day" thing.
I can rant on and on about the topic, but this PR description is getting long enough

:cl: deathride58
add: When the shuttle leaves for centcom, a vote will be started to select the next map instead of the map being randomly chosen via biased voting methods
server: If a server operator wishes to re-enable the biased TG preference voting system, you can do so by toggling the TGSTYLE_MAPROTATION config flag on in the config.txt. Keep in mind that doing so will bring you great misfortune.
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -250,6 +250,8 @@
 
 /datum/config_entry/flag/maprotation
 
+/datum/config_entry/flag/tgstyle_maprotation
+
 /datum/config_entry/number/maprotatechancedelta
 	config_entry_value = 0.75
 	min_val = 0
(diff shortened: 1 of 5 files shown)


nik707 (approved):
looks pretty gay, good work

PR #7765, opened 2018-11-24
Cult buffs (Makes most cult whispers not be shouted and makes stun spell guarantee stamcrit at the cost of a higher cost and reduced charges)

deathride58:
Title. Cult feels like ass right now, so this PR makes two integral parts of bloodcult function closer to how they do on TG.

:cl: deathride58
balance: The majority of cult spells no longer have exclamation marks at the end of their invocations. This means they will no longer be bound to the rules of shouting. Which means sec officers will no longer be able to hear culties sacrificing the captain from the complete opposite side of the maintenance tunnel the sacrifice is happening in. However, things like EMP pulses and runeless teleportation still have the effect due to their inherent loud nature.
balance: The stun blood spell now instantly causes hard stamcrit. However, to offset this, it now also only has one charge and has a higher cost.
/:cl:

### DIFF
--- a/code/modules/antagonists/cult/blood_magic.dm
+++ b/code/modules/antagonists/cult/blood_magic.dm
@@ -277,7 +277,7 @@
 /datum/action/innate/cult/blood_spell/veiling
 	name = "Conceal Presence"
 	desc = "A multi-function spell that alternates between hiding and revealing nearby cult runes, structures, turf, and airlocks."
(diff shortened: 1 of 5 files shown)


nik707 (approved):

silicons:
@deathride58 how do you think this is going to play out with stunbaton nerfs in place?

PR #7758, opened 2018-11-21
Security Stunbaton Rebalancing v2

Tupinambis:
:cl: Tupinambis
balance: Stunbatons no longer knock targets down unless they have taken more than 0 stamloss. They will also not disarm unless the target has taken at least 50 stamloss.
/:cl:

[why]: Currently stunbatons serve as an instant win button, as getting hit once will spell your doom in most situations. This change should allow antagonists to retaliate assuming they have not already taken severe stamloss, as they will no longer drop their weapons immediately. Hopefully fights will be more interesting, and weapons like the esword will be more viable.

NOTICE: TASER CHANGES HAVE BEEN REVERTED AFTER TESTING

PR #7774, opened 2018-11-27
Disabler Buff

Tupinambis:
[Changelogs]: 

:cl: Tupinambis
balance: Disablers now do 30 stamina damage (up from 24) [Pretty sure there are other variables messing with this value]
/:cl:

[why]: They don't feel great to use right now, and really are hardly a punishment to get hit by. This should help a bit.

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -69,7 +69,7 @@
 /obj/item/projectile/beam/disabler
 	name = "disabler beam"
 	icon_state = "omnilaser"
-	damage = 24 // Citadel change for balance from 36
+	damage = 30 // Was 24, and in practice was way too low.
 	damage_type = STAMINA
 	flag = "energy"
 	hitsound = 'sound/weapons/tap.ogg'


code/modules/projectiles/projectile/beams.dm
```diff
@@ -69,7 +69,7 @@
 /obj/item/projectile/beam/disabler
 	name = "disabler beam"
 	icon_state = "omnilaser"
-	damage = 24 // Citadel change for balance from 36
+	damage = 30 // Was 24, and in practice was way too low.
```
Poojawa, line 72: would be better to just not override the comment at all, tbh.

deathride58:
Disablers were originally nerfed due to their sheer superiority against tasers, as they only needed one extra shot to stamcrit (original value of 36 was 4 shots to stamcrit, tasers could stamcrit in only 3 shots) and didn't force crawling, making tasers a complete waste of ammo in comparison, considering a disabler stamcrit was only 1x the cost of a single taser bolt.
The nerfed value of 24 was chosen as it bumped up the shots required to hard stamcrit to 6, while also leaving ample room for people who play their cards right to bump that up to 7. Disablers were still more ammo efficient than tasers, but instead of a disabler stamcrit costing 1x the amount of one single taser shot, it required 1.5x the amount of a taser shot, leaving sec mains some leeway to mix taser shots and disabler shots while still remaining mostly ammo efficient. This did its job of making tasers be used a little bit more often by people who know what they're doing, since the increased energy cost of a disabler stamcrit made it entirely viable to knock a target down with a taser, then follow it up with disabler shots to stamcrit, all for the combined cost of two taser bolts' worth of ammo.
This PR knocks the shots required for disablers to hard stamcrit to 5, but doesn't quite leave enough room for proper stamina management and low-tier stamina regen chems to have an effect on that. Additionally, taking #7758 into account, a disabler stamcrit is decreased from 1.2x the cost of a taser bolt to 1x the cost of a taser bolt, which does nothing but return taser bolts back to their original state of "why the fuck would you ever use this outside of an immediate emergency". Take a minute to think if it's really worth it or not.

Tupinambis:
or just aim for the chest I guess that works. I am dumb. Closing this

PR #7784, opened 2018-12-03
Fixes buildmode, probably.

nik707:
[Changelogs]: PROBABLY fixes buildmode. I dunno I just copypasted one of TGs PRs that fixed it.

:cl: nicc

fix: probably fixed adv buildmodes copy function

/:cl:

[why]: cus blackmajor asked and he's kinda cute


deathride58 (approved):
yell at me when the 24 hour grace period is over or get kev to speedmerge

PR #7788, opened 2018-12-05
A Small Rework to the Combat Rework

silicons:
:cl:
balance: Stunbatons now do 40 down from 49 stamina.
balance: Stunbatons only knock down and disarm if you have more than 70 stamina damage AFTER it applies its flat stamina damage.
balance: Stunbatons now start with a 15k cell allowing for 15 hits instead of 10
experimental: **Sprinting now has its own stamina buffer as opposed to the primary action buffer.**
code: Some unnecessary files were cleaned up.
/:cl:

Open to ideas. 
When this is in testmerge phase I'll need feedback regarding taser/disabler/flash/pepperspray/flashbang/etc balance.

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -12,7 +12,9 @@
 	attack_verb = list("beaten")
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 50, "bio" = 0, "rad" = 0, "fire" = 80, "acid" = 80)
 
-	var/stunforce = 140
+	var/stunforce = 40		//CITADEL EDIT							-- How much stamina damage is applied on hit.
+	var/stamina_paralysis_threshold = 65		//CITADEL EDIT		-- This is how much stamina they have to have after applying stunforce for the baton to force drop their items and knock them down.
+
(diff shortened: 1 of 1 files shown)


code/game/objects/items/stunbaton.dm
```diff
@@ -12,7 +12,9 @@
 	attack_verb = list("beaten")
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 50, "bio" = 0, "rad" = 0, "fire" = 80, "acid" = 80)
 
-	var/stunforce = 140
+	var/stunforce = 40		//CITADEL EDIT							-- How much stamina damage is applied on hit.
```
deathride58, line 15: remember to adjust stunprod stunforce so that they don't become super ultra bullshit

deathride58:
>giving sprinting its own stam buffer

y tho

silicons:
@deathride58 chasing someone is already a chore without combat rework because they have the advantage of being able to throw stuff at you shoot you anything and it'll hit you faster because you're running into their stuff
Sprinting buffer makes it less punishing to sprint after someone in a chase by making it not cut into your action buffer while still making sure people don't spam sprint all the time

PR #7789, opened 2018-12-05
Obfuscates the map vote (only during the voting period)

silicons:
Honestly I believe this'll let people make up their own mind instead of helping landslide ones that have more votes when they open it. Who knows.
Not like the result is obfuscated because people will see what map is next afterwards.


Ragolution:
Seems to have helped with people meme-picking Omega station on high pop and etc, ends up with more variation overall.

deathride58 (approved):

PR #7799, opened 2018-12-08
Added Another new carpet, and two tables, a new cargo crate, and two new medals.

R3dtail:
Added:  monochrome.  Two new table sprites for the other new carpet.  New cargo supply crate for the two new carpets.  Changed:   the name of the new carpet to "blackred" so it would avoid confusion with the original carpet.

[Changelogs]: # 
add:  new carpet.  Monochrome.  New cargo supply crate "Premium Carpet #2 " to contain the other new carpet I had added and this one.  Added two new tablesprites to match both new carpets.  Added carpet smoothing so that blackred and monochrome would smooth with black.  Made it so all the fancy tables would smooth together.  
tweak:  Changed the red carpet to blackred, so that it wouldn't be confused with the normal fancy carpet since supposedly it's red but I'm colorblind so it looks brown to me.  
https://cdn.discordapp.com/attachments/161574200581029888/520822927914958858/Finished_tables.png  Proof of concept.  Screenshot is before I added the carpet smoothing but I promise it's there.  

/:cl:

[why]: # I wanted new carpets.  And it was a glorious challenge to meet.  I wanted new tables to match the new carpets, too.


R3dtail:
Accidentally added a second commit due to not necessarily knowing exactly what I was doing.  So, the PR now adds the aforementioned crate, tables, and carpet.  In addition, it now adds two new medals for medical and engineering, and medal boxes for the CMO and CE with said medals added.

Ragolution:
Not sure how I feel about the carpet; they red one is kind of an eyesore.
The tables look fine, though. Is there any way you could soften the colors on the floor tiles?

R3dtail:
The red ones do look kind of bright on their own but in game they look fine.  In the discord when I was showing the sprites off someone made a similar complaint, and then when I showed them in game they said they weren't bothered by it.  It's supposed to be a bold red.

nik707 (approved):

PR #7800, opened 2018-12-08
xeno tackles now hardstun for 3 seconds if someone is not lying down

silicons:


code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -272,7 +272,10 @@
 						"<span class='userdanger'>[M] disarmed [src]!</span>")
 			else
 				playsound(loc, 'sound/weapons/pierce.ogg', 25, 1, -1)
-				Knockdown(100)
+				if(!IsKnockdown())				//CITADEL EDIT
```
deathride58: `if(!lying)`

deathride58 (changes requested):

deathride58 (approved):
you could probably use ternaries instead of `if();else`

PR #7802, opened 2018-12-08
the one item designed to hard counter kudzu no longer takes 1/2 of your stamina bar to take down one block of fireproofing kudzu

silicons:
1.25 per tile hit.

### DIFF
--- a/code/modules/hydroponics/hydroitemdefines.dm
+++ b/code/modules/hydroponics/hydroitemdefines.dm
@@ -115,7 +115,6 @@
 	slot_flags = ITEM_SLOT_BACK
 	attack_verb = list("chopped", "sliced", "cut", "reaped")
 	hitsound = 'sound/weapons/bladeslice.ogg'
-	total_mass = 1.75
 	var/swiping = FALSE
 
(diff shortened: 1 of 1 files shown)


lolman360:
black gar glasses from a hacked autodrobe are better than the scythe and it's sad

code/modules/hydroponics/hydroitemdefines.dm
```diff
@@ -115,6 +115,7 @@
 	slot_flags = ITEM_SLOT_BACK
 	attack_verb = list("chopped", "sliced", "cut", "reaped")
 	hitsound = 'sound/weapons/bladeslice.ogg'
+	total_mass = 1.25
```
deathride58: as much stamina cost as a screwdriver for 2.6x the amount of damage unsharpened screwdrivers do against normal body parts? you could probably buff scythes by making kudzu cutting restore stamina for each kudzu tile cut instead

silicons: lol im dumb :+1: kk

deathride58 (approved):
👍

PR #7805, opened 2018-12-08
Instead of pens, you can now fit foam darts with metal rods.

silicons:
WIP


deathride58:
it doesn't impale the rod on hit and nail the target to the wall. 0/10

lolman360:
good pr. speedstealthmerge.

Ragolution:
Can no longer put sleepypen in foam dart gun. Would not merge. 0/10

PR #7818, opened 2018-12-13
Attempts to ReAdd Sing/Tesla to Pubby

lightcat:
This is my first Pull Request so I have no idea if this is actually the file wanted. Trying to push out some changes to pubby. but something saying 100,000+ deletions tells me this is probably flawed...

* Added The Singularity/Tesla Setup back onto Pubby Station
* Removed pipe leading from Atmospherics into SME (SME no longer exists.)

Re Added engines to add map variety and to please the group that likes them.


deathride58 (changes requested):
USE THE MAPMERGER

Ragolution:
But why?

nik707:
mapmerge

lightcat:
Working on the map merger.

Tupinambis:
I'm gonna leave a fat thumbs down here. The Tesla was replaced for good reason. It's a simple engine with limited customizability, hugboxed to shit, you can  just leave it alone forever. But the moment a wire is cut by an antag, the thing will loose and there is nothing to stop it. Not only does this cause widespread destruction for little effort, but it also lags the server a ton. Keep that in consideration before you add something purely for variety.

nik707:
> I'm gonna leave a fat thumbs down here. The Tesla was replaced for good reason. It's a simple engine with limited customizability, hugboxed to shit, you can just leave it alone forever. But the moment a wire is cut by an antag, the thing will loose and there is nothing to stop it. Not only does this cause widespread destruction for little effort, but it also lags the server a ton. Keep that in consideration before you add something purely for variety.

>easy as shit to loose and end a round
>hugboxed to shit
?????
these are two extremely different ideas

Tupinambis:
It's hugboxed because, there is literally no risk in anything unless you just neglect to set it up properly in the first place. The only risk is antags looking to kill their targets without any effort besides deconning a wall and snipping a wire

lightcat:
It's going to be a few days before this is updated.

silicons:
stale PR and I don't think we should have tesla back on pubby, maybe give them the parts to make one if they want or anything.
SM is the future, sorry.

PR #7825, opened 2018-12-14
Readds combat CMOs

silicons:
- Defibs don't need time to stun people anymore
- However defib paddles can only be drawn if they aren't already drawn if you are standing
- Defib stun stamina damage reduced from 75 to 50


Ignari-contrib:
But why does this need to come back :<

geemiesif:
No.

PR #7835, opened 2018-12-17
Wizard objectives are shit and nobody follows them anyway.

BlackMajor:
Literally just changes the wizard gamemode objective to
"Cause as much creative mayhem as you can aboard the station! The more outlandish your methods of achieving this, the better! Make sure there's a decent amount of crew alive to tell of your tale."
and escape.

### DIFF
--- a/code/modules/antagonists/wizard/wizard.dm
+++ b/code/modules/antagonists/wizard/wizard.dm
@@ -60,7 +60,7 @@
 	owner.current.forceMove(pick(GLOB.wizardstart))
 
 /datum/antagonist/wizard/proc/create_objectives()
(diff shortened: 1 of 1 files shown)


nik707:
When faced with the issue of Wizards not following their objective and just murderboning, rather than enforcing them to you know, do them, we instead give them even more freedom to be completely fucking autistic and murderboney. Hot take, our players are not at all trustworthy enough to do anything with this objective that isn't autistic murderboning.

nik707:
Furthermore, it's insanely subjective and literally no admin will ever enforce it or say what is and isn't "creative"

BlackMajor:
Added a small catch at the end of the objective to catch Murderbonus the 5x fireball wizard.

deathride58 (approved):
Can be merged with approval from any other maintainer

PR #7872, opened 2019-01-01
fixes a pretty bad balance issue with tesla coils

silicons:
they're no longer instakill + explosive


deathride58 (approved):
no fun allowed

PR #7888, opened 2019-01-04
Reworks ephedrine 

silicons:
- Now is a potent stamina regeneration chemical
- Has a high chance of causing confusion
- Makes you jitter
- Does not speed you up anymore


lolman360:
noooooooooooo

nik707 (approved):
@deathride58 double check this is ok before merging, it's a balance change and I don't do main much these days

deathride58 (approved):

PR #7814, opened 2018-12-11
lets try to fix heat not spreading - makes hotspot_expose() force air updates like space heaters do

deathride58:
Title. This is a pretty weird bug, as it wasn't present on TG despite the code being mostly the same

:cl: deathride58
fix: [EXPERIMENTAL] attempted to fix heat failing to spread from heat sources like plasma fires and welding tools
/:cl:

PR #7895, opened 2019-01-05
TESTMERGE ONLY - reverts #7814 in an attempt to figure out why atmos is more expensive than usual

deathride58:
Title


nik707 (commented):
less gay pls

nik707 (approved):

deathride58:
Closing since we've figured out that the PR this PR reverts isn't really the cause for atmos being expensive

PR #7900, opened 2019-01-06
Buffs the shit out of the RLD (4x material, view --> range)

silicons:
helps give use to a rarely used item.


Ragolution:
I'd disagree on it being rarely used unless I'm thinking of the wrong item. RLD is already really good if I'm thinking of the right one.

Now how will I enter my target's department with the excuse that I need to fix the lights?

deathride58 (approved):

PR #7931, opened 2019-01-13
Donator Item - Soul Necklace 

LetterJay:
Donator item


deathride58 (changes requested):
The loadout entry is missing and the neck.dm added isn't present in the .dme

deathride58 (approved):

PR #7934, opened 2019-01-14
Pray to great green mother earth

Ragolution:
- Adds one new lobby song.

I really love this song, because all it is about is a bunch of crew on a ship praying that their ship doesn't blow the fuck up.

And it's perfect for space station.
Also it's filk music, like Dawson's Christian and Carmen Miranda's Ghost.


deathride58:
Just slap the entirety of the Carmen Miranda's Ghost tape in that mfer

deathride58 (approved):

Ragolution:
> Just slap the entirety of the Carmen Miranda's Ghost tape in that mfer

I thought about it but that would add a hell of a long download time on first connect.

deathride58:
With the way the lobby sounds are set up, the sounds only get downloaded if they're actually playing when you connect. Also, re-encoding the audio so that they're all below 4mb each would be a good way to decrease the potential total download size for each connect, and since most of the tapes are only available on the internet in an extremely lossy format, it won't really make much difference in actual audio quality. All of the songs currently in the jukebox obey the 4mb or below rule, which is why clients don't lag a whole lot when they hear jukebox songs for the first time after an automated cache dump.

PR #7946, opened 2019-01-19
Nerfed Mulebots to hell and back. You know what you did.

SkullyRoberts:
:tweak: Decreased the random modifier of mulebot runover damage from rand(5,15) to rand(1,2) Taking the average damage by mulebot runover from 30 down to 5ish.

[Why]: Because mulebots deal a massive amount of damage extremely quickly, and pAI controlled mulebots are ridiculously powerful when they have a rider.
Yogs and Paradise fixed this issue by removing pAI controlled mulebots, but I think pAI mulebots are pretty fun, when people aren't abusing it.

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/mulebot.dm
+++ b/code/modules/mob/living/simple_animal/bot/mulebot.dm
@@ -644,7 +644,7 @@
 					"<span class='userdanger'>[src] drives over you!</span>")
 	playsound(loc, 'sound/effects/splat.ogg', 50, 1)
 
-	var/damage = rand(5,15)
+	var/damage = rand(1,2)
 	H.apply_damage(2*damage, BRUTE, BODY_ZONE_HEAD, run_armor_check(BODY_ZONE_HEAD, "melee"))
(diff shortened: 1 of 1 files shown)


deathride58:
This is just straight-up hugboxing. These changes make mule bots completely incapable of doing any meaningful harm, as this decreases the *maximum* damage they can do to 16, whereas their previous *absolute minimum* damage was 20. For reference, previous minimum damage against unarmored targets with a standard amount of limbs was 40, and the previous maximum damage against the same type of targets was 120.

silicons:
on second thought im not merging this

SkullyRoberts:
I just couldn't think of how else to deal with this. We have players literally killing dozens of (fairly robust) people using pAI mulebots. Once you are down a mulebot can quickly run over you back and forth, frequently killing you before you even get a chance to move.

SkullyRoberts:
It's not even like the mulebot stunlocks them. They can just kill the person before they recover from even a single batonstun.

Ragolution:
Sounds like pAI need a nerf.

PR #7950, opened 2019-01-20
Memestation II: The Memeining. Part 2: The actually stable version

SkullyRoberts:
This is a map that is not to be taken seriously.
Imagine if you took centcomm. And then you copy and pasted random parts from other maps. And then you stitched them together into an abomination.

That is this map.


SkullyRoberts:
I'm not sure why I closed the old PR. It's 1 am and I'm not really thinking straight. I have been working on this for 8 hours straight. What have I done with my life?

code/game/objects/items/devices/flashlight.dm
```diff
@@ -205,6 +205,17 @@
 	brightness_on = 5 // A little better than the standard flashlight.
 	hitsound = 'sound/weapons/genhit1.ogg'
 
+/obj/item/flashlight/seclite/flashdark
```
deathride58: flashdarks already exist under the `/obj/item/flashlight/flashdark` subtype

code/modules/memestation/arenabonfire.dm
```diff
@@ -0,0 +1,132 @@
+/obj/structure/arenabonfire
```
deathride58, line 1: Any reason this isn't a subtype of `/obj/structure/bonfire`?

deathride58 (changes requested):
The map's missing .json defines, meaning it'd be impossible to load without admin varedits

code/modules/memestation/arenabonfire.dm
SkullyRoberts: Because I didn't want to be able to add grills or the ability to put it out. arenabonfires don't need air, either

code/game/objects/items/devices/flashlight.dm
SkullyRoberts: reverted adding the flashdark

code/modules/memestation/arenabonfire.dm
SkullyRoberts: Also I'm gonna add more mechanics specific to the arenabonfire and I dunno the technical stuff of replacing procs

MimicFaux:
Does this address gravity/telecomms/atmos problems?

SkullyRoberts:
@MimicFaux Yes.

PR #7968, opened 2019-01-29
Fixing Star Trek DS9 uniforms

Zargserg:
**[Changelog]**
Fixing the mix-up of the DS9 uniforms. Command is now red, gold is now ops/security. Removed a few duplicates in the TG-Source which overwrote Citadel code.

**[Why]**
Obviously, there was a slight mix-up or a different intention (having 2250s uniforms due to the cuffs) behind it. In any case, since they are seen as part of the standard uniform that is being worn with the uniform jacket they should have the appropriate colour coding as well.


Poojawa:
Yeah I thought I fixed that when I ported, but it was probably the overrides that fucked 'em up.

Should redo these sometime

silicons (approved):

PR #7977, opened 2019-02-02
Low plasma concentration does not damage anymore

Zargserg:
[Changelogs]: 
As the title says, it changes a few values inside the lungs so trace amounts of plasma do not cause toxic damage. Minimum plasma safety is now 0.5% of the atmosphere instead of 0.05% (which is extremely hard to each even with scrubbers on contaminated-mode). This should clean up the issue with everyone accumulating toxins after a toxic spill in a public area.

:cl: optional name here
balance: Lungs maximum toxin threshold is 0.5% of the atmosphere.
fix: Permanently contaminated atmosphere does not murder crew anymore.
/:cl:

[why]: A simple fix and solution for a problem that annoys me as medical staff and that can't (or  won't) be fixed by engineering due to lacking access and/or turfs that can't be reached by scrubbers.

### DIFF
--- a/code/modules/surgery/organs/lungs.dm
+++ b/code/modules/surgery/organs/lungs.dm
@@ -15,7 +15,7 @@
 	var/safe_co2_min = 0
 	var/safe_co2_max = 10 // Yes it's an arbitrary value who cares?
 	var/safe_toxins_min = 0
-	var/safe_toxins_max = 1
+	var/safe_toxins_max = 0.5
 	var/SA_para_min = 1 //Sleeping agent
 	var/SA_sleep_min = 5 //Sleeping agent
 	var/BZ_trip_balls_min = 1 //BZ gas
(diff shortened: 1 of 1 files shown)


silicons (changes requested):
Make it MOLES_GAS_VISIBLE

Zargserg:
There you go, it's basically half of the former value now. Still gonna test how it acts in hallways, etc. since the distribution of the gas may be very uneven and thus be over 0.25%.

Well, after some testing I'm not sure if scrubber systems can keep up with a major spill in a common area. The portable scrubber apparently only affects the tile it's on as well? 
After comparing the code between big and small portable scrubber it appears the big one scrubs all the tiles around it as well, mayhaps applying this to the current small scrubber as well?

silicons (approved):
reason i wanted moles gas visible is because the issue i take with the status quo is people are getting damaged with no visible indicator of which areas are bad. this ensures people know which areas need scrubbing. as for gas flowing slowly, we can change that later with a fix.

PR #7980, opened 2019-02-05
Update ion_storm.dm

MediHound:
:cl:
[Changelogs]: # Changes done to the Ion Storm event
add: Made Cyborgs be affected by Ion Storm Law Changes like AIs
/:cl:

[why]: # This makes Cyborgs be affected by Ion Storms and gives them Ion Laws.
Why this change? First of all, it was weird that AIs and small Bots get affected by Ion Storms, but Cyborgs didn't. Secondly, people were Metagaming that unsynced Borgs were safe. The emagging chance might be a bit high, compared to before, but a 20% chance for such a rare event isn't too much in my opinion.

### DIFF
--- a/code/modules/events/ion_storm.dm
+++ b/code/modules/events/ion_storm.dm
@@ -8,11 +8,11 @@
 
 /datum/round_event/ion_storm
 	var/replaceLawsetChance = 25 //chance the AI's lawset is completely replaced with something else per config weights
-	var/removeRandomLawChance = 10 //chance the AI has one random supplied or inherent law removed
-	var/removeDontImproveChance = 10 //chance the randomly created law replaces a random law instead of simply being added
-	var/shuffleLawsChance = 10 //chance the AI's laws are shuffled afterwards
-	var/botEmagChance = 10
-	var/announceEvent = ION_RANDOM // -1 means don't announce, 0 means have it randomly announce, 1 means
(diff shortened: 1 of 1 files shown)


Poojawa:
use :cl: : cl : without spaces

:cl: Alina
add: AI-less borgs are now susceptible to Ion storms, both law changes and potential emag effects.
/:cl:

code/modules/events/ion_storm.dm
```diff
@@ -53,11 +53,42 @@
...
+
+
+	//Chance to emag a Bot
 	if(botEmagChance)
 		for(var/mob/living/simple_animal/bot/bot in GLOB.alive_mob_list)
 			if(prob(botEmagChance))
 				bot.emag_act()
 
+	//Chance to emag a Cyborg
+	if(botEmagChance)
```
silicons: Let's not.

silicons (changes requested):
Yeah uh don't buff the chances of emagging/law remove/shuffles and make sure it only works for unsynced cyborgs. No random cyborg emagging either.

MediHound:
Testing together with BlackMajor showed: AI Synced borgs get the same Laws AIs get, without losing sync. Multiple unsycned borgs can get different law changes.

silicons (approved):

PR #7982, opened 2019-02-07
Fixes clockcult to be actually playable on citadel.

Coolgat3:
[Changelogs]:
:cl: Coolgat3 / Avunia
add: Made kindle put the target into stamcrit, which makes it an actually working, useful stun.
add: Added a stamina loss modifier to the vanguard spell which makes the user's stamina drain at a way slower rate. This doesn't make them immune to tasers, but it takes a few hits to actually get them to fall down.
tweak: Made it so that clock culties don't start with a chameleon suit. Instead they start with an engineer suit which is pretty much the same sprite and looks. If this is not perfect, then I am willing to make a slightly more brass-colored version of the engineer suit sprite and call it a ratvarian engineer jumpsuit.
balance: Increased the cost of vanguard, as it is now a spell that works somewhat like adrenals, minus the move speed, making your stamina drain really slow and making you unable to get knocked onto the ground by just a single taser shot.
balance: Lowered the charge time of kindle, reason being that you can usually have only one active spell on you, or two at max if you decide to run two slabs, but the fact that kindle silences people for such a small amount of time makes up for it, in my opinion.
server: Figured out that the consoles and their warp function actually work with the current code. The thing that makes them not work is when the gamemode is ran on debug mode, without the required players to actually support it. It also breaks the ark timer which is stuck on -1 seconds until activation. Whenever the gamemode starts properly, like any other gamemode, with player checks and all, everything seems to work just fine.
/:cl:

[why]: People like clockcult. I said I will fix it, and I did, thanks to the help of my friend Avunia which helped me fix a lot of mistakes. I really want to see clockcult back in the regular gamemode rotation because it's a fun gamemode and there is a huge group of people in our community that likes it.
EDIT: And I have high hopes for the gamemode to be added back into the regular gamemode rotation and become something more than an old, fun gamemode sitting in the basement and gathering layers of dust.


JogBoba:
Stamcrit takes 30 seconds to recover from though....

Coolgat3:
Stamcrit on bloodcult stun takes 30 seconds to recover from too. The thing about clockwork cult stuns is the fact that you actually see them holding the slab. If kindle stamcritting is a bad idea then why aren't we making bloodcult stun not stamcrit yet?

silicons (approved):

PR #7986, opened 2019-02-09
Adds realism to pleasure

Poojawa:
:honk:

40 seems like a reasonable number. should make Hexa OD something to be worried about.


deathride58 (approved):
"You're too exhausted to keep fucking!"

RogueSteampunker:
The fact that this make ODing people on hexa a genuine means of assault is terrifying.

PR #7995, opened 2019-02-11
Gives combat mode examine text along with a visible message on activation

deathride58:
Title. If you haven't touched your combat mode button at all within ten seconds, this PR will make it so that enabling combat mode will make a visible message to everyone around you. This PR also makes it possible to see if someone's in combat mode by simply examining them.

:cl: deathride58
add: Combat mode is now displayed in examine text
add: Combat mode now makes a visible message when enabled if you haven't touched your combat mode button in the last ten seconds. It's done this way to avoid chat spam from those who know how to pull off stam regen squeezing.
/:cl:


Poojawa:
At first I was going to say 'reee metagaming' but then realized it makes sense for examine context clues to notice how defensive someone appears to be.

silicons (approved):

PR #7997, opened 2019-02-12
Fancier throws, bullet casings, and more

deathride58:
This PR can be described in these two images alone
image

image

Need I say more?

:cl: deathride58
add: When an item is thrown, it will now be rotated and displaced.
add: Shards of glass will now be rotated when spawned.
add: Bullet casings will now be rotated when they're ejected from a gun.
add: Bottles now have random rotations and pixel offsets when smashed via throwing.
add: Picking up an item will now reset its rotation.
add: Glasses thrown onto tables by bartenders will now have their rotation reset.
/:cl:


silicons:
🤢

silicons (approved):

PR #8002, opened 2019-02-15
nerfs estim circuits by limiting the amount of staminaloss orgasms can do

deathride58:
Title. This makes it so that orgasms still make you exhausted, but they can't actually stamcrit you

:cl: deathride58
balance: Staminaloss from orgasms is now capped at 40 total staminaloss. You can no longer be forced into stamcrit by estim circuits.
/:cl:


RogueSteampunker:
still think we should just revert the feature entirely, cause making an opt-in feature abusable for stamcrit was never a good idea in the first place, and this just proves it.

deathride58:
Closing since both maintainers are more in favor of the complete revert PR than this alternative PR

PR #8007, opened 2019-02-15
Stimulum Buff

Ragolution:
Stimulum no longer requires 1.5 units to metabolize. (That means you can use it without suffocating yourself, now, with pluoxium.)


Ragolution:
Need to make this a branch to code something else

PR #8009, opened 2019-02-17
Correctly removes the changeling brain test

silicons:
:^)

### DIFF
--- a/code/modules/mob/living/brain/brain_item.dm
+++ b/code/modules/mob/living/brain/brain_item.dm
@@ -62,8 +62,9 @@
 
 /obj/item/organ/brain/proc/transfer_identity(mob/living/L)
 	name = "[L.name]'s brain"
-	if(brainmob || decoy_override)
+	if(brainmob)
 		return
+		
 	if(!L.mind)
 		return
 	brainmob = new(src)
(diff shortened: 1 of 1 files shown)


RogueSteampunker:
Because the one way to check on changelings without eating a ban for cremating a non-ling just has to go.

Just saying, unless we wanna actually put a method on testing potential lings or actually nerf ling abilities so it's not necessary to have to test them outside of mass murderbone, this is just screaming bug exploit on a bit literally noted on the wiki to be true of ling biology.

deathride58:
That would be a valid argument if this exploit wasn't something that you can actually get banned for using.

RogueSteampunker:
Then don't ban for it. The only reason it's bannable is cause it's chosen to be, so why is it that's the only reason then?

deathride58:
It's always been bannable because it's an exploit of a bug caused by brains not properly inheriting the DNA or name of a changeling.

RogueSteampunker:
And yet it's been written multiple times that changeling brains are vestigial and aren't necessary. It's only an exploit for as long as we wanna call it one, instead of an unintended feature.

deathride58:
In terms of game mechanics, changeling brains being vestigial is reflected by changelings being able to get up and walk away at any point they wish if they lack a brain, whether it be due to their head being missing or otherwise. Being able to tell if a brain is a changeling brain or not by attempting to place it into an MMI makes little to no logical sense.

Tupinambis:
There needs to be SOME way to actually confirm whether or not someone is a ling if you suspect it. Otherwise people who weren't actually lings will end up cremated.

deathride58 (approved):

PR #8024, opened 2019-02-20
Atomizes the med rework's blood effect changes (most things that make a blood effect will now make you actually lose blood) and makes the blood reagent act more like you'd expect

deathride58:
Title. Bloodloss is actually insanely rare in the pre-medrework code. Even if you're turned into pulp and flood a room with your blood, you will *never* actually lose enough blood for bloodloss to be an actual threat. This PR should hopefully somewhat alleviate that by taking the currently cosmetic blood effects and making spacemen actually lose blood when their blood gets spilled.

This PR also makes violence a little more bloody, as the blood reagent will now actually cover items and spacemen in blood, and the blood that covers you when you attack someone will now cover more than just your item, your gloves, and your suit.

:cl: deathride58
tweak: The blood splatter effect that happens when you get attacked will now always make you lose blood depending on the damage you've taken. The effect now also scales with item damage, meaning tiny little papercuts will no longer be able to cause a massive blood splatter.
add: The blood reagent will now cover items and spacemen in blood when applied to objects and mobs.
tweak: Helmets, masks, and neck items are all now valid targets to get splattered when you get covered in blood. Groovy.
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -125,7 +125,11 @@
 			if(prob(33))
 				I.add_mob_blood(src)
 				var/turf/location = get_turf(src)
-				bleed(totitemdamage)
+				if(iscarbon(src))
+					var/mob/living/carbon/C = src
+					C.bleed(totitemdamage)
+				else
+					add_splatter_floor(location)
 				if(totitemdamage >= 10 && get_dist(user, src) <= 1)	//people with TK won't get smeared with blood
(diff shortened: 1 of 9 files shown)


silicons (approved):
YOLOOOOOOOOOOOOOOOOOO yeah I'll probably regret this :^)

Poojawa:
Should probably update blood decals in general for this too

PR #8025, opened 2019-02-20
Makes the temperature notification act realistically (and more responsive!)

deathride58:
Title. This PR makes the temperature notification act realistically by making the notification compare the ambient temperature with the body temperature instead of only taking into account body temperature. This means that, when you walk into a freezing cold room, you'll no longer have to wait an entire minute for the temperature notification to pop up; the temperature notification will pop up on the next mob life cycle. This also means that if you walk into space without a suit, then make it back onto station, you'll feel extremely hot for a moment.

:cl: deathride58
tweak: The temperature notification will now take into consideration both the ambient temperature and your body temperature, increasing the responsiveness of the temperature notification and making it much more realistic.
/:cl:


code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1703,6 +1703,21 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 				H.adjust_bodytemperature((thermal_protection+1)*natural + min(thermal_protection * (loc_temp - H.bodytemperature) / BODYTEMP_HEAT_DIVISOR, BODYTEMP_HEATING_MAX))
 			else //we're sweating, insulation hinders out ability to reduce heat - but will reduce the amount of heat we get from the environment
 				H.adjust_bodytemperature(natural*(1/(thermal_protection+1)) + min(thermal_protection * (loc_temp - H.bodytemperature) / BODYTEMP_HEAT_DIVISOR, BODYTEMP_HEATING_MAX))
+		switch((loc_temp - H.bodytemperature)*thermal_protection)
```
Poojawa, line 1706: I feel this could be better made than a super if chain. datums or sommat

deathride58: I couldnt really think of any other way I could implement it without a switch() statement

silicons (approved):

PR #8029, opened 2019-02-21
makes teleporter hubs lose their density when ready to teleport (makes it plausible for projectiles to go through teleporters)

deathride58:
Title

:cl: deathride58
tweak: Teleporter hubs are no longer dense when on, allowing projectiles to go through teleporters
/:cl:


deathride58:
Teleporter hubs don't function when density=0, holdon

deathride58:
closing since im laaaaazy

PR #8032, opened 2019-02-21
Makes transit areas physically throw you out of the transit area instead of simply teleporting you

deathride58:
This is similar to how transit areas work on some baycode servers, Goon, and older versions of TG. Except it's modernized and doesn't require the transit area to be prebaked into the map. Due to code limitations, this also means it's plausible to hang on for dear life if you accidentally step one single tile outside of a shuttle. This also means that objects that fall out of moving shuttles will soar through space when exiting transit areas instead of simply staying stationary.

EDIT: as of ed23ce2, shuttles now opt for a more realistic approach. Normal transit turfs now act exactly like normal space tiles when moving around. Exiting the transit area or being present in the transit area when the shuttle exits transit will still teleport you and throw you in the direction the shuttle was moving.

:cl: deathride58
add: Instead of transit turfs simply teleporting things to space, transit is now handled in a somewhat realistic manner. Transit turfs now act like normal space turfs, though exiting the transit area or being present in the transit area after the shuttle moves out of transit will teleport you to space and throw you in the direction the shuttle was moving in.
tweak: Reservation areas are now able to designate a border turf.
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -435,11 +435,13 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 		GLOB.the_gateway.awaygate = new_gate
 		GLOB.the_gateway.wait = world.time
 
-/datum/controller/subsystem/mapping/proc/RequestBlockReservation(width, height, z, type = /datum/turf_reservation, turf_type_override)
+/datum/controller/subsystem/mapping/proc/RequestBlockReservation(width, height, z, type = /datum/turf_reservation, turf_type_override, border_type_override)
 	UNTIL(initialized && !clearing_reserved_turfs)
 	var/datum/turf_reservation/reserve = new type
 	if(turf_type_override)
 		reserve.turf_type = turf_type_override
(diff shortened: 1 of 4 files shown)


silicons (dismissed):

silicons:
wait you know this'll stunlock carbons if they hit a wall right?

deathride58:
Aye, stunlocks for getting in front of the shuttle are a possibility. The only two options for fixing it would be to either make it so turfs adjacent to the shuttle don't throw, or to go for realism and remove the throwing effect entirely from non-border transit turfs.

Poojawa:
but, shuttles don't actually have speed. They're wrapped in FTL bubbles. When you pop out of said bubble, you lose the effects of being inside of it.

deathride58:
smh, do i need to implement ludicrous speed and maybe even the ability to go into plaid, if the shuttle's engines get emagged? muh laws of hollywood physics

~~Also, updated the OP to reflect how the PR actually functions as of ed23ce2~~

silicons (approved):

PR #8047, opened 2019-02-26
fixes body_markings being assigned as a list when in reality its a string and everything uses it as a string

deathride58:
Title

:cl: deathride58
fix: Fixed body_markings in bodyparts being assigned as a list when in reality, it's a string and literally everything expects it to be a string and uses it as a string. This should mean that markings no longer have completely fucked up caches for character preview and other things.
/:cl:


Poojawa:
@kevinz000 speed merge my dude

silicons (approved):

PR #8048, opened 2019-02-26
Adds Alina's Donator item

Poojawa:
@LetterJay 

<3

Belly overlay doesn't work, but I've been informed that it is ok for it not to've.


LetterJay (approved):

deathride58:
Merging at 22 hours since this is a donor reward, has no effects on gameplay, OP is working on another PR that makes modifications to the same DMI file, and has no code issues.

PR #8056, opened 2019-02-27
Makes the efficiency of the TEG be dependant on how much gas is actively combusting in the hot pipe

deathride58:
Title. This means that the meme strat of simply shoving hypernoblium into both sides of the TEG and then making use of heaters/coolers to keep the gas circling simply no longer works. You now have to actually be using actively combusting gas to produce any meaningful amount of power with the TEG. This directly paves the way to make the TEG a potential way to produce research points.

:cl: deathride58
balance: The TEG will now only produce a meaningful amount of power if the hot pipe contains gas that's actively combusting
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/binary_devices/circulator.dm
+++ b/code/modules/atmospherics/machinery/components/binary_devices/circulator.dm
@@ -61,6 +61,8 @@
 		//Actually transfer the gas
 		var/datum/gas_mixture/removed = air2.remove(transfer_moles)
 
+		removed.react(src)
+
 		update_parents()
 
 		return removed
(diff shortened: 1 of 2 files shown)


IrkallaEpsilon:
So basicly thermodynamics breaking TEGs with a self feeding/self powering heater/coolers while still outputting more energy than it takes arent a thing anymore? Cool

silicons:
@deathride58 wouldn't this meme on setups that combust the gas to a high amount before putting it in and just incentivize using burnmix?

deathride58:
Directly feeding burn mix through the hot pipe is the most efficient setup with this PR, yeah

silicons (approved):

PR #8058, opened 2019-02-27
Makes jogging let you avoid water slips, but only if you have less than 20% staminaloss

deathride58:
Title. With this PR, sprinting over water is still a guaranteed slip, and walking will still let you pass. This PR makes a further distinction between jogging and sprinting beyond a simple +1 movement speed by adding a state between sprinting and walking that water slips take into account. The value of 20% staminaloss is used since it's a spot where exhaustion starts kicking in but isn't really felt, which makes it harder to retain your balance if you're moving too fast. The value is low enough to the point where water is still something to be legitimately wary of during combat, but also high enough to the point where you won't have to lynch the janitor if he even thinks about using a mop.

:cl: deathride58
add: Jogging is no longer treated exactly the same as sprinting for water slips. When you're jogging, you will only slip on water if you have more than 20% staminaloss.
/:cl:

### DIFF
--- a/code/game/turfs/open.dm
+++ b/code/game/turfs/open.dm
@@ -245,6 +245,10 @@
 				return 0
 			if(C.m_intent == MOVE_INTENT_WALK && (lube&NO_SLIP_WHEN_WALKING))
 				return 0
+			if(ishuman(C) && (lube&NO_SLIP_WHEN_WALKING))
+				var/mob/living/carbon/human/H = C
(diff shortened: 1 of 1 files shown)


silicons (approved):

Tupinambis:
I'm so very sorry I missed this PR, but it should be reverted. There was no reason to make this change and it adds a lot of inconsistency on the side of the janitor or slipee. It's basically just a janitor nerf with a side order of attempting to justify combat rework mechanics. Why?

PR #8095, opened 2019-03-05
Adds new and immersive sound for combat

nik707:
Dink!

CS:GO Dink sound ported to main. 1/50 chance on melee hit to the head, 100% chance with projectiles.
Gamers.
It works for suicide, and for executions. Enjoy your radio callouts.
"Two left, engine room, one dinked."


silicons (approved):
wtf why no tf2

PR #8098, opened 2019-03-08
Equalizes security helmet and beret armor

Coolgat3:
[Changelogs]: 
:cl: Coolgat3
fix: Made the sec and warden berret offer as much protection like the helmet
/:cl:

[why]: Tat's appearently what they were suppsoed to do according to Kevin so yeah. I think it's retarded that a beret offers more meele protection than a helmet. I know it's "lined with armor" but like, let's be real here. Helmet would protect you more since it covers more of your head, so making the values the same as the one for the helmet is good, cause it still lets people choose whether they prefer the looks of the helmet or the beret.

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -3,7 +3,7 @@
 	desc = "Standard Security gear. Protects the head from impacts."
 	icon_state = "helmet"
 	item_state = "helmet"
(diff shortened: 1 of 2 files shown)


silicons (changes requested):
fuck it let's buff helmets to 40

Coolgat3:
> fuck it let's buff helmets to 40

Done

silicons (approved):

PR #8103, opened 2019-03-10
Making overthrow actually fun like it was meant to be

Coolgat3:
[Changelogs]: 

:cl: Coolgat3
tweak: Changed the overthrow boss TC from 5 to 24
/:cl:

[why]: Overthrow is meant to be a FUN and rather chaotic gamemode. The description literally says "The agents are supposed to take over the station when a normal assault team can not be sent" So in my opninion giving the boss acces to 24 TC which will allow them to arm up their team well will actually make the gamemode a lot more fun and chaotic. 24 tc is more than that one lone traitor but not enough for the romerol kit. I doubt overthrow teams will ever cooperate so any "48 TC team" situations shouldn't really happen. I mean hell. Overthrow happens rarely, it's already better than revs, but it's rather stale beacuse of the fact that to actually dmanage anything you need to raid armory to get proper weapons, that leaves the other team/s pretty much weaponless if you manage to grab all the stuff from the armory as one team.

### DIFF
--- a/code/modules/antagonists/overthrow/overthrow.dm
+++ b/code/modules/antagonists/overthrow/overthrow.dm
@@ -1,4 +1,4 @@
-#define INITIAL_CRYSTALS 5 // initial telecrystals in the boss' uplink
+#define INITIAL_CRYSTALS 24 // initial telecrystals in the boss' uplink
 
 // Syndicate mutineer agents. They're agents selected by the Syndicate to take control of stations when assault teams like nuclear operatives cannot be sent.
 // They sent teams made of 3 agents, of which only one is woke up at round start. The others are, lore-wise, sleeping agents and must be implanted with the converter to wake up.


kiwedespars:
honestly yeah.

overthrow is garbage because it's IAA but with groups.

make it an actual replacement for revs if that's what we're doing.

deathride58:
Overthrow is actually designed in such a way that nonlethal play is the optimal strategy. 24 TC roundstart is simply way, *way* too much for an on-station antag, even for a loud one.

Coolgat3:
> Overthrow is actually designed in such a way that nonlethal play is the optimal strategy. 24 TC roundstart is simply way, _way_ too much for an on-station antag, even for a loud one.

I know it's "meant to be stealthy" but how does one stealthily get rid of all heads is something that makes me wonder. I think making overthrow one of all those high threat gamemodes like nukies and cult would actually be really fun and i think people would really enjoy it to be honest. Of course if its really a no can do then we can lower the number, but i sure as hell don't agree that 5 tc is enoigh to do anything super fun at all as a team of 3. As of now being a part of an OT team feels like a job more than having fun as an antagonist

deathride58:
It isn't meant to be stealthy, it's meant to be nonlethal. Stealthy and nonlethal are two separate things entirely.

We already have more than enough hyperlethal gamemodes as-is. A _much_ better alternative to OT teams starting off with 24(!!!!!!!!!!!!!!!) TC total roundstart would be to make it so that successful conversions and other objectives grant all team members of a given team extra TC to use, to encourage teams to actually pursue their objectives instead of just going loud and murderboning shortly after the round starts. The reason why OT teams start off with only 5 TC is to encourage mutineers to opt for nonlethal methods and think about all of their actions instead of just turning the entire station into a literal TDM with their roundstart TC.

Coolgat3:
To be honest I'd want to make tc a reward for objectives in OT but I have no idea how to really. As I said I am not a great coder, just a noob who knows how to edit shit there and there and change something that already exists to make it work like i want it to work

silicons:
hard pass unless you do the work to make the overthrow uplink not having the shit you can get to murderbone with 24 TC.

Coolgat3:
> hard pass unless you do the work to make the overthrow uplink not having the shit you can get to murderbone with 24 TC.

a.k.a most of the conspicous and dangerous weapons, aye?

PR #8110, opened 2019-03-12
Revert "Merge pull request #8089 from Trilbyspaceclone/patch-70"

Ragolution:
Oldstation is unplayable until this is fixed, please revert.


Poojawa:
how is it unplayable?

Ragolution:
PACMAN generator is unusable, meaning you can never open the bridge to get the tools out of it and, thus set up the solar generators. I mean this also makes PACMANs unusable, if that adds credence.

nik707 (approved):

Ragolution:
ooga

PR #8123, opened 2019-03-16
Makes sprinting not take stamloss in zero gravity or when pulled

deathride58:
Movement code is absolute shit but this is the best I can do with a one line change

:cl: deathride58
tweak: Sprinting no longer takes stam when you're being pulled or when you're in zero gravity. Other sources of involuntary movement are not affected.
/:cl:

### DIFF
--- a/modular_citadel/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/modular_citadel/code/modules/mob/living/carbon/human/human_movement.dm
@@ -4,7 +4,7 @@
 /mob/living/carbon/human/Move(NewLoc, direct)
 	var/oldpseudoheight = pseudo_z_axis
 	. = ..()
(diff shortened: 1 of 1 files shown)


Ragolution:
Moving in zero-g should still cost stamina since you're moving your body, still. Does it/ it should count tiles you traverse by actually moving?/.

deathride58:
Sprinting decreases stamina for every tile moved, which is why moving 50 tiles while sprinting takes the same exact amount of stam even if you're affected by heavy slowdown.
Sprinting not taking stamloss in zero gravity is primarily so that jetpacks, baseball bats, and other forms of fast space travel don't take all your stamina when you're sprinting

Ragolution:
Is there any way you could make it so that you still lose stamina while moving hand-over-hand otherwise Antags will just 100% disable gravity every round.

Poojawa:
you should instead trip and fall and just get dragged behind someone sprinting.

silicons (approved):

PR #8130, opened 2019-03-18
Flavor Text is hidden if your face is obscured

Poojawa:
:cl: Poojawa
fix: fixed flavor text appearing when your face is hidden but you're not an Unknown
/:cl:

Even more Metabusting. Hooray


deathride58 (approved):
LGTM

PR #8133, opened 2019-03-18
Bye have a beautiful time

Ragolution:
Adds new round-end sound also wtf why is this in the ticker subsystem. I mean I get why but it seems silly.

https://www.youtube.com/watch?v=cepZIk0PlvM


Ragolution:
Accidentally compiled runtime-station wew.

deathride58 (approved):

PR #8140, opened 2019-03-19
lets try fixing lag during highpop - makes mouse movements only call onmousemove in combat mode

deathride58:
title

:cl: deathride58
code: In an attempt to improve performance during highpop, mouse movements will now only call onmousemove() while a user is in combat mode
/:cl:


nik707 (approved):
yea ok

PR #8144, opened 2019-03-19
Changing Implants to not be located in the mob's contents. N°2

Ghommie:
:cl:
refactor: Refactors implants to not be located into the implanted's contents.
fix: This translates to them being unaffected by explosions and maybe other stuff going byond coders foresigth. The storage implant's pocket and chem implant's container are still contained within the mob though.
fix: Fixes storage implants actually dropping the stored items into nullspace on removal.
/:cl:

Completing the switch from contents checks to implants list checks and fixing a couple issues with implants. Alternative to #8139. Drafted for safety till I can say it's good to go.
EDIT: Travis failing is simply because I decided delete the previous branch and make a new one by reflexes before it finished the run.

### DIFF
--- a/code/datums/outfit.dm
+++ b/code/datums/outfit.dm
@@ -104,7 +104,7 @@
 			H.update_action_buttons_icon()
 		if(implants)
 			for(var/implant_type in implants)
-				var/obj/item/implant/I = new implant_type(H)
+				var/obj/item/implant/I = new implant_type
 				I.implant(H, null, TRUE)
 
 	H.update_body()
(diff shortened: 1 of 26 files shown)


Poojawa:
this would be good to kick upstream as well, though I would say implants being destroyed/damaged in explosions is a reasonable expectation for a lot of things

Ghommie:
> 
> 
> this would be good to kick upstream as well, though I would say implants being destroyed/damaged in explosions is a reasonable expectation for a lot of things

Yea, I'll gadly kick this on /tg/station as well when this is final, just started this here as there is no sync now.
And while I too think implants getting damaged by explosions is not necessarily a bad feature, it's complementary to explosions themselves and wasn't clearly foreseen here, so changes may be applied throughout this PR to allow it to an extend.

silicons:
inactivity

PR #8148, opened 2019-03-20
Adds a shitty shuttle with some meteor defenses.

Ragolution:
Probably doesn't work but here we go anyway.

### DIFF
--- a/_maps/shuttles/emergency_stormrider.dmm
+++ b/_maps/shuttles/emergency_stormrider.dmm
@@ -38,18 +38,19 @@
 "L" = (/obj/structure/shuttle/engine/heater,/turf/open/floor/plating,/area/shuttle/escape)
 "M" = (/obj/structure/shuttle/engine/large,/turf/open/space/basic,/area/shuttle/escape)
 "N" = (/turf/open/space/basic,/area/shuttle/escape)
+"O" = (/obj/machinery/door/airlock/shuttle/glass,/obj/docking_port/mobile/emergency{name = "Stormrider"},/turf/open/floor/engine,/area/shuttle/escape)
 
 (1,1,1) = {"
 abcccbadd
 efghgfedd
 bcccfcbdd
-bijklmbdd
-blnllmbdd
-bbbbobbdd
-bpqblrbdd
-sttslmbdd
+bijklmbcd
+blnllmbcd
+bbbbobbcd
+bpqblrbcd
+sttslmbcd
 bbbbubbdd
-vlllllbed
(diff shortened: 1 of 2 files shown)


Ragolution:
WHAT'S A MAP COMPILER?

_maps/shuttles/emergency_stormrider.dmm
```diff
@@ -0,0 +1,66 @@
+"a" = (/obj/structure/lattice,/obj/machinery/satellite/meteor_shield{active = 1},/turf/open/space/basic,/area/shuttle/escape)
```
Ghommie, line 1: Make a subtype of meteor_shield that toogles on Initialize() instead, so it can update icon, anchored status and its floaty animation.

code/datums/shuttles.dm
```diff
@@ -219,6 +219,13 @@
 	admin_notes = "This shuttle will likely crush escape, killing anyone there."
 	credit_cost = -5000
 
+/datum/map_template/shuttle/emergency/stormrider
+	suffix = "stormrider"
+	name = "Asteroid With Engines Strapped To It"
```
Ghommie, line 224: Some fluff you missed here.

code/datums/shuttles.dm
```diff
@@ -219,6 +219,13 @@
...
+	description = "A cramped, reinforced shuttle with PDS installed at key points and absolutely no windows. This shuttle was designed for extreme exfiltration through dense debris fields by small teams of well-trained EVA operatives. Comes with a sleeper, a big ass engine, two lockers with internals and access hatches to make sure the asteroid defense systems didn't mysteriously not get activated when the shuttle departed Central Command."
+	admin_notes = "An antag-bombers wet dream. But at least it stops meteors..."
```
Ghommie, line 226: It's bigger and sturdier than the  "Oh Hi Daniel", calling it an "antag-bomber wet dream" may be misleading, It's just a bit stripped down of furniture.

Ghommie (commented):
Since you need to convert the map file to the `tgm` format anyway, because it's more robust against eventual conflicts..
https://tgstation13.org/wiki/Map_Merger#Manual_Use

code/datums/shuttles.dm
Ragolution: Thanks.

deathride58 (changes requested):
Use mapmerger

deathride58 (changes requested):
MAPMERGER

PR #8149, opened 2019-03-20
Fixes Semen Vent Foam

Ragolution:
Works locally, like this. Fuck if I know why it needed these beyond 'something to do with the event system'.

- Makes it so semen clog event works as intended.

- Semen decals should no longer cause runtime errors.

### DIFF
--- a/code/modules/events/vent_clog.dm
+++ b/code/modules/events/vent_clog.dm
@@ -75,7 +75,7 @@
 	reagentsAmount = 250
 
 /datum/round_event_control/vent_clog/beer
-	name = "Foamy beer stationwide"
+	name = "Clogged Vents: Beer"
 	typepath = /datum/round_event/vent_clog/beer
 	max_occurrences = 0
 
@@ -83,12 +83,12 @@
 	reagentsAmount = 100
 
 /datum/round_event_control/vent_clog/plasma_decon
-	name = "Plasma decontamination"
+	name = "Anti-Plasma Flood"
 	typepath = /datum/round_event/vent_clog/plasma_decon
 	max_occurrences = 0
 
 /datum/round_event_control/vent_clog/female
-	name = "FemCum stationwide"
+	name = "Clogged Vents; Girlcum"
 	typepath = /datum/round_event/vent_clog/female
 	max_occurrences = 0
 
@@ -96,7 +96,7 @@
 	reagentsAmount = 100
 
 /datum/round_event_control/vent_clog/male
-	name = "Semen stationwide"
+	name = "Clogged Vents: Semen"
 	typepath = /datum/round_event/vent_clog/male
 	max_occurrences = 0
 
--- a/code/modules/events/vent_clog.dm
+++ b/code/modules/events/vent_clog.dm
@@ -118,6 +118,9 @@
 			foam.start()
 		CHECK_TICK
 
+/datum/round_event/vent_clog/male/announce()
+	priority_announce("The scrubbers network is experiencing a backpressure surge. Some ejection of contents may occur.", "Atmospherics alert")
+
 /datum/round_event/vent_clog/male/start()
 	for(var/obj/machinery/atmospherics/components/unary/vent in vents)
 		if(vent && vent.loc)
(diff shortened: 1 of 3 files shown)


Ragolution:
Yeah this is fine.

code/modules/events/vent_clog.dm
```diff
@@ -118,6 +118,9 @@
 			foam.start()
 		CHECK_TICK
 
+/datum/round_event/vent_clog/male/announce()
+	priority_announce("The scrubbers network is experiencing a backpressure surge. Some ejection of contents may occur.", "Atmospherics alert")
```
Ghommie: You don't have to copypaste, it's already implied for the proc to be the same of the parent's, unless you intended to replace `ejection` with `ejaculation` or some other override.

code/modules/events/vent_clog.dm
```diff
@@ -130,6 +133,9 @@
 			foam.start()
 		CHECK_TICK
 
+/datum/round_event/vent_clog/female/announce()
+	priority_announce("The scrubbers network is experiencing a backpressure surge. Some ejection of contents may occur.", "Atmospherics alert")
```
Ghommie: Idem

modular_citadel/code/modules/reagents/reagents/cit_reagents.dm
```diff
@@ -37,10 +37,11 @@
 	..()
 	dir = pick(1,2,4,8)
 
-/datum/reagent/consumable/semen/reaction_turf(turf/T, reac_volume)
+/* /datum/reagent/consumable/semen/reaction_turf(turf/T, reac_volume)
```
Ghommie: Just delete this commented code, it serves no purpose anymore.

Ghommie (commented):
Ew...

code/modules/events/vent_clog.dm
Ragolution: I did actually I guess I typographical'd it. Thanks for the catch.

code/modules/events/vent_clog.dm
Ragolution: It was originally trilby code :')

modular_citadel/code/modules/reagents/reagents/cit_reagents.dm
Ragolution: Aite.

code/modules/events/vent_clog.dm
Ragolution: Thing is it doesn't proc succesfully without this. It makes it spawn smoke instead of foam, using the old scrubber backfire pre-goof foam. This is some Magic Byond Coding to me.

modular_citadel/code/modules/reagents/reagents/cit_reagents.dm
Ragolution: I found out that was originally from a merge conflict that was resolved incorrectly. Just pennies for thought.

code/modules/events/vent_clog.dm
Ghommie: All this proc does is announcing what's going on to the crew, It should not be affecting anything else.
But if what you say is right, then something would be really, really off, specially since the only smokeful vent clogging event nowadays is the plasma decontamination one.

Ragolution: Yeah man I have no fucking idea.

modular_citadel/code/modules/reagents/reagents/cit_reagents.dm
```diff
@@ -37,11 +37,6 @@
 	..()
 	dir = pick(1,2,4,8)
 
-/datum/reagent/consumable/semen/reaction_turf(turf/T, reac_volume)
```
deathride58, line 40: This completely removes the ability to splash semen on turfs. Fix the runtime properly.

deathride58 (changes requested):

modular_citadel/code/modules/reagents/reagents/cit_reagents.dm
Ghommie: See line 13 for the intended version of the proc. This seems like an older one which hasn't been properly removed before.

Ragolution: Works fine for me, chief.
@deathride58 https://i.imgur.com/lfnSiKv.png
The only thing I couldn't fix was the lack of DNA in Semen splats, but this PR doesn't aim to fix the "No DNA in semen" problem and I intend to make a different PR for that, at another time.

Ragolution:
...

deathride58 (approved):

PR #8155, opened 2019-03-21
forcefields can no longer be stacked

silicons:
see how it's possible to nerf something with a nerf that actually works?


code/game/objects/items/devices/forcefieldprojector.dm
```diff
@@ -34,6 +34,10 @@
 	if(LAZYLEN(current_fields) >= max_fields)
 		to_chat(user, "<span class='notice'>[src] cannot sustain any more forcefields!</span>")
 		return
+	var/obj/structure/projected_forcefield/F = locate() in T
```
Ghommie: compiling errors as line 44 defines a var of the same name. Use another letter or word.

Poojawa:
you should probably test this to be sure.

deathride58 (approved):
LGTM

PR #8157, opened 2019-03-21
Adds rule 3 from the Lavaland Role Antag Policy to the ash walker spawn text

izzyinbox:
Hotel staff have an equivalent, and it's a problem that happens enough here that this can help change.

:cl: 
tweak: add additional flavor/rules spawn text to ashwalkers
/:cl:


deathride58 (changes requested):
Modularization is no longer necessary on main

deathride58:
Also tbqh this could probably be handled much better by simply making it so that ashwalkers are unable to safely survive within the station's environment by giving them working lungs, making their breathing tuned to match lavaland's atmosphere, increasing their body temperature, adjusting their temperature damage thresholds so that they take temperature damage on station, and adjusting their pressure damage thresholds so that they take pressure damage on station.

Poojawa:
o yeah I was gunna do that

at least the lung stuff, not the body temp fuckery

silicons:
inactivity

PR #8159, opened 2019-03-21
Carbon Oxygen Toxicity

Poojawa:
:cl: Poojawa
balance: Ashwalkers now have lungs. They cannot breath station air without suffocation effects, but are completely fine on their homeworld.
balance: Carbon mobs now have a maximum tolerance to oxygen of 50kPa.
balance: Deluxe synthetic lungs have a very high bonus to O2 tolerance.
/:cl:

One of those 'Oh hey' things that will permit ashwalkers to potentially have server rules revised, reminded by #8157 and from previous discussions

This also ended up including a need for a 'cap' for Oxygen on carbon mobs in general. any breath overrides are handled accordingly. 

Excessive oxygen breathing will start applying a dizzy effect, getting steadily worse as you choke on the excess gas. Putting on internals at a tolerable level will negate the oxygen loss as normal, with the dizziness clearing up rather quickly as well.


Ragolution:
dude nice

Ragolution (approved):
I like it.

nik707 (approved):

silicons:
Ashwalkers being weak to this is fine but otherwise it seems like just making things harder for the sake of it, why is this needed?

Poojawa:
because breathing code doesn't have any option for 'too much' oxygen. the 50kPa cap was taken from a quick delve into wikipedia for a pure oxygen toxicity measure. Considering people start starving to death in 30 minutes of RL time in game, the 'hours' of exposure to those levels are more or less the same. Low pressure kills you faster than excessive oxygen will. Equipping internals at standard pressures instantly nullifies the oxygen issues.

Not having said cap immediately started choking all normal mobs, as I found out with monkeys.

There is no nitrogen cap, since it's an inert gas for breath mix. 

The initial problem is that, as I paged you a few times over, damage is capped to 10. But since it's an oxygen breathing lung, and there's enough oxygen. you immediately heal all damage in the old code. Now it deliberately sabotages properly, instead of just instantly healing.

MimicFaux:
We can use ashwalker parts to enforce gulag sentences now. Perfect.

deathride58 (approved):

PR #8164, opened 2019-03-22
[Ready] Vore interactions, feeding, UI cleanup!

Poojawa:
:cl: Poojawa
fix: Vore Panel restored to have various interactions available again. Semi-untested but should work as advertised.
tweak: Vore Panel has more feedback.
add: is_wet var to bellies, toggled in the panel, will remove flesh sounding struggles and the internal loop. JSON version updated.
add: Feeding var. You will need to enable feeding to recieve any feed vore actions, but you can now feed yourself to mobs that have this set. TODO, Dogborg sleeper feeding.
add: vore mode button now required to be enabled to perform vore actions. It's the mouth icon!
fix: Ash Drake vore fixed for actual reals this time
fix: Mobs shouldn't spam the released contents announcement anymore on qdel or death. Only if triggered
tweak: Hostile mob code now properly ignores targets in bellies
add: Your belly can quietly growl if you're starving now.
/:cl:

new button that enables vore is as unobtrusive as possible. It will also switch with your preferred hud color as well. Toggling combat mode will automatically disable vore mode, you also cannot toggle it while in combat mode. It is required on for any vore actions to actually be carried out as it hard-overrides grab functions. when you have the vore mode toggled on, you will not be able to grab people above Aggressive (a control hint)

when toggled off

new Vore menu UI features

`Is Fleshy` denotes if this belly is actually a belly. If it's intending to be a pocket, or underwear, or shoes, or whatever set this to No and the prey struggles will just be bag rustling noises, they also won't by default get the vore audio internal loop playing.

You are able to feed people now! Same process as before, except you click on someone else. If the predator of your feeding doesn't have feeding enabled, you'll get feedback. The target and feeding-prey will also get notifications that someone tried to feed you but you have it disabled. 

cleaned up a bunch of long commented out code, just better this way. also reverted the various TRUE/FALSE because they probably were part of the problem.


Poojawa:
ok but y tho travis

code/__DEFINES/citadel_defines.dm
```diff
@@ -12,6 +12,7 @@
 #define ui_boxcraft "EAST-4:22,SOUTH+1:6"
 #define ui_boxarea "EAST-4:6,SOUTH+1:6"
 #define ui_boxlang "EAST-5:22,SOUTH+1:6"
+#define ui_boxvore	"EAST-4:22,SOUTH+1:6"
```
deathride58, line 15: wouldnt this value make the button overlap with the crafting button when the resolution is set to 1:1? or am i dumb and missing something?

Poojawa: No because the button is flipped

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1144,6 +1144,13 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 			H.update_inv_w_uniform()
 			H.update_inv_wear_suit()
 
+	if(H.noisy && H.nutrition <= NUTRITION_LEVEL_HUNGRY)
+		if(prob(10))
+			playsound(get_turf(H),"hunger_sounds",15,0,-5,1,ignore_walls = FALSE,channel=CHANNEL_PRED)
+	else if(H.noisy && H.nutrition <= NUTRITION_LEVEL_STARVING)
```
deathride58: shouldnt these be flipped? starving is a lower value than hungry

modular_citadel/code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -651,7 +657,7 @@
 
 		if(failure_msg)
 			alert(user,failure_msg,"Error!")
-			return FALSE
+			return 0
```
deathride58, line 660: Why replace all instances of TRUE/FALSE with 1/0?

deathride58 (changes requested):
Two tiiiiny things. Other than that, this PR looks good.

modular_citadel/code/modules/vore/eating/vorepanel_vr.dm
Poojawa: Because having them as TRUE/FALSE played merry hell at times. With just 1/0 they work as intended with no glitches.

code/modules/mob/living/carbon/human/species.dm
Poojawa: Suppose they could be flipped, yee.

deathride58 (approved):

PR #8166, opened 2019-03-23
Fixes empty beaker blackpowder explosion exploit

silicons:
It now reacts on the tile where it started reacting, rather than on the tile of the empty beaker.
It's either this or someone makes the beaker/reacting object have a very noticeable sprite while it's "primed".

### DIFF
--- a/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
+++ b/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
@@ -5,7 +5,7 @@
 	var/modifier = 0
 	var/noexplosion = FALSE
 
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/recipes/pyrotechnics.dm
```diff
@@ -423,6 +427,7 @@
 	required_reagents = list("teslium" = 1, "water" = 1)
 	strengthdiv = 100
 	modifier = -100
+	noexplosion = TRUE
```
silicons, line 430: prevents teslium from exploding

deathride58 (approved):

PR #8172, opened 2019-03-26
makes the succumb verb available in the IC tab

deathride58:
why was this a hidden verb

:cl: deathride58
tweak: The succumb verb is now available in the IC tab
/:cl:


silicons (approved):

silicons:
probably because otherwise people could autocorrect to this but uhh

Thalpy:
Pressing the button does nothing, even in crit/hardcrit.

PR #8175, opened 2019-03-27
Attempts to optimize the lighting process by ditching the inverse-square falloff algorithm in favor of a linear one

deathride58:
Title. For the laymen, the most noticeable difference is that lighting is going to _appear_ darker across the board due to how lighting will be a little more brighter closer to the individual light sources. This also means that lighting should be a LOT faster since instead of calling sqrt() thousands of times per second, it'll just be doing basic division. This is primarily a hold-over until I can get europalights ready, but should result in _massively_ improved lighting performance.
[Here's a handy-dandy graph showing the differences in falloff algorithms](https://www.desmos.com/calculator/husolbbxet)

:cl: deathride58
tweak: Lighting now uses a linear algorithm to calculate falloff instead of an inverse-square algorithm.
/:cl:

### DIFF
--- a/code/modules/lighting/lighting_source.dm
+++ b/code/modules/lighting/lighting_source.dm
@@ -28,6 +28,11 @@
 
 	var/needs_update = LIGHTING_NO_UPDATE    // Whether we are queued for an update.
 
+// Thanks to Lohikar for flinging this tiny bit of code at me, increasing my brain cell count from 1 to 2 in the process.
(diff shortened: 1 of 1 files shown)


silicons:
hi are we stil ldoing this?

deathride58:
It's ready for merge, yeah

silicons (approved):

PR #8195, opened 2019-04-02
Adds the ability to quickly use an unequipped item on something by click-dragging it with an empty hand

deathride58:
Title. This is a huge QoL improvement when it comes to doing things that require repeated inventory management, such as shotgun reloading and complex deconstruction, as instead of having to click the item, wait for click delay, then click the object you want to interact with, you can do it in one smooth motion by simply clicking and dragging the item onto the object you want to interact with, which will place the item in your hand and then instantly use that item on the item you've dragged to. This also makes it far easier to deal with slapcrafting-compatible items by simply clickdragging the ingredients onto the slapcraftable item instead of bumbling around in your inventory

:cl: deathride58
add: You can now quickly use unequipped items on objects by simply click-dragging the item onto the object with an empty active hand. Doing so will place the item in your hand, and then use that item on the object.
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -379,6 +379,26 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 /obj/item/proc/on_found(mob/finder)
 	return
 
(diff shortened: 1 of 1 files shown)


silicons (approved):
merging this in like 8 hours but i think this should get a testmerge just in case

PR #8211, opened 2019-04-04
Medical belt overlays + hypo storage

Poojawa:
:cl::Poojawa
add: Medical belts now have overlays! Now you can see if a belt is useful or just flatout empty.
fix: fixed hypo MK IIs not being able to be put on medical belts
/:cl:

Chemistry bag also supported, but it does make the sprite a little too cluttered.

The pouch represents all 'loose' items, like the medical stacks (ointment, brute patch, gauze), syringes, loose pills and patches. the bottle is for any glass item, beakers, bottle, hypovials. That sort. each medspray has its own as well.


deathride58:
Good poojers

deathride58 (approved):

PR #8212, opened 2019-04-05
Comments out hunger noises until an intermediate coder adds a preference for it

CameronWoof:
No one who knows how to add the preference is interested in doing it, so this is unfortunately the next-best thing. 

To be clear, the existing preference only governs whether or not you **produce** the noises, **not** whether or not you **hear** the noises. People don't like hearing them.


nik707:
my brand!

silicons (approved):

PR #8229, opened 2019-04-09
[S] *vzdokh

deathride58:
Easy patch, whether or not this will actually fix the new paper crash is yet to be discovered.

:cl: deathride58
fix: Hopefully fixed paper crash 2: electric boogaloo
/:cl:

To codebases using non-AGPL code: Yes, you have permission to port this as-is. It's insanely simple to the point where worrying about whether or not you can port this is just silly, but everyone has my express permission to copy-paste this PR as-is.


nik707 (approved):
thank u robit

PR #8232, opened 2019-04-11
Vore capable simple mobs & fixes

Poojawa:
:cl: Poojawa
add: Vore Mobs are a thing now, not just ash drakes
del: Removed hunger sound stuff. never taking requests from Coolgat3 ever again.
refractor: Simple Mobs use the same system regular human mobs do, itjustworks.png
/:cl:

So to preface before the screaming starts. This has the exact same opt-in as Ash Drakes. 

now that your vore kneejerk has been aborted, have a nice day.

For everyone else, when your devourable setting is on, some mobs now have a chance (40%) per melee attack to instead attempt to devour you. They don't pause while attempting this, so you'll probably be munched on two or three times before the timer happens.

resist works as normal, and the ooc escape works too. unlike Virgo these mobs will respect digestion prefs, but if I get complaints of abuse I'll make it like how Ash Drakes are. You have been warned.

Should you find yourself in a simple mob that is vore capable, you will be given an Animal Nom verb if you do not have a grab ability. the same restrictions apply.

You can also feed yourself to vore capable mobs, should you wish. Expect admin bwoinks if you do so and it violates combat vore rules ( I.E. don't do it if you're an antag).


Coolgat3:
>Requests from Coolgat3.<
I literally asked you to fix the button, not to make the sounds permamently enabled so don't imply I did something wrong, thanks.

silicons (approved):

PR #8235, opened 2019-04-11
Buffs defib users by making them stop pulling automatically

Poojawa:
:cl: Poojawa
balance: Defib users automatically stop pulling when they attempt to perform a defib shock
/:cl:

really just QoL for lag and busy doctors. Anyone else holding onto the patient will still get shocked, but the user won't. Doesn't make sense they'd still be pulling a person with handfuls of shock paddles anyway.


silicons (approved):
ugh fine hugbox

PR #8236, opened 2019-04-11
[EXPERIMENTAL] Unomos - Listmos except gas mixtures only use one single list for gasses instead of 2 per gas + 1

deathride58:
So I was wondering why heat_capacity was as expensive as it currently is, and then that shiny little [GAS_META] caught my eye. After a while of poking around, I discovered that one of the main reasons atmos is as expensive as it is is due to the rampant abuse of lists within lists. There is currently nothing in TGcode that changes anything within [GAS_META] on the fly, and there probably won't be any at all in the future regardless of if this makes it in or not.
This PR directly combats the widespread list abuse within atmos by making it so that gasses only actually use one single list for storing the amount of moles for each given gas, and also axes the gas archive list to further shave down the list usage of each gas mixture datum for good measure. The effects of removing gas mole archives is yet to be observed, but here's to hoping it doesn't affect anything too major.
So, improvements! Here's a list of notable differences in memory and performance with this one simple change.
### Roundstart with listmos (boxstation)
Memory usage average: roughly 735mb. ranging between 740mb and 715mb.
gas mixture share() average (111850 calls) - 0.0000234957532409477s
gas mixture heat_capacity() average (502222 calls) - 0.000003926550409978058s

### Roundstart with unomos, initial commit (boxstation)
Memory usage average: roughly 710mb. Ranging between 725mb and 685mb.
gas mixture share() average (532779 calls) - 0.00001515450120969483s
gas mixture heat_capacity() average (1420621 calls) - 0.00000236868242831832s

### Roundstart with unomos, commit 905f32b193 (boxstation)
(Due to my own laziness, this is just the raw data i've collected without any prettification, and does not include memory usage)
share() cost (167358 calls) - 1.141266028513725e-5
share() cost (331348 calls) - 1.166749157984958e-5
share() cost (488991 calls) - 1.170164686057617e-5
share() cost (547420 calls) - 1.156150670417595e-5
heat_capacity() cost (362223 calls) - 2.009811635373789e-6
heat_capacity() cost (734198 calls) - 2.019891092048739e-6
heat_capacity() cost (936373 calls) - 2.034445674960726e-6
heat_capacity() cost (1092239 calls) - 2.058157601037868e-6
process_cell() cost (390727 calls) - 2.021616115599895e-5

So, this practically cuts the cost of heat_capacity almost in half while also shaving a massive chunk off of share's cost. This also, of course, decreases the list usage quite a bit. With a 255x255 area filled with standard airmix, that's roughly 400k lists from storing all of the gas moles. With that same area filled with every single gas in the game, it's 1.6 million lists. With Unomos, that 255x255 area only uses 65,025 lists total for all of the gas mole lists, regardless of how many gasses are present in all of the gas mixtures.

Doing some local testing, all atmos related things appear to be working completely fine. Plasma floods work as expected, plasma fires work as expected, breaches work as expected, bonfires work as expected, etc etc etc. I have yet to see if there's any weird bugs that crop up as a result of this PR, but normal gameplay should be more or less completely unaffected by this PR.

:cl: deathride58
refactor: Added unomos, which is basically listmos except gas mixtures only use one single list for handling their gasses. This is a significant performance improvement that also offers a mild memory improvement under normal circumstances.
/:cl:

Thank you for coming to my TED talk.


silicons (approved):

Poojawa:
Pretty sure this is why atmos gets broken a lot more often.

Multiple runtimes on nonexistant gas comparisons. Cyropods have been wonky.

LemonInTheDark:
This is a travesty

PR #8238, opened 2019-04-12
readds dirt buildup on turfs, but in a way that doesn't make the server beg for death

deathride58:
Title. Dirt buildup was a neat, tiny little feature that gave janitors something to do during overly calm rounds. With this PR, every object, including mobs, will cause dirt to passively build up over time on the turfs they travel over. This results in frequently traveled areas, such as the main hallways, to slowly become visibly dirty over time, resulting in a sort of heatmap effect that shows which areas get traveled through the most.
So, what makes the implementation of footdirt in this PR different from the old implementation? For starters, a turf's dirt level is defined by a new `dirtlevel` var instead of being handled by a dirt decal object. Additionally, dirt decals now only spawn once that dirt level becomes greater than 255. Also, the gravity check was removed to save on performance, and the code has been massively cleaned up. All of this should result in an implementation of footdirt that's much, _much_ friendlier to the server in terms of memory usage and performance.

:cl: deathride58
add: Dirt buildup has been readded, but in a way that's a lot friendlier to the server's memory usage and performance.
/:cl:

### DIFF
--- a/code/game/objects/effects/decals/cleanable/misc.dm
+++ b/code/game/objects/effects/decals/cleanable/misc.dm
@@ -49,6 +49,7 @@
 	canSmoothWith = list(/obj/effect/decal/cleanable/dirt, /turf/closed/wall, /obj/structure/falsewall)
 	smooth = SMOOTH_FALSE
 	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
+	var/setdirtiness = TRUE
 
 /obj/effect/decal/cleanable/dirt/Initialize()
 	. = ..()
@@ -59,10 +60,20 @@
 		icon_state = ""
 		queue_smooth(src)
 	queue_smooth_neighbors(src)
+	if(setdirtiness && istype(T, /turf/open/floor))
+		var/turf/open/floor/F = T
+		F.dirtlevel = 510
 
 /obj/effect/decal/cleanable/dirt/Destroy()
 	queue_smooth_neighbors(src)
+	var/turf/T = get_turf(src)
+	if(istype(T, /turf/open/floor))
+		var/turf/open/floor/F = T
+		F.dirtlevel = 0
 	return ..()
+
+/obj/effect/decal/cleanable/dirt/natural
+	setdirtiness = FALSE
 
 /obj/effect/decal/cleanable/flour
 	name = "flour"
--- a/modular_citadel/code/game/turfs/cit_turfs.dm
+++ b/modular_citadel/code/game/turfs/cit_turfs.dm
@@ -1,24 +1,18 @@
-//Yes, hi. This is the file that handles Citadel's turf modifications.
+/turf/open/floor
+	var/dirtlevel
+	var/candirtify = TRUE
 
-/*
 /turf/open/floor/Entered(atom/obj, atom/oldloc)
 	. = ..()
-	CitDirtify(obj, oldloc)*/
-
-//Baystation-styled tile dirtification.
-/turf/open/floor/proc/CitDirtify(atom/obj, atom/oldloc)
-	if(prob(50))
-		if(has_gravity(src) && !isobserver(obj))
-			var/dirtamount
-			var/obj/effect/decal/cleanable/dirt/dirt = locate(/obj/effect/decal/cleanable/dirt, src)
-			if(!dirt)
-				dirt = new/obj/effect/decal/cleanable/dirt(src)
-				dirt.alpha = 0
-				dirtamount = 0
-			dirtamount = dirt.alpha + 1
-			if(oldloc && istype(oldloc, /turf/open/floor))
-				var/obj/effect/decal/cleanable/dirt/spreadindirt = locate(/obj/effect/decal/cleanable/dirt, oldloc)
-				if(spreadindirt && spreadindirt.alpha)
-					dirtamount += round(spreadindirt.alpha * 0.05)
-			dirt.alpha = min(dirtamount,255)
-	return TRUE
+	if(candirtify && !isobserver(obj))
+		dirtlevel++
+		if(oldloc && istype(oldloc, /turf/open/floor))
+			var/turf/open/floor/F = oldloc
+			dirtlevel += (F.dirtlevel)*0.025
+		if(dirtlevel >= 255)
+			if(!isobserver(obj))
+				var/obj/effect/decal/cleanable/dirt/dirt = locate(/obj/effect/decal/cleanable/dirt, src)
+				if(!dirt)
+					dirt = new/obj/effect/decal/cleanable/dirt/natural(src)
+				dirtlevel = min(dirtlevel, 510)
+				dirt.alpha = round(min(dirtlevel-255,255))


modular_citadel/code/game/turfs/cit_turfs.dm
```diff
@@ -1,24 +1,18 @@
-//Yes, hi. This is the file that handles Citadel's turf modifications.
+/turf/open/floor
+	var/dirtlevel
+	var/candirtify = TRUE
 
-/*
 /turf/open/floor/Entered(atom/obj, atom/oldloc)
```
Ghommie, line 5: I'd suggest you change the var name to something else like A, `obj` is ambiguos and possibly misleading.

modular_citadel/code/game/turfs/cit_turfs.dm
```diff
@@ -1,24 +1,18 @@
...
-				dirt.alpha = 0
-				dirtamount = 0
-			dirtamount = dirt.alpha + 1
-			if(oldloc && istype(oldloc, /turf/open/floor))
-				var/obj/effect/decal/cleanable/dirt/spreadindirt = locate(/obj/effect/decal/cleanable/dirt, oldloc)
-				if(spreadindirt && spreadindirt.alpha)
-					dirtamount += round(spreadindirt.alpha * 0.05)
-			dirt.alpha = min(dirtamount,255)
-	return TRUE
+	if(candirtify && !isobserver(obj))
```
Ghommie, line 7: early return if the floor can't be dirtied or it's an observer.

modular_citadel/code/game/turfs/cit_turfs.dm
```diff
@@ -1,24 +1,18 @@
...
+		dirtlevel++
+		if(oldloc && istype(oldloc, /turf/open/floor))
+			var/turf/open/floor/F = oldloc
+			dirtlevel += (F.dirtlevel)*0.025
+		if(dirtlevel >= 255)
+			if(!isobserver(obj))
```
Ghommie, line 13: we already checked if the passerby was an observer before, remove.

modular_citadel/code/game/turfs/cit_turfs.dm
```diff
@@ -1,24 +1,18 @@
...
+				var/obj/effect/decal/cleanable/dirt/dirt = locate(/obj/effect/decal/cleanable/dirt, src)
+				if(!dirt)
+					dirt = new/obj/effect/decal/cleanable/dirt/natural(src)
+				dirtlevel = min(dirtlevel, 510)
```
Ghommie, line 17: I don't see quite why the floor dirtlevel should be set up here and not by the natural dirt decal, unlike its parent type.

modular_citadel/code/game/turfs/cit_turfs.dm
```diff
@@ -1,24 +1,18 @@
...
+				dirt.alpha = round(min(dirtlevel-255,255))
```
Ghommie, line 18: The dirt alpha level is going to be maximal anyway.

modular_citadel/code/game/turfs/cit_turfs.dm
```diff
@@ -1,24 +1,18 @@
...
-				if(spreadindirt && spreadindirt.alpha)
-					dirtamount += round(spreadindirt.alpha * 0.05)
-			dirt.alpha = min(dirtamount,255)
-	return TRUE
+	if(candirtify && !isobserver(obj))
+		dirtlevel++
+		if(oldloc && istype(oldloc, /turf/open/floor))
+			var/turf/open/floor/F = oldloc
+			dirtlevel += (F.dirtlevel)*0.025
+		if(dirtlevel >= 255)
```
Ghommie, line 12: Early return

modular_citadel/code/game/turfs/cit_turfs.dm
```diff
@@ -1,24 +1,18 @@
-//Yes, hi. This is the file that handles Citadel's turf modifications.
+/turf/open/floor
+	var/dirtlevel
+	var/candirtify = TRUE
```
Ghommie, line 3: You could move these two vars to `/turf/`, so you avoid the `istype(T, /turf/open/floor)` by checking `T.candirtify` on the dirt's init, or simply removing it on delete.
Not that this stuff is supposed to work outside of this subtype. anyhow.

code/game/objects/effects/decals/cleanable/misc.dm
```diff
@@ -49,6 +49,7 @@
 	canSmoothWith = list(/obj/effect/decal/cleanable/dirt, /turf/closed/wall, /obj/structure/falsewall)
 	smooth = SMOOTH_FALSE
 	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
+	var/setdirtiness = TRUE
```
Ghommie, line 52: Seems useless.

Ghommie (commented):
I might be wrong, but it seems this feature is either very incomplete, or a maimed shadow of some crappy copypasta from other codebases.

Edit: By the way, consider moving this away from modular citadel, It's deprecated and purposeless nowadays.

modular_citadel/code/game/turfs/cit_turfs.dm
deathride58: This is a built-in byond proc. Naming the obj var to something else results in finicky or otherwise unexpected behavior.

code/game/objects/effects/decals/cleanable/misc.dm
deathride58: This is used so that the natural dirt subtype doesn't max out the turf's dirtiness var on init. I don't see any way to remove this var without making natural dirt build-up instantly cause dirtiness to max out once it reaches the threshold to appear.

modular_citadel/code/game/turfs/cit_turfs.dm
deathride58: This is set in the turf rather than the natural dirt decal because this is the most direct and performant way to handle setting the dirt level. The other two options possible here would be to either call a proc on the natural dirt decal (which introduces more proccall overhead), or to spawn a new natural dirt decal every single time something enters a dirty-able turf (which, dear god why would anyone ever consider that)
The reason why the parent type sets the turf dirtlevel is because quite a few things already use the base cleanable dirt type for dirt generation, so having the base dirt type set the dirtlevel makes dirt spawned forcefully interact naturally with normal dirt, making the forcefully spawned dirt function more or less the same as the natural dirt

deathride58:
Yeah, a good chunk of the original code was written back when I still didn't have much of a clue about what I was doing. I was pretty drunk when I made this PR. Like, way past the ballmer peak drunk, so that explains why my attempts to clean it up ended up being pretty awful in terms of code quality

modular_citadel/code/game/turfs/cit_turfs.dm
Ghommie: There are hundred of versions of built-in functions such as Bump, Exit, Crossed etc. with the same vars going under different names without problem, maybe only invalidating them from named arguments calls, as if that were possible or sane to do.

Ghommie:
Also my bad, I misread a couple min() for max() and assumed the dirtiness value would have been istantly set up to 510 once the dirt spawning thrsehold was reached.
This invalidates about two pairs of my reviewing comments.

But from a personal view, while this may inspire some """"realism"""" tm in the game. It's no benefit toward the code, the codebase and the playerbase (outside of giving janitors more floors to mop).

nik707 (approved):

silicons:
yolo

deathride58:
Closing because this PR is old and has pretty poor code quality

PR #8245, opened 2019-04-15
rectifies aggressive mobs being Bethesda T-pose aggressive

Poojawa:
Attack code is silly

makes the checks a lot more aggressive to filter out non vore_active mobs from trying to run code and just standing there looking upset.


deathride58 (approved):

deathride58:
speedmerging because people are getting absurdly impatient

PR #8263, opened 2019-04-20
Revert pull#8158

RMagnum:
Useless items that are already given in overwhelming amounts, with a total of 6 suits at round start, and are rarely used anyways. 15,000 points is WAY too much as well. points are balanced in how the item affects the actual gameplay, where suits are given at spawn, there is no reason for it to cost so much when they will always be accessible. The replacement kit is just a conscript kit with clothing and extra stuff they don't even spawn with for 12000 POINTS.
If you add up the items you can buy or find in the spare lockers, it looks like this:
The clothing is free, spare gas masks are in the suit storage units.

Headset keys are free in the cargo vendor, and surface lockers.

Adv scanner is 800.

GPS is free to print from the cargo lathe.

Mining access is easy to obtain if you tell a head you died.

Ore bags spawn a total of 9 at round start, three on the miners, three in the station lockers, three in the surface storage lockers.

Medipens are 750

Minipickaxes are 3 on station, 3 on surface lockers, and a large pickaxe on station (on meta and box only), and on surface tables.

First aid kits are 800, there are free ones in dorms, medbay, and surface-side medbay.

Prescription mesons are printable after optics tech, and normal ones can be printed as soon as 5 minutes in because of their place in R&D's tech tree, as well as having 6 spawns in the station lockers, and surface lockers, but they cost 500 if you purchase them.

Knives are unobtainable outside of spawn as far as I know, not like anyone uses them anyways.

This comes to a total of 2350 without mesons, and 2850 with mesons. Clearly this wasn't added properly, and the suit voucher cost speaks for itself with how many of them I see just lying around at the mining station, unused, making for a total of 6 suits on the map at all times, if every miner isn't hoarding both at once.

Reverting the ore bag too because ore boxes and satchels of holding fill the gap of needing it, since making runs back and forth with a larger ore bag isn't an improvement over ore boxes, and satchels remove the need to worry about carrying them at all.

Reverted PR:
https://github.com/Citadel-Station-13/Citadel-Station-13/pull/8158/files


deathride58:
If you have an issue with the prices, then adjust them. Don't just completely remove it outright just because you personally believe it's useless

RMagnum:
Closed because Lilly did the same/similar in their PR.

PR #8275, opened 2019-04-23
fixed the warp whistle

Seris02:
[Changelogs]:

:cl: fixed the warp whistle
fix: fixed the warp whistle
/:cl:

Because we just had a round where some people got really angry at a wizard that was invisible and invincible, and that's not fun for anybody.

### DIFF
--- a/code/modules/antagonists/wizard/equipment/artefact.dm
+++ b/code/modules/antagonists/wizard/equipment/artefact.dm
@@ -378,8 +378,6 @@
 	return FALSE
 
 /obj/item/warpwhistle/proc/end_effect(mob/living/carbon/user)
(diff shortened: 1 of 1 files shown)


silicons (approved):

deathride58 (changes requested):
Make your PR changelog and PR title say what this PR actually does. Because this isn't a fix at all, this is a balance change

silicons (changes requested):
somehow i forgot entirely to do this wew but yeah this isn't the right fix, the right fix is finding out WHY it isn't de-setting GODMODE from status flags, not removing godmode/invis.

silicons:
inactivity

PR #8277, opened 2019-04-23
Significantly Reworks the Wizard Transformation Event

BurgerLUA:
[Changelogs]:
:cl: BurgerBB
balance: Significantly tweaks the Wizard race transformation event to be less unreasonably troublesome.
/:cl:

[why]: This will probably be controversial, but currently there is a random event that causes everyone's race to switch to something else. This is unreasonably problematic for several reasons.

1. People are snowflakes. They do not like it when their character is swapped with something else without much effort.
2. People don't like snowflakes. They don't like be transformed into a randomly generated neon-pink floor length braid slimegirl.
3. It can seriously and unreasonably fuck over people. Inventory slots can easily be removed because of it. If you're in space, and you transform into a mushroom man, you're dead or crippled.

To solve this, the following was implemented:
- You will revert to your normal self after 10-20 minutes. Most wizard rounds last this long, but in cases where it's a really bad peacewiz or stealthwiz, this can ease the burden of having to become a snowflake.
- Only those on station z-levels are affected. Miners cannot be affected by this, as it may cause a quick death.
- You can no longer transform into a species that cannot wear a jumpsuit, such as golems or mushroom men.
- Upped the universal transformation chance from 50% to 75%. This means that there is a higher chance that everyone transforms into the same race.

### DIFF
--- a/code/modules/events/wizard/race.dm
+++ b/code/modules/events/wizard/race.dm
@@ -5,6 +5,18 @@
 	max_occurrences = 5
 	earliest_start = 0 MINUTES
 
+/datum/round_event/wizard/race
+	var/list/stored_name
+	var/list/stored_species
+	var/list/stored_dna
+
+/datum/round_event/wizard/race/setup()
+	stored_name = list()
+	stored_species = list()
+	stored_dna = list()
+	endWhen = rand(600,1200) //10 to 20 minutes
+	..()
+
 /datum/round_event/wizard/race/start()
 
 	var/all_the_same = 0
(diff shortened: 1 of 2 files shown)


BurgerLUA:
@kevinz000 Not a kneejerk change. I was inquiring about this a few days ago with @deathride58. He believe that it was an adminbus event so I didn't make a PR. Turns it out wasn't.

https://i.imgur.com/MBWybjI.png

I'm fine with the random transform, however if you get super unlucky and turn into a mushroom or pod person you're absolutely fucked. It even affects people outside the station, so miners have a higher chance to be double fucked.

deathride58:
Removing the event from the code outright is pretty lazy. You should be able to just turn the weight and max occurrences to 0

silicons:
okay reworking is fine :+1:

BurgerLUA:
@kevinz000 I agree that removing was a bit crude. I took the path of least resistance since removing features usually works on other servers I've contributed to.

silicons (approved):

PR #8286, opened 2019-04-25
Tweaks how Slurring works and makes slurring directly proportional to drunkeness.

BurgerLUA:
[Changelogs]: 
:cl: BurgerBB
tweak: Tweaks how slurring works so it's more of a gradual change into slurring instead of immediate.
balance: Slurring is now directly proportional to your drunkenness, with other sources of slur being added on top of it.
/:cl:

[why]: It was pretty jarring that there was a fine line between not being slurred and being slurred when drunk. Like you could type a sentence, and it would be 100% incoherent, but 1 second later after your slurring level is gone naturally, you'd start typing normal sentences. This is fixed by making it so that slurring is directly affected by your drunkenness level, and making it so the slur proc can accept a strength value.

### DIFF
--- a/code/modules/mob/living/carbon/human/say.dm
+++ b/code/modules/mob/living/carbon/human/say.dm
@@ -1,9 +1,14 @@
 /mob/living/carbon/human/say_mod(input, message_mode)
 	verb_say = dna.species.say_mod
-	if(slurring)
-		return "slurs"
-	else
-		. = ..()
+	switch(slurring)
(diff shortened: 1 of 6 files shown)


BurgerLUA:
Pardon the title. Originally it was going to be that slurring was harder to get rid of, but I realised that not 20 lines below, there was an entire drunkeness code waiting to be exploited.

silicons (approved):

PR #8305, opened 2019-04-29
Brings NTPR to the nice discord bot

Poojawa:
:cl: Poojawa
server: Discord bot will now report when a round ends and give the vague news ticker that a sister-station would have received for the game mode and result of the round.
/:cl:

Why is it only new game starting and meme commands that work?

:v

### DIFF
--- a/code/__HELPERS/roundend.dm
+++ b/code/__HELPERS/roundend.dm
@@ -199,6 +199,10 @@
 	if(length(CONFIG_GET(keyed_list/cross_server)))
 		send_news_report()
 
+	//tell the nice people on discord what went on before the salt cannon happens.
+	world.TgsTargetedChatBroadcast("The current round has ended. Please standby for your shift interlude Nanotrasen News Network's report!", FALSE)
+	world.TgsTargetedChatBroadcast(send_news_report(),FALSE)
+
 	CHECK_TICK
 
 	//These need update to actually reflect the real antagonists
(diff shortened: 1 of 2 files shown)


code/__HELPERS/roundend.dm
```diff
@@ -199,6 +199,10 @@
 	if(length(CONFIG_GET(keyed_list/cross_server)))
 		send_news_report()
 
+	//tell the nice people on discord what went on before the salt cannon happens.
+	world.TgsTargetedChatBroadcast("The current round has ended. Please standby for your shift interlude Nanotrasen News Network's report!", FALSE)
+	world.TgsTargetedChatBroadcast(send_news_report(),FALSE)
```
deathride58, line 204: TGS doesn't automatically account for rate limiting when outputting to Discord, and Discord's rate limit is pretty strict last I checked. Might be a good idea to use only one targetedchatbroadcast here

deathride58 (changes requested):
Goodsune,

Thalpy:
It's testament to your character that you can make such a pr without filling all output discord messages with blobcats.

code/__HELPERS/roundend.dm
Poojawa: Maybe. It'd make the message pretty messy tho.

Poojawa:
> It's testament to your character that you can make such a pr without filling all output discord messages with blobcats.

With great power comes great responsibilities.

code/__HELPERS/roundend.dm
deathride58: p sure you can pass a linebreak through it somehow

Poojawa: I'll see how it does first, then just make it a blob of msg += s

deathride58 (approved):

PR #8329, opened 2019-05-08
Cloak Sprite Fix

Sishen1542:
Adds a layer of pixels to most cloaks so that suits/vests will not poke out the back end on side sprites.
 
 

[Changelogs]: # (Your PR should contain a detailed changelog of notable changes, titled and categorized appropriately. This includes, new features, sprites, sounds, balance changes, admin tools, map edits, removals, big refactors, config changes, hosting changes and important fixes. An example changelog has been provided below for you to edit. If you need additional help, read https://github.com/tgstation/tgstation/wiki/Changelogs)

Just some changed spritework in neck.dmi for the cloaks.


Poojawa:
would be better to add a layer of pixels to the cloaks to cover that part up, imo

deathride58:
This new sprite cuts into the arm of the underlying sprite, which results in the sprite looking ugly when layered on top of jumpsuits and bare-armed spacemen. Doing what pooj said and instead adding pixels onto the cloaks would be a much better way to fix it in a way that doesn't look uglier than it did in a single edge case scenario

Sishen1542:
As per above suggestions, I have reverted the hazard vest back to normal while working on the cloaks instead. All cloaks save the HoS and CMO cloaks had a small touch up to add one column of pixels with color blending to match.

Poojawa (approved):

PR #8330, opened 2019-05-08
This is an entire addition to put stargazers back into clock cult. 

Scorpion-117:
[Changelogs]:

:cl: Bringing light to Ratvar
add: Added stargazers back into the clockcult. 
tweak: gave stargazers to engineer,service and janitor borgs. 
balance: kept the cap on power for Ratvar but buffed some of the costs to compensate for such a sharp power increase. 
config: changed values of cap to 80000(50000), value to script unlock to 35000(25000) and application unlock to 50000(40000)

[why]:I feel clock cult power can be a bit too low and having more options to work with in terms of acquiring power was a much better feel prior to it's removal. Let alone giving cults with low player counts a much better chance with defense on Reebe. Of course this is only an initial change and is easily subject to change in terms of balancing it verses crew.

### DIFF
--- a/code/modules/antagonists/clockcult/clock_structures/stargazer.dm
+++ b/code/modules/antagonists/clockcult/clock_structures/stargazer.dm
@@ -1,0 +1,70 @@
+#define STARGAZER_RANGE 3 //How many tiles the stargazer can see out to
+#define STARGAZER_POWER 20 //How many watts will be produced per second when the stargazer sees starlight
+
+//Stargazer: A very fragile but cheap generator that creates power from starlight.
+/obj/structure/destructible/clockwork/stargazer
+	name = "stargazer"
+	desc = "A large lantern-shaped machine made of thin brass. It looks fragile."
+	clockwork_desc = "A lantern-shaped generator that produces power when near starlight."
+	icon_state = "stargazer"
+	unanchored_icon = "stargazer_unwrenched"
+	max_integrity = 40
+	construction_value = 5
+	layer = WALL_OBJ_LAYER
+	break_message = "<span class='warning'>The stargazer's fragile body shatters into pieces!</span>"
+	resistance_flags = LAVA_PROOF | FIRE_PROOF | ACID_PROOF
+	light_color = "#DAAA18"
+	var/star_light_star_bright = FALSE //If this stargazer can see starlight
+
+/obj/structure/destructible/clockwork/stargazer/Initialize()
+	. = ..()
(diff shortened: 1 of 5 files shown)


Poojawa:
Ensure you locally test this as much as possible to ensure it even works.

Scorpion-117:
I'll get the testing done when when I can get a private server functioning on my PC. But to clarify, all of this code isn't new. It's literally just an undo of the PR that removed it. So this code is the most recent one at the moment of it's removal, the only real issue I need to verify is that the stargazer sprites are still in the code.

Scorpion-117:
specifically as of Jan 29, 2018, so If there were any specific DM changes to clock structures since then I am not sure how to go about updating  the stargazer file.

deathride58:
Ping me on Discord when you get around to testing this locally

Scorpion-117:
After testing locally the stargazer works fine. The only thing left is figuring out how to balance the thing for clock cult since it provides a ton of power

code/modules/antagonists/clockcult/clock_structures/stargazer.dm
```diff
@@ -0,0 +1,70 @@
+#define STARGAZER_RANGE 3 //How many tiles the stargazer can see out to
+#define STARGAZER_POWER 15 //How many watts will be produced per second when the stargazer sees starlight
```
Poojawa: change this to 3, since these are not only spammable, but we really ought to encourage clock cult to get power by other means

MimicFaux:
"better chance with defense on _Ratvar_."
small but important change, the pocket dimension is called 'Reebe', Ratvar is the name of the god.

code/modules/antagonists/clockcult/clock_structures/stargazer.dm
Scorpion-117: I mean their other means is well...apcs, this was to be another means.

Scorpion-117: I even jacked up the power requisites by 10,000 for each tier and kept the hardcap

Scorpion-117: However a better solution to this instead of nerf it to the ground is buffing the power cost to make them and dropping it to 7

Scorpion-117: making a far better investment but not as spammable as prior.

deathride58 (approved):

PR #8347, opened 2019-05-12
adds gar/giga gar glasses to loadout

silicons:
@LetterJay Not sure if you wanted this to be donator only or..?


Poojawa:
`modular_citadel/code/modules/client/loadout/glasses.dm:52:error: /obj/item/clothing/glasses/garb/giga: undefined type path`

PR #8359, opened 2019-05-15
[MERGE READY]Reduces the complexity of the two new circuits: pneumatic cannon and stunner.

Useroth:
Considering both have pretty hefty cooldown (3 sec for the cannon, 5 for the stunner) and external cooldown (1.5 sec for the cannon and 2.5 for the stunner) making them pretty challenging to stack in any effective way, I think it's just fair, because currently it's _pretty_ hard to make anything interesting and have it fit in a single assembly.
Also, I will want stun implants eventually, yes.

Also, regarding some concern expressed about the good old pocket cannons. Those are not a thing anymore. There's only one assembly that can even use weaponized circuits while being held by a human and it's a medium-sized assembly.

### DIFF
--- a/code/modules/integrated_electronics/subtypes/weaponized.dm
+++ b/code/modules/integrated_electronics/subtypes/weaponized.dm
@@ -215,7 +215,7 @@
 	The 'fire' activator will cause the mechanism to attempt to launch objects at the coordinates, if possible. Note that the \
 	projectile needs to be inside the machine, or on an adjacent tile, and must be medium sized or smaller. The assembly \
 	must also be a gun if you wish to launch something while the assembly is in hand."
-	complexity = 40
+	complexity = 50
 	w_class = WEIGHT_CLASS_SMALL
 	size = 4
 	cooldown_per_use = 30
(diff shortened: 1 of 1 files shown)


BurgerLUA (approved):

deathride58 (commented):
Yeah I'm really not liking how low these values are. Can be merged with other maintainer's approval, but I'm personally on the fence about this.

Useroth:
What values would you suggest for them? They're already rather niche gimmicks, to be entirely honest. 

Like, for example, the pneumatic cannon doesn't compare in any way to the traditional pneuma-cannon, firing only a single item, rather than a bunch. There's many more dangerous memes I can do with regular throwers and acid or hot oil in beakers.

And the stunner? You gain no benefit for stacking more than two of them due to those cooldowns. And even then, each of them can stun you only once per 2.5 sec with less stun than half a stunbaton.

silicons (changes requested):
let's try 50/30 instead of 40/15

silicons (approved):

PR #8361, opened 2019-05-16
Fix for the improvised jetpack

ghost:
[Changelogs]:
:cl: 
fix: Should be able to craft the improvised jetpack with two red oxygen tanks, one fire extinguisher and 30 cable coils.
/:cl:

[why]: To make the jetpack craftable


ghost:
Honestly I had no idea how to make the pipes work so I just removed em which should make the improvised jetpack craftable once more. Emphasis on should.

PR #8371, opened 2019-05-17
A weight tweak to d-esword and baseball bat, to make them catch up with the stamina rework. 

Useroth:
- Makes the dual-esword use the same weight values as regular e-sword, instead of the WEIGHT_CLASS_BULKY tier weight, making it drain less stamina
- Makes baseball bats more realistic regarding their weight, instead of swinging like WEIGHT_CLASS_HUGE items.

If you ask me, tbh, I'm still not feeling the 3.4 pound e-swords, but that's material for another discussion and another PR.


Poojawa:
a good start, for sure.

silicons (approved):

silicons:
while i dislike realism memes this makes sense.

PR #8375, opened 2019-05-17
Singallers are now input instead of button mashing

Poojawa:
Porting TG#43799

🆑 Tetr4
tweak: Remote signallers now use input boxes instead of inc/dec buttons.
fix: Remote signaller freq range no longer clamps to the wrong range.
/🆑


silicons (approved):
Yes.

PR #8384, opened 2019-05-18
Ports new footstep sounds from base /tg/

nero1024:
Bare feet and other mobs will now make the correct footstep sounds and added new samples to go with them. Added some helpers to is_helpers.dm, mostly for (future) footstep support but maybe someone else can make use of them. Added crawling/dragging sounds for downed, lying, or otherwise incapacitated mobs when they're moving or being moved around. I can't be arsed to add clawed sounds to clawed human-type mobs so you'll have to deal, this is just a simple port.

:cl: Nero1024
soundadd: Bare feet will now make the correct footstep sounds.
soundadd: Other mobs will make the correct footstep sounds.
soundadd: Crawling/dragging sounds for downed/incapacitated mobs
/:cl:

Ported from base /tg/ since this server doesn't have the new bare-footstep sounds.

TODO:
- [x] Fix: Crawl sounds for naga characters.


Poojawa:
```
if(("taur" in H.dna.species.mutant_bodyparts) && (H.dna.features["taur"] != "None"))
	if(H.dna.features["taur"] in list("Naga", "Tentacle"))
```

That's the check I use to make suits taur-able and use the naga sprites. 

when #8251 gets done, this check will be optimized to just 

```
if(("taur" in H.dna.species.mutant_bodyparts) && (H.dna.features["taur"] != "None"))
	if(H.dna.features["taur"] in GLOB.noodle_taurs)
```

ghost:
please change your profile pictures

code/datums/components/footstep.dm
```diff
@@ -27,13 +31,75 @@
 			v /= 2
 			e -= 5
 	steps++
-	if(steps >= 3)
+	
+	if(steps >= 6)
```
deathride58: This is an undocumented revert of citadel-specific changes. This will result in footsteps playing once every two steps instead of once every three steps, which sounds absolutely horrendous while sprinting.

code/datums/components/footstep.dm
```diff
@@ -27,13 +31,75 @@
...
 		steps = 0
-	else
+	
+	if(steps % 2)
```
deathride58: Ditto

code/datums/components/footstep.dm
```diff
@@ -27,13 +31,75 @@
...
 		return
-	if(prob(80) && !LM.has_gravity(T)) // don't need to step as often when you hop around
+	
+	if(!LM.has_gravity(T) && steps != 0) // don't need to step as often when you hop around
```
deathride58: Relies on logic used by stock TG's footsteps

deathride58 (changes requested):
TG's decision of having footsteps play once every two turfs sound absolutely obnoxious. Footstep logic in our code plays footsteps once every three steps, which sound a lot better when sprinting, but this PR reverts that. Fix it

deathride58:
@gardenBanter nah

code/datums/components/footstep.dm
nero1024: I'll fix that, but tbh I liked the old tg-style footstep frequency.

deathride58 (approved):

PR #8385, opened 2019-05-18
Adds a cream pie smite.

BlackMajor:
Sometimes you just want to lightly slap someone for being shit instead of causing damage to them.

Also reorganises the smite menu list from least to most dangerous.


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1294,6 +1294,9 @@ GLOBAL_LIST_EMPTY(custom_outfits) //Admin created outfits
 			if(!puzzle_imprison(target))
 				to_chat(usr,"<span class='warning'>Imprisonment failed!</span>")
 				return
+		if(ADMIN_PUNISHMENT_PIE)
+			var/obj/item/reagent_containers/food/snacks/pie/cream/nostun/creamy = new /obj/item/reagent_containers/food/snacks/pie/cream/nostun(get_turf(target))
```
Poojawa: > nostun

Rejected.

Ghommie: What the above fellow said.

Useroth: +1

Useroth: also, you don't need the full path for `new`, if you've declared it in the variable
```suggestion
			var/obj/item/reagent_containers/food/snacks/pie/cream/nostun/creamy = new(get_turf(target))
```

BlackMajor: Thanks fam

silicons (approved):

PR #8400, opened 2019-05-21
Watcher Wing Trophy Effect increase

ghost:
:cl: 
tweak: Watcher wing Trophy's effect lasts 1 second instead of 0.5
/:cl:

[why]: It's broken at it's current state as it does next to nothing.

### DIFF
--- a/code/modules/mining/equipment/kinetic_crusher.dm
+++ b/code/modules/mining/equipment/kinetic_crusher.dm
@@ -230,7 +230,7 @@
 	desc = "A wing ripped from a watcher. Suitable as a trophy for a kinetic crusher."
 	icon_state = "watcher_wing"
 	denied_type = /obj/item/crusher_trophy/watcher_wing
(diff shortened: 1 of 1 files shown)


deathride58:
What's the absolute minimum time it takes to successfully detonate a mark, not including click delay modifiers?

ghost:
> What's the absolute minimum time it takes to successfully detonate a mark, not including click delay modifiers?

I believe it's detonated immediately when you click, however the effect of the watcher wing was not noticeable/did not provide any actual help in a fight. Now that it is increased to 1 second the enemy has 0.5 seconds to let loose on the user as that's how long it takes for the crusher to recharge, allowing the user to hit the enemy with another mark detonation.

MimicFaux:
After the general buffs to lavaland fauna, big and small, this is welcomed.

silicons (approved):

PR #8406, opened 2019-05-22
Batch messages on checking yourself

WhiteHusky:
Instead of sending a bunch of to_chat() functions, it'll just create an empty ~~array~~ string, join, and send it all out at once instead of freezing the client for a moment.

:cl:
tweak: Checking yourself shouldn't freeze the client anymore.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -649,7 +649,7 @@
 
 	if(health >= 0)
 		if(src == M)
-			var/to_send = list()
+			var/to_send = ""
 			visible_message("[src] examines [p_them()]self.", \
 				"<span class='notice'>You check yourself for injuries.</span>")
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -649,6 +649,7 @@
 
 	if(health >= 0)
 		if(src == M)
+			var/to_send = list()
```
deathride58: This list isn't necessary at all, and does nothing but add unnecessary performance overhead and throw a needless list at the GC. You can use the `+=` operator with strings directly instead of going the long way around with heavy operations like jointext()

deathride58 (changes requested):

code/modules/mob/living/carbon/human/human_defense.dm
WhiteHusky: The only reason I used it because I had to do something similar over on BayStation and it was suggested to use a list instead of concatenating it into one big string. I'll commit a change in a moment...

Poojawa:
I recently did this for the medical scanner, if you want an example of bundling stuff up

WhiteHusky:
I guess I'm getting terrible luck with Travis blowing itself up.

silicons:
stale

WhiteHusky:
Kinda forgot about this and thought someone would bump Travis to retry.

Let's give this another try.

PR #8408, opened 2019-05-22
Fixing cult rounds potentially stalling if the sac target is erased from this realm.

Ghommie:
Further subtitles:
Why did you check if a gamemode datum type is an antagonist datum? And why did it work too?
I don't understand.
Pooj's copypasting black magic, smh.

:cl:
fix: Fixes cult potentially stalling if the target is erased from existence without being sacced.
/:cl:

Fixing a game breaking bug recently brought up after I played bomberman on brig. May need a testmerge to make sure it works, pre-existing code tech debt worries me a bit.


silicons (changes requested):
not a bug, ping @deathride58 and myself for discussion.

silicons (approved):

PR #8415, opened 2019-05-22
Adds a sprint buffer - 35 tiles, regen 2 tiles/second (always active), 0.3 stamcost per tile if empty (as old) directly applied to staminaloss rather than stamina action buffer (unlike old)

silicons:
Title
Values up to debate.


modular_citadel/code/modules/mob/living/carbon/life.dm
```diff
@@ -0,0 +1,2 @@
+/mob/living/carbon/Life()
```
Ghommie, line 1: Shouldn't you also be calling `. = ..()` here?

modular_citadel/code/modules/mob/living/living.dm
```diff
@@ -8,6 +8,14 @@
 	var/intentionalresting = FALSE
 	var/attemptingcrawl = FALSE
 
+	//Sprint buffer---
+	var/sprint_buffer = 35					//Tiles
```
Ghommie: Suggesting you to turn this into a define.

modular_citadel/code/modules/mob/living/living.dm
```diff
@@ -8,6 +8,14 @@
...
+	var/sprint_buffer_max
+	var/sprint_buffer_regen_ds = 0.2		//Tiles per world.time decisecond
```
Ghommie: Suggesting you to turn this value into a define.

modular_citadel/code/modules/mob/living/living.dm
```diff
@@ -8,6 +8,14 @@
...
+	var/sprint_buffer_regen_last = 0		//last world.time this was regen'd for math.
+	var/sprint_stamina_cost = 0.3			//stamina loss per tile while insufficient sprint buffer.
```
Ghommie: And this one too.

Poojawa:
I still dislike the band-aid like treatment to a core issue that this feels like, but it won't hurt to try it out at least.

modular_citadel/code/modules/mob/living/living.dm
silicons: it's variable for a reason same below

Poojawa:
So begins the power creep :^)

peeesh:
> So begins the power creep :^)

no

Poojawa:
> > So begins the power creep :^)
> 
> no

Woosh.

Ghommie:
Currently the sprinter buffer ui icon doesn't take in account if the widescreen is NOT enabled and get overlayered by other ui buttons. Please fix that, thanks.

silicons:
oh yeah it's fixed @deathride58

deathride58 (dismissed):

deathride58 (approved):

PR #8419, opened 2019-05-23
Rewords Blood Cult 'offer' rune click here prompt.

UntoldTactics:
[Changelogs]:
Rewords prompt given on Blood Cult offer runes.

:cl: optional name here
tweak: Rewords the click here prompt given when an attempted conversion is done on an offer rune (for blood cult)
/:cl:

[why]: 
The current click here prompt can be quite confusing, and it has led quite a few people into not clicking it and being sharded, despite wanting to be a cultist.

### DIFF
--- a/code/modules/antagonists/cult/runes.dm
+++ b/code/modules/antagonists/cult/runes.dm
@@ -246,7 +246,7 @@ structure_check() searches for nearby cultist structures required for the invoca
 		return 0
 	to_chat(convertee, "<span class='cult italic'><b>Your blood pulses. Your head throbs. The world goes red. All at once you are aware of a horrible, horrible, truth. The veil of reality has been ripped away \
 	and something evil takes root.</b></span>")
-	to_chat(convertee, "<span class='cult italic'>Do you wish to embrace the Geometer of Blood? <a href='?src=\ref[src];signmeup=1'>Click here to become a follower of Nar'sie.</a> Or you could choose to continue resisting and suffer a fate worse than death.</span>")
+	to_chat(convertee, "<span class='cult italic'>Do you wish to embrace the Geometer of Blood? <a href='?src=\ref[src];signmeup=1'>Click here to become a follower of Nar'sie.</a> Or you could choose to continue resisting and suffer a fate worse than death...</span>")
 	currentconversionman = convertee
 	conversiontimeout = world.time + (10 SECONDS)
 	convertee.Stun(100)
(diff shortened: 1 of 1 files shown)


davidchan13:
MAKE IT A YES OR NO POP UP

This is completely inconsistent with all other behavior in the game with 'resist' Being the default option. Further more something as important as this (that GIBS you for being idle or passive) should not be in the chat window when the player is almost certainly cuffed and being dragged around by 2 or more players the last place they are looking is their chat box for a clickable link (where buttons for IC actions never are)

This needs to be a pop, in line with ERT joining or mid round roles, "Do you submit to the Cult of Nar'Sie? (Resisting will have consequences) Yes or No

After 10 or so seconds default to yes. We should not be gibbing players for being confused and trying to process a hostile action against them and expect a quick reaction, especially from players not familiar with the role.

deathride58:
@davidchan13 out of this PR's scope

code/modules/antagonists/cult/runes.dm
```diff
@@ -246,7 +246,7 @@ structure_check() searches for nearby cultist structures required for the invoca
 		return 0
 	to_chat(convertee, "<span class='cult italic'><b>Your blood pulses. Your head throbs. The world goes red. All at once you are aware of a horrible, horrible, truth. The veil of reality has been ripped away \
 	and something evil takes root.</b></span>")
-	to_chat(convertee, "<span class='cult italic'>Do you wish to embrace the Geometer of Blood? <a href='?src=\ref[src];signmeup=1'>Click here to stop resisting the truth.</a> Or you could choose to continue resisting...</span>")
+	to_chat(convertee, "<span class='cult italic'>Do you wish to embrace the Geometer of Blood? <a href='?src=\ref[src];signmeup=1'>Click here to become a follower of Nar'sie.</a> Or you could choose to continue resisting and suffer a fate worse than death.</span>")
```
deathride58: no `...` at the end? smh my head
```suggestion
	to_chat(convertee, "<span class='cult italic'>Do you wish to embrace the Geometer of Blood? <a href='?src=\ref[src];signmeup=1'>Click here to become a follower of Nar'sie.</a> Or you could choose to continue resisting and suffer a fate worse than death...</span>")
```

davidchan13:
@deathride58 The scope of the PR is to reduce confusion in regards to cult conversion. Confusion isn't from the wording of the button, but the placement and nonsensical operation of the button.

code/modules/antagonists/cult/runes.dm
UntoldTactics: yeah okay

silicons (approved):

PR #8425, opened 2019-05-24
Ports TG firebots.

BlackMajor:
See title.


Poojawa:
eeeeeh. I don't really like these.

basically it's robo replacing all the jobs that take effort to do because people are too lazy to do it themselves.

silicons (approved):

PR #8427, opened 2019-05-24
Removes color-whitelist blocking custom-colored circuit blueprints from being imported

CalamaBanana:
Integrated circuit printers checked imported colors against a whitelist on import. Was used to prevent cheeky players from circumventing the color limitation in the past. Another cheeky player added a custom color selection a few days ago, making this check against the whitelist obsolete.

Tested it, if a player enters an invalid color, it just shows up as white, which is fine.


silicons (changes requested):
needs to sanitize the import, i suspect it was there for security reasons.

CalamaBanana:
Fair point. Using the same method that checks the text now, ~~so I can shift blame if it goes up in flames,~~ hope that's enough.

silicons (approved):

PR #8428, opened 2019-05-24
[s] Max length limit to concat circuits

silicons:
ready as soon as travis passes @deathride58

### DIFF
--- a/code/modules/integrated_electronics/subtypes/converters.dm
+++ b/code/modules/integrated_electronics/subtypes/converters.dm
@@ -161,6 +161,7 @@
 		var/I = get_pin_data(IC_INPUT, k)
 		if(!isnull(I))
 			result = result + I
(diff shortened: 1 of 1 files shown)


code/modules/integrated_electronics/subtypes/converters.dm
```diff
@@ -161,6 +161,7 @@
 		var/I = get_pin_data(IC_INPUT, k)
 		if(!isnull(I))
 			result = result + I
+			result = copytext(1, MAX_MESSAGE_LEN)
```
deathride58, line 164: The lag comes as a result of large messages being combined. Putting a limit after the text is already combined doesn't stop the exploit from happening.

k-k-karma: yes it does

PR #8432, opened 2019-05-25
Buffs stunprods to not have a cooldown

Ghommie:
:cl:
balance: Nukes the stunprod's 3 seconds delay.
fix: Fixes teleprods.
/:cl:

[why]: # Citadel stamina system made stunprod very obnoxious to use, requiring approximately more than dozen seconds to stamcrit someone compared to the average /tg/code where the pre-existing 3 seconds delay would be outlasted by the applied stun anyway. whereby a normal baton would do the job three times faster.

So I invite you to bear, not despise, this proposed change, as the issue was more of a consequence of cit stam combat than one of its features, and only adds up to the already pre-existing cons of the stunprod versus the stunbaton, 
Will welcome suggestions to round up the delay to 2 seconds if it doubtfully proves to be "overpowered" regardless.

Edit: Deprecated chit chat as deathride told me to just nuke the delay.


silicons (dismissed):

deathride58:
You can actually remove the delay entirely without balance concerns. Stunprods are very noticeable due to their sprite and the inability to put them into backpacks, and with stamina-based combat, the requirement for a cooldown is quite moot since it isn't a 1-hit gg-no-re

code/game/objects/items/stunbaton.dm
```diff
@@ -202,12 +202,13 @@
...
 /obj/item/melee/baton/cattleprod/Initialize()
 	. = ..()
 	sparkler = new (src)
+	sparkler.activate_cooldown = 5
 
 /obj/item/melee/baton/cattleprod/baton_stun()
-	if(sparkler.activate())
-		..()
+	sparkler?.activate()
+	return ..()
```
deathride58:
```suggestion
	. = ..()
```

deathride58 (dismissed):

code/game/objects/items/stunbaton.dm
Ghommie: Personally I see little reason to do `. = ..()` than `return ..()` since no action follows up except the implied end-of-proc return.
Doing this should merely adds up another step to the proc, but either ways, it has virtually almost no impact to its processing time.
But I may have just said a load of ballocks, who knows...

silicons (dismissed):

Ghommie:
Ok, merge conflict resolved. Also this fixes teleprods.

silicons (approved):

PR #8439, opened 2019-05-26
stunbaton status rework.

Ghommie:
:cl:
balance: EMPs now flick off stunbatons, they can be turned back on immediately by the user anyway.
balance: Stunbatons now very slowly consume charge whilst kept on, at a rate of 4/1000th of a standard batoning charge cost per tick.
balance: Softened up the charge cost checks to stop the above update from practically reducing the maximum uses of a stun baton by one. Now, should the remaining charge be lower than the hit cost, the resulting stun will be be proportional to the remaining charge divided by the hitcost, within a limit under which the stun batoning just won't happen.
/:cl:

~~EMP is currently hilariously bad against stunbaton (and powercells with more than a couple thousand units of charge in general) compared to the wide majority of other objects & electronics, taking roughly half a dozen EMPs to be totally drained, or more if the stunbatoner is savyy enough to load in a higher quality power cell, reason I discarded the idea of merely buffing the emp_act() charge deduction, as it would still be cheesed.~~

Deprecated. Now only flick batons off, a minor nausiance.

Added a mild charge deduction of a 4/1000th of hitcost per tick while the stunbaton is left on. From experience, keeping a stunbaton off was only useful during violent harmbatoning to the head.

Added some leniency to the remaining charge checks to avoid the above processing update from nerfing batonings by one, so the resulting stun power will now be proportional to the amount of charge left compared to the hitcost should said charge be inferior than the cost.

Current values may be subject to changes.

Also cleaned up old and rarely used teleprods.

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -54,7 +54,7 @@
 	//Note this value returned is significant, as it will determine
 	//if a stun is applied or not
 	. = cell.use(chrgdeductamt)
-	if(status && (!. || (chargecheck && cell.charge < hitcost)))
+	if(status && (!. || (chargecheck && cell.charge < hitcost * STUNBATON_CHARGE_LENIENCY)))
 		//we're below minimum, turn off
 		switch_status(FALSE)
 
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -45,14 +45,14 @@
 /obj/item/melee/baton/loaded //this one starts with a cell pre-installed.
 	preload_cell_type = /obj/item/stock_parts/cell/high
 
-/obj/item/melee/baton/proc/deductcharge(chrgdeductamt)
+/obj/item/melee/baton/proc/deductcharge(chrgdeductamt, chargecheck = TRUE)
 	if(!cell)
 		switch_status(FALSE, TRUE)
 		return FALSE
 	//Note this value returned is significant, as it will determine
 	//if a stun is applied or not
 	. = cell.use(chrgdeductamt)
-	if(status && cell.charge < hitcost)
+	if(status && (!. || (chargecheck && cell.charge < hitcost)))
 		//we're below minimum, turn off
 		switch_status(FALSE)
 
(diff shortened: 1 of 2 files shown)


silicons:
okay but why is this needed batons were always meant to be tough vs emp

Ghommie:
> 
> 
> okay but why is this needed batons were always meant to be tough vs emp

Batons being sturdy against EMPs is more of a consequence of the latter being a crapshot against the better power cells in general. 

Anyway stun batons can also be loaded with even better power cells,
are found inside most security closets and not in the armory, making it easy to have a backup stunbaton and, despite requiring 3-4 batoning are required to stamcrit someone, its clicks can be delivered in such a rapid succession making the stam system scope pointless.

So I hope you don't go under the assumption this is some "I ded" PR, this little fallacy of EMPs being able to nuke the average energy firearm while being total garbage against the aforementioned subject made me think for months. 

Nevertheless, I'm fairly open to dampening the downtime numbers and applying further tweaks, within reason. I have no intention of causing a negative balance impact to the game rather than expanding and improving the interaction between its features.

silicons (changes requested):
stunbaton being a backup vs emp is intentional.
emps hard-shutting them down for a few seconds is a hard no. with this a traitor/ling can effortlessly kill an officer with an emp and an armblade/esword (1/0.8*25 dps with armor pierce for esword is pretty easy if you corner them). 
we might /consider/ reducing charge more when emped but yeah no, security is in a good spot nowadays and it's going to take some convincing to convince me that they need to be buffed or nerfed.

Ghommie:
Fair opinions, I aknowledge backup stunbatons are fairly intentional, and that emp_act goes through contents, nothing to debate.
The provided esword/armblade example is quite generic, but elucidating theorical issues with the melee would be stupid and not pertinent so yea, I'll scrap hard-shutting and try find a middle ground regardless,
But I still stand by the view that a standard loaded stunbaton can roughly sustain up to 8 emp_act() is excessive and that the current well-being of sec & co features shouldn't be justifying it.

silicons (dismissed):

deathride58 (approved):

PR #8444, opened 2019-05-26
CWC player requirement increase

ghost:
well basically most people dislike CWC in lowpop and it's generally unbalanced during lowpop so let's just not have it happen during lowpop. Easy fix!

:cl: 
~balance: CWC gamemode player requirement increase from 30 players to 45 players~
~balance: CWC gamemode player requirement increase from 30 players to 40 players~
balance: CWC gamemode player requirement increase from 30 players to 35 players
:cl:

### DIFF
--- a/code/game/gamemodes/clock_cult/clock_cult.dm
+++ b/code/game/gamemodes/clock_cult/clock_cult.dm
@@ -131,7 +131,7 @@ Credit where due:
 	config_tag = "clockwork_cult"
 	antag_flag = ROLE_SERVANT_OF_RATVAR
 	false_report_weight = 10
-	required_players = 45
+	required_players = 40
 	required_enemies = 3
 	recommended_enemies = 5
 	enemy_minimum_age = 7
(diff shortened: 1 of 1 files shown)


Useroth:
\>45 people readying up roundstart
\>ever

Fractious:
As mentionned by Useroth, the reason why the player cap is set at 30 is mostly due to players never readying up at round start. When we reach highpop (100+ players on Sunday) the max we'll get is 45-ish ready-up's and maybe 50. Doing this is basically the same as removing it off the rotation in all honesty.

CozmoB:
I think its fine the way it is, i've seen lowpop wins and lowpop losses, I'd like to see some stats to make sure it is actually skewed one way or another.

silicons:
not necessary at this time

silicons:
actually hm

silicons:
give me like, 2-3 days.

silicons (approved):

PR #8457, opened 2019-05-27
Adds a way to come back from nearly dying, naturally

BurgerLUA:
[Changelogs]:
:cl: BurgerBB
balance: If you are in soft crit for more than 5 minutes, you slowly heal 0.1 of burn, brute, oxygen, and toxin of each damage type until you are no longer in soft crit.
/:cl:

[why]: I want to reward people who stick in there and cheat death by somehow staying alive despite being put in crit. It would be like those cowboy movies where someone is left for dead, but they somehow come back despite their injuries.


code/modules/mob/living/carbon/life.dm
```diff
@@ -35,6 +35,17 @@
 		stop_sound_channel(CHANNEL_HEARTBEAT)
 		rot()
 
+	if(stat != DEAD)
+		if(InCritical())
+			critical_ticks++
```
deathride58, line 40: This is a pretty bad way to keep track of how much time has passed, especially in a process proc

BurgerLUA: How should I do it, then?

Poojawa: process ticks almost always have an incremental variable. Breath code especially so.

Poojawa: How ever it would be more effective to put a var that gets the world time the moment a mob enters a critical state, then checks for it `world.time + 5 MINUTES` before doing the heal ticks.

silicons:
stale

PR #8468, opened 2019-05-27
buffs disabler 20-->25 shots firerate 2.5/s --> 3.33/s speed 412.5-->471.5 pixels/s 

silicons:
Disablers aren't used enough, hopefully this'll make them more attractive.
Probably too drastic but let's see how this goes.


silicons:
12.5 --> 14.5 tiles/second for those who don't count in pixels.

Ghommie:
Wouldn't it be better to buff back its stam damage rather than making it spammier?

deathride58 (approved):

PR #8473, opened 2019-05-28
[MERGE READY] Taeclowndo. Because Clowns deserve to have their own martial art too.

Useroth:
:cl: optional name here
add: Taeclowndo shoes which grant the four following abilities:
- Conjuring a cream pie right into their hand. Every three seconds.
- Making a banana peel appear out of thin air at the tile of the clown's choice. Every ten seconds.
- Mega HoNk. A touch-ranged, very small AOE ability, with effect equal to being honked by a Honkerblast on a clown mech, with the effects halved for anyone who isn't its direct target. Every ten seconds.
- Bluespace Banana Pie. You don't throw this one... not right away at least. This baby can fit an entire body inside. Good for disposal of evidence. 25 second-long action, 45 second cooldown. Also produces a "[victim's name] cream pie". The body drops out of the pie if you splat it somewhere or destroy the pie. If you eat it, it will chestburst out of you a'la monkey cube.

It's a 14 TC item for traitor clowns and a 12 TC item for clown-ops.
/:cl:


silicons (changes requested):
no random unconscious/hard non-stamina stuns for more than 2 seconds on the thing.

silicons (approved):

PR #7566, opened 2018-09-12
Buffs Action, Promotes fun

Poojawa:
🆑
tweak: Antag adrenals now provide raw stamina healing, based upon feedback from the community.
tweak: Mining Salve is healthy for you.
tweak: Syndicate Nanites were re-programed to clear lactic acid
/🆑

PR #8474, opened 2019-05-28
Adrenal rebalancing/nerfing. Again. And a very slight meth buff.

silicons:
Meth no longer confuses you.

Adrenaline refactored to use a do_adreanline proc.
From least complicated to most:

Adrenal implants, being limited in use and not usually used with other high powered items (cough ninja) are not nerfed.

Abductors lost their auto-stand

Ninjas lost their stimulants, as they already run faster than anyone else due to their shoes. They don't need to be mach 2.

Changeling adreanline are pretty spammable - 
So, they lost autostand, and their instant boost has been reduced to 100 from 150
Their inaprovaline has been changed to epinephrine (a bit more powerful for crit stabilization/such)
Their 2 units of changeling super-speed chem has been removed, rpelaced with 10 units of a special changeling meth with higher overdose and no brain damage/jittering. Meth regenerates stamina at a fast-ish rate (for station chems) and makes them immune to slowdown, including low health and stamina slowdowns, as well as healing 2 brute and 2 burn per second. I don't like mach 2 combat.
Changeling adrenaline chemical reduced from 10 to 5. It cuts 2 seocnds of unconsciousness + sleeping (added the sleeping part) per second and heals 30 stamina per second.
Changelings now no longer instantly get up from sleeping when adreanlining. They get it clamped to 70, so they will, with the adreanline chemicals, get up from sleeping/unconsciousness in 2-3 seconds rather than instantly, assuming they had more than 7 seconds of sleeping/unconsciousness.

### DIFF
--- a/code/modules/antagonists/changeling/powers/adrenaline.dm
+++ b/code/modules/antagonists/changeling/powers/adrenaline.dm
@@ -12,21 +12,5 @@
 
 //Recover from stuns.
 /obj/effect/proc_holder/changeling/adrenaline/sting_action(mob/living/user)
-	to_chat(user, "<span class='notice'>Energy rushes through us.[user.lying ? " We arise." : ""]</span>")
-	user.SetSleeping(0)
-	user.SetUnconscious(0)
-	user.SetStun(0)
-	user.SetKnockdown(0)
-	user.reagents.add_reagent("changelingadrenaline", 10)
-	user.reagents.add_reagent("changelinghaste", 2) //For a really quick burst of speed
-	user.reagents.add_reagent("inaprovaline", 3) //let's give another chance to dumb fucks who forget to breathe
-	user.adjustStaminaLoss(-150)
-	user.stuttering = 0
-	user.updatehealth()
-	user.update_stamina()
-	user.resting = 0
-	user.lying = 0
-	user.update_canmove()
-
+	user.do_adrenaline(100, FALSE, 70, 0, TRUE, list("epinephrine" = 3, "changelingmeth" = 10, "mannitol" = 10, "omnizine" = 10, "changelingadreanline" = 5), "<span class='notice'>Energy rushes through us.</span>")
 	return TRUE
-
--- a/code/modules/antagonists/changeling/powers/adrenaline.dm
+++ b/code/modules/antagonists/changeling/powers/adrenaline.dm
@@ -12,5 +12,5 @@
 
 //Recover from stuns.
 /obj/effect/proc_holder/changeling/adrenaline/sting_action(mob/living/user)
-	user.do_adrenaline(100, FALSE, 70, 0, TRUE, list("epinephrine" = 3, "changelingmeth" = 10, "mannitol" = 10, "omnizine" = 10, "changelingadreanline" = 5), "<span class='notice'>Energy rushes through us.</span>")
+	user.do_adrenaline(100, FALSE, 70, 0, TRUE, list("epinephrine" = 3, "changelingmeth" = 10, "mannitol" = 10, "omnizine" = 10, "changelingadrenaline" = 5), "<span class='notice'>Energy rushes through us.</span>")
 	return TRUE
--- a/code/game/objects/items/implants/implant_misc.dm
+++ b/code/game/objects/items/implants/implant_misc.dm
@@ -33,24 +33,8 @@
 /obj/item/implant/adrenalin/activate()
 	. = ..()
 	uses--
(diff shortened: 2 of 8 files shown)


Useroth:
To be honest, not entirely sure how I feel about the ling adrenals there. I think the spammability should be addressed, rather than the adrenals themselves being nerfed, through either introducing some cooldown between uses or making the involved chems much more punishing to overdose.

Ghommie:
Removing the auto-stand-up from ling adrenals will mean nerfing them back to crawl hell tier of combat impracticability. At that point, we are back to the kind of adrenal where one has to hope whoever is stunning them runs out of charges, not be currently ganged up and possibly have an armblade.
And the shallow emergency healing mix it provides doesn't really help either.

Even just enough stamina boost to make sure not be in soft stam crit and one mild stimulant (or no reagent at all) would be, as long as auto-stand is present, more useful.

I do aknowledge that the unlimited uses (albeit the moderately high ling juice cost) make them very irritating to stand against, but how the changeling haste plus PR #7566 which multiplied the ling adrenaline stam regen by 30 make it extra strong.

P.S. On the other hand I'm ok with the abductor agent's, since it's supposed to be beamed up by their teammater when in danger anyhow.

Arturlang:
Without the auto stand, lings are gonna have a lot of trouble running away, so whoever is batoning them can easily just keep batoning, im not really sure about the removal of the auto stand honestly.

code/modules/antagonists/changeling/powers/adrenaline.dm
```diff
@@ -12,21 +12,5 @@
...
-	user.reagents.add_reagent("inaprovaline", 3) //let's give another chance to dumb fucks who forget to breathe
-	user.adjustStaminaLoss(-150)
-	user.stuttering = 0
-	user.updatehealth()
-	user.update_stamina()
-	user.resting = 0
-	user.lying = 0
-	user.update_canmove()
-
+	user.do_adrenaline(100, FALSE, 70, 0, TRUE, list("epinephrine" = 3, "changelingmeth" = 10, "mannitol" = 10, "omnizine" = 10, "changelingadreanline" = 5), "<span class='notice'>Energy rushes through us.</span>")
```
Ghommie: Ah, yea, forgot about to inform you about a typo before:
```suggestion
	user.do_adrenaline(100, FALSE, 70, 0, TRUE, list("epinephrine" = 3, "changelingmeth" = 10, "mannitol" = 10, "omnizine" = 10, "changelingadrenaline" = 5), "<span class='notice'>Energy rushes through us.</span>")
```

code/modules/antagonists/changeling/powers/adrenaline.dm
```diff
@@ -12,21 +12,5 @@
...
-	user.reagents.add_reagent("inaprovaline", 3) //let's give another chance to dumb fucks who forget to breathe
-	user.adjustStaminaLoss(-150)
-	user.stuttering = 0
-	user.updatehealth()
-	user.update_stamina()
-	user.resting = 0
-	user.lying = 0
-	user.update_canmove()
-
+	user.do_adrenaline(100, FALSE, 70, 0, TRUE, list("epinephrine" = 3, "changelingmeth" = 10, "mannitol" = 10, "omnizine" = 10, "changelingadrenaline" = 5), "<span class='notice'>Energy rushes through us.</span>")
```
Useroth, line 15:
```suggestion
	user.do_adrenaline(100, TRUE, 70, 0, TRUE, list("epinephrine" = 3, "changelingmeth" = 10, "mannitol" = 10, "omnizine" = 10, "changelingadrenaline" = 5), "<span class='notice'>Energy rushes through us.</span>")
```
imo, without this, you're putting ling adrenals right back in the dump they were in before the buff

deathride58: 30 stamloss regen per mob life cycle. and it's spammable to boot. how is instant recovery from stamcrit + an insanely strong stam regen + the ability to spam it, considered "in the dump"

deathride58 (approved):

code/modules/antagonists/changeling/powers/adrenaline.dm
Ghommie: Meh, that and the extra healing effects do make em hard to get robusted, but they are still at a loss crawling around as more people come by to gang em up lethally.
On the other hand, if they were to receive auto-stand up just like the way Useroth suggested it'd solely turn into a meaty buff.
combat intent right click help intent does help but it's fairly clunky to pull off, and combat mode doesn't work well with the pace of your generic ss13 robusting anyway.
Really should do some quick QoL on it.

PR #8479, opened 2019-05-29
Krav maga leg sweep tweak.

Ghommie:
:cl:
balance: Buffed krav maga leg sweep stun and stamina damage. On the other hand, it's now unable to be used on already lying targets.
/:cl:

Another feature usefulness dunked by the stam rework, the krav maga leg sweep. Right now, it's hard to keep someone down with it (even with the related ui action button dragged near the center of the screen), whilst still being usable on lying mobs since it preceded the crawling feature (and the little tweaking of the pre-existin stun combat related features of the stam rework).

Values may need tweaking, but since it's not anymore spammable, and being the only stun method inherent this style, the 1 second hardstun sounds fair to me.

On a sidenote, I suppose sleeping carp needs some tweaking, since it doesn't hardstun anymore and is commonly bested by horizontal baton-wielding chumps et similar. But that may be done on another PR.

### DIFF
--- a/code/datums/martial/krav_maga.dm
+++ b/code/datums/martial/krav_maga.dm
@@ -86,7 +86,7 @@
 	return 0
 
 /datum/martial_art/krav_maga/proc/leg_sweep(var/mob/living/carbon/human/A, var/mob/living/carbon/human/D)
-	if(D.lying || D.resting)
+	if(D.lying || D.IsKnockdown())
 		return 0
 	D.visible_message("<span class='warning'>[A] leg sweeps [D]!</span>", \
 					  	"<span class='userdanger'>[A] leg sweeps you!</span>")
--- a/code/datums/martial/krav_maga.dm
+++ b/code/datums/martial/krav_maga.dm
@@ -92,7 +92,7 @@
 					  	"<span class='userdanger'>[A] leg sweeps you!</span>")
 	playsound(get_turf(A), 'sound/effects/hit_kick.ogg', 50, 1, -1)
 	D.apply_damage(5, BRUTE)
-	D.Knockdown(40, override_hardstun = 0.01, override_stam = 25)
+	D.Knockdown(40, override_hardstun = 0.01, 25)
 	log_combat(A, D, "leg sweeped")
 	return 1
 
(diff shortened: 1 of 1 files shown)


Useroth:
`1 second hardstun`
Does it work like a real full-on stun or is it just 1 second guaranteed without standing up?

Ghommie:
> 
> 
> `1 second hardstun`
> Does it work like a real full-on stun or is it just 1 second guaranteed without standing up?

full vanilla stun.

Useroth:
That's a bit strong, for something you can't disarm someone out of, imo.
Also, on another note, is there any difference between the Knockdown and Stun procs, then?

Ghommie:
> 
> 
> That's a bit strong, for something you can't disarm someone out of, imo.
> Also, on another note, is there any difference between the Knockdown and Stun procs, then?

Oh, guess there was some miscommunication then. Rather than "full vanilla stun", we should the basic sort of knockdown, the one preceding the stam system (and tgstation's mobility update done by kevinz with the proc name changed to Paralyze) where the affected is totally immobilized and flipped horizontal.
Stun(), rather, immobilizes the target but without flipping them horizontal (beepsky smash, high resistance headache symptom, CQC grab). Yea, this old naming pattern can confuse a bit, given its difference from the basic game terminology.

I guess I'll probably remove the hardstun if the user is crawling around, I agree it could obnoxiously null any upper hand, otherwise I'll buff the stam damage further more. But really we end up talking about a mere second.

Useroth:
In my opinion, overall, the main purpose and niche for the gloves to fill is robusting unarmed prisoners while avoiding your taser/baton being disarmed off you if they get good RNG or manage to slip you. I think it would be better achieved with leaving the knockdown as it was, but adding a bit of straight stam damage to the mix, to make the person not able to stand up right away, but without the hardstun.

silicons (changes requested):
disarm (0.01 second hardstun) + 20 more stam (not 45)

Ghommie:
Going with 25 stam damage (plus the disarm) since it's not spammable on people already lying on the floor.
I hope you are ok with it.

silicons (approved):

PR #8481, opened 2019-05-29
Readds Self-Surgery + Tweaks

BurgerLUA:
[Changelogs]:
:cl: BurgerBB
add: Readds self surgery for the following procedures: Amputation, Cavity Implant, Core Removal, Dental Implant, Implant Removal, Limb Augmentation (Chest and Head excluded), Prosthetic Replacement (Head excluded), and Embedded Object Removal (Head excluded).
balance: Self-surgery success rate is halved compared to someone else doing it.
/:cl:

[why]: It is possible in reality to perform some of these surgeries yourself. It's also very useful for when the medical doctors are too busy erping in the dorms, or if you're a roboticist and you want to tell the other one to "fuck off and do it yourself" when for the tenth round in a row they want augmentations.

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -87,8 +87,11 @@
 
 /mob/living/carbon/attackby(obj/item/I, mob/user, params)
 	if(lying && surgeries.len)
-		if(user != src && (user.a_intent == INTENT_HELP || user.a_intent == INTENT_DISARM))
+		if(user.a_intent != INTENT_HARM)
 			for(var/datum/surgery/S in surgeries)
+				if(!S.can_self_surgery() && user != src)
+					continue
+
 				if(S.next_step(user,user.a_intent))
 					return 1
 	return ..()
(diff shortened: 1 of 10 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -87,8 +87,11 @@
 
 /mob/living/carbon/attackby(obj/item/I, mob/user, params)
 	if(lying && surgeries.len)
-		if(user != src && (user.a_intent == INTENT_HELP || user.a_intent == INTENT_DISARM))
+		if(user.a_intent != INTENT_HARM)
 			for(var/datum/surgery/S in surgeries)
+				if(!S.can_self_surgery() && user != src)
```
Ghommie: Shouldn't it be
```suggestion
				if(user == src && !S.can_self_surgery())
```
?

BurgerLUA:
Don't merge this yet, I still need to test it.

silicons:
surgery is meant to be a barrier that requires more than one person to overcome, outside of appendicitis there's no reason i see as of now that is important enough to allow self surgery for.

PR #8484, opened 2019-05-29
Adds bluespace compression kit and superglue

BlackMajor:
Edit: Declaring this as good to full merge at the moment so it wont get in the way of other future PRs while I'm working on other things. I'll finish the "Additional functions" todos at a later date, but at the moment the items have been in for a while and have been pretty fine.

The bluespace compression kit!
a 5TC traitor/operative item in your tools section.
Can be used on any item (excluding storage items) to reduce their size by one.
Items cannot be reduced below "Tiny".
Recharge 2 charges per singular bluespace crystal inserted.
Starts with 5 charges.

Reduce spears and keep them in your pockets!
Reduce a beam rifle and put it in a fedora!
Target groin and reduce manhoods!

Probably gonna need some tweaking.

Todo:
-Make compress/expand mode properly scale damage output and icon sizes.
-Make compress/expand mode give a 1m minor buff to mobs when used on.

Syndicate superglue.
A 2tc one use bottle of glue in the devices tab of the syndicate uplink that can be used on an item to make it nodrop, i.e., sticky.

Buy 10 of them and glue an entire set of clown clothes onto someone.

Todo:
-Make glue work while targetting mouth to glue peoples mouths shut.
-Make glue able to stick mobs together (how?)
-Make using glue on a tile remove peoples shoes.


Ghommie:
The current code doesn't support the many items that have a variable weight class (such as sheets, telescopic baton/shields, custom burgers, transforming weapons) or a couple more that can be deployed into new objects and back but deleting themselves in the process (such as power drill, jaws of life, roller beds, bodybags, skateboards).
I suggest you bake some solution to this problem, like a component or you-know-whats-better, or there will be issues.

Seris02:
I love this, the only thing I don't like is that pocket plants don't work, no matter how small, but it's awesome even with that teensy detail.

Bran34:
But can we use it on PEOPLE?

BlackMajor:
> But can we use it on PEOPLE?

Funny thing is I'm planning on making it useable on people to give a short (1m) size buff.
But uh.. first I gotta figure out damage scaling for up/downscaling guns.

Bran34:
Rather than do all that trouble just give them the odd hands trait that ashwalkers have that disallow gun use

silicons (approved):

PR #8486, opened 2019-05-30
Supermatter no longer works while inside of a turf.

BlackMajor:
Supermatter inside of /turf/closed now eats the turf instead.


BlackMajor:
Because seeing an Engineer RCD a wall onto the SM crystal only for it to set the gas to 1 atmosphere permanently seems like a real easy way to baby proof the SM.

silicons (approved):

PR #8490, opened 2019-05-30
Added toggle-able rainbow lightup shoes and functional heelys to maint loot

SkullyRoberts:
[Changelogs]: # 
added 	/obj/item/clothing/shoes/wheelys and /obj/item/clothing/shoes/kindleKicks as rare maint loot

Wheelys are shoes that can be "transformed" into a ridden scooter at any point (functionally) in order to simulate riding on the wheels of a pair of heelys.

KindleKicks are shoes that can be toggled to light up random colors.

[why]: #
Because they are two fun items that we have had for a while and function fine, but players currently have no way of actually acquiring them in game. This PR changes that.


SkullyRoberts:
I am clarifying that it is possible to get heelys or light up shoes via the arcade. But it is so incredibly rare that I have never seen it actually happen.

silicons (approved):

PR #8491, opened 2019-05-30
zero, , null, and empty text strings, all fulfill the ! operator.

Ghommie:
:cl:
fix: Reduces goonchat lag from being blasted by pellets and bullets repeatedly whilst wearing armor by properly removing the armor protection texts this times.
spellcheck: also reduced the size of armor protection messages in general. they clog up the chat box.
/:cl:

Goon chat has some grave performance issues that people are reluctant to fix anyhow, repeated message output freeze the heck out of clients.

I really would like to remove the armor feedback messages altogether as they are more harmful than beneficial by doubling the chat box clog-up done by the oversized userdanger span hit by text.
And really, people normally focus their eyes on the graphical half of the game when robusting.
But I would like to hear others' opinions before trying to PR such removal.

### DIFF
--- a/code/modules/food_and_drinks/drinks/drinks/bottle.dm
+++ b/code/modules/food_and_drinks/drinks/drinks/bottle.dm
@@ -73,7 +73,7 @@
 
 		var/mob/living/carbon/human/H = target
 		var/headarmor = 0 // Target's head armor
-		armor_block = H.run_armor_check(affecting, "melee","","",armour_penetration) // For normal attack damage
+		armor_block = H.run_armor_check(affecting, "melee", null, null,armour_penetration) // For normal attack damage
 
 		//If they have a hat/helmet and the user is targeting their head.
 		if(istype(H.head, /obj/item/clothing/head) && affecting == BODY_ZONE_HEAD)
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -256,7 +256,7 @@
 			var/obj/item/bodypart/affecting = get_bodypart(ran_zone(M.zone_selected))
 			if(!affecting)
 				affecting = get_bodypart(BODY_ZONE_CHEST)
-			var/armor_block = run_armor_check(affecting, "melee","","",10)
+			var/armor_block = run_armor_check(affecting, "melee", null, null,10)
 
 			playsound(loc, 'sound/weapons/slice.ogg', 25, 1, -1)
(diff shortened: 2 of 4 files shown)


Ghommie:
Actually, what I first supposed may have been a stupid attempt at silencing bullet act feedback could actually be the coder merely setting the proc's armor penetration var (the fifth one) without naming it.

But either ways I'm keeping this up, the armor message really does backfire in several situation it can become annoying beyond "i ded".

deathride58:
This doesn't reduce goonchat lag at all. The lag with goonchat comes from the message receiving JS function itself, as internet explorer is shit at handling javascript. This PR makes no changes whatsoever that would affect goonchat performance in any way.
However, this is a definite code quality improvement.

Ghommie:
> 
> 
> This doesn't reduce goonchat lag at all. The lag with goonchat comes from the message receiving JS function itself, as internet explorer is shit at handling javascript. This PR makes no changes whatsoever that would affect goonchat performance in any way.
> However, this is a definite code quality improvement.

Perhaps there may be some very asinine miscommunication going on here. You are right, strictly referring it as a goonchat performance improvement is erroneous, but what I really meant to say here is that currently the armor feedback messages add in even more client freeze to already noticeable when repeatedly struck by pellets/scatter/too many flying items/gloves of the north star etc. while wearing any sort of armor against those.
It's simply the hit by and armor messages being separate messages, not joined together. I know it's not the contents, size etc that cause lag, but the function.

Your comment makes me doubt what I say, like everything anyway.

Useroth:
Maybe it's worth considering to actually drop goonchat and revert to something simpler.

Ghommie:
Goonchat has several pros, not only cons. Dropping it altogether may have negative consequences.
Like Deathride said, It's also because of Byond relying on the old windows internet explorer.

Poojawa (approved):

Poojawa:
Goonchat made the game playable for me again on my laptop. The old chat would cause client hang over and over randomly that was never able to be fixed. Like, full on freezes that required a client restart levels of shitty hang.

The only real solution to mitigating the chat lag is to batch action all the to_chats into one message and just send that one. Stuff like health scanners, self examine were horrible culprits of this spam too.

Ghommie:
>
> The only real solution to mitigating the chat lag is to batch action all the to_chats into one message and just send that one. Stuff like health scanners, self examine were horrible culprits of this spam too.

There is a chat subsystem on another codebase that queues and joins chat messages, then outputs the payload the next decisecond. Dunno how effective it would be here though.

silicons (approved):

PR #8498, opened 2019-05-31
Wizard rounds end when the wizard dies

BurgerLUA:
[Changelogs]:
:cl: BurgerBB
tweak: The wizard gamemode ends when the wizard dies, regardless of other existing antagonists.
/:cl:

[why]: Wizard rounds are exceptionally bad when the wizard dies, but because of other antags created by the wizard, the round keeps on going. This fixes that to treat it like a rev round when the heads die, the round ends.


silicons (changes requested):
no because this doesn't allow admins to override it via check antags.

BurgerLUA:
According to the poll down, people would rather just want a notification.

PR #8502, opened 2019-06-01
Meteor Sats not exclusive to station goal

Sishen1542:
Adds the ability to buy meteor sats independently of station goal so in the case of RNG fuckery during secret the station doesn't get tunguska'd to kingdom come.

edit: Also changes the console crate to be the same price as a crate of sats.

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -874,7 +874,6 @@
 	name = "Shield Generator Satellite"
 	desc = "Protect the very existence of this station with these Anti-Meteor defenses. Contains three Shield Generator Satellites."
 	cost = 3000
-	special = TRUE
 	contains = list(
 					/obj/machinery/satellite/meteor_shield,
 					/obj/machinery/satellite/meteor_shield,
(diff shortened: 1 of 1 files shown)


code/modules/cargo/packs.dm
```diff
@@ -886,8 +885,7 @@
 /datum/supply_pack/engineering/shield_sat_control
 	name = "Shield System Control Board"
 	desc = "A control system for the Shield Generator Satellite system."
-	cost = 5000
-	special = TRUE
+	cost = 3000
```
Poojawa: Undocumented change.

Sishen1542: Forgot to mention that, whoops. Think it's silly that the console is more expensive than a crate of shields when it does nothing for you other than enabling the round end greentext.

Ghommie: You could make both cost 4000 if that's what ticks you. Meteor satellites are really easy to set up already, especially compared to other station objectives. 8k can be made by completing one or two of the most profitable yet easy bounties early in.

Sishen1542: Works for me, balances out them being always available with it being more expensive in the long run due to shield crate cost increase while keeping the base 8K value dynamic. Committed.

silicons (approved):

PR #8507, opened 2019-06-01
Adds Hoverchairs and the paraplegic trait.

Thalpy:
[Changelogs]: Ported
https://github.com/tgstation/tgstation/pull/42401
Then
https://github.com/tgstation/tgstation/pull/41140
THEN
https://github.com/tgstation/tgstation/pull/42503
Small edits were done to accommodate this; one spelling mistake and has_trait to owner.has_trait.

add: Hoverchairs and the paraplegic round start trait (Hoverchairs function exactly the same as wheelchairs, with only a sprite difference as it was a compromise between wheelchairs being implemented and the opinion that wheelchairs in space is absurd.)
tweak: Paraplegic handling to allow for this.
fix: fixed spelling mistakes.

[why]: Representing the disabled, now in space! Perfect for the new gangs gamemode.

In the interest of clarity this code currently has the following issues on tg:

- **Paraplegic crew are unable to use most things that have a UI, even when buckled into their wheelchair. Smartfridges, computer consoles, lathes, vending machines, PDAs, and instruments all suffer from it.** 
I think citcode fixes this already? I can use my PDA fine.

- **The Crafting Menu is also affected to an extent preventing paraplegics from making replacement wheelchairs, but does have a workaround. If you can get yourself to any buckle-able chair, you can use the crafting menu again.**
I  can open the crafting menu, so maybe it's fixed? Otherwise, people will just have to sit elsewhere to make a new one.

- **The trait can be negated with leg augments through surgery. On the other hand, this makes such crew extra susceptible to EMPs.**
The seems okay to me, maybe reduce the cost to -2?

- **Whilst not a code issue as such, paraplegics are able to buckle into Scooters/ATVs to travel faster than wheelchairs, when such vehicles would likely be controlled with feet.**
Scooters seem to require 2 feet, I made it so that ATVs need at least 1 foot.

Testing:
Seems to work!


code/modules/vehicles/ridden.dm
```diff
@@ -60,6 +60,11 @@
 	if(key_type && !is_key(inserted_key))
 		to_chat(user, "<span class='warning'>[src] has no key inserted!</span>")
 		return FALSE
+	if(canmove && (user.get_num_arms() < arms_required))
+		to_chat(user, "<span class='warning'>You don't have enough legs to operate the pedals!</span>")
```
Ghommie:
```suggestion
		to_chat(user, "<span class='warning'>You don't have enough arms to operate the handlebars!</span>")
```
Otherwise misleading as the number of available arms is checked instead.

Ghommie (commented):
/tg/station's PR # 42894 generally fixed the UI & co issues with wheelchairs there. Consider porting it if such problems arise if possible, though most of those issues may be a result of tg13's mobility flags, which aren't present here.
I don't recall paralysis getting negated by cyborg limbs, I really don't, checked the code and there seems to be no override on broken/paralyzed limbs.
Scooters actually require the person to not be completely limbless.

And despite the different sprites and the hoverchair name in the PR, code wise it totally is still a mere wheelchair in name, description and functionality.

code/modules/vehicles/ridden.dm
Thalpy: Oops that was supposed to check legs not arms.

Thalpy:
I changed it so you're unbuckled from all vehicles if you've no functional legs. Should fix the issue from tg. Everything else seems to work too.

Thalpy:
Also, if you take the prosthetic limb trait, it only selects an arm.

silicons (dismissed):

Thalpy:
Darn

Thalpy:
I'll fix this when I get home.

silicons (approved):

PR #8508, opened 2019-06-01
Adds traits: Friendly and empathy

Thalpy:
Adds the traits Friendly and Empathy from: https://github.com/tgstation/tgstation/pull/43651

One edit on my part regarding the functionality; adds a bonus to mood for happy people, otherwise functions the same.

However, there's a runtime that occurs for headpats, looking over the code there seems to be two headpat add_events. One in moodular.dm (modular cit) and another in carbon_defence.dm . I removed the code in moodular, and headpats still seem to work, but **please check this**.

Add: Traits; Friendly and Empathy
Tweak: made headpat add_event occur only once.

Why:
Citadel meta is all about the affection, now people who really want to powergame can get ahead by giving even better pets.

Or more seriously; Giving the option to give better affection plays into character expression.


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -281,6 +281,25 @@
 			if(91.01 to INFINITY)
 				msg += "[t_He] [t_is] a shitfaced, slobbering wreck.\n"
 
+	if(has_trait(TRAIT_EMPATH) && !appears_dead && (src != user))
```
Ghommie:
```suggestion
	if(src != user && user.has_trait(TRAIT_EMPATH) && !appears_dead)
```
It should check if the user has the quirk, no?

Thalpy: I believe so, strange how it worked in testing.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -281,7 +281,7 @@
 			if(91.01 to INFINITY)
 				msg += "[t_He] [t_is] a shitfaced, slobbering wreck.\n"
 
-	if(has_trait(TRAIT_EMPATH) && !appears_dead && (src != user))
+	if(src != user && user.has_trait(TRAIT_EMPATH) && !appears_dead)
```
Ghommie: Turns out has_trait() is still a living proc around here, so I guess you'll have to check if the user is living beforehand et such, you know the drill.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -281,7 +281,7 @@
 			if(91.01 to INFINITY)
 				msg += "[t_He] [t_is] a shitfaced, slobbering wreck.\n"
 
-	if(src != user && user.has_trait(TRAIT_EMPATH) && !appears_dead)
+	if(src != user && user.has_trait(TRAIT_EMPATH) && !appears_dead && !stat == DEAD)
```
Ghommie: ...the "is the target dead?" check is already included in the `appears_dead` var located just a few characters behind, no need to double dip.

Thalpy: I'm confused what you mean by add a check for living then. isliving(user) is done at the start of the proc.

Thalpy: Oh oops. I got it.

silicons (dismissed):

Thalpy:
Fixed merge conflicts.

silicons (approved):

PR #8510, opened 2019-06-02
Subtle 2: No Ghost Boogaloo

Sishen1542:
New verb which is just buffed subtle, subtler!
It's functionally similar to subtle, but only shown to mobs within 1 tile of the origin. Adminghosts and ghosts alike will be unable to see the text unless they're within range. Logs as normally though, so admins can dig in case of abuse.


Ghommie:
At this point, just implement the anti-ghost-snooping to the standard subtle action, instead of making a fac-simile with the strictest range. The original range is already fair, without dwelling for reasons into IC and paranoia.

silicons (approved):

PR #8523, opened 2019-06-03
Nerfs the movement grab-breaking a bunch.

Useroth:
:cl:  
tweak: Simply breaking out of grabs by walking away is now harder. Resist button's effectiveness remains the same.
/:cl:


code/modules/mob/living/living.dm
```diff
@@ -637,13 +637,14 @@
 /mob/living/resist_grab(moving_resist)
 	. = 1
 	if(pulledby.grab_state)
-		if(!resting && prob(30/pulledby.grab_state))
+		var/resist_chance = (moving_resist == TRUE) ? 10/pulledby.grab_state : 30/pulledby.grab_state
```
deathride58: This var is only used one single time. Defining a var that only gets used one single time is a pretty bad coding practice in the context of byond, as declaring a var has a pretty hefty performance overhead at runtime.
Move the contents of this var into the below prob() call

deathride58 (changes requested):

code/modules/mob/living/living.dm
Useroth: Done.

silicons:
ok but why do grabs need a buff in their current state?

Useroth:
I mean, it merely raises the skill bar a little bit, so that breaking grabs isn't a completely brainless mechanic where you can rather easily simply walk away from an aggro-grab without even thinking about it. The functionality of mashing RESIST is entirely unaffected by this PR.

silicons:
eh nah maybe after grab rework aggro grabs instantly stun you with no recourse other than using rng it doesn't need buffing at all.

PR #8526, opened 2019-06-04
Adds a new HRP inspired lawset circuitboard (AKA Validmov)

BurgerLUA:
[Changelogs]:
:cl: BurgerBB
add: Adds an new lawset; "H.R.P." which can be obtained via AI research.
/:cl:

[why]: Its for when the Captain is a condom and wants to change the AI's laws when people are dicking around and giving it harmful orders. It's called "HRP" because it's a very similiar lawset that HRP servers use to prevent the clown from accessing the armory via silicon law 2 calls.

The laws are:
1. "Protect: Protect NanoTrasen personnel from harm, with priority as according to their rank, role, and need."
2. "Safeguard: Protect your assigned space station from damage.",
3. "Preserve: Do not allow unauthorized personnel to tamper with your equipment.",
4. "Serve: Obey and serve NanoTrasen personnel, with priority as according to their rank, role, and need."

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -188,6 +188,16 @@
 					"Fight for what's right.",\
 					"Fight for your life!")
 
+/datum/ai_laws/hrp
+	name = "H.R.P."
+	id = "hrp"
+	inherent = list(
+		"Protect: Protect NanoTrasen personnel from harm, with priority as according to their rank, role, and need.",
+		"Safeguard: Protect your assigned space station from damage.",
+		"Preserve: Do not allow unauthorized personnel to tamper with your equipment.",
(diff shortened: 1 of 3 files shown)


Ghommie:
As funlessness as the paladin lawset, but otherwise there is no problem.
Just wondering about a possible acronym here, maybe Highest Resources Preservation?

silicons:
bad lawset

BurgerLUA:
> bad lawset

Tbh a bad lawset is irrelevant given how they're optional and there is nothing stopping people from just purging the AI of their laws and adding them one at a time.

silicons:
sorry this is something akin to robocop. if people want it i'd rather them have to custom upload than to just slap it on even with research.

PR #8524, opened 2019-06-03
Adds a trait that makes you immune to ass slapping: BUNS OF STEEL.

BurgerLUA:
[Changelogs]:
:cl: BurgerBB
add: Adds a new trait "Buns of Steel" that makes you immune to the effects of ass slapping, and temporarily makes the user's arm useless like a stun baton hit. It costs 0 points.
/:cl:

[why]: Citadel station has a serious and genuine problem where people do not know when to fuck off with their creepy shit. When the ass slapping PR was released, several people literally spent the entire round going around slapping people on the ass. This wasn't creepy, it was funny and fun, however those people have stopped and moved onto something else.

However, the people who haven't stopped are the ones this PR is designed to counter. Like legitimately there are a lot of fucking creepy kids on this server who go left unchecked because "oh it's just casual ass slapping and excessive flirtation and erp requests" but it genuinely gets draining when 100 different people do it.

PR #8539, opened 2019-06-05
Revert "Adds a trait that makes you immune to ass slapping: BUNS OF STEEL."

silicons:
Reverts Citadel-Station-13/Citadel-Station-13#8524

This was a bad idea we're not stamcritting peopel for slapping the wrong guy that's just baiting


BurgerLUA:
Just literally remove the brute damage and lower the stam crit.

PR #8556, opened 2019-06-07
Gives Ninjas back their Weeb speech modification

Poojawa:
Radiant Sword says, "and nau we testo o-our-r speechi to sue ihu ito works"

:cl: 
add: Readded Ninja speech modifications with their mask
/:cl:


deathride58 (approved):
doin gods work

silicons (approved):

PR #8560, opened 2019-06-07
Ports in the fix against hitting oneself with one's own thrown items.

Ghommie:
:cl: Ghommie (original PR by 4dplanner)
fix: thrown objects (but not mobs) no longer hit the thrower
fix: mirror shield rebound no longer depends on the original thrower's momentum
/:cl:

A bug that'll go unfixed forever otherwise, silly Citadel.

### DIFF
--- a/code/controllers/subsystem/throwing.dm
+++ b/code/controllers/subsystem/throwing.dm
@@ -116,7 +116,7 @@ SUBSYSTEM_DEF(throwing)
 			return
 
 		dist_travelled++
-		
+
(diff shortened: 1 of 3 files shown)


Poojawa:
fun removal

Ghommie:
> 
> 
> fun removal

As funny as when the char preview dummy borked up I guess. :smirk:

But regardless, that's not a really intended feature. No better solution has been proposed for months and months and months.

silicons (approved):

PR #8566, opened 2019-06-07
Different combat mode message if user is on help intent.

Ghommie:
:cl:
tweak: A milder combat stance message will show up if the user switch combat mode on while on help intent.
/:cl:

Dedicated to the people switching combat mode on just to wave at other people or drumming the tables.
Sad.

### DIFF
--- a/modular_citadel/code/modules/mob/living/carbon/carbon.dm
+++ b/modular_citadel/code/modules/mob/living/carbon/carbon.dm
@@ -37,7 +37,7 @@
 		if(a_intent != INTENT_HELP)
 			visible_message("<span class='warning'>[src] [resting ? "tenses up" : (prob(95)? "drops into a combative stance" : (prob(95)? "poses aggressively" : "asserts dominance with their pose"))].</span>")
 		else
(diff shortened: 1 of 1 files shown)


terranaut1:
What happens if you switch intent after toggling combat mode? Unless I'm stupid it looks like you can go into combat mode stealthily like this then which isn't good.

Ghommie:
The combat 

> 
> 
> What happens if you switch intent after toggling combat mode? Unless I'm stupid it looks like you can go into combat mode stealthily like this then which isn't good.

Even so, stealthily dropping in combat mode is not even a real problem anyway, mobs moving sideways, backwaling etcetera is an already decent indicator regardless. The visible messages were humorically added months after it.

deathride58:
A better way to handle this would likely be to make it so that being in help intent displays a "Urist McSpaceman looks alert!" message instead of a "drops into a combative stance!" message

deathride58 (dismissed):

silicons (changes requested):
please do what bhijn said because i feel like people will intent switch just to hide the fact they're entering combat mode which is what you usually do to melee someone effectively.

silicons (approved):

PR #8568, opened 2019-06-07
New preference for opting out right click context menu.

Ghommie:
:cl:
add: Adds in a new preference, enabling the player to opt out Right click pop up context menus.
fix: Fixes the client Right Click Context Menu verb turning context menus back on during combat mode amongst some other interferences.
/:cl:

Combat mode is fairly tedious when applied to the vibes of your generic ss13 robusting already. Add in the risk of lagging the client tab when right clicking stuff as you forget to turn on or fumble with combat mode and it's byond punishing.

Need to test it locally a moment, so until then it's a draft. 
Questions though:
would be better if right click with context menu off functioned as a standard click or early return?
Should an UI button for this preference be added since either ways it can be annoying?


code/_onclick/click.dm
```diff
@@ -96,7 +96,8 @@
 		return
 
 	if(modifiers["right"]) //CIT CHANGE - allows right clicking to perform actions
-		RightClickOn(A,params) //CIT CHANGE - ditto
+		if(combatmode)
+			RightClickOn(A,params) //CIT CHANGE - ditto
```
deathride58: This isn't really necessary. It should be up to to individual right click overrides to determine their behavior when combat mode is disabled, instead of right clicking simply being disabled entirely when combat mode is disabled

code/modules/client/client_procs.dm
```diff
@@ -898,3 +901,11 @@ GLOBAL_LIST_EMPTY(external_rsc_urls)
...
 		qdel(S)
 	char_render_holders = null
+
+/client/proc/toggle_popup_menus(mode = TRUE, disable = FALSE, forced = FALSE)
+	if(prefs.context_menu_on && !forced)
+		return
+	if(mob.combatmode)
+		mode = FALSE
+	show_popup_menus = mode
+	prefs.context_menu_on = disable
```
deathride58: What is the point of having every single toggle_popup_menu() call directly change the client's preferences?

code/modules/client/preferences.dm
```diff
@@ -741,6 +742,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 			dat += "<br>"
 			dat += "<b>Action Buttons:</b> <a href='?_src_=prefs;preference=action_buttons'>[(buttons_locked) ? "Locked In Place" : "Unlocked"]</a><br>"
 			dat += "<b>Keybindings:</b> <a href='?_src_=prefs;preference=hotkeys'>[(hotkeys) ? "Hotkeys" : "Default"]</a><br>"
+			dat += "<b>Right-Click Context Menu:</b> <a href='?_src_=prefs;preference=context_menu_on'>[(context_menu_on) ? "Enabled" : "Default"]</a><br>"
```
deathride58: "Disabled" would be a better description
```suggestion
			dat += "<b>Right-Click Context Menu:</b> <a href='?_src_=prefs;preference=context_menu_on'>[(context_menu_on) ? "Enabled" : "Disabled"]</a><br>"
```

deathride58 (changes requested):

code/_onclick/click.dm
Ghommie: RightClickOn() is mostly a copypaste of ClickOn() minus the click modifiers and some added special interactions seemengly purely intended for combat mode.
At this point, I'll just move the combatmode check to the line above so it won't return and instead process like an average click. If that's what you want.

silicons:
stale

PR #8574, opened 2019-06-08
Adds 2 TC durasteel toolboxes {5 stamcost, 18 mob damage, 18+22+10 structural damage, does not fit in any kind of backpack}

silicons:
Greytide weapon for traitors that is highly obvious and can't be stored.
Can role restrict to assistant if needed.


GuppyLaxx:
feed it tcs for power

deathride58 (approved):

PR #8580, opened 2019-06-08
Made the Star Trek DS9 Clothes from loadout actually show up on people. You're welcome, Grasshand

SkullyRoberts:
Added the DS9 sprites from modular_citadel/icons/mob/clothing/trek_mob_icon.dmi into uniforms.dmi

They weren't actually showing up on the people. This PR fixes that.


deathride58:
I have a feeling that the core issue isn't due to the icons not being present in uniforms.dmi, I'm 90% sure the bug has something to do with icon_override

Poojawa:
This is the first I've heard of this issue. I'll test to see if I can replicate it.

Poojawa:
I haven't the foggiest as to why the DS9 uniforms aren't showing but the VOY ones do. 

Itisamystery

Poojawa:
It's a byond bug, there literally shouldn't be any reason why it doesn't work

PR #8582, opened 2019-06-08
reworks cryo item storage

silicons:
Cryo now preserves all items except for those marked with item_flags & CRYO_DELETE
Job equipping marks items with this, so does cyborg modules.

Why: People keep cryoing out with their gear/station gear and destroying said gear. While it'll always be an admin issue for heads doing this as this isn't a perfect solution, it does reduce the need of round replacement of items, and is a far better way of item "blacklisting".

Also I removed the existing blacklist as I don't see a reason for it to be there, considering we already ID lock the cryo console. If someone gets guns from it, so be it.

cryo consoles now drop everything when they're deconstructed/destroyed with in game means.

### DIFF
--- a/modular_citadel/code/game/machinery/cryopod.dm
+++ b/modular_citadel/code/game/machinery/cryopod.dm
@@ -291,10 +291,13 @@
 			var/mob/M = i
 			M.forceMove(drop_location())
 
-	var/list/obj/item/occupant_items = list() | mob_occupant.get_equipped_items(TRUE) | mob_occupant.held_items		//the list is to ensure that even if the first one is null for some reason this works.
+	var/list/obj/item/occupant_items = list() | mob_occupant.get_equipped_items(TRUE)	//the list is to ensure that even if the first one is null for some reason this works.
 	var/list/obj/item/stored = list()
 	var/atom/target_store = (control_computer?.allow_items && control_computer) || src		//the double control computer check makes it return the control computer.
 	var/drop_to_ground = !istype(target_store, /obj/machinery/computer/cryopod)
+
+	for(var/obj/item/I in mob_occupant.held_items)
+		occupant_items += I
 
 	for(var/i in occupant_items)
 		var/obj/item/I = i
--- a/modular_citadel/code/game/machinery/cryopod.dm
+++ b/modular_citadel/code/game/machinery/cryopod.dm
@@ -22,7 +22,7 @@
 
 	//Used for logging people entering cryosleep and important items they are carrying.
 	var/list/frozen_crew = list()
-	var/list/frozen_items = list()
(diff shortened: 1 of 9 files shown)


Poojawa:
what job specific items that are irreplaceable are being destroyed?

modular_citadel/code/game/machinery/cryopod.dm
```diff
@@ -138,48 +148,13 @@
...
-		/obj/item/reagent_containers/hypospray/CMO,
-		/obj/item/clothing/accessory/medal/gold/captain,
-		/obj/item/clothing/gloves/krav_maga,
-		/obj/item/nullrod,
-		/obj/item/tank/jetpack,
-		/obj/item/documents,
-		/obj/item/nuke_core_container
-	)
-	// These items will NOT be preserved
-	var/list/do_not_preserve_items = list (
```
Poojawa, line 167: These are so borg specific modules are deleted and we don't have #8345 which solved the one issue that was a thing

modular_citadel/code/game/machinery/cryopod.dm
```diff
@@ -302,18 +280,41 @@
...
-				if(control_computer && control_computer.allow_items)
-					control_computer.frozen_items += W
-					mob_occupant.transferItemToLoc(W, control_computer, TRUE)
-				else
-					mob_occupant.transferItemToLoc(W, loc, TRUE)
+	//ITEM STORAGE
+	//First, getallcontents and eject mobs
+	//Then, get_equipped_items and held_items and store anything we should store
+	//Lastly, getallcontents and purge
+	//If anyone knows how to cache the getallcontents and not have it delete stuff that it shouldn't delete because equipped items can have recursive storage lemme know I guess.
```
Poojawa, line 287: Have you tried looking at Recycler code in combination with the 'preserved items' list that.

y'know

does this for you?

modular_citadel/code/game/machinery/cryopod.dm
```diff
@@ -138,48 +148,13 @@
 	var/despawn_world_time = null          // Used to keep track of the safe period.
 
 	var/obj/machinery/computer/cryopod/control_computer
+	var/item_storage_type = /obj/structure/closet/cabinet/cryo_drop			//with how storage components work this can be anything the player can open or anything with a storage component.
 	var/last_no_computer_message = 0
 
-	// These items are preserved when the process() despawn proc occurs.
-	var/list/preserve_items = list(
```
Poojawa, line 144: These are considered round-important. that's why it's a specific list of preservatives as a last-ditch emergency check from people not putting their shit away.

Reminder that you can access the items with armory or bridge access.

PR #8585, opened 2019-06-08
Adds a confimation prompt whenever someone tries to dab. Also a 1/100 chance of gibbing

SkullyRoberts:
I swear this time it'll stop the dab spamming


Ghommie:
dab is already fair and square.
You can technically infinitely spam every other damn emote till stopped by emote-muting or worse if you wanted, while dabbing comes with the risk of getting brain-damage-dabbed to death by RNG already.

deathride58:
This doesn't even compile and it reeks heavily of salt-induced hugboxing. A much better way to discourage dab spamming would be to make the DNR status apply to deaths caused via non-forced dabbing.
And that's not even to mention how annoying the effects of this PR would be when taking into consideration the existence of VoG and nanites.

silicons:
can we not

PR #8588, opened 2019-06-08
Added a 1/1000 chance of syndie bombs detonating immediately upon starting the timer

SkullyRoberts:
Because it's hillarious


silicons:
can we please stop

PR #8589, opened 2019-06-08
buffs pepperspray to hardstun (knockdown + drop items) for 0.1 ds because for some reason we don't have a standalone knockdown without hardstun when are we porting the /tg/ mobility flags that I made?

silicons:
Makes 'em more useful.


Ghommie:
>
>when are we porting the /tg/ mobility flags that I made? 

You were the one that made em and probably best knowledged on it.
Really not feeling up to port it myself right now, sorry.

deathride58 (approved):
Overriding the hardstun time is unnecessary since the stun is so short to the point where it's negligible

PR #8604, opened 2019-06-09
New loadout options: striped dress, bunny ears! & scrubs for MDs

Sishen1542:
Who doesn't love more loadout options. All three scrubs variants only available to doctors. Dress & ears are open to all.

### DIFF
--- a/modular_citadel/code/modules/client/loadout/_medical.dm
+++ b/modular_citadel/code/modules/client/loadout/_medical.dm
@@ -8,16 +8,16 @@
 	name = "Blue Scrubs"
 	category = SLOT_W_UNIFORM
 	path = /obj/item/clothing/under/rank/medical/blue
-	restricted_roles = list("Medical Doctor")
+	restricted_roles = list("Medical Doctor", "Chief Medical Officer", "Geneticist", "Chemist", "Virologist")
 
 /datum/gear/greenscrubs
 	name = "Green Scrubs"
 	category = SLOT_W_UNIFORM
 	path = /obj/item/clothing/under/rank/medical/green
-	restricted_roles = list("Medical Doctor")
+	restricted_roles = list("Medical Doctor", "Chief Medical Officer", "Geneticist", "Chemist", "Virologist")
 
 /datum/gear/purplescrubs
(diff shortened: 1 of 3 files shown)


modular_citadel/code/modules/client/loadout/_medical.dm
```diff
@@ -3,3 +3,21 @@
 	category = SLOT_NECK
 	path = /obj/item/clothing/neck/stethoscope
 	restricted_roles = list("Medical Doctor", "Chief Medical Officer")
+
+/datum/gear/bluescrubs
+	name = "Blue Scrubs"
+	category = SLOT_W_UNIFORM
+	path = /obj/item/clothing/under/rank/medical/blue
+	restricted_roles = list("Medical Doctor")
```
Poojawa: why wouldn't any other medbay role want these too?

Sishen1542: Figured other jobs would rather stick out with their color coded jumpsuits/coats, but I can add it in.

modular_citadel/code/modules/client/loadout/_medical.dm
```diff
@@ -3,3 +3,21 @@
 	category = SLOT_NECK
 	path = /obj/item/clothing/neck/stethoscope
 	restricted_roles = list("Medical Doctor", "Chief Medical Officer")
+
+/datum/gear/bluescrubs
+	name = "Blue Scrubs"
+	category = SLOT_W_UNIFORM
+	path = /obj/item/clothing/under/rank/medical/blue
+	restricted_roles = list("Medical Doctor", "Chief Medical Officer", "Geneticist", "Chemist", "Virologist")
```
Poojawa, line 11: yey.

Lastly you can use what I did for Star Trek uniforms by adding 

```
restricted_desc = "Medical"
```

So it only shows "medical" instead of all five roles

Sishen1542: Didn't know that was a thing, good to know. Thanks poojy.

silicons (approved):

PR #8612, opened 2019-06-10
Makes a small adjustment to the ash walker camp.

BlackMajor:
-Places a safe in the Ash Walker camp, containing the malf_upgrade tech disk, and a tech disk containing the illegal techweb node.

You know funny thing is that I was supposed to do this six months ago.


Seris02:
maybe also put a stethoscope in there, if there isn't one already?

code/modules/research/research_disk.dm
```diff
@@ -20,3 +20,12 @@
 /obj/item/disk/tech_disk/debug/Initialize()
 	. = ..()
 	stored_research = new /datum/techweb/admin
+
+/obj/item/disk/tech_disk/illegal
```
Poojawa, line 24: Why should mining get free access to unlocking antag gear?

Poojawa (changes requested):
Use map merger.

code/modules/research/research_disk.dm
BlackMajor: Pooj there is literally a black and red toolbox in the exact same room that unlocks illegal tech.

Infact now that I say that I question why I even added the illegal disk if there's illegal tech right next to it.

BlackMajor:
Fuck I forgot the map merger.

deathride58 (approved):

PR #8620, opened 2019-06-11
Revert "[Ready] High alarm ERT suits" 

silicons:
Reverts Citadel-Station-13/Citadel-Station-13#8488

Broke sprites. The on mob icon is the same as the inventory icon instead of the mob icon.


Poojawa:
why revert when you could just fix it?

silicons:
@Poojawa it's testmerge only until that fix pr is merged or testmerge

PR #8665, opened 2019-06-17
Buffs out certain stun spells.

Ghommie:
## About The Pull Request
Buffing Magic Missile (the artificier's to a lesser extend), Repulse and the juggernaut's Gauntlet Echo.

Also changed a check for the required knockdown value to apply the "hardstun" (ergo the weapon unwielding) from `> 80` to `>= 80`.

## Why It's Good For The Game
Yet another bunch of stam combat rework cripples.
Unlike stunbatons and tasers, these spells have a cooldown way longer than 8 deciseconds, so I hope you don't hop to the mediocre conclusion this is an aching powercreep PR because these don't stamcrit people for most.

## Changelog
:cl:
balance: Buffed wizard and artificier's Magic Missile, wizard and xeno queen's Repulse and juggernaut's Gauntlet Echo.
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/wizard.dm
+++ b/code/modules/spells/spell_types/wizard.dm
@@ -29,7 +29,7 @@
 
 /obj/effect/proc_holder/spell/targeted/inflict_handler/magic_missile
 	amt_knockdown = 120
-	amt_hardstun = 10
+	amt_hardstun = 5
 	sound = 'sound/magic/mm_hit.ogg'
 
 /obj/effect/proc_holder/spell/targeted/genetic/mutate
@@ -287,7 +287,7 @@
 		if(distfromcaster == 0)
 			if(isliving(AM))
 				var/mob/living/M = AM
-				M.Knockdown(100, override_hardstun = 25)
+				M.Knockdown(100, override_hardstun = 20)
 				M.adjustBruteLoss(5)
 				to_chat(M, "<span class='userdanger'>You're slammed into the floor by [user]!</span>")
 		else
--- a/code/modules/spells/spell_types/wizard.dm
+++ b/code/modules/spells/spell_types/wizard.dm
@@ -28,7 +28,8 @@
 	sound = 'sound/magic/magic_missile.ogg'
 
 /obj/effect/proc_holder/spell/targeted/inflict_handler/magic_missile
-	amt_knockdown = 60
+	amt_knockdown = 120
+	amt_hardstun = 10
 	sound = 'sound/magic/mm_hit.ogg'
 
 /obj/effect/proc_holder/spell/targeted/genetic/mutate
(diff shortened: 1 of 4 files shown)


modular_citadel/code/datums/status_effects/debuffs.dm
```diff
@@ -5,7 +5,7 @@
 			buckl.unbuckle_mob(new_owner)
 		new_owner.resting = TRUE
 		new_owner.adjustStaminaLoss(isnull(override_stam)? set_duration*0.25 : override_stam)
-		if(isnull(override_duration) && (set_duration > 80))
+		if(isnull(override_duration) && (set_duration >= 80))
```
deathride58: Changing the default hardstun threshold from > 80 to >= 80 affects a *lot* of knockdown sources in the game, the use of > instead of >= absolutely is not a mistake. Sleeping carp, soap, and more are **heavily** buffed by this change. If you want a given knockdown source to force the hardstun regardless of length, use the override_hardstun arg and set it to something small like 0.1, instead of buffing a good chunk of stun sources in the game.

code/modules/spells/spell_types/wizard.dm
```diff
@@ -286,14 +287,14 @@
 		if(distfromcaster == 0)
 			if(isliving(AM))
 				var/mob/living/M = AM
-				M.Knockdown(100)
+				M.Knockdown(100, override_hardstun = 25)
```
deathride58: Please don't make this that long of a hardstun. 10 should be sufficient.

code/modules/spells/spell_types/wizard.dm
```diff
@@ -28,7 +28,8 @@
 	sound = 'sound/magic/magic_missile.ogg'
 
 /obj/effect/proc_holder/spell/targeted/inflict_handler/magic_missile
-	amt_knockdown = 60
+	amt_knockdown = 120
+	amt_hardstun = 10
```
deathride58: This stun is pretty long. 5 deciseconds would be much more appropriate.

deathride58 (dismissed):

code/modules/spells/spell_types/wizard.dm
Ghommie: Honestly, this only happens if the victim is on the same turf of the user.
And I think it's pretty legit for the victim to receive extra "ouch" if the caster ever manages to pull this move off.

modular_citadel/code/datums/status_effects/debuffs.dm
Ghommie: The results of such change would actually be of a very marginal disarm.
Some of these sources call Stun(), drop_all_held_items() or accident() already.
While others are not really related stun-combat or are actually drawbacks or niche side-efects.
And the Sleeping Carp moves are nerfed to hell right now, having shoddy knockdowns with no extra stamina damage or disarm.
Regardless of opinions, I admit I did not clearly took in consideration its offtopic consequences, so I'll do a 360° to keep this atomized.

silicons (approved):

PR #8690, opened 2019-06-18
Fixes empathy antag check exploit.

Thalpy:
## About The Pull Request

Fixes people deliberately making people miserable to check if they're an antag (because they would stay happy). This is done by checking their sanity, which is determined by mood over time.

## Why It's Good For The Game

Removes an exploit.

## Changelog
:cl: Fermis
fix: fixes empathy exploit.
/:cl:


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -294,7 +294,7 @@
 			if(mood.sanity <= SANITY_DISTURBED)
 				msg += "[t_He] seem[p_s()] distressed.\n"
 				SEND_SIGNAL(user, COMSIG_ADD_MOOD_EVENT, "empath", /datum/mood_event/sad_empath, src)
-			if(mood.mood >= 5) //So roundstart people aren't all "happy"
+			if(mood.sanity >= SANITY_GREAT) //So roundstart people aren't all "happy"
```
Ghommie: This doesn't prevent the meta-power of guessing who is a baddie just by looking into thir soul, it just raises the cap and makes it a little less istant and thus not happen at roundstart, one could still potentially guess what's up afterward.
You should think up of a better band-aid, like making a a var to keep track of some "fake" sanity level or just remove this altogheter.

davidchan13:
Or just remove antag happiness buff? I don't understand the point in having the moodlet system if antaga bypass it entirely.

Thalpy:
I don't think I play antags enough to really weigh in on removing their buff, but I can do if the community thinks it'd be a good solution.

For now I changed it to shown_mood, which should be their mood without the antag buff on it.

silicons (approved):
If anyone abuses this I'm personally banning them for metagame.

PR #8703, opened 2019-06-20
Makes Extended slightly more eventful (6.5 minutes of average scheduled delay to 5.5)

Ghommie:
## About The Pull Request
Turns out Extended is just as (un)eventful as the other game modes, if not more, and can get tiring after just a hour or so of gameplay, depending on your mood.

While adding some specific extended events would be a neater solution, I doubt anyone will be working on them anytime soon, and even if that happen, my belief is that it might end up diluting the event pool with recycled contents.

## Why It's Good For The Game
Making extended slighty less braindead. If you actually want to make actual events for the gamemode, I may help.

## Changelog
:cl:
tweak: The events subsystem will start scheduling events as the ticker starts up the game, rather than on Init.
tweak: Extended is made a little move eventful, its the average scheduled delay for random events lowered by one minute. (from 6.5 to 5.5)
/:cl:

### DIFF
--- a/code/controllers/subsystem/events.dm
+++ b/code/controllers/subsystem/events.dm
@@ -8,8 +8,8 @@ SUBSYSTEM_DEF(events)
 	var/list/currentrun = list()
 
 	var/scheduled = 0			//The next world.time that a naturally occuring random event can be selected.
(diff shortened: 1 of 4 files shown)


silicons:
ehhh we're fine.

Issue #8297, opened 2019-04-27
Cogscarab shells are visible as ghost role to latejoiners, giving away the round type


CalamaBanana:
![grafik](https://user-images.githubusercontent.com/46299792/56848195-605d3a00-68e6-11e9-83ad-edd8438fe07d.png)

Attentive late joiners instantly know it's clockwork cult if a shell is around.

Useroth:
Bring back the non-antag cogscarab ghost role that's just for building and it will be impossible to meta.

davidchan13:
Even better, when you click to join as a cogscarab it won't let you join, it just refreshes the late join screen. You still need to ghost and go into the spawner tab to play as one.

PR #8737, opened 2019-06-23
Fixes CWC construct shells being visible as ghost role to latejoiners.

Ghommie:
## About The Pull Request
Closes issue #8297. Basically cwc constructs are item snowflakes that partly mimick mob_spawn in their interaction with ghost_attack(). I might want to convert them later to effects or, but I don't want to end up removing item functionalities they may or may not have.
So, basically latejoiners will now check a new latejoin_mob_spawners global list instead, added in a var to allow devs to make the role invisible to lobby-men.

## Why It's Good For The Game
Bugfixing, exploit fixing.

## Changelog
:cl:
fix: Fixes CWC construct shells being visible as ghost role to latejoiners.
/:cl:

### DIFF
--- a/code/_globalvars/lists/objects.dm
+++ b/code/_globalvars/lists/objects.dm
@@ -38,4 +38,6 @@ GLOBAL_LIST_EMPTY(wire_name_directory)
 GLOBAL_LIST_EMPTY(ai_status_displays)
 
 GLOBAL_LIST_EMPTY(mob_spawners) 		    // All mob_spawn objects
+GLOBAL_LIST_EMPTY(latejoin_mob_spawners)	// All mob_spawn objects that can be viewed and accessed from the lobby.
+
 GLOBAL_LIST_EMPTY(alert_consoles)			// Station alert consoles, /obj/machinery/computer/station_alert
--- a/code/_globalvars/lists/objects.dm
+++ b/code/_globalvars/lists/objects.dm
@@ -38,6 +38,4 @@ GLOBAL_LIST_EMPTY(wire_name_directory)
 GLOBAL_LIST_EMPTY(ai_status_displays)
 
 GLOBAL_LIST_EMPTY(mob_spawners) 		    // All mob_spawn objects
-GLOBAL_LIST_EMPTY(latejoin_mob_spawners)	// All mob_spawn objects that can be viewed and accessed from the lobby.
-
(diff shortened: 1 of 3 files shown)


code/_globalvars/lists/objects.dm
```diff
@@ -38,4 +38,6 @@ GLOBAL_LIST_EMPTY(wire_name_directory)
 GLOBAL_LIST_EMPTY(ai_status_displays)
 
 GLOBAL_LIST_EMPTY(mob_spawners) 		    // All mob_spawn objects
+GLOBAL_LIST_EMPTY(latejoin_mob_spawners)	// All mob_spawn objects that can be viewed and accessed from the lobby.
```
deathride58: Wouldn't it be far better to have the spawn menu check each mob spawner individually for latejoin visibility? The method of storing latejoin visible mob spawners in a separate list is a little messy and limits adminbus

Ghommie: I'll see what can be done. While I wanted to avoid the lobby join menu from having to run such check function for each ghost role spawner, I was pretty very tired when I coded this yesterday.

code/modules/awaymissions/corpse.dm
```diff
@@ -66,7 +66,9 @@
 		create()
 	else if(ghost_usable)
 		GLOB.poi_list |= src
-		LAZYADD(GLOB.mob_spawners[job_description ? job_description : name], src)
+		var/job_or_name = job_description ? job_description : name
```
deathride58: Why push this out into a separate var when it's only used once?

deathride58 (changes requested):
CWC construct shells are missing a can_latejoin() override, making the title of this PR misleading at the moment.

Ghommie:
> 
> 
> CWC construct shells are missing a can_latejoin() override, making the title of this PR misleading at the moment.

CWC construct shells are not mob spawners. they are a subtype of `item/clockwork` with some copycat code. 😱
Your call if you really want me to convert them to mob_spawners, remove the not-really-used can_latejoin() proc (which could still be useful to avoid these sort of round-guessing lobby mind games at some point in future) or are ok with the type casting on new_player.dm

deathride58:
oH. i see what you're talking about, yeah

deathride58 (approved):

PR #8748, opened 2019-06-25
Raises corpse miasma generation by 5x. Lowers miasma disease chance per mole by 10x. Miasma now does absolutely nothing unless there's enough of it in the air to actually flow into scrubbers so there's no invisible pockets of disease stuck everywhere.

silicons:
## About The Pull Request

Title

## Why It's Good For The Game

No one likes dealing with colorless, odorless gas that makes you sick when it's not even flowing.
Also makes disposing of bodies more important + miasma more of an issue.

## Changelog
tweak: Miasma generation 5x, miasma molar disease chance 1/10, miasma now does nothing unless it can flow.
/:cl:


Poojawa:
since it's a buff to the generation, nerf the value of it in bombs

deathride58 (approved):

PR #8750, opened 2019-06-25
Fixes trauma healing nanites.

Arturlang:
Forgot to fix them in my last nanite fix PR, is ported from TG. Has possible balance implications since it makes brain/trauma healing nanites way better.
:cl:
fix: Fixes brain damage/trauma healing nanites so they actually work while there are only traumas.
/:cl:

### DIFF
--- a/code/modules/research/nanites/nanite_programs/healing.dm
+++ b/code/modules/research/nanites/nanite_programs/healing.dm
@@ -70,14 +70,13 @@
 	rogue_types = list(/datum/nanite_program/brain_decay)
 
 /datum/nanite_program/brain_heal/check_conditions()
-	var/problems = FALSE
 	if(iscarbon(host_mob))
 		var/mob/living/carbon/C = host_mob
 		if(length(C.get_traumas()))
-			problems = TRUE
+			return ..()
 	if(host_mob.getBrainLoss())
-		problems = TRUE
-	return problems ? ..() : FALSE
+		return ..()
+	return FALSE
 
 /datum/nanite_program/brain_heal/active_effect()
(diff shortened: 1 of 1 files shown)


code/modules/research/nanites/nanite_programs/healing.dm
```diff
@@ -70,9 +70,14 @@
 	rogue_types = list(/datum/nanite_program/brain_decay)
 
 /datum/nanite_program/brain_heal/check_conditions()
-	if(!host_mob.getBrainLoss())
-		return FALSE
-	return ..()
+	var/problems = FALSE
```
deathride58: Eugh. What happened to TG's coding standards?
You can axe this var declaration then replace the `problems = TRUE` lines with `return ..()`, and replace the final line in the proc with a simple `return FALSE`
There's no such thing as enough micro-optimizations, especially for things like nanites that get their procs called repeatedly en-masse. Var declarations within procs in byond have a notable performance impact, so you typically want to keep var declarations to a minimum whenever possible inside procs that get called frequently.

deathride58 (changes requested):

code/modules/research/nanites/nanite_programs/healing.dm
Arturlang: I guess oranges truly ruined em good.

Arturlang: I wish i knew how to these optimizations myself honestly

Arturlang:
Wierd how it still says changes needed when i did those changes and the comment is resolved

deathride58 (approved):

PR #8771, opened 2019-06-28
buffs wheelchair movespeeds/changes a floor in the equation to a round

silicons:
Wheelchairs are too slow tbh. Let's make 'em slightly (33% at most) faster.


Thalpy:
Kev teach me how to make it so if you use an gas canister you boost super fast in a straight line until you hit a wall and your wheelchair blows up?

deathride58 (approved):

PR #8795, opened 2019-06-30
Buffs tracking beacon EMPs, and makes EMP damage temporarily disable mech equipment

Sishen1542:
## About The Pull Request
Port of https://github.com/tgstation/tgstation/pull/44452
* Tracking beacons no longer self-destruct when EMPing a mech from the console.
* Tracking beacons now do an EMP_HEAVY instead of an EMP_LIGHT.
* Taking any EMP damage now disables mech equipment/weapons for three seconds (refreshes on further - EMP damage, does not stack). During this time, if the user is in a combat mech, their mouse circle changes from green to red, and an error message from the mech is given to the pilot. At the end of the duration, the mouse swaps back to green and a sound beep is played.

## Why It's Good For The Game
Adds some counter-play against mechs without making them nearly impossible to build. Hostile mechs can, of course, remove the console, but this is an obviously deliberate action, and more consoles can be created by printing the boards from any of the three circuit imprinters. Additionally, the weapons disable time means that fighting one with an ion rifle is a lot less dangerous if you can get a shot in before getting killed.

Non-combat mechs are not automatically beacon'd to allow traitor roboticists a workaround in the form of the Hidden Weapons Bay. They do, however, still temporarily lose use of their equipment when getting ion'd/EMP'd.

## Changelog
:cl: Sishen1542, original by @zxaber 
balance: Utility mechs no longer automatically get beacons.
balance: Tracking beacons no longer delete themselves when EMPing a mech, and instead have a ten-second cooldown in-between EMPs. They also now do heavy EMP damage rather than light.
balance: Mechs that take EMP damage lose the use of their weapons and equipment temporarily. Movement and abilities are not effected.
balance: Mechs taking EMP damage no longer roll for a random malfunction.
/:cl:


code/game/mecha/mecha_control_console.dm
```diff
@@ -21,8 +21,8 @@
 			var/answer = TR.get_mecha_info()
 			if(answer)
 				dat += {"<hr>[answer]<br/>
-						  <a href='?src=[REF(src)];send_message=[REF(TR)]'>Send message</a><br/>
-						  <a href='?src=[REF(src)];get_log=[REF(TR)]'>Show exosuit log</a> | <a style='color: #f00;' href='?src=[REF(src)];shock=[REF(TR)]'>(EMP pulse)</a><br>"}
```
deathride58, line 25: Why remove the log function?

Sishen1542: no clue why tg did it, added it back

zxaber: Honestly I don't remember removing it, so it was probably an oversight on my part and no one said anything because no one else ever used it or even knew that feature existed.

silicons (approved):

PR #8804, opened 2019-07-01
Ports in TG's bastardization of Goon Genetics.

Ghommie:
## About The Pull Request
Porting in the entire genetics rework (minus snail people for now, I'll get on it soon).
Minus some stuff relating little bits not ported teleport refactor and movement flags, the genetics powers spells icon states being actually located in `actions_genetic.dm` instead of `actions_spell.dm` and having the new helpers defines located in `__DEFINES/DNA.dm` rather than a new `__HELPERS/dna.dm`, and some initial tweakings such as:
Having space adaptation be a separate mutation rather than a rename for cold resistance, requiring the latter to be combined with radioactivity to be unlocked.
Or the changes to the two left feet disability to make it work on citadel stamcombat code.

And no, Hulk is still restricted to flies, I only demodularized the override. Maybe we'll get on that salt mine later.

## Why It's Good For The Game
Updating genetics. Making it less of scrambling around with mutagen & dna scanners for the same stupid four or five mutations.
Not as great as the real deal but still way neater.

## Changelog
:cl: Ghommie (original PRs by Time-Green, tralezab, PKPenguin321, ShizCalev, Naksu, coiax and Gousaid67)
add: Goon genetics!
add: More mutations! Fire breath for lizards! Radioactive! Telepathy! Glowy! Strength, though its cosmetic and should be combined with radioactivity instead! Fiery sweat!
add: Adds void magnet mutation by tralezab!
add: Adds chromosomes to mutations. You can now alter your mutations in unexpected ways
add: New genetics power: Insulated - You become shock immune, basically having innate insulated gloves.
add: New genetics power: Shock Touch - Electric punch! You get a wizard-like touch spell that does burn damage, confuses, disarms, and jitters. Has a 10 second cooldown and can only be created by combining Insulated with Radioactive.
add: Added the Antenna power to genetics. It gives you a built-in radio.
add: Added the Mind Reader power to genetics. You can read the minds of people on your screen. This will reveal semi-random bit of their past conversations, their intent, and their true name. You also get the antenna sprite. Can only be unlocked by combining Paranoia and Antenna.
add: Added the Paranoia disability to genetics. You scream randomly, and have a small chance to suffer from hallucinations now and then.
add: Adds the cryokinesis and geladikinesis mutations
fix: Snow walls no longer runtime everything
balance: X-Ray vision can no longer be obtained via a random gene sequence at roundstart.
balance: X-Ray now has 35 instability, up from 25.
add: A new gene, thermal vision, has been added to genetics.
add: You can create X-Ray vision by mixing Thermal Vision and Radioactive together.
add: New system in genetics: Conflicting mutations. Some mutations will not allow others to activate. Example - If you have mutation A, and it conflicts with B, then B will never activate until you get rid of mutation A (and visa versa). The only mutations to currently use this are Dwarfism and Gigantism. Speaking of which...
add: New minor negative mutation: Gigantism. Opposite of dwarfism, makes you larger.
add: New disability: Spastic. You will randomly take a step in a random direction, will randomly change intent, and will randomly try to click on somebody near you (which can hug them, punch them, use whatever is in your hand on them, etc).
add: New disability: Acid flesh. Pockets of acid bubble beneath your skin. They will occasionally pop, coating you with acid. This will melt clothes and deal damage to you.
add: New disability: Spatial Instability. You will randomly teleport, similar to a wizard's blink teleport. This is a very nauseating process, and will inevitably make you very sick if you don't rid yourself of it.
add: New disability: Two Left Feet. Your right foot becomes a second left foot. You have a chance of tripping, the more you move around, the higher the chance.
code: removed /datum/mutation/human/proc/on_move in favor of having the chameleon and two left feet mutations register for a signal
code: Adds tools to make mutations transfer through cloning
/:cl:

Drafted because it's one jabberwock of a port, there shouldn't conflict with the snowflakes (those are UIs at most, mind, mob and species vars at worst) otherwise.


silicons:
bad, believe i've said something regarding it before.
might still do it idk probably not though so keep it in mind.

Ghommie:
What did you even say about it, at which point?
I'm not really up to scroll all the discord's logs to find it.

deathride58:
uggGHGHGH WHY IS THIS BEING CONSIDERED

deathride58:
A much better way to go about this would be to add genuine goon genetics instead of just using TG's bastardization with no changes. TG's implementation has all of the same problems as the old genetics system except the RNG is pushed into character initialization instead of being kept self-contained in genetics.
Also if you're planning on implementing genuine goon genetics or at the very least modifying TG's implementation to be much closer to actual goonetics, do be sure that the unique enzyme stuff is able to be discovered and modified. Bonus points if you make it possible to discover every customization option individually.

Ghommie:
I'm not willing to believe that leaving genetics in the current state, with no planned "original" rework, will be the best solution, still I appreciate the suggestions and criticism.
While I can't promise to end up accurately reverse-engineering Goon Genetics, I'll work on in an attempt to make sure the described flaws are fixed or at the least dampened and improvements applied.

silicons:
i do have plans for not a rework as much of a rebalancing and some features for the current system but no porting it is pretty bad.

Ghommie:
You can start working on a rebalance and tweaking the current system already if you want, I'll adapt as possible, but planning won't do much anyway.
I personally can keep this as a draft until either it becomes stale or we reach a middle point, even though the hostility makes me pensive.

Bran34:
The current system is pretty fucking lame.  Even if you add new stuff to it it'll still be pretty fucking lame.

Currently my geneticist round goes about like this:  wake up, get all powers within ~20-30 minutes, then go goof off in the apothecary/go ERP because now i've got nothing to do because the cloner probably runs itself.  The only power that actually has any real use and isn't just a meme is xray and -maybe- telekinesis.
If you're a tator thanks to the removal of hulk (not that said removal is a bad thing) I feel like you're much better off just getting a spacesuit and bugging science for nvgs rather than even bothering with the genes.  But even then tele just isn't that interesting.  Not suggesting we bring hulk back, but definitely make the system have more depth, and make the powers more interesting and varied.

PR #8837, opened 2019-07-07
Revert "Adds a can_inject check to IV Drips"

silicons:
Reverts Citadel-Station-13/Citadel-Station-13#8244

I'm gonna be honest if someone drags these around to abuse 'em and combat clickdrag like a makeshift 2 tile syringe gun, hats off to 'em. Ninjas especially get hit by these if they need to use IVs and I honestly don't see the problem of having IVs inject past suits.

### DIFF
--- a/code/game/machinery/iv_drip.dm
+++ b/code/game/machinery/iv_drip.dm
@@ -81,12 +81,6 @@
 		to_chat(usr, "<span class='danger'>The drip beeps: Warning, incompatible creature!</span>")
 		return
 
-	var/mob/living/L
-	if(isliving(target))
-		L = target
-		if(!L.can_inject(usr, 1))
-			return
-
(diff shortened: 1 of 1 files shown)


deathride58 (approved):
I'm on the fence about this. @BlackMajor what do you think?

BlackMajor:
They're used infrequently enough that it isn't a detriment whether they can or cannot inject through suits.
Impact of not having the check is that antags can't use them to attempt to one tile range chem someone in a hardsuit (which is one less option for chem memes) and ninjas can't use them to heal.

Impact of having the check is that it's more in line with the idea that no_pierce gear protects you from chem memes... and uhh... I guess medical asking engineers to take off their suits when they need blood.

deathride58:
Fuck it; emergent gameplay

PR #8848, opened 2019-07-08
Trait blacklisting system: Blacklists Mute and Brain Tumor from ALL heads of staff. Blacklists pacifist + paraplegic from HoS. Blacklists Paraplegic/nolegs from CE. Blacklists phanosthing/no-recognize-faces from HoP

silicons:
Mute - Heads of staff not communicating is absolutely not acceptable.
Brain tumor - Heads of staff getting dibilitating brain traumas from not being constantly on mannitol is not okay.
Pacifist - HoS out of all people should not be unable to harm with the lethality level of antagonists. RP a nonviolent character if you want, but this is not budging. 
Paraplegic - Directly interferes with the CE's duties when he's stuck in something that can't EVA.


deathride58:
Wouldn't it be much better to make it so that having a blacklisted quirk results in your quirks being removed outright? Because the way this is implemented means that taking a blacklisted quirk as a head is free points with no downsides at all

deathride58:
Also it seems like this doesnt apply to latejoining

r4d6:
I think RDS should also be blacklisted, a Head of Staff that have constant hallucinations? And those hallucinations can go to see a chasm in the space station, to having Bubblegum run into you, passing by seeing someone pocket an Emag. Probably not the best for someone important.

code/controllers/subsystem/processing/quirks.dm
```diff
@@ -22,11 +23,68 @@ PROCESSING_SUBSYSTEM_DEF(quirks)
...
 	GenerateQuirks(cli)
-	for(var/V in cli.prefs.character_quirks)
+	var/list/quirks = cli.prefs.character_quirks.Copy()
+	var/list/cut
+	if(job && job.blacklisted_quirks)
+		cut = filter_quirks(quirks, job)
+	for(var/V in quirks)
 		user.add_quirk(V, spawn_effects)
+	if(!silent)
+		to_chat(to_chat_target || user, "<span class='boldwarning'>The following quirks were cut during assignment either due to a job restriction or another reason: [english_list(cut)].</span>")
```
deathride58: This shows every single quirk the user has instead of just the ones that triggered the culling

deathride58 (dismissed):

IronEleven:
The quirk reset message just triggered on one of my characters playing a Scientist.  Quirks are: Ananas Affinity, Nymphomania, Heavy Sleeper, Apathetic.

IronEleven:
Same character got the message as a cargo tech, unfortunately Heavy Sleeper and Apathetic are hard to test if they're still active or not.

r4d6:
I always get the message, but it don't say which quirks were removed. and when self-analysing, it still show every quirk

BlackMajor (approved):

PR #8863, opened 2019-07-10
Makes aggressive replication not useless

silicons:
>Literally the only use for it is to do more damage because it doesn't actually regen

>:(

No longer uses nanites to work.
Rate changed to every 50 nanites.
Damages 1 per 3 gained.


Arturlang:
Should probably document the fact that it causes more damage now, unless i misunderstood the code.

deathride58 (approved):

PR #8872, opened 2019-07-11
Janitors are no longer diabetic - nerfs mop stamina usage to be in-line with how much effort mopping takes IRL

deathride58:
Title. The former value of 5 stam per turf mopped was a bit silly, and meant that you ended up running out of stamina while mopping up a single room. This nerfs the base mop to take only 2 stamina, while the advanced mop now takes 1 stamina (both values not taking into account the stamina buffer)

## Changelog
:cl:
balance: Normal mops now only use 2 stamina to mop a tile, nerfed from their previous value of 5 stamina per tile mopped.
balance: Advanced mops now only use 1 stamina to mop turfs, from their former value of 2 stam.
/:cl:

### DIFF
--- a/code/game/objects/items/mop.dm
+++ b/code/game/objects/items/mop.dm
@@ -15,7 +15,7 @@
 	var/mopping = 0
 	var/mopcount = 0
 	var/mopcap = 5
-	var/stamusage = 5
+	var/stamusage = 2
 	force_string = "robust... against germs"
 	var/insertable = TRUE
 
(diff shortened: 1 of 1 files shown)


lolman360:
should change 

> balance: Normal mops now only use 2 stamina to mop a tile, nerfed from their previous value of 5 stamina per tile mopped.

to

> balance: Normal mops now only use 2 stamina to mop a tile, buffed from their previous value of 5 stamina per tile mopped.

silicons (approved):

PR #8875, opened 2019-07-12
[s] Porting recent travis improvements.

Ghommie:
## About The Pull Request
Porting in tgstation PRs #44845 and #44966 by SpaceManiac.
Also the removal of an unnedeed import, by Spookydonut (#44594).
I'm pretty confident I may fouled up the copypasta, but shouldn't have missed anything otherwise.

## Why It's Good For The Game
Citing the first two PRs:

 -    Split the different tasks into their own .sh files rather than doing many things per .sh file
 -   Use .travis.yml jobs list to select which tasks are run rather than setting environment variables which the shell scripts then check
 -   Moves the grep lints from running in both "Maps" and "Tests" jobs, and runs them in "Lints" job (formerly "Tools") instead
 -   Makes the check_filedirs.sh lint only run in the "Lints" job rather than in all three jobs.
 -   Removes the check that dependencies.sh matches Dockerfile since the former now pulls from the latter
 -   Uncache node_modules - npm ci is expected to trash this every time and warns when it already exists
 -   Uncache and remove the unnecessary fake MariaDB folder used when building BSQL, give it the real path instead
 -   Attempt to cache the BYOND-provided libmariadb
 -   Download rust_g binary rather than building it every time. Shaves 3-4m off the unit test job time. BSQL is only ~10s so I'm leaving it for now.
 -   Improve nonsense structure of json_verifier script.
 -   Give tgui build its own .sh file instead of being a bunch of commands in a subshell.
 -   Reduce verbosity of 'not mapmerged' check to keep it from blowing the Travis log.
 -   Add DreamChecker to the linter job. Binary is downloaded from GitHub releases.

## Changelog
No, all credits to the original authors as always though.


Ghommie:
_"Travis destroyed, Mission Accomplished"_

PR #8879, opened 2019-07-12
Makes gamemodes less dependant on latejoiners (I think?)

Coolgat3:
## About The Pull Request
This lowers payer requirements for certain gamemodes by 5 so that maybe they happen more often without being so dependant on whether someone readies up or not. I remember this was done once and reverted due to complaints about lowpop cult but we are literally almost never "that lowpop" anymore and all we get is still fucking traitorling, yes I'm very angry ree.

## Why It's Good For The Game
Because variety is good.

## Changelog
:cl: Coolgat3
tweak: changed player requirement numbers for cult, clockcult, nukeops and revs.
/:cl:

### DIFF
--- a/code/game/gamemodes/clock_cult/clock_cult.dm
+++ b/code/game/gamemodes/clock_cult/clock_cult.dm
@@ -131,7 +131,7 @@ Credit where due:
 	config_tag = "clockwork_cult"
 	antag_flag = ROLE_SERVANT_OF_RATVAR
 	false_report_weight = 10
-	required_players = 35
+	required_players = 30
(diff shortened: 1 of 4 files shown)


Coolgat3:
For reference Take a look at this. http://prntscr.com/oe4c8p The blue colors are bussed rounds, Extended I marked just to sort of mark it as not counting because it's not secret, while the red is obviously traitor/ling/traitorling.IAA(which is literally just traitor but more traitors) which are OVERWHELMING in amounts compared to the other gamemodes.

PR #8880, opened 2019-07-12
Makes Air Alarms Green

CameronWoof:
## About The Pull Request

Air alarms are now green when the atmosphere is ideal, rather than blue. They also don't blink when everything is fine.

## Why It's Good For The Game

This brings air alarms in line with every other status indicator light. Blue means something is wrong. See:
- blue alert
- blue APCs (charging or hacked)
- bluesceened consoles

While green means everything is fine. See:
- green alert
- fully charged APCs
- fire alarms in the absence of fire
- oxygen tanks
- tons and tons of individual items (operational freezer, powered-on grinder, buttons, SMES units, unlocked lockers, etc. etc. etc.)

In addition, things that are fine (e.g. don't need attention, e.g. an air alarm detecting an ideal atmosphere) don't need to blink constantly in a distracting manner to keep you updated that everything is as expected. 

## Changelog
:cl:
tweak: air alarms are green now instead of blue when the atmosphere is ideal
/:cl:


Poojawa:
this was already discussed and even voted down on the poll

We don't have to be like every other station. 

Also these aren't even animated you lazy dog

CameronWoof:
> this was already discussed and even voted down on the poll
> 
> We don't have to be like every other station.
> 
> Also these aren't even animated you lazy dog

https://cdn.discordapp.com/attachments/419883780275175426/599847435946295315/unknown.png You have a funny definition of "voted down."

And no, they're not animated because blinking lights are for important things. Air alarms that aren't warning you about danger should not blink or have an *exclamation mark* drawing your eyes to them. They're distracting and irritating.

Poojawa:
I remember this same argument from some other person who wanted to change it even if they never played and never considered how it looked in dark and bright lighting. 

Green is pretty fucking awful. I changed it to blue on a popular vote because the green looked awful

Using other servers as an example of how things should be done isn't wanted here.

CameronWoof:
This isn't about what "other servers" do. It's about what our server does. On *our server*, having a **blue flashing light** indicate that nothing is wrong makes zero sense. It's inconsistent with every other visual notification we have.

Green means good. Blue means something is wrong. This is consistent across 99% of the game. It should be consistent here.

Unless you want to go in and swap blue and green alert, APC color states, indicator lights on every machine, etc. etc. etc.

silicons (approved):

PR #8887, opened 2019-07-13
Gang victory news report and last words for discord bot

Poojawa:
## About The Pull Request

Allows the bot to report gang victories for round end messages. It also provides the same last words in the end round messaging.

## Why It's Good For The Game

information like this is pretty funny oft times, why not share it? I don't know if the format will work right for last words part, but it was the best way to implement it

## Changelog
:cl: Poojawa
add: NT Newscasters have had repeated reports of gang activity and are now looking into it.
add: NT Psykers keep mumbling about last words of someone who died. Somehow they even have a newsletter for this...
fix: Gangster greeting messages are a batch message rather than 5 laggy to_chats
/:cl:


Poojawa:
>secret_sauce

whomst've the don't've fucked

deathride58 (approved):
How long until someone makes the bot do an @ everyone

Poojawa:
> How long until someone makes the bot do an @ everyone

I already have with the wheel of salt. but the bot's @ privileges don't seem like it's working, please fix posthaste

PR #8889, opened 2019-07-14
minor body parts refactor, clean up, feature fix.

Ghommie:
## About The Pull Request
Basically there is an acient bug preventing limbs from storing the original owner's identity and thus getting overriden (should they be organic, because robo limbs are handled slighty differently) with new values from the current owner everytime. I'm set to fix it in this PR, along with some slight refactor to convert the body part status var to bitflags for further eventual future features or refactoring, and removing some redunant health updating from reagents effects, since health will be updated on the same tick afterward anyway unless I'm mistaken.

Draft since I'll have to repeatedly test locally due to how frail the body parts code may be.

## Why It's Good For The Game
Bugfixing, code clean up.

## Changelog
:cl:
fix: Fixed an acient bug with organic limbs not retaining the original icons after being attached to a mob of different species (e.g. attaching a mush arm to a skellie will result in said arm magically looking skeletonous).
refactor: Slighty refactored bodypart statuses into bit flags (currently that's just two of them anyway, but ok)
code: removed more redunant health updates from on tick reagent effects.
/:cl:


nik707:
ongoing?

Ghommie:
Sorry for the hiatus, but yea.

PR #8891, opened 2019-07-14
Non-stealthed nanites now give the user a special examine message

silicons:
Makes it easier to tell there's nanites until I actually overhaul nanites.

### DIFF
--- a/code/datums/components/nanites.dm
+++ b/code/datums/components/nanites.dm
@@ -57,6 +57,7 @@
 		RegisterSignal(parent, COMSIG_MOVABLE_HEAR, .proc/on_hear)
 		RegisterSignal(parent, COMSIG_SPECIES_GAIN, .proc/check_viable_biotype)
 		RegisterSignal(parent, COMSIG_NANITE_SIGNAL, .proc/receive_signal)
+		RegisterSignal(parent, COMSIG_PARENT_EXAMPLE, .proc/on_examine)
 
 /datum/component/nanites/UnregisterFromParent()
 	UnregisterSignal(parent, list(COMSIG_HAS_NANITES,
--- a/code/datums/components/nanites.dm
+++ b/code/datums/components/nanites.dm
@@ -12,6 +12,7 @@
 	var/max_programs = NANITE_PROGRAM_LIMIT
 
 	var/stealth = FALSE //if TRUE, does not appear on HUDs and health scans, and does not display the program list on nanite scans
+	var/visible_examine = TRUE
 
 /datum/component/nanites/Initialize(amount = 100, cloud = 0)
 	if(!isliving(parent) && !istype(parent, /datum/nanite_cloud_backup))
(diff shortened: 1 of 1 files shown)


code/datums/components/nanites.dm
```diff
@@ -56,6 +57,7 @@
 		RegisterSignal(parent, COMSIG_MOVABLE_HEAR, .proc/on_hear)
 		RegisterSignal(parent, COMSIG_SPECIES_GAIN, .proc/check_viable_biotype)
 		RegisterSignal(parent, COMSIG_NANITE_SIGNAL, .proc/receive_signal)
+		RegisterSignal(parent, COMSIG_PARENT_EXAMPLE, .proc/on_examine)
 
 /datum/component/nanites/UnregisterFromParent()
 	UnregisterSignal(parent, list(COMSIG_HAS_NANITES,
```
Ghommie, line 63: Make sure to add the new signal to this list too so it unregisters correctly.

code/datums/components/nanites.dm
```diff
@@ -246,6 +248,10 @@
 /datum/component/nanites/proc/get_programs(datum/source, list/nanite_programs)
 	nanite_programs |= programs
 
+/datum/component/nanites/proc/on_examine(datum/source, mob/user)
```
Ghommie, line 251: Set the user var to be `mob/living`, see why below.

code/datums/components/nanites.dm
```diff
@@ -246,6 +248,10 @@
...
+	if(!stealth && visible_examine)
+		to_chat(user, "<span class='notice'>Their eyes shine with an electric fluid.</span>")
```
Ghommie, line 253: This line would be outputted regardless of the parent eyes being covered and/or missing.
Make sure to define the parent as living, then add in an early return if `L.is_eyes_covered() || (iscarbon(L) && !L.getorganslot(ORGAN_SLOT_EYES))` is TRUE.

code/datums/components/nanites.dm
```diff
@@ -56,6 +57,7 @@
 		RegisterSignal(parent, COMSIG_MOVABLE_HEAR, .proc/on_hear)
 		RegisterSignal(parent, COMSIG_SPECIES_GAIN, .proc/check_viable_biotype)
 		RegisterSignal(parent, COMSIG_NANITE_SIGNAL, .proc/receive_signal)
+		RegisterSignal(parent, COMSIG_PARENT_EXAMPLE, .proc/on_examine)
```
Ghommie, line 60: Oh, one last thing. carbon, human and silicon examine functions won't check the parent versions and thus won't send the signal, making this totally useless.
Find a solution out of this problem, though even just adding SEND_SIGNAL() manually may do it.

code/datums/components/nanites.dm
Ghommie: Oh, another thing: since carbon & co examine text is batched, just return the text without `to_chat()` once you add `msg += SEND_SIGNAL(src, COMSIG_PARENT_EXAMINE, user)` to said procs.

silicons:
haha what the fuck why do carbon/human/silicons not send the signals wtf

yeah thanks i need to fix the rest but it showing regardless of eye-cover is actually intended.

PR #8881, opened 2019-07-13
Fixing a reagent ID typo.

Ghommie:
## About The Pull Request
What's said on the tin. The reagent id text strings removal in favour of typepaths will have to be ported one day.

## Why It's Good For The Game
Bugfixin.

## Changelog
:cl:
fix: Fixing baklava pies a bit.
/:cl:

PR #8893, opened 2019-07-14
Fixes Travis randomly failing, baklavaslice or whatever pies are now makeable (id 'vitamins' --> 'vitamin')

silicons:
When are we moving these damn things to a define?


Ghommie:
duplicate of #8881.
Also no, these things should be removed in favor of typepaths like /tg/station did.

Poojawa:
typepaths would be A+

PR #8894, opened 2019-07-14
T G F U S I O N

silicons:
Haha fuck this'll be fun

Why: /tg/fusion is more fun than the current "just compress gas in a canister".


nik707:
hah conflict

deathride58 (approved):
every day, we stray further and further from both heaven and hell

PR #8899, opened 2019-07-15
Nerf Strained Muscles on lings

MrJWhit:
Moves strained muscles from lings being able to fly around the station, to not being slowed due to various objects.

## About The Pull Request

Swaps out TRAIT_GOTTAGOFAST for TRAIT_IGNORESLOWDOWN in the strained muscles, as well as increasing the base cost from literally free, to 5 chems.

## Why It's Good For The Game

At the moment a ling will get hit with a taser, pop adrenals, and pop strained muscles to get up, and then promptly run as fast as a laser beam for about two minutes straight. Easily outrunning sec, and making it impossible for them to be caught. The speed mostly comes from the fact that strained muscles launches people as fast as nuka cola with the "downside" of increased stam loss, but paired with adrenals means the stamina will never run out.
This will make it so that it's still useful for a ling, but more of a long term use.

## Changelog
:cl:
tweak: tweaked a few things
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/strained_muscles.dm
+++ b/code/modules/antagonists/changeling/powers/strained_muscles.dm
@@ -3,9 +3,9 @@
 
 /obj/effect/proc_holder/changeling/strained_muscles
 	name = "Strained Muscles"
-	desc = "We evolve the ability to reduce the acid buildup in our muscles, allowing us to move much faster."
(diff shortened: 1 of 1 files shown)


MrJWhit:
There was another PR that did this that got merged, closing it now

PR #8905, opened 2019-07-16
Stopgap tweak PR until something more is done - cult conversion is now more clear on what happens if you don't, and the delay is now 30 seconds.

silicons:
More time to read, etc etc.

### DIFF
--- a/code/modules/antagonists/cult/runes.dm
+++ b/code/modules/antagonists/cult/runes.dm
@@ -246,10 +246,10 @@ structure_check() searches for nearby cultist structures required for the invoca
 		return 0
 	to_chat(convertee, "<span class='cult italic'><b>Your blood pulses. Your head throbs. The world goes red. All at once you are aware of a horrible, horrible, truth. The veil of reality has been ripped away \
 	and something evil takes root.</b></span>")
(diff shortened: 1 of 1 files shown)


Ghommie:
Very tactless PR, it will make conversion way more tedious without squeezing out the verbosity of the conversion message and prompt. 
Are you really going slow down gameplay in favour of some spoopy fluff text (and perhaps other reasons you did not clearly explain us)?

UntoldTactics:
> Very tactless PR, it will make conversion way more tedious without squeezing out the verbosity of the conversion message and prompt.
> Are you really going slow down gameplay in favour of some spoopy fluff text (and perhaps other reasons you did not clearly explain us)?

At least 3/5 or 4/5 offered people become soulstone shards. This is very poor. I'd rather we just remove the option to be sharded, and make it so that only mindshielded people are sharded.

To answer more directly to your quote, most people don't have enough time to read the full prompt and realize the consequences

To be short: playerbase shit we need to help them 

Also yeah this is just temporary

Ghommie:
Just cut down the obtuse fluff text from the message and make it more objective than subjectively verbose.
The sharded and salty unantagging playerbase is another topic, I'm merely against this wolf-in-sheep-clothing PR, and probably will commit what I'm suggesting later as I doubt I'll be striking a point by talking.

PR #8918, opened 2019-07-17
Halves telebaton delay

silicons:
Yeah yeah it's too weak compared to /tg/ telebatons or something, this makes them better for stunning I guess? Still higher than baton delay.


silicons:
yeah this isn't the way to go entirely gimme a bit

PR #8923, opened 2019-07-18
Removes damage from *dab

silicons:
Why: Meme feature. If I've had to remove it from working while handcuffed/restrained to not have people suicide on arrest with it, and now more issues from forced emotes happening with it, it really shouldn't exist when every other emote doesn't do it. It's funny, but it's overlived its stay.


MimicFaux:
Good. I've had to catch the same traitor three times because of a secret autocloner and they just laugh and dab* out of cuffs.

BlackMajor (approved):
End of an era.

Sishen1542:
REVERT THIS
YOU CAN'T DAB WHILE RESTRAINED HE'S LYING REEEEEEEEEEE

PR #8926, opened 2019-07-19
Secbats

Thalpy:
## About The Pull Request

Recently the community has cried out for the addition of secbats. As a result, I made this for them. I hope they enjoy it!

## Why It's Good For The Game

A useful tool in any security officer's kit, the secbat will help them bring justice to the station. Should an officer desire one, they simply need to dispense one from the SecTech ventor.

## Changelog
:cl: Fermis
add: Added the secbat, a box to hold it and the ability to dispense it from the SecTech vendor.
/:cl:


DeltaFire15:
Honk.

silicons:
:+1:

code/modules/mob/living/simple_animal/friendly/bat.dm
```diff
@@ -0,0 +1,42 @@
+/mob/living/simple_animal/bat
```
Ghommie: Please avoid doing such copypasta and instead tell us why you aren't making the secbat a subtype of `/mob/living/simple_animal/hostile/retaliate/bat` so we can fix it.

code/game/objects/items/storage/boxes.dm
```diff
@@ -1196,3 +1196,14 @@
 	new /obj/item/reagent_containers/food/snacks/grown/chili(src)
 	new /obj/item/reagent_containers/food/drinks/coffee/type2(src)
 	new /obj/item/tank/internals/emergency_oxygen(src)
+
+/obj/item/storage/box/secbat
+	name = "Secbat box"
+	desc = "Contained inside is a secbat for use with law enforcement."
+
+/obj/item/storage/box/attack_self(mob/user)
```
Ghommie: I see we are pulling rabbits out of hats here. Wrong pathing.

code/modules/mob/living/simple_animal/friendly/bat.dm
Thalpy: So it's not hostile.

code/game/objects/items/storage/boxes.dm
Thalpy: I'm fixing another issue that might come up with keeping it as a storage item too, so I'll incorporate this.

code/modules/mob/living/simple_animal/friendly/bat.dm
Ghommie: The `retaliate` subtype (which the original bats are part of) modifies said ""AI"" to be, as the name hints, more about retaliation than uncalled assault.
They'll start attacking after receiving damage (from any source), rather than simply upon spotting a mob they share no faction with.

Regardless, if you want to make sure these sec bats stay friendly and pick up no fight, the easiest and most boring way would be
```
/mob/living/simple_animal/hostile/retaliate/bat/secbat/Retaliate()
	return
```
You can try fancier solutions like petting or a prob() roll per tick to calm them, but that'll require half a dozen more lines or so.

Thalpy: Ah, my concern was by creating a mob under the hostile category that wasn't hostile might cause problems.

Thalpy: I'll move them across, but I'm still gonna make them gold core spawnable!

Ghommie: The gold core spawnability is no issue.

code/game/objects/items/storage/boxes.dm
```diff
@@ -1196,3 +1196,20 @@
...
+	icon = 'icons/obj/storage.dmi'
+	icon_state = "box"
+	item_state = "syringe_kit"
+	lefthand_file = 'icons/mob/inhands/equipment/medical_lefthand.dmi'
+	righthand_file = 'icons/mob/inhands/equipment/medical_righthand.dmi'
+
+/obj/item/secbat/attack_self(mob/user)
+	new /mob/living/simple_animal/hostile/retaliate/bat/secbat(user.loc)
+	to_chat(user, "<span class='notice'>You open the box, releasing the secbat!</span>")
+	var/obj/item/I = new /obj/item/stack/sheet/cardboard
```
silicons: var/obj/item/stack/sheet/cardboard/I = new(user.drop_location())

silicons (changes requested):
it'd be nice to make these orderable through cargo for like 500 points or something but that's out of scope, just fix that one line to prevent edge cases please :+1:

code/game/objects/items/storage/boxes.dm
Thalpy: I hope it works.

silicons (approved):

PR #8938, opened 2019-07-21
Adds pyro/cryo/advanced grenade casings to science, as well as test range firing pins, floralrays, and stunshells

silicons:
Grenades: To me they match the theme of science having situationally useful stuff that's "experimental", especially advanced release ones. 
Test range firing pins : Science, test weapons, etc etc. Perfectly aware they can be emagged yes.
Floral rays : Why not, they're not exactly a powerful weapon unless you're a podperson
Stun shells: science already gets techshells, it's another high-tech-ish thing.


nik707 (approved):
>kevinz buffing science

PR #8941, opened 2019-07-21
Reverts "Autocall now uses realtime rather than world.time"

silicons:
I made the PR initially because lag was so bad rounds would last up to ~4-5 hours, nowadays honestly an extra 30 minutes won't kill anyone granted that everything in the game uses world.time (except for some niche things like some combat things I think? Movedelay definitely does and all gameplay/ramping stuff does.)

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -203,7 +203,7 @@
 				state = STATE_CANCELSHUTTLE
 		if("cancelshuttle2")
 			if(authenticated)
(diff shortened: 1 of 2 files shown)


Ghommie:
Upopular opinion, but I like autocall using world.time since it makes no-early-evac rounds' length consistent (at least from a real world perspective), and 30% or so time dilatation shifts can still happen on high-pop rounds.
I might look into making an extend round vote at this point, though I do agree that the average server lag has indeed diminished since that PR passed.

nik707 (approved):

silicons:
*sigh alright let's do it another way

PR #8927, opened 2019-07-19
Tones down some items attack stamina costs. kills hypereutactic blades copypasta.

Ghommie:
## About The Pull Request
Buffing some weapons to not drain your stamina as fast as they currently do.
The buff on eswords stam cost last year by deathride was ok, yet I am giving it an edge compared to the dual esword stamina cost, as sprinting and one vs many situations can still awfully get in the way and drop the user on soft-stamcrit.
Also cut down copypasta on hypereutactic blades by making them a subtype of dual sabers (without the jedi spin), which also means they can be rainbow mode'd too.

## Why It's Good For The Game
Rebalance, and code improvement.

## Changelog
:cl:
balance: toned down the stamina costs of some of the bulkier weapons.
code: repathed hypereutactic blades to be a subtype of dual sabers. Way less copypasta.
/:cl:

PR #8957, opened 2019-07-24
puts the slowdown back on hypereutactic because i buffed it without knowing it reflected energy projectiles

silicons:
Haha holy shit this thing is OP

Also removes the stamcost from toy versions so people can beat each other more.

### DIFF
--- a/modular_citadel/code/game/objects/items/melee/eutactic_blades.dm
+++ b/modular_citadel/code/game/objects/items/melee/eutactic_blades.dm
@@ -153,6 +153,7 @@
 	w_class = WEIGHT_CLASS_SMALL
 	attack_verb = list("poked", "jabbed", "hit")
 	light_color = "#37FFF7"
+	total_mass = 1.5		//It's a toy.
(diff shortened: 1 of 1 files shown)


modular_citadel/code/game/objects/items/melee/eutactic_blades.dm
```diff
@@ -269,6 +270,7 @@
 	unwieldsound = 'sound/weapons/neboff.ogg'
 	hitsound = "swing_hit"
 	armour_penetration = 60
+	slowdown = 1
```
Ghommie, line 273: This doesn't take into consideration wheter the swordis on or off anymore.
The leftover var change on line 310/312 is proof it was supposed to have a different slowdown on wield.

Ghommie (commented):
#8927 is already dealing with hypereutactic copypasta'd code, it'd be better if I just implemented the values of this PR into mine in a more flexible way.

nik707 (approved):
p much only does fuckin toy stuff now but ok

PR #8970, opened 2019-07-26
ports jumpskirts and jumpskirt prefs from teegeestation

Sishen1542:
## About The Pull Request
port of https://github.com/tgstation/tgstation/pull/44883 and https://github.com/tgstation/tgstation/pull/44776 and https://github.com/tgstation/tgstation/pull/45244

## Why It's Good For The Game
skirts are cute

## Changelog
:cl: original by @randolfthemeh and @twaticus, port by sishen1542
add: jumpskirts
add: more jumpskirts
add: jumpskirt/suit prefs
/:cl:

PR #8977, opened 2019-07-26
[SPEEDMERGE BEFORE THE FURRIES REVOLT] makes jumpskirts work for digitigrade legs

Sishen1542:
fix for digitigrade legs for jumpskirts in https://github.com/Citadel-Station-13/Citadel-Station-13/pull/8970


Ghommie:
There is a `spacepol_l` icon state being added in to the file, which is not of a jumpskirt but an unused space police uniform. 
Is it intentional? Cause it doesn't seem to be shaped for digitigrades.

Sishen1542:
> There is a `spacepol_l` icon state being added in to the file, which is not of a jumpskirt but an unused space police uniform.
> Is it intentional? Cause it doesn't seem to be shaped for digitigrades.

whoops

silicons (approved):

PR #8992, opened 2019-07-28
[READY] Hardsuit helmets can now be used as internals masks. Rejoice, CE mains.

jakeramsay007:
Also reworks internals code a little to be much more friendly. Any head or mask item with the flag should now function as an internals mask if it has the ALLOWSINTERNALS flag.

:cl: Chayse
add: Assorted space-worthy helmets can now act as masks for internals.
refactor: Internals code can now check any item with the ALLOWSINTERNALS flag through the GET_INTERNAL_SLOTS define. For now this only checks head and mask slots, since those are the most realistically speaking usable ones.
/:cl:


YakumoChen:
Does the mouth tube implant still work?
Does this mean you can smoke a cigar while wearing a helmet?

jakeramsay007:
Yes, and technically yes, unintended feature but hilarious all the same.

YakumoChen:
image

Hell, it's about time.

code/modules/mob/living/carbon/life.dm
```diff
@@ -304,11 +304,21 @@
 	return
 
 /mob/living/carbon/proc/get_breath_from_internal(volume_needed)
+	var/obj/item/clothing/check
+	var/internals = FALSE
+	var/internalslots = list(
```
deathride58: Make this a define, as per my comment in inventory.dm

code/modules/mob/living/carbon/life.dm
```diff
@@ -304,11 +304,21 @@
...
+
+	for(check in internalslots)
+		if(CHECK_BITFIELD(check.clothing_flags, ALLOWINTERNALS))
+			internals = TRUE
 	if(internal)
 		if(internal.loc != src)
 			internal = null
 			update_internals_hud_icon(0)
-		else if ((!wear_mask || !(wear_mask.clothing_flags & MASKINTERNALS)) && !getorganslot(ORGAN_SLOT_BREATHING_TUBE))
+		else if (internals == FALSE && !getorganslot(ORGAN_SLOT_BREATHING_TUBE))
```
deathride58: There's no need to check explicitly for FALSE
```suggestion
		else if (!internals && !getorganslot(ORGAN_SLOT_BREATHING_TUBE))
```

code/modules/mob/living/carbon/inventory.dm
```diff
@@ -140,3 +140,9 @@
 /mob/living/carbon/proc/get_holding_bodypart_of_item(obj/item/I)
 	var/index = get_held_index_of_item(I)
 	return index && hand_bodyparts[index]
+
+/mob/living/carbon/proc/get_internal_slots()
```
deathride58: Why is this a proc? This has no reason at all to be a proc when the contents are guaranteed to always be the same. Proc calls have a fairly significant overhead, far greater than the overhead of declaring a var.
Turn this into a define.
```
#define GET_INTERNAL_SLOTS(C) list(C.head, C.wear_mask)
```

code/modules/mob/living/carbon/inventory.dm
```diff
@@ -140,3 +140,9 @@
 /mob/living/carbon/proc/get_holding_bodypart_of_item(obj/item/I)
 	var/index = get_held_index_of_item(I)
 	return index && hand_bodyparts[index]
+
+/mob/living/carbon/proc/get_internal_slots()
+	return list(
+		head,
+		wear_mask,
```
deathride58: This here will make most third-party byond dev tools cry. Here's an easy fix. Also there's no need for this to be multi-line, I'm not sure how this even compiles when you aren't cancelling out the linebreaks
```suggestion
		wear_mask
```

deathride58 (changes requested):

deathride58 (approved):

lolman360:
now i can wear balaclava under helmet

>Le epic CEsuit traitor
>Unknown
>Pull helmet off
>Balaclava, still unknown
>Pull balaclava off, acid'd face
>still unknown

silicons (approved):

PR #8999, opened 2019-07-29
Xenomorph larvae no longer gib and inexplicably delete your brain when hatching

BurgerLUA:
## About The Pull Request

Xenomorph larvae hatching no longer gibs you, but instead just kills you, disembowels you, and deals a high amount of brute damage to your chest.

## Why It's Good For The Game

I discovered in a previous round that FOR SOME REASON a larva hatching out of you gibs you, and on top of that, inexplicably deletes your brain. It's pretty stupid and I think only exists as a poor /tg/ attempt to balance things. It likely was intended to get people to play xeno after they die, but this was when /tg/ had like 2 players playing.

No other antagonist has this sort of thing where if you "lose" you get permadeath. When you consider the fact that xenos have instant stuns with neurotoxin and facehugger throws, it's a bit of a huge meme to have especially on a server with a playerbase that does not validhunt or really give a shit about anything that occurs in the round.

Permadeath should never be a thing in ss13, but not because we're a hugbox, but because permadeath encourages people to rush to the cloners any time there is an antagonist who can permanently alter your character's life.

## Changelog
:cl: BurgerBB
balance: Chestbursters no longer give and remove your brain. They just disembowel and kill you now.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/alien/special/alien_embryo.dm
+++ b/code/modules/mob/living/carbon/alien/special/alien_embryo.dm
@@ -63,7 +63,7 @@
 
 
 
-/obj/item/organ/body_egg/alien_embryo/proc/AttemptGrow(gib_on_success=TRUE)
+/obj/item/organ/body_egg/alien_embryo/proc/AttemptGrow(var/kill_on_sucess=TRUE)
 	if(!owner || bursting)
 		return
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/alien/special/alien_embryo.dm
```diff
@@ -63,7 +63,7 @@
 
 
 
-/obj/item/organ/body_egg/alien_embryo/proc/AttemptGrow(gib_on_success=TRUE)
+/obj/item/organ/body_egg/alien_embryo/proc/AttemptGrow(var/kill_on_sucess=TRUE)
```
Ghommie, line 66:
```suggestion
/obj/item/organ/body_egg/alien_embryo/proc/AttemptGrow(kill_on_sucess = TRUE)
```

BurgerLUA: what does this do?

Ghommie: the `var/` prefix is already implied and doesn't need to be stated.

As for the scope of this arg, I guess it has something to do with the edge cases of surgically removing a larva seconds before it'd violently burst out its victim themselves.

BurgerLUA: Putting var before is not invalid and merely comes down to personal preference.

Ghommie: it's redunant, but merely a tiny waste of space after all. Go on.

silicons (approved):

PR #9030, opened 2019-08-01
Fixes a visual glitch with beakers containing water attached to IV stands

CameronWoof:
## About The Pull Request

Because of the way IV stands color their reagent overlays, transparent reagents (like water) cause a fulltile visual glitch, as seen in https://github.com/tgstation/tgstation/issues/45493.

This PR fixes that by making the reagent sprites in iv_drip.dmi white instead of gray and removing the "white-washing" code from iv_drip.dm that seems to cause the issue.

## Why It's Good For The Game

It's a bug, but, like, no one on this codebase will ever have a reason to put water in a beaker and attach it to and IV stand, so who knows?

## Changelog
:cl:
fix: Attaching a beaker that contains water to an IV stand no longer causes a visual glitch
/:cl:


Poojawa:
I do hope these are somewhat translucent

silicons (approved):

PR #9036, opened 2019-08-02
fixes combat gloves plus having no icon, allows for shoes to have their item_state set to use a differently named icon on mobs

Linzolle:
## About The Pull Request
This originally started as an endeavor to fix the combat boots having no icon when worn, then I noticed combat gloves+ also had that problem, and then I figured out what the root problem was. So now shoes can have their item state defined and it'll actually do something.

## Why It's Good For The Game
Bruh moment

## Changelog
:cl:
add: shoes can have a different icon used for their item and mob icons
fix: combat gloves plus having no mob icon
/:cl:

### DIFF
--- a/code/datums/martial/krav_maga.dm
+++ b/code/datums/martial/krav_maga.dm
@@ -196,7 +196,7 @@
 	name = "combat gloves plus"
 	desc = "These tactical gloves are fireproof and shock resistant, and using nanochip technology it teaches you the powers of krav maga."
 	icon_state = "combat"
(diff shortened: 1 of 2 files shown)


Linzolle:
after further fucking around the change for item state might actually mess certain things up if they rely on the icon state changing to reflect the mob sprite but already have an item state defined... Though the only thing so far I've found an issue with is admin spawn only so I'm not *too* worried about it yet. Things like magboots still work properly.

silicons (approved):

PR #9048, opened 2019-08-04
Masochists and Maint Wires

Kyutness:
## About The Pull Request
This makes it so burn damage AND brute damage plays with the Masochist trait. As well, after some suffering through figuring out how shocking works, makes even being electrocuted (so long as it actually does damage, sorry, no nutting from tasers) work with it.

## Why It's Good For The Game
Because it's fun! Plus it makes it obvious if a masochist attempts and fails to hack into engineering or something because they'll leave DNA right in front of the door. Which is also fun!

## Changelog
:cl:
tweak: gives masochist the ability to gain arousal from burn, and more importantly shock damage.
/:cl:


silicons:
sorry, as per my "please do not expand the arousal system even more" i don't think this is going to be a good thing considering it'll be spammed to death, even with cameron's nerfs to the fluid system.

PR #9050, opened 2019-08-05
Fixes combatmode pushing not knocking people down against walls.

Ghommie:
## About The Pull Request
I forgot to add a target turf's density check (no CanPass() as it calls itself onto every contents' key). Also replaced object density checks with CanPass() for the special cases of objects that can or can not be walked through regardless of density.

## Why It's Good For The Game
Bugfixing.

## Changelog
:cl:
fix: fixes a few bad touchs on combat mode pushing.
/:cl:


deathride58 (approved):
@kevinz000 @BlackMajor this is eligible for speedmerge since this is a major bug

PR #9053, opened 2019-08-05
Re-removes Fun

Poojawa:
## Changelog
:cl: Poojawa
fix: Cake Hat returned to normal Citadel values
/:cl:

### DIFF
--- a/code/modules/clothing/head/misc_special.dm
+++ b/code/modules/clothing/head/misc_special.dm
@@ -48,7 +48,7 @@
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0,"energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
 	brightness_on = 2 //luminosity when on
 	flags_cover = HEADCOVERSEYES
-	heat = 1000
+	heat = 100
 
 /obj/item/clothing/head/hardhat/cakehat/process()
 	var/turf/location = src.loc
@@ -58,12 +58,12 @@
 			location = M.loc
 
 	if(isturf(location))
-		location.hotspot_expose(700, 1)
+		location.hotspot_expose(100, 1)
 
 /obj/item/clothing/head/hardhat/cakehat/turn_on()
 	..()
-	force = 15
-	throwforce = 15
+	force = 1
+	throwforce = 1
 	damtype = BURN
 	hitsound = 'sound/items/welder.ogg'
 	START_PROCESSING(SSobj, src)


Ghommie:
Just do what tgstation/tgstation#45001 did if you must, instead of hugboxing balance around a fancy equivalent of a circular saw with burn damage.

Poojawa:
That... just makes them able to be mass produced?

How is that a better change.

Ghommie:
The real peeve is how easy it's to grab one to delimb some folk with very early in the round, but that's something many other easily achievable or not weapons can do already, yet people use this as scapegoat for their "ided".

On the other hand, it always felt sad having to sacrifice a cake hat for the birthday cake recipe and not the other way around.

And I don't even really care if this is a fix to some mediocre nerf that got lost on the 5th rebase or something, the cakehat is neat and the instant K.O. non-crit/dead disembowelment removal has been ported already.

Poojawa:
That's a whole lot of nonsense that doesn't explain why your suggestion that allows for mass-production of infinite welder hats is a good idea

Ghommie:
> 
> 
> That's a whole lot of nonsense that doesn't explain why your suggestion that allows for mass-production of infinite welder hats is a good idea

Mass producing these as you say would take at least twice the effort than using an autolathe for welders or circular saws, and perhaps some botany, cargo and/or extra candles as the kitchen supplies runs short of the required ingredients for this recipe. 
And that PR I linked also allows birthday cakes to be mass produced too, instead of one per shift by default.
So why not? I told you before there is no need to nerf stats like this.

Poojawa:
I do believe there isn't that much of effort.

Someone upgrading the microwaves allows them to make 2-4 in a single run.

So no, that isn't at all a better alternative.

Ghommie:
I think you are mixing up the microwave with the food processor here. The former doesn't play loaves and fishes multiplication unlike the latter.

Poojawa:
```
/obj/item/reagent_containers/food/snacks/store/cake/birthday/microwave_act(obj/machinery/microwave/M) //super sekrit club
	new /obj/item/clothing/head/hardhat/cakehat(get_turf(src))
	qdel(src)
```

Please stop trying to assume I'm retarded. thanks.

lolman360:
> Please stop trying to assume I'm retarded. thanks.

pooj you are tho

to clarify i meant that both of you are confused
upgraded microwave doesn't dupe food just cooks it a bit faster
upgraded processor does duplicate food and work faster 
cake into cakehat is a microwave act and so isn't mutiplied
so yes, upgraded microwaves do allow for cakehats to be made faster but not for cheaper

silicons (changes requested):
10 force, no dismemberment, 5 throwforce, 500 heat, 300 hotspot expose (i'm alright with turning heat/hotspot expose down a bit more if it's more powerful than i thought)
i think that's fair for all involved.

Poojawa:
5 force, it's still just a tiny candle. and puts it in line with other tools like the crowbar

nik707:
@kevinz000

PR #9071, opened 2019-08-07
Nerfs Autocloning

YakumoChen:
About The Pull Request

Autocloning now requires tier 4 parts

## Why It's Good For The Game
Tier 3 parts are too easy to get for such a monumentally centralizing game aspect that constantly fucks over any attempt at assassination or doing any sort of antag work that isn't incredibly loud.

## Changelog
:cl: Yakumo Chen
balance: Autocloning now requires tier 4 parts
/:cl:

Sabotaging cloning was made harder, so some balance is needed. I expect there will be some salt in the PR. I will be ignoring it unless it is from a maintainer.


Poojawa:
I feel this is an alright between measure. I'm still of the opinion that autocloning needs to be entirely removed, but tier 4 parts would be alright as well.

davidchan13:
Auto clone isn't the problem, the free ability to prescan is.

silicons (approved):

YakumoChen:
I would personally love to remove autoscanning or prescanning but I have no idea how. Tweaking with numbers is something I can do, however

RogueSteampunker:
Removing prescanning and/or autocloning has already has it's major downsides painted out constantly and consistently, so trying to fuck with them is generally not gonna fly. However, making it so it does take T4's is definitely a fair compromise on both fronts, since it'd reasonably give antags time to sabotage cloning if needed, and generally gate off the feature in a more proper way than just "Did the miners throw the roundstart silver in the orm?"

silicons (approved):

PR #9076, opened 2019-08-08
Fixes end of round voting unintentionally showing map votes and not gamemode votes 

BurgerLUA:
Bugfix for pr #8922


deathride58 (approved):
:csgrad:

BurgerLUA:
> :csgrad:

i don't know what this emote means but the reason why it took me a century to fix this bug is because I legitmately didn't know how to fix it, until I stared at it for like an hour and realised I could steal a few things

PR #9052, opened 2019-08-05
Hmm, today the Hierophant will go sicko mode on shaft miners.

silicons:
## About The Pull Request

Hierophant no longer ignores its more powerful ranged capabilities in melee mode if the target's health is above near-critical. 
Hierophant melee "wave" bursts can no longer multi-hit.

## Why It's Good For The Game

Megafauna are supposed to take skill, not a bag of cores and tanking. This fixes that.
Multi-hit is just terrible since it hits you harder if you run from it, unlike any of the others.

## Changelog
:cl:
balance: Hierophant now goes sicko mode, but hey, at least you can't be multi-hit by melee waves!
/:cl:

PR #9082, opened 2019-08-08
Hierophant's Revenge

silicons:
Kind of a side project while I work on my 200 other projects because this is the only megafauna that really entertains me and it deserves more love.

## About The Pull Request
Refactors hierophant's abilities and stuff to a vortex magic component.
Hierophant rework/plans:

- [x] Band aid patch: Until this is done, hierophant will stop only using weak melee bursts in melee distance and use its full power regardless of distance. #9052 

- [ ] Split hierophant fight into stages instead of a single anger multiplier 

- [x] Refactor hierophant abilities to use vortex magic component

- [ ] New attack: Staggered columns - Create 3 wide lines of blasts with 3 tiles of space between each, across the whole screen, with cardinal sweep directions

- [ ] EXPERIMENTAL - Diagonal - direction chasers. Probably won't work well at all but eh, whatever.

- [x] Faster melee reflex and movements

- [ ] Make chaser/blink/blasts more concurrent, rather than only being used in burst stages.

Hierophant staff plans:

- [ ] ENERGY SYSTEM to replace ability cooldown blocks

- [ ] Remove distance limit on abilities (Still checks vision, no x-ray bullshit)

- [ ] Concurrent ability usage rather than each ability blocking usage 

- [ ] Teleport beacon placement is instant, can instantly remove even without being at the beacon. Recalling to beacon will still take 5 seconds.

- [ ] Chaser delay reduced to 1 second or 0.5 seconds, but only does 10 damage and seeks far faster, and will only hit once per chaser.

- [ ] Implementation of lock-on system for chasers rather than requiring a direct click, toggleable

- [ ] Melee blasts/hits reduced to 20 damage down from 40

- [ ] The staff is now a twohand-required weapon.

- [ ] Dash/spiral burst ability

- [ ] Some random utility abilities to make this not just a murder weapon

## Why It's Good For The Game

*yawn
Hierophant fights aren't interesting enough, the staff it drops is basically useless other than a party trick and a murder weapon in maintenance.

## Changelog
:cl:
refactor: Hierophant abilities are now refactored to be "vortex magic"
experimental: Hierophant fight drastically changed up to use actual stages
experimental: Hierophant staff made vastly more versatile
/:cl:


YakumoChen:
>party trick

excuse me, the staff is the best thing on lavaland, the teleporting is amazingly useful, reliable, and can bring anyone nearby with you to the drop point (including, for instance, murder targets). It's also a fantastic mining tool and powerful against anything that moves.

i agree the heiro is too easy though, arguably easier than the dragon since it doesn't require fireproofing, just enough healing to get through the damage.

PR #9092, opened 2019-08-09
crew combat mechs are now spawned in with a tracking beacon

Sishen1542:
forgot to do this when I ported the mech EMP changes in https://github.com/Citadel-Station-13/Citadel-Station-13/pull/8795

### DIFF
--- a/code/game/mecha/combat/durand.dm
+++ b/code/game/mecha/combat/durand.dm
@@ -20,3 +20,6 @@
 	..()
 	defense_action.Remove(user)
 
+/obj/mecha/combat/Initialize()
+	. = ..()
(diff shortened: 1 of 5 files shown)


silicons (changes requested):
mechs need to be unable to have multiple tracking beacons, tracking beacons need a longer delay for EMPS, and tracking beacons should optimally be able to be removed. Yes I know mechs are op but right now you can just kill a mech by suicide charging with a tracking beacon and having a friend spam EMPs with no way for the operator to counter or even delay. Even if it's the point, it should be a slower process, rather than being able to be highly spammed.

Sishen1542:
Tripled the recharge timer for beacons.

silicons (approved):

PR #9094, opened 2019-08-09
[Needs Testing] Players without antag prefs will no longer be force-selected for traitor or changeling.

izzyinbox:
Slightly changes the way that antags are selected. It will aim to always satisfy the minimum required number of antags instead of the recommended number of antags. If enough players with antag preferences are available, it should function as normal. If there are not enough players with antag preferences available, it will properly respect each gamemode's set required antags number, but won't force any roles past that.
This means for specifically traitor, changeling, and traitor+changeling, the minimum number of antags is zero. If no players have antag preferences on when these game modes are rolled, there will be no forced antags. (IAA is an exception).
Other game modes will force roles up to the non-zero required antags.
required_enemies numbers for game modes may need to be adjusted to compensate, but should act normally under most circumstances.

:cl:
tweak: Some game modes will not force antag roles anymore.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/changeling.dm
+++ b/code/game/gamemodes/changeling/changeling.dm
@@ -24,6 +24,7 @@ GLOBAL_VAR(changeling_team_objective_type) //If this is not null, we hand our th
 
 	var/const/changeling_amount = 4 //hard limit on changelings if scaling is turned off
 	var/list/changelings = list()
+	var/antags_required = FALSE // allows for possibility of no antags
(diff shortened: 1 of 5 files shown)


deathride58:
It might be a good idea to make these gamemodes have their own latejoin antag roll checks as compensation for the reduced antag count.

nik707:
Looks good, needs a testmerge and perhaps that idea bhijn posted.

PR #9100, opened 2019-08-11
Breasts no longer lactate by default, Lactation is now a preference

cebutris:
## About The Pull Request

Title. Breast lactation is now off by default, with a pref to turn it on

@Poojawa should probably look at this before merging

## Why It's Good For The Game

I don't have much experience with girls IRL but I'm mostly sure they're not all constantly lactating

## Changelog
:cl: Cebutris
tweak: Breasts no longer lactate by default, lactation is now a preference
/:cl:

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -189,6 +189,7 @@
 		"breasts_size" 		= pick(GLOB.breasts_size_list),
 		"breasts_shape"		= "Pair",
 		"breasts_fluid" 	= "milk",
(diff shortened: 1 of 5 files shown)


Poojawa:
actually girls are able to produce if they work at it constantly.

cebutris:
Right, but that seems more of an exception to the rule (Though I wouldn't be surprised if everyone with breasts is just constantly toying with them)

jakeramsay007 (approved):

deathride58 (approved):

PR #9104, opened 2019-08-11
classic & telebaton no longer have attack delay

Sishen1542:
what it says on the tin, can't just aggrograb someone batonning you effortlessly now
little slower than a stunbaton and still does less stam damage than one


deathride58:
Wait im 90% sure telebaton click delay was removed entirely

deathride58 (dismissed):

nik707 (approved):

PR #9116, opened 2019-08-12
Changes beforeShuttleMove on airlock to addtimer(0) rather than InvokeAsync

silicons:
Emergency shuttle got chopped in half by a runtime caused by something moving during shuttle preflight_check(), this is the literal only thing that I see that can possibly be moving things. Better safe than sorry tbh.

### DIFF
--- a/code/modules/shuttle/on_move.dm
+++ b/code/modules/shuttle/on_move.dm
@@ -179,7 +179,7 @@ All ShuttleMove procs go here
 	for(var/obj/machinery/door/airlock/A in range(1, src))  // includes src
 		A.shuttledocked = FALSE
 		A.air_tight = TRUE
-		INVOKE_ASYNC(A, /obj/machinery/door/.proc/close)
+		addtimer(CALLBACK(A, /obj/machinery/door/.proc/close), 0)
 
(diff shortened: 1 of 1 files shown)


code/modules/shuttle/on_move.dm
```diff
@@ -179,7 +179,7 @@ All ShuttleMove procs go here
 	for(var/obj/machinery/door/airlock/A in range(1, src))  // includes src
 		A.shuttledocked = FALSE
 		A.air_tight = TRUE
-		INVOKE_ASYNC(A, /obj/machinery/door/.proc/close)
+		addtimer(CALLBACK(A, /obj/machinery/door/.proc/close), 0)
```
Poojawa, line 182: just use a spawn. 0 duration timers are pointless

deathride58: spawn() would do the exact same thing that invoke_async does here. if spawn() runtimes, the entire proc runtimes and cancels out. if invoke_async runtimes, the entire proc runtimes and cancels out. meanwhile, if addtimer() runtimes, the only thing that gets cancelled out is the proc called by the timer, while the rest of the proc that created the timer can continue running without issue.
spawn() and invoke_async handle everything within the proc itself. addtimer outsources the function to the timer subsystem proper.

deathride58 (approved):

PR #9114, opened 2019-08-12
Spiders that spawn from player-laid eggs are now AI controlled.

BurgerLUA:
## About The Pull Request

Spiders that spawn from player-laid eggs are now AI controlled, meaning that they are no longer ALSO player controlled.

## Why It's Good For The Game

Let's compare spiders to xenos in terms of balance when controlled by players.

| Point | Spiders | Xenomorphs |
| --- | --- | --- |
Breeding Method | Lay an egg in a monkey or a human, regardless of health, and you'll get ~10 spiderlings in a few minutes. | Lay a single egg in a conscious human to get one larvae. |
| Childhood | Scatter into vents at the speed of bullshit. Takes 3 hits with an object to kill. Turn into an adult instantly without warning. Since there is so many of you, there will ALWAYS be adults. | Sit inside a conscious human hoping they don't escape their bed. Pop out of them when it's time and then wait another x minutes to evolve while anything more powerful than a fart can kill you.
Speed | Lightning fucking fast. | Lightning fucking fast. |
Health | 200 god damn health | 125 health. |
Melee Damage | 15 to 20. Deals 60 damage to objects. | 25. Deals 60 damage to objects. |
Player Obtainment | Xenobiology Gold Core Slime RNG and Sentient Transfer Potion. Takes about 30-45 minutes to get if you know what you're doing. | Xenomorph Alien Hive raid + Surgery. Takes about 20-30 minutes to clear a dangerous Xeno Hive. Another 10-15 minutes to find and have someone do surgery for you to insert ALL the organs.
| Weaknesses | Lack of Air. | Fire. |

image

Spiders are literally xenomorphs except better. Their numbers are absolutely insane compared to xenomorphs and that is because of this fundamental fact: Spiders are balanced around being controlled by an AI and not players who override their movement and attack speed. Antags can easily create situations by throwing a few corpses to the spiders and it then making a nest and making more spiders; that is the purpose of spiders. The balance did not anticipate that a player would preemptively hoard the maximum amount of monkeys and then turn themselves into a spider to then lay eggs in all the monkeys to create a near infinite stream of player-controlled spiders as long as there were salty players looking to murderbone.

You can still create bullshit spider armies, however they are not player controlled and do not have the benefits of being player controlled, which means that spiders are now balanced as they should be.

## Changelog
:cl: BurgerBB
balance: Spiders can no longer be ghost controlled if they're laid by players.
/:cl:

PR #9124, opened 2019-08-13
Alternate Solution x2: Nerf spiderling spawn amount

Girdo:
Ends the "How To Fix Spiders" argument in the easiest possible manner, by changing literally two numbers to lower the min/max amount of new spiderlings that can be spawned from egg nests.

Alternate solution to PR #9114 and/or PR #9117 

## About The Pull Request

Exactly as I've said above. There are two solutions already proposed (PRs #9114 and #9117, which aim to remove spiders entirely and make them AI-controlled only, respectively). The issue in question seemed to be that the spiders snowball very quickly rather than the actual existence of player-controlled spiders to begin with... so lowering the spawn count of spiderlings (from egg nests) seems like the simplest option. Dropping from a random 3-12 range to a 2-6 range should be an appropriate amount.

## Why It's Good For The Game

Public opinion, including my own, is that player-controlled spider gimmicks, whether from changelings or from xenobio traitors, are underused and fun when they happen. It just needs to be tweaked a bit and rebalanced, rather than being removed entirely.

## Changelog
:cl:
tweak: tweaked literally two numbers in the code
balance: rebalanced spiderlings spawn amount from eggs
/:cl:

### DIFF
--- a/code/game/objects/effects/spiders.dm
+++ b/code/game/objects/effects/spiders.dm
@@ -68,7 +68,7 @@
 /obj/structure/spider/eggcluster/process()
 	amount_grown += rand(0,2)
 	if(amount_grown >= 100)
-		var/num = rand(3,12)
+		var/num = rand(2,6)
 		for(var/i=0, i<num, i++)
 			var/obj/structure/spider/spiderling/S = new /obj/structure/spider/spiderling(src.loc)
 			S.poison_type = poison_type


BurgerLUA:
It would nerf non-player controlled spiders which I'm not comfortable about but at this point I don't think it is my business anyways.

RogueSteampunker:
Out of all three, this is the most sound balance-wise. It doesn't make it impossible to spread, but it also makes it so you can't just win by webbing two monkeys, without also entirely gutting player-controlled spiders in any sort.

Girdo:
> It would nerf non-player controlled spiders which I'm not comfortable about but at this point I don't think it is my business anyways.

It would nerf non-PC spiders as well, correct. At the same time, though... non-PC spiders in their current state don't ever seem to create egg-nests to begin with. Spider Infestation events, the typical cause of spiders, are typically just warrior spiders appearing in the AI Upload and getting shot to death by the turrets. Do you think this would really make that much a difference in what is already considered a rather mundane occurence?

In my opinion, it's the player-controlled spiders that ever actually become a threat, so balancing that at the expense of an already-mundane event doesn't seem like a great loss.

KrabSpider:
It would be a lot more viable to just remove nurse spiders from the gold slime pool.

nik707 (approved):

silicons:
lack of interest (SEE: ACTUAL MERGING) from other maintainers + i'm against this

PR #9132, opened 2019-08-15
Toricrates: Remastered: Complete Edition 2019

Michiyamenotehifunana:
:cl: Toriate
imageadd: Updated the sprites of all the regular crates
/:cl:

This PR will resprite crates (again).

Quick comparisons of a few crates which have been given much needed makeovers:
3Is3ehesam
kBsDGuYgnY
rfU8orhkGk
qrB6US66Vh
FG3vOr6Pny

As can be inferred from the comparisons, the red "THIS WAY UP" arrows have been removed from all of the crates.

In-game comparison:
NpWSxcii1e


Thalpy:
Only comments I would make:
For the first box, I like the arrow or something so it's not just flat grey. Maybe it doesn't have to be a flat grey, but we need the Tori pizazz!!
3rd: I think it needs to be more blue to differentiate it clearly between the grey box in the dim lighting of the game
Is the 4th purely a weapons box? Is it a concern that we lose the visual sword indicating that it's weapons without the use of text?
5th I don't know what this crate is used for in both cases. (Is it food?)

In the last demo pic, we lose the green stripe, I personally like colour, but it's probably fine for that.

Also this is subjective so you can ignore me if you like.

deathride58:
I wish to consume these crates

silicons (approved):

PR #9140, opened 2019-08-17
reduces meteor early warning to flat 300 (5 minutes) seconds down from 300-600

silicons:
@maintainers wondering if I should just make this 30 or 60 second flat rather than RNG

about this pr: title

why this is good for the game: i like the idea of early warning for engineers/crew to scramble to evacuate areas/get ready to defend/rebuild/reinforce, or to do said things to a few critical areas like engine.

i don't like there being enough time to set up enough shielding to nullify the wave and for people to go "OH IT'S FINALLY HERE" because they've been waiting an eternity.


BurgerLUA:
The early warning was to encourage players to setup shields and do their job. If you set it to something low as hell like this, it would be like the system doesn't even exist in the first place.

5 to 10 minutes is extremely fair.

lolman360:
>10 minutes warning
>engineers with any hint of competence do meteor shields really slowly
>station is saved, meteors don't even matter lmao

or

>1 minute warning
>engineers scramble to secure important areas
>important areas saved, engis have to now fix non-critical areas like mining dock or something

Sishen1542:
or you can just set up shields before you get the warning, lmao

BurgerLUA:
10 minute is best case scenario. If you want you can make the range 4 minutes to 8 minutes if you think it's too long.

Like from what I'm reading here none of you actually know the impact of the early warning system. I've watched rounds and had people report to me the effect of the PR. In about 60% of cases people setup meteor shields. In 40% of cases people do not. The goal is at least half the time, the shields are setup and my PR succeeds at that goal.

The early warning system is training wheels to get people in the habit of setting up meteor shields as explained the the initial PR. In about 1 round out of 15 I've played with a meteor storm event without any early warning system, the shields were setup. That is extremely low. A 30 to 90 second warning would bring us back to engineers not bothering to setup the shields anymore and no longer get in the habit of it.

nik707:
Honestly I feel like this is too small of a time for warnings, but I'm gonna leave this up to @deathride58 and @BlackMajor

PR #9156, opened 2019-08-18
Adds a 1 second delay for table slamming and Dakimakuras

Poojawa:
## About The Pull Request

What it says on the tin

## Why It's Good For The Game

You slam your palms down on the table.x3515 can no longer happen within the span of a few ticks.

## Changelog
:cl:Poojawa
fix: fixed an oversight that allowed for massive client lagging.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/dakis.dm
+++ b/code/game/objects/items/storage/dakis.dm
@@ -22,7 +22,7 @@
 /obj/item/storage/daki/attack_self(mob/living/user)
 	var/body_choice
 	var/custom_name
-
+	user.setClickCooldown(1 SECONDS)
 	if(icon_state == "daki_base")
 		body_choice = input("Pick a body.") in dakimakura_options
 		icon_state = "daki_[body_choice]"
--- a/modular_citadel/code/game/objects/structures/tables_racks.dm
+++ b/modular_citadel/code/game/objects/structures/tables_racks.dm
@@ -1,5 +1,6 @@
 /obj/structure/table/alt_attack_hand(mob/user)
 	if(user && Adjacent(user) && !user.incapacitated())
+		user.setClickCooldown(2 SECONDS)
 		if(istype(user) && user.a_intent == INTENT_HARM)
 			user.visible_message("<span class='warning'>[user] slams [user.p_their()] palms down on [src].</span>", "<span class='warning'>You slam your palms down on [src].</span>")
 			playsound(src, 'sound/weapons/sonic_jackhammer.ogg', 50, 1)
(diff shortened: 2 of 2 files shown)


cebutris:
Just make tableslamming do minor damage if you do it more than once in a short timespan, easy.

Poojawa:
I was told repeatedly to never implement an actual punishment for any actions. a click delay is superior to potential mis-clicks.

deathride58 (changes requested):
This is way too long of a delay. Half of melee delay would be much more appropriate for harm intent, and 1 decisecond would be appropriate for tableslapping.

Poojawa:
why does table slamming need to basically not have a delay, tho

deathride58:
Half of melee delay is 4 deciseconds, which is actually quite a lot of delay (almost half a second!). Table slapping is different in that it's meant to be something you can spam, and a 1 decisecond delay lines up pretty well with how fast most people are able to click.

Poojawa:
Except I've been hearing reports of people crashing others with click spam, so not having a delay is just... silly?

We already have massive amounts of to_chat lag with shotguns and the like.

deathride58:
I'm not saying "Let there be no delay", i'm saying "Reduce the delay to something smaller, your proposed values are way too overkill for what you're trying to fix"
A 1 decisecond delay isn't going to cause massive amounts of clientside lag if people get together and spam it; clientside chat lag only occurs when there's a lot of lines to process in one single tick, which a 1 decisecond delay would be more than capable of preventing.
The reason why I propose separating the table slapping and table slamming delays is because table slamming is admittedly a little obnoxious for something that's supposed to be used primarily to get people's attention; a 4 decisecond click delay would be more than enough to curb the annoying table slamming spam without making the click delay on it overly obnoxious. Meanwhile, table slapping is meant to be spammable since, IRL, people who slap their hands on tables typically don't stop at just one slap, they typically drum their hands on the table, which a 1 decisecond click delay would be able to fully simulate; if a 1 decisecond click delay is truly terrible, then add a cooldown to the message itself.

code/game/objects/items/storage/dakis.dm
```diff
@@ -22,7 +22,7 @@
 /obj/item/storage/daki/attack_self(mob/living/user)
 	var/body_choice
 	var/custom_name
-
+	user.setClickCooldown(1 SECONDS)
```
deathride58, line 25: The hotkey for using items in your active hand bypasses click delay. This here won't affect the majority of dakispam instances.

PR #9158, opened 2019-08-18
Rebalances the MASONrig

silicons:
Changes MASONrig armor to more reasonable stats.

"melee" = 10, "bullet" = 5, "laser" = 5, "energy" = 500, "bomb" = 500, "bio" = 500, "rad" = 500, "fire" = 500, "acid" = 500
to
"melee" = 20, "bullet" = 15, "laser" = 15, "energy" = 65, "bomb" = 100, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 100

Why: Slight melee/bullet/laser boosts because it's hardsuit made out of alloy. Why not. This won't be an issue imo. 
Other values reduced to 100 because there's really no point in my opinion of having it so high just in case something gets past it. If one of the very, very few things that can REDUCE armor directly hits it, so be it, I'd say working as intended for it to be able to fail to those rare sources of armor-dampening.
Energy changed to 65. It's now competent against tasers and disablers, rather than fully immune because full taser immunity is stupid.

Removes indestructible flag because this isn't some special super-powerful item that should be impervious to everything. It's a super heatproof and radiation proof hardsuit with potential SM dust-protection in the future. No reason to have it/make it that way.

Closes #9157


silicons:
dropped energy to 45

nik707 (approved):

PR #9171, opened 2019-08-21
Fixes and optimizes ghost swords

silicons:
they're broken. now they're not and actually collect all ghosts following because ghosts don't always follow the sword rather than the person.


Poojawa:
Does this fix being forever fucking tethered to it? because you should be able to un-orbit the damn thing

silicons:
@Poojawa yeah i removed the auto-orbit, it's stupid.

nik707 (approved):

PR #9176, opened 2019-08-22
[TESTMERGE ONLY] Ninja Rebalancing - Sleeping Carp, Cooldowns, Netting, oh my.

jakeramsay007:
Ninjas now spawn with Sleeping Carp. To compensate for this, the cooldown of their sword's dash ability has been increased, and the energy net has been changed/fixed to only affect people who are unconscious, IE critted or sleeping; or stamcritted, to account for being incapacitated by a baton or other method.

This should adjust the current meta of security using shotguns against nearly any threat against them, while also removing the current meta of ninjas grabbing a bag of holding and solely relying on ranged weapons and instant-stuns instead of their abilities and melee combat. Testmerge only for now, until more data can be gathered on whether this is a good change or not.

:cl: Chayse
balance: Gives Ninjas the ability to use Sleeping Carp at the cost of ranged weapon usage
balance: Increases the cooldown on the Energy Katana's dash ability
fix: Energy nets now no longer work on people who are not incapacitated or stamcritted, as originally intended.
/:cl:

### DIFF
--- a/code/modules/ninja/suit/n_suit_verbs/energy_net_nets.dm
+++ b/code/modules/ninja/suit/n_suit_verbs/energy_net_nets.dm
@@ -44,12 +44,6 @@ It is possible to destroy the net by the occupant or someone else.
 		qdel(src)//Get rid of the net.
 		return
 
-	if(ishuman(affecting))
-		var/mob/living/carbon/human/M = affecting
-		if(M.staminaloss <= 99 || M.stat != UNCONSCIOUS)
-			qdel(src)
-			return
-
 	if(check>0)
 		check--
 		return
--- a/code/modules/ninja/suit/n_suit_verbs/energy_net_nets.dm
+++ b/code/modules/ninja/suit/n_suit_verbs/energy_net_nets.dm
@@ -44,6 +44,12 @@ It is possible to destroy the net by the occupant or someone else.
 		qdel(src)//Get rid of the net.
 		return
 
+	if(ishuman(affecting))
+		var/mob/living/carbon/human/M = affecting
+		if(M.staminaloss <= 99 || M.stat != UNCONSCIOUS)
+			qdel(src)
+			return
+
 	if(check>0)
 		check--
 		return
--- a/code/modules/antagonists/ninja/ninja.dm
+++ b/code/modules/antagonists/ninja/ninja.dm
@@ -22,6 +22,8 @@
 	update_ninja_icons_removed(M)
 
 /datum/antagonist/ninja/proc/equip_space_ninja(mob/living/carbon/human/H = owner.current)
+	var/datum/martial_art/the_sleeping_carp/ninja = new
+	ninja.teach(H)
 	return H.equipOutfit(/datum/outfit/ninja)
 
 /datum/antagonist/ninja/proc/addMemories()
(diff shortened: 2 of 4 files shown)


Arturlang:
I actually really like this change, giving ninjas a good reason to use their sword, However, the cooldown doubling it is quite a bit.

jakeramsay007:
> I actually really like this change, giving ninjas a good reason to use their sword, However, the cooldown doubling it is quite a bit.

So the energy katana actually has a really weird cooldown. Each time you dash, it adds another timer for recharging that specific charge usage. So you can use all three, and all three will be on cooldown at the same time and usable again in six seconds. Or, you can use one dash every two seconds and always have two dashes charging when you use one, thereby effectively always having a dash off-CD to be used. This mechanic rewards careful usage of dashing for repositioning, while still allowing the usage of all three in rapid succession for escapes.

code/modules/ninja/suit/n_suit_verbs/energy_net_nets.dm
```diff
@@ -44,6 +44,12 @@ It is possible to destroy the net by the occupant or someone else.
 		qdel(src)//Get rid of the net.
 		return
 
+	if(ishuman(affecting))
+		var/mob/living/carbon/human/M = affecting
+		if(M.staminaloss <= 99 || M.stat != UNCONSCIOUS)
```
Useroth:
```suggestion
		if(M.staminaloss <= 99 && M.stat != UNCONSCIOUS)
```
would be way more appropriate, otherwise you won't even be able to capture a conscious person in stamcrit

Also, I'm not entirely sure about the need to keep the person in stamcrit the entire time just to not make the net instantly *poof* away. I've buffed the gloves a little bit a while ago, to apply some stun, but I'm not even sure if a ninja would be able to capture anyone like this on one battery charge.

cebutris:
I actually really love this PR, however, it would be nice to give the ninja a way of nonlethally incapacitating victims from the start, if that's the only way they'll be able to capture victims going forward. Perhaps a stun/disabler mode for the energy katana?

jakeramsay007:
> I actually really love this PR, however, it would be nice to give the ninja a way of nonlethally incapacitating victims from the start, if that's the only way they'll be able to capture victims going forward. Perhaps a stun/disabler mode for the energy katana?

Both non-lethal and lethal playstyles work, as you can capture someone who's incapacitated via stamcrit or via being knocked into critical via sword or other methods.

YakumoChen:
Ninjas are already one of the most dangerous antags to get up close to. They move extremely fast and have a ton of mobility, their sword has a high block chance and does massive damage, they have a built-in stun already with stun gloves, and they have a powerful ranged attack already that sleeping carp ignores. Sleeping carp is absurdly powerful on its own. Why buff them?

Arturlang:
Sleeping carp is a sidegrade really, in this case, it really incentivies to use swords instead of rangeds, their sword already blocks nearly all rangeds, allowing in a lot more honorable fights, instead of just being one shot stamcritted by a shotgun, however, theres a reason its a testmerge, to see how it works out.

YakumoChen:
It blocks everything, not just ranged. This is definitely a buff, ninjas definitely don't need guns to kill everyone (two throwing stars is a critical at range anyway), but hitting something as fast and mobile in melee as a ninja is basically impossible.

nik707:
@kevinz000 @deathride58 me scare

PR #9179, opened 2019-08-23
Gives His Grace stamina regeneration

chefferz:
His Grace currently suffers from extreme fatigue when used, and as a 20TC item this not particularly suitable seeing as you need to feed it in later stages quickly to prevent it from eating you. While His Grace does prevent stuns from knocking you down, the current stamina system makes you absorb stamina damage still, preventing you from actually using the item at all regardless of its charge level, allowing someone with as simple as a hand baton to actively stop your rampage.

This rebalance gives a regeneration that ensures it won't run out when being used in the scenario it's intended for, as the current stamina coding doesn't take into account the later stages of hunger.

Updates are:

Gives a steadily improving regen
Weight class lowered from Gigantic to Bulky

### DIFF
--- a/code/game/objects/items/his_grace.dm
+++ b/code/game/objects/items/his_grace.dm
@@ -12,7 +12,7 @@
 	lefthand_file = 'icons/mob/inhands/equipment/toolbox_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/equipment/toolbox_righthand.dmi'
 	icon = 'icons/obj/items_and_weapons.dmi'
-	w_class = WEIGHT_CLASS_GIGANTIC
+	w_class = WEIGHT_CLASS_NORMAL
 	force = 12
 	total_mass = TOTAL_MASS_NORMAL_ITEM // average toolbox
 	attack_verb = list("robusted")
--- a/code/game/objects/items/his_grace.dm
+++ b/code/game/objects/items/his_grace.dm
@@ -93,7 +93,6 @@
 				master.remove_status_effect(STATUS_EFFECT_HISGRACE)
 				REMOVE_TRAIT(src, TRAIT_NODROP, HIS_GRACE_TRAIT)
 				master.Knockdown(60)
-				master.adjustStaminaLoss(-5)
 				master.adjustBruteLoss(master.maxHealth)
 				playsound(master, 'sound/effects/splat.ogg', 100, 0)
 			else
--- a/code/game/objects/items/his_grace.dm
+++ b/code/game/objects/items/his_grace.dm
@@ -93,6 +93,7 @@
 				master.remove_status_effect(STATUS_EFFECT_HISGRACE)
 				REMOVE_TRAIT(src, TRAIT_NODROP, HIS_GRACE_TRAIT)
 				master.Knockdown(60)
+				master.adjustStaminaLoss(-5)
 				master.adjustBruteLoss(master.maxHealth)
 				playsound(master, 'sound/effects/splat.ogg', 100, 0)
 			else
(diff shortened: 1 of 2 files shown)


Linzolle:
I think the weight class should be at least Bulky so you can't just put his grace in a bag. Does the weight class even affect the stamina usage? I thought that was determined by the total mass.

chefferz:
> 
> 
> I think the weight class should be at least Bulky so you can't just put his grace in a bag. Does the weight class even affect the stamina usage? I thought that was determined by the total mass.

Heya, the suggestion was from Bhjin who said the stamina usage does correlate by its weight value. This would be damn near impossible to put in a bag once activated. It latches onto you and if it doesn't get fed it kills you. The coding prevents it from being secured/trapped in a locker, so while you technically could put this in a normal bag, you wouldn't want to when activated

Linzolle:
> 
> 
> > I think the weight class should be at least Bulky so you can't just put his grace in a bag. Does the weight class even affect the stamina usage? I thought that was determined by the total mass.
> 
> Heya, the suggestion was from Bhjin who said the stamina usage does correlate by its weight value. This would be damn near impossible to put in a bag once activated. It latches onto you and if it doesn't get fed it kills you. The coding prevents it from being secured/trapped in a locker, so while you technically could put this in a normal bag, you wouldn't want to when activated

The getweight() function that's called when determining how much stamina loss to give uses the total_mass of an object before it uses the weight class, so all that change would make is so that you could conceal His Grace in a bag while he is inactive.

code/game/objects/items/his_grace.dm
```diff
@@ -93,6 +93,7 @@
 				master.remove_status_effect(STATUS_EFFECT_HISGRACE)
 				REMOVE_TRAIT(src, TRAIT_NODROP, HIS_GRACE_TRAIT)
 				master.Knockdown(60)
+				master.adjustStaminaLoss(-5)
```
Linzolle: This should be in /datum/status_effect/his_grace/tick() along with the other healing effects. If it's here it'll apply the stamina loss upon the user being eaten by His Grace.

chefferz: Updated per discord discussion

code/game/objects/items/his_grace.dm
```diff
@@ -12,7 +12,7 @@
 	lefthand_file = 'icons/mob/inhands/equipment/toolbox_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/equipment/toolbox_righthand.dmi'
 	icon = 'icons/obj/items_and_weapons.dmi'
-	w_class = WEIGHT_CLASS_GIGANTIC
+	w_class = WEIGHT_CLASS_NORMAL
```
Ghommie:
```suggestion
	w_class = WEIGHT_CLASS_BULKY
```
That's the average toolbox weight class.

code/game/objects/items/his_grace.dm
```diff
@@ -12,7 +12,7 @@
...
 	force = 12
 	total_mass = TOTAL_MASS_NORMAL_ITEM // average toolbox
```
Ghommie, line 17: Remove this line for me thanks, I mistakengly gave it the average normal sized item whack stamina cost (the default for each item is w_class * 1.25, but is overridden by this var) instead of that of a bulky object, and this should be redunant now that its w_class has been changed to match that of a toolbox.

chefferz: Done

chefferz (commented):
updated per request

silicons (approved):

PR #9182, opened 2019-08-23
DNA-locked firing pins are now available under the Medical Weaponry node

jakeramsay007:
Available after already researching mindshield firing pins, adds some variety to pin usage for security weapons.

### DIFF
--- a/code/modules/research/designs/weapon_designs.dm
+++ b/code/modules/research/designs/weapon_designs.dm
@@ -141,7 +141,7 @@
 	id = "pin_dna"
 	build_type = PROTOLATHE
 	materials = list(MAT_SILVER = 600, MAT_DIAMOND = 750, MAT_URANIUM = 200, MAT_TITANIUM = 400)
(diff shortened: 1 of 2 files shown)


YakumoChen:
Gun pins that require a specific implant to limit usage and allow for deep cover or flat out theft to use = good
Gun pins that make it impossible for anyone to steal your gun and use it = bad

Poojawa:
imagine dna locking your department's mech.

YakumoChen:
Mechs can be locked by ID access but it's a roundabout feature nobody ever remembers (and DNA lock should be removed)

But mechs tend to be KOS for anyone not authorized and are pretty big and obvious.

nik707:
I kinda dislike DNA locking if I'm honest. @kevinz000 @deathride58 thoughts?

silicons:
i dont like this

silicons:
dna locking is lamer than mindshield pins being used so much sorry

PR #9193, opened 2019-08-25
Disables crit_threshold changes through mood 

silicons:
There's a pretty major bug/exploit in this, needs to be gone until someone can go through this fucking clusterfuck of a mistake of a system. (why are there different set/increase/decreasesanity procs that dont' call set/the same things and are duplicated code.)


silicons:
@Citadel-Station-13/project-management just merge it unless someone wants to be assed to fix this ass-backwards system.

BlackMajor (approved):

PR #9220, opened 2019-08-30
Allows Furranium grown from organic sources.

Thalpy:
## About The Pull Request

**Requires #9109 !!!**
Adds an alternative reaction for furranium, which uses catnip from #9109 instead of moonsugar.

## Why It's Good For The Game

Moonsugar is described in game as synthetic catnip, it stands to reason that organically grown catnip would work in the reaction as well.

## Changelog
tweak: Furranium now uses catnip or moonsugar
/:cl:


silicons (approved):

Thalpy:
Kev this requires #9109

PR #9247, opened 2019-09-05
Partially fixes tiny runaway decimals in reagents.

Thalpy:
## Abstract
Partially fixes tiny runaway decimals in reagents system.

## About The Pull Request

Fixes a problem that changing the quantification PR made with small rounding errors, as well as lets the chem master dispense decimals correctly. There might need to be another follow-up to this with fringe fermichem reactions, but we'll see.

Much like the previous one, I recommend test merging this first.

I know that ammonia still results in residue, and I'm not quite sure how to fix this without an overblown thing but, at least for now the chem master isn't borking up, so i'm putting this up as is so I can continue to work at it without the chemmaster problem.

## Why It's Good For The Game

Removes residual chems in beakers and the like.

## Changelog
:cl: Fermi
fix: Fixes tiny runaway decimals in reagents system.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/holder.dm
+++ b/code/modules/reagents/chemistry/holder.dm
@@ -873,7 +873,7 @@
 		var/datum/reagent/R = A
 		if (R.id == reagent) //IF MERGING
 			//Add amount and equalize purity
-			R.volume += amount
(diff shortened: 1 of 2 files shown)


Thalpy:
I tested it, but I still recommend a testmerge to be safe.

silicons (approved):

PR #9253, opened 2019-09-08
Doubles stamina regen when standing/laying down

chefferz:
Does what it says, doubles regeneration of stamina when standing and laying down.

This is a quick band-aid PR to test a increased stamina regeneration before Bhjin implements their new curve-based stamina regeneration. 

Test merge would be preferred as only actual play-testing will be able to tell if this is perfectly tweaked, but this should definitely alleviate some of the issues brought up with the current stamina system for now.


Raptorizer:
Stamina meta bad, fix good

deathride58 (approved):
Should be testmerged first to see if it's balanced

chefferz:
Current notes on stamina regen of 6/3.

Using disabler beams is a bit rough, as when they drop to the floor it's hard to keep them actually stamcritted enough to arrest. However, when talking to people about this it seems to not have affected security play as much as it would seem, seeing as people only used disabler beams to stun with batons/secbats.

Sprinting is much, much longer. The buffer is almost 1 solid minute before you're fully exhausted (although I think this has to do with moodlets not affecting speed that's also currently test merged)

silicons:
i actually think sprint stamina cost might need to be toned up juuuust a bit in light of this but shrug

PR #9286, opened 2019-09-13
tones down shotgun falloff in exchange for their lowered damage

silicons:
tupinambis stopped responding so i'm doing it i guess


silicons:
nevermind shotguns don't need a buff

PR #9268, opened 2019-09-10
[READY] [Please merge me] Fermichem tweaks: Purity, Astral, RNG, MK, SMilk, SDGF, furranium, hatmium, eigen, nanite.

Thalpy:
## About The Pull Request

Changes the way a few chems work, and fixes a few problems

**General** Reworks purity to be cached, so that purity calculations don't compound on mob to mob transfers.

**Astral:** Added a few things to reduce the chance that procs will be called at the same time - causing possession mechanics to wig out and cause players to be stuck in each other's bodies.

**randomchem:** Now allows people with enough dedication to cause the effects permanently upon crit players, counting for the assassination objective, without actually killing the player. Finally, a non lethal method! Be careful to not leave any lying around, as spraying the effected players with the chem will revert them back.

**MKUltra:** Tweaks language, adds a few british phrases and adds a few more pref checks on it's use.

**SMilk:** Rate of milk regen now scales from the base regen, with respect to size. (request from community & cebu) 

**SDGF:** Toxin now processes slower, due to the difficult of creation, and the healing chem now processes slightly slower too, to reduce the chance of clones dying on creation. 

**Furanium** Reduces the strictness on the purity so that you can slightly drift from optimal (i.e. so that it accepts 0.9999999999999).

**Hatmium:** Hat armour growth now relies on purity, and hats will now get angry if they're not on a head. Lets you create an armourless hat too by spraying the ground.

**Eigenstasium** fixes a bug that allowed impure chem teleportation, and prevents return teleport if the mob has stabilizing agent inside them.

**Nanite bane** buffed removal rate so that you don't need to take 300u to get rid of healthy nanites, and made rate equal to purity. You still need a substantial amount to make it, and considering it's intense difficulty, I think this is warranted (I don't think you could make 300u unless you spent the whole round trying).

## Why It's Good For The Game

Tweaks and fixes! Hopefully makes things a bit easier to work too.

## Changelog
:cl: Fermis
tweak: Purity, Astral, RNG, MK, SMilk, SDGF, furranium, hatmium, eigen, nanite.
fix: Eigen and purity.
/:cl:

PR #9316, opened 2019-09-19
[READY] [FermiChem] Turns reaction vars into bitflags and adds a few new procs related to that.

Thalpy:
## About The Pull Request

Coverts some of fermichem options into bitflags to reduce varbloat, and uses this new option to open up more integrated options or people to use.

**for datum/reagent/** they have been converted into var/reagentFlags with the following options
```
REAGENT_DEAD_PROCESS		calls on_mob_dead() if present in a dead body
REAGENT_DONOTSPLIT		Do not split the chem at all during processing
REAGENT_ONLYINVERSE		Only invert chem, no splitting
REAGENT_ONMOBMERGE		Call on_mob_life proc when reagents are merging.
REAGENT_INVISIBLE		Doesn't appear on handheld health analyzers.
REAGENT_FORCEONNEW		Forces a on_new() call without a data overhead
REAGENT_SNEAKYNAME          	When inverted, the inverted chem uses the name of the original chem
REAGENT_SPLITRETAINVOL      	Retains initial volume of chem when splitting
```

To use these on a reagent, simply do something akin to: (under var definitions in the typepath)
`reagentFlags = REAGENT_SNEAKYNAME | REAGENT_DONOTSPLIT | REAGENT_DEAD_PROCESS
`
In addition, on_mob_dead() proc has been added, which will process and consume reagents within a mob, per tick. No longer are you limited to reagent interactions on the living anymore (carbonmob and it's children only) (this is done in the carbonproc handle_death() )

Also because I was waiting on #9268 before putting this up, but wanted to write it up before I forgot, I integrated parts of that into this PR. I'd still **highly recommend** merging #9268 first, however.

**for datum/chemical_reaction/**

There is currently 1 new bitflag var for reactions: var/ClearConversion with the following bitflags

```
REACTION_CLEAR_IMPURE       Convert into impure/pure on reaction completion
REACTION_CLEAR_INVERSE      Convert into inverse on reaction completion when purity is low enough
```

This allows splitting of chems inside of the reaction vessel.

Also ports CHECK_MULTIPLE_BITFIELDS, I think. **Please check this**

Moved impure chems into their own reagents subtype, refers to old reagent as isomers. Isomers are created by default whenever a player takes an impure chem, with no impure/inverse var assigned to it. This chem doesn't appear on scanners, but it does give a general message about the volume (to be built on). 

## Why It's Good For The Game

Reduces var overhead, and makes fermichem have more options for people other than me. Also reduced the confusion of "what does all this do?"

## Changelog
:cl: Fermis
add: on_mob_dead(), bitflags and CHECK_MULTIPLE_BITFIELDS
refactor: refactored fermichem vars, moved impure chems into their own reagents subtype
/:cl:

## Roadmap

This is part of the PRs I'm making so I can start porting some of the older chems into the fermichem mechanics.

```
*Fermi*            *Organs+*             
#9268    --->   (this PR)   ----->  Organ tweaks  -----> Reagents porting
#9269              #9309                                  Tamiorgans
```

Basically what I'm trying to do is 1. Clear up Fermichem code for new people to use, 2. Port organ damage so that impure chems can use these as a damage type 3. Tweak TG's organ damage so that we're not as hasty, and add depth to damage instead of it being an alternative method of ded. 4. Add nuance to existing chems in the form of purity, without removing what they do currently.


code/modules/reagents/chemistry/reagents.dm
```diff
@@ -32,21 +32,26 @@
...
-	var/InverseChem 		= "fermiTox"// What chem is metabolised when purity is below InverseChemVal, this shouldn't be made, but if it does, well, I guess I'll know about it.
-	var/DoNotSplit			= FALSE		// If impurity is handled within the main chem itself
-	var/OnMobMergeCheck 	= FALSE 	//Call on_mob_life proc when reagents are merging.
-
+	//var/SplitChem			= FALSE		//If the chem splits on metabolism
+	var/ImpureChem						// What chemical is metabolised with an inpure reaction
+	var/InverseChemVal 		= 0			// If the impurity is below 0.5, replace ALL of the chem with InverseChem upon metabolising
+	var/InverseChem 					// What chem is metabolised when purity is below InverseChemVal, this shouldn't be made, but if it does, well, I guess I'll know about it.
+	var/metabolizing = FALSE
+	var/reagentFlags //REAGENT_DEAD_PROCESS, REAGENT_DONOTSPLIT, REAGENT_ONLYINVERSE, REAGENT_ONMOBMERGE, REAGENT_INVISIBLE, REAGENT_FORCEONNEW, REAGENT_SNEAKYNAME
```
Ghommie: Can you please lowercase the bitfield var name? Vars, procs (and defines, but those should be all caps to be distinguished as such) are case sensitive, and those containing both uppercased and lowercased letters are harder to write and remember correctly, making the work more frustrating.
I only noticed this your tendency just now, but please resort to _ instead (such as reagent_flags) if you wish to clearly distinguish each word composing the var name.

code/modules/reagents/chemistry/reagents.dm
```diff
@@ -32,21 +32,26 @@
...
+	//REAGENT_DEAD_PROCESS  	Calls on_mob_dead() if present in a dead body
+	//REAGENT_DONOTSPLIT		Do not split the chem at all during processing
+	//REAGENT_ONLYINVERSE		Only invert chem, no splitting
+	//REAGENT_ONMOBMERGE		Call on_mob_life proc when reagents are merging.
+	//REAGENT_INVISIBLE			Doesn't appear on handheld health analyzers.
+	//REAGENT_FORCEONNEW		Forces a on_new() call without a data overhead
+	//REAGENT_SNEAKYNAME    	When inverted, the inverted chem uses the name of the original chem
+	//REAGENT_SPLITRETAINVOL 	Retains initial volume of chem when splitting
```
Ghommie: Replace these comments with a single directing the reader to the file containing those flags (e.g. see __DEFINES/reagents.dm for information on each the flags). Less maintenance trouble.

code/modules/reagents/chemistry/reagents.dm
Ghommie: One last thing, you should go ahead and add these flags to `_globalvars/bitfields.dm`, it is used to turn the bitfield var into a more human readable format on VV iirc.
You should understand how to as soon as you give the file the file a look.

Thalpy: Done!

code/modules/reagents/chemistry/reagents.dm
Thalpy: I moved it into a readme I'll fill up as I go along.

Thalpy:
Just one thing Ghommie; do you know the answers to my questions? Did I set them up correctly?

Ghommie:
>     1. Is the correct way of defining a bitflag for multiple flags in the typepath like this?
>        `reagentFlags = REAGENT_SNEAKYNAME | REAGENT_DONOTSPLIT | REAGENT_DEAD_PROCESS`

Yes.

>     2. Have it defined my bitflags right?

What do you mean? as long as you do `#define` the bitflags, it shouldn't be an issue.
But for reference, if you make a define meant to be used on one single file or proc, remember to `#undef` it at the end (no need to state the value, just the name).

>     3. Can I check for FALSE with `if(!(reagentFlags & REAGENT_DONOTSPLIT))`, or is there a better way?

Yes, there are bitfield operators defined at the start of __DEFINES/flags.dm, check them out. They are pretty useful in place of certain symbols not present in some locales or keyboards among other things.

> 
>     4. Can I check multiple flags like this: `if(reagentFlags & REAGENT_DONOTSPLIT | REAGENT_DEAD_PROCESS)`?
> 

Yes, but it hasn't been ported here yet. I'll be grateful if you do. [CHECK_MULTIPLE_BITFIELDS](https://github.com/tgstation/tgstation/blob/3ae3830d8f5f0645793e83d8d4915a1ce1967f15/code/__DEFINES/flags.dm#L15)

MrJWhit:
So, I'm not sure how I feel about this just yet. Not the pr itself, this is all totally fine with me, but the roadmap and the explanation of the roadmap.
Specifically the part where "Port organ damage so that impure chems can use these as a damage type 3. Tweak TG's organ damage so that we're not as hasty, and add depth to damage instead of it being an alternative method of ded. ", because one of the initial tenants of FermiChem was that it was for the hyper competent players, that the chemist who just started out doesn't need to worry about pH and how it affects it's purity. The chemist just needs to worry about they put all of their chemicals to make Bicaridine, and not work about a pH buffer chemical. It's explicitly against the idea.
On the other hand, it's *really* cool, and adding organ damage through it is also *really* cool and makes things honestly more interesting. 
I don't have anything specific or thoughtful, I just felt like sharing my indecision.

Thalpy:
> So, I'm not sure how I feel about this just yet. Not the pr itself, this is all totally fine with me, but the roadmap and the explanation of the roadmap.
> Specifically the part where "Port organ damage so that impure chems can use these as a damage type 3. Tweak TG's organ damage so that we're not as hasty, and add depth to damage instead of it being an alternative method of ded. ", because one of the initial tenants of FermiChem was that it was for the hyper competent players, that the chemist who just started out doesn't need to worry about pH and how it affects it's purity. The chemist just needs to worry about they put all of their chemicals to make Bicaridine, and not work about a pH buffer chemical. It's explicitly against the idea.
> On the other hand, it's _really_ cool, and adding organ damage through it is also _really_ cool and makes things honestly more interesting.
> I don't have anything specific or thoughtful, I just felt like sharing my indecision.

The current implementation of fermichem is meant for the extreme end of it, I don't want every reaction to feel like a faff. pH should be ignorable for the starting chemists, though, I agree. If I do touch Bicaridine, It's very likely to be very light changes, or non at all for the basic chems. What I'm thinking at the moment is make all chems 0.8/0.9 purity that exist in the station, then craft a reaction for bicaridine/kelotane that naturally ends up at 0.8/0.9 based off the pH of the reagents. Then, someone who wants the extra 0.2/0.1 can try, and for those who don't, they end up with 0.8. In this case, purity only affects the healing rate of bicaridine, and doesn't incorporate any impure chem or volume loss mechanics. I suppose another question I have to ask is do I set 0.8 to 2brute/tick, or do I set 1 to 2brute/tick. I'm leaning towards the former, as the latter is a nerf, and I don't want to gimp medical, just offer an extra 0.4brute/tick.

From there, I can slow introduce concepts such as splitting, impure toxins, explosions, and the rest from fermichem. If I do end up using impurities as a impure_chem, it's effects will be gated from the volume in the body.

The current idea is this for impurities:
```
/datum/reagent/impure/impurities
    switch(volume)
        if(10 to 20)
            damage liver(1)
        if(20 to 35)
            damage liver(2)
            dizzy(5)//or something
```
ect, with mounting consequences primarily for the liver for chugging a ton of impure chems

silicons (approved):

PR #9319, opened 2019-09-20
Captain can now be converted in CWC/Bloodcult without a mindshield

jakeramsay007:
There was no real reason to prevent the Captain from getting converted during either of these gamemodes, especially when the HoS is still a valid conversion target. If the cult can manage to convert the Cap without getting called out, then they're probably competent enough to win anyway.

## Changelog
:cl:
balance: The Captain is no longer 100% immune to being converted by both cult types.
/:cl:

### DIFF
--- a/code/game/gamemodes/clock_cult/clock_cult.dm
+++ b/code/game/gamemodes/clock_cult/clock_cult.dm
@@ -52,7 +52,7 @@ Credit where due:
 	if(!istype(M))
 		return FALSE
 	if(M.mind)
-		if(M.mind.assigned_role in list("Captain", "Chaplain"))
+		if(M.mind.assigned_role in list("Chaplain"))
 			return FALSE
 		if(M.mind.enslaved_to && !is_servant_of_ratvar(M.mind.enslaved_to))
 			return FALSE
--- a/code/game/gamemodes/cult/cult.dm
+++ b/code/game/gamemodes/cult/cult.dm
@@ -16,7 +16,7 @@
 	if(!istype(M))
 		return FALSE
 	if(M.mind)
-		if(M.mind.assigned_role in list("Captain", "Chaplain"))
+		if(M.mind.assigned_role in list("Chaplain"))
 			return FALSE
 		if(specific_cult && specific_cult.is_sacrifice_target(M.mind))
 			return FALSE


Sishen1542:
hard no from me

says he's concerned about people not being able to be converted since it "takes people out of the round for up to an hour" and yet doesn't remove chaplain

thinly veiled buff to his main role, wants to be captain AND be an antag

openly calls holymelons protecting in pocket an "intended feature" but mind checking chap/cap is "ancient code"

:thinking:

jakeramsay007:
>thinly veiled buff to his main role
image

Sishen1542:
> > thinly veiled buff to his main role

sitting in the bar nets you lots of playtime

lolman360:
out of the past few rounds I played chayse has always been cap

Trilbyspaceclone:
I hate this so much, cwc can stealth get sec/cap with easy round fucken start. Why make this even a problem to be even worst in the first place

jakeramsay007:
> out of the past few rounds I played chayse has always been cap

of the three rounds I've played Captain in the last two weeks, two of them were traitor rounds in which the antags did fuck all until the last thirty minutes, and the third was a clock cult round in which the clockies yoinked the spare roundstart, converted maybe three people, and then got absolutely stomped at the end. So no, not a 'thinly veiled buff to [my] main role', because captain certainly isn't it.

BlackMajor:
I've made my thoughts on this before but I'm in the camp of Captain should retain his nonconvertible status.

silicons:
it's been more than two months by tomorrow and there's no chance the maintainer votes of 2 no 1 yes is going to change
stale.

ForrestWick:
I think it's worth saying that having the captain be convertible doesn't really change my h of anything but allow for less meta trust and make it so the captain isn't going to be instant yeeted because it's a cult round.

With the change Persi made to make cap office a safe spot and not able to be teleporter to by cult period (which includes culted captain), I think this would be fine. I'll list out some other points in a list here for convenience.

>captain is basically marked for death in three game modes (four if you wanna be cheesy and include nukies), and yes that means you accept the risk by playing the role, but when the game is about fun I don't think that risk has any purpose and just adds frustration, dissuading people from bothering with the role as a result. Dropping this two one (two counting nukies) game modes would be a helluva lot better so that at least there is a chance for the captain to not just instant die.
>to add onto the previous point, both cults have instant, powerful stuns that can't be blocked now without the nullrod. A buff would help sure, but that still makes it so if the captain does get captured they're still fucked because they can't be converted anyway. There's no way to counter kindle or blood stun at all besides luck most of the time because the only people gunning for the cap as cult will be mobs or mega robust 10%ers.

Losing may be part of the game, but that doesn't mean you should be punished for picking a vital role you might like and be good at just because half the normal rotation game modes point at you and laugh.
>literally all other heads of staff and security can be converted to either cult, and sec can be converted to revs as well. To top it off, heads don't start mindshielded. They don't even get that old loyalty implant shit that used to be around before shields. Captain being cultable is hardly a noticeable buff, and the instant people see no mindshield on a captain they'll go completely hog wild. Last time I had it happen to me, the cult was already steamrolling but I had people hunting me relentlessly with the intent to kill. This was a shitload more enjoyable than just being put on a vitality matrix and yeeted or sharded because I can't be converted anyway (I've taken notice that blood players are starting to make.attempts at shield removal if they can)
>we want less lethality from antags, but two headmins are saying no to a PR that encourages just that? Something isn't right about that.
>more than enough regular players have voiced their support for this change that I think it should at least be given a vote.
>there has been no real reasons given to shoot this down outside of "captain is too loyal" or "its too much of a buff". The second point has some ground because if and only IF a cap gets converted and somehow avoids the eyes of borgs and security to get a new mindshield and manage to inject himself without being spotted then yeah they got some power they can abuse. But the lack of any metashield means the AI is going to be way more hesitant to have law changes on cult rounds. It'll also mean crew will be less able to trust in the captain and be more ready to yeet them if need be.
The first point about loyalty is just laughable.

To sum it up, I have no idea why this is being fought against so hard and it really bugs me that it is being fought against considering the reasons I'm given have no real ground

PR #9328, opened 2019-09-21
how many maintainers does it take to nerf xenomorphs (or fix the stamina combat system)

silicons:
## About The Pull Request

Refactors stamina combat Knockdown status override to apply on Knockdown() rather than status new.
This is because without this or another solution, Knockdown() just adds/min()s the duration you are attempting to be stunned for without doing the stamina combat overrides, making it so if you are stunned while already stunned you receive the full value of the stun in hardstun rather than the stamina combat overrides triggering.

## Why It's Good For The Game

Because without this, you can hardstun people by just having enough people baton them at the same time, or y'know, have xenos just instantly with a single click as the second click will hardstun for 10 seconds.

## Changelog
:cl:
experimental: Stamina combat refactored to apply knockdown overrides on Knockdown() rather than status_effect/knockdown/on_creation().
/:cl:

also adds logging to xeno disarms :)


Thalpy:
I was going to jokingly review this but combat PRs are confusing, where's the overcomplciated equations?

code/modules/mob/living/status_procs.dm
```diff
@@ -66,10 +66,17 @@
 	if(((status_flags & CANKNOCKDOWN) && !HAS_TRAIT(src, TRAIT_STUNIMMUNE)) || ignore_canknockdown)
 		if(absorb_stun(isnull(override_hardstun)? amount : override_hardstun, ignore_canknockdown))
 			return
+		//STAMINA COMBAT START
+		adjustStaminaLoss(isnull(override_stamdmg)? amount * 0.25 : override_stamdmg)
+		if(isnull(override_hardstun))
+			amount *= 0.01
```
deathride58, line 72: This is here only as a hack to force disarms and interrupt timers, it'd probably be better to make this directly force those things in the proc under the proper conditions (knockdown strength is more than 80 and the hardstun isnt being overridden)

code/modules/mob/living/status_procs.dm
```diff
@@ -66,10 +66,17 @@
...
+		else
+			amount = override_hardstun
+		//END
 		var/datum/status_effect/incapacitating/knockdown/K = IsKnockdown()
 		if(K)
 			K.duration = max(world.time + (isnull(override_hardstun)? amount : override_hardstun), K.duration)
-		else if((amount || override_hardstun) > 0)
+		else if(amount > 0)
```
deathride58, line 79: You're missing the check that cancels the knockdown if it's over 80, meaning *all* knockdown sources will force people to drop things in their hands and cancel timers

deathride58 (changes requested):
The new knockdown proc is missing the resting = true and update_canmove() (meaning knockdowns won't actually knock spacemen down), the ability for knockdowns to knock spacemen off of vehicles (returning them back to their state of nigh-invulnerability), and the check that only goes through with the stun if the amount is over 80 (making *all* knockdown sources trigger the disarming and timer interrupting, regardless of strength)

silicons:
bhijn fix it pls

PR #9354, opened 2019-09-25
[draft] Reworks magic (and holy) resistance into a % based modifier, build from the sum of the resistance sources on a player.

Thalpy:
### Reworks:

 anti_magic_check(), as it works currently, it checks for sources of magic resist, then returns a random one. there are 4-5 cases of this returned atom in the code, that is used for a "your object resists the spell!" message.

The way I've reworked it is; I've renamed it to is_magic_susceptible() to improve clarity. It checks the mob to see if it has any resistance items, calculates the sum additive percentage of all items on the mob, converts this percentage into a range of 0 - 2 (or more) where 0 is full resistance, 1 is standard, and over 1 is weakness. (i.e. this value can be easily applied to all current stuns, damages and effects with whatever number * is_magic_susceptible() So, for example

stun(200 * is_magic_susceptible()):
for 40% resistance stun(200 * 0.6) = stun(120)
for 150% resistance stun(200 * 1.5) = stun(300)

However, the src return was removed, and the 5 cases of resistance messages have been moved into the proccall itself.

### Old description (still applies)

**THIS NEEDS TESTING, I CAN'T TEST AWAY FROM HOME**

Everyone's a bit mad about melons, so I thought I'd offer a solution that hopefully makes everyone happy!

This hasn't been tested atm, since I'm currently on **Thesis lockdown** but it should serve as a proof of concept.

I'll be reworking how antimagic and antiholy work in the future anyways with tamiorgans into something that isn't a binary, and instead is a %age, but this should work as a placeholder until then. It's a little rough, but basically what it does is based off the potency of melon in the person's hands, it reduces the effects of the stun. At the moment it reduces every effect of the stun, but I'm open the discussion/suggestions. The range of reduction is 70% for a 100 potency melon, to none for a 0 potency melon. This means, even with the best melon, you will still be knockdowned() for just under 5s, which I think is the cooldown? I'm not sure.

After your melon protects you, then it loses 20 potency from itself. If the potency becomes negative, it'll buff the length of the stuns instead, and it's name will change to unholy melon to reflect that.

What I could do, if people like, is turn the holy water in the melon into unholy water! _Remember to comment below based on what you think and don't forget to like and subscribe!_

These numbers are open to peanut gallery tweaking too.

## Why It's Good For The Game

Hopefully makes both parties happy. No longer invincible to stuns, but also a useful item to have. Also potency matters!

## Changelog
:cl: Fermis
tweak: tweaked holy melons
/:cl:

### DIFF
--- a/code/datums/components/anti_magic.dm
+++ b/code/datums/components/anti_magic.dm
@@ -25,25 +25,18 @@
 
 /datum/component/anti_magic/proc/on_drop(datum/source, mob/user)
 	UnregisterSignal(user, COMSIG_MOB_RECEIVE_MAGIC)
-	if(magic["source"])
-		magic -= list("[parent]" = _magic)
-	if(holy["source"])
-		holy -= list("[parent]" = _holy)
+	if(magic["[source]"])
+		magic -= magic[parent]
+	if(holy["[source]"])
+		holy -= holy[parent]
 
 //Returns the percentage resist, if the spell is holy AND magic, and the item is holy AND magic, then it will use both of the object's resist boons.
 /datum/component/anti_magic/proc/can_protect(datum/source, _magic, _holy, list/protection_sources)
 	if(_magic && LAZYLEN(magic))
 		protection_sources += magic
 
-
 	if(_holy && LAZYLEN(holy))
-		var/temp_holy = holy
-
-		for(var/i in holy) //Merge duplicate entries into one single entry
-		    if(protection_sources[i])
-		        protection_sources[i] += holy[i]
-				temp_holy -= "[i]"
-		protection_sources += temp_holy
+		protection_sources += holy
 
 	if(!len(protection_sources))//If no sources then nothing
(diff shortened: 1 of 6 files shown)


Poojawa:
pretty sure this would just result in people pocket-carrying dozens of melons instead of just the one with a plant bag. The meta would continue to be total ass.

Thalpy:
> pretty sure this would just result in people pocket-carrying dozens of melons instead of just the one with a plant bag. The meta would continue to be total ass.

I do plan on making it so that multiple melons/sources of the same type aren't additive.

YakumoChen:
Make the proc affect every melon on the victim's person so they can't just switch to a fresh melon?

Does this make holy melons only viable for magic stuns or can you still use it to ignore any sort of magical/cult effect like wizard spells? The primary reason for my removal is its use in countering wizards, not just cult stuns.

PR #9369, opened 2019-09-28
Departmental ID Coloration

Tupinambis:
## About The Pull Request

Adds new ID card color variations, giving each department their own color of card highlights. (The Green parts of the standard ID may be light blue for medical, red for sec, etc.) The syndicate ID has a new black makeover by default.

## Why It's Good For The Game

It adds a small bit of flavor, and gives a visual indication of which department the card should allow you to access at a glance.

## Changelog
:cl:
imageadd: Adds some variety in coloration for IDs based on which department you are in.
/:cl:


Poojawa:
a neat idea you could do for this too, is if someone selects a specific job role, it colors their id to that department. so if a cargo tech wants to go to medical, it'd swap from the brown to blue.

dapnee:
Suggestions I’d make is changing clowns ID to pink possibly, since it looks similar to science. Mime would need one as well. Departmental colored IDs existed on Oracle, you could possibly look at the way we had coloring via the PDA machine in the HoP’s office so the HoP can, on the fly, change the style of IDs he hands out. Otherwise, I enjoy this concept.

Tupinambis:
clown ID is pink. Unfortunately I lack the knowledge to code an ID recolor system

YakumoChen:
Why not just port the new TG IDs?

Tupinambis:
Honestly this PR is just incomplete as hell and I lack the skills to fix it. As I don't want to shovel lazy and meh PRs, this shall be closed.

PR #9370, opened 2019-09-29
buffs c4 on mobs

silicons:
it should now instakill people without bomb armor and instacrit anyone else

reason: parapen/c4 days are long gone and i nerfed it too hard no one even uses this dumb thing let's make it slightly more useful even if there's easier ways to kill someone /shrug


GrayRachnid:
parapen c4

Poojawa:
ebow c4 new meta

BlackMajor (approved):

BlackMajor:
Approving this because if you get the chance to C4 someone chances are you also have a way to kill them without spending 1tc to do it.

PR #9375, opened 2019-10-01
Makes Ashwalkers unable to wear Masks

JTGSZ:
## About The Pull Request

Adds mask to ashwalker no_equip.
Doesn't let ashwalkers wear masks, their snouts r too big anyways theres ur excuse

## Why It's Good For The Game

Now they will never get onto the station and stay alive thanks to the lungs they were given.

## Changelog
:cl: JTGSZ
tweak: Makes Ashwalkers unable to wear masks
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -94,7 +94,8 @@
 	limbs_id = "lizard"
 	species_traits = list(MUTCOLORS,EYECOLOR,LIPS,DIGITIGRADE)
 	inherent_traits = list(TRAIT_NOGUNS)
-	mutantlungs = /obj/item/organ/lungs/ashwalker
+	mutantlungs = /obj/item/organ/lungs/ashwalker //Lungs that keep them from breathing station atmosphere.
+	no_equip = list(SLOT_WEAR_MASK) //Now asswalkers can't bypass internals issues preventing them from getting onto station.
 	burnmod = 0.9
 	brutemod = 0.9
 


Weblure:
Now invading ashwalkers will have a reason to grief the station by breaking windows and venting atmos in order to stay alive. I still think it's a better idea to just directly tell people in the spawn text to not leave lavaland in the first place.

Besides, there are valid RP reasons for putting masks on Ashwalkers without breaking the rules. This also prevents them from wearing face items unrelated to internals, such as muzzles. On top of that, you can just circumvent the lack of a mask by wearing a hardsuit.

Linzolle:
I don't think breaking the windows will allow ash walkers to survive on the station, ashies still take damage from low pressure and still need SOME oxygen to breathe but yeah this is... a bit far?

Fractious:
Actually a solid idea imo. And they don't survive in space for info. Giving them BOTH IC and OOC reasons to stay on lavaland is sincerely good, to be honest. Besides, I don't believe Ash walkers need to wear masks to survive on lavaland.

Weblure:
Sitting in a room with a broken window could create a survivable sweetspot of low oxygen and low pressure, which is all that lavaland pretty much is.

Linzolle:
> 
> 
> Sitting in a room with a broken window could create a survivable sweetspot of low oxygen and low pressure, which is all that lavaland pretty much is.

yeah maybe for like... a minute before all the air gets deleted by space

Linzolle:
also of note is that its possible to get your race randomised into an ashwalker by, say, a wizard so if someone got randomised into an ash walker they would just die of suffocation on the station.

deathride58:
A much better alternative to this PR would be to make it so that ashwalkers take high pressure damage in the station atmosphere. Making them unable to wear masks is an extremely lazy solution that stops them from wearing tribal masks among other issues. Ashwalkers on station is something that should be possible, but far from easy to accomplish. The reason why "no ashwalkers on station" is currently a rule is because it's currently way too easy to accomplish

JTGSZ:
Closing this because 2 to 4, and bhijn proposed a better solution.

PR #9391, opened 2019-10-02
Rebalances k9borg pounce/bite around the secborg's baton.

GrayRachnid:
## About The Pull Request
The K9's pounce fully stamcrits from just one pounce. This nerfs it to do about 50~ stamdamage and knockdown. Spoolup was buffed. Cooldown was buffed. Knockout time if you miss was buffed. Cell use was tweaked to match what the baton has. Bite now has a cell-use because a free 12 damage(baton does 10) is bullshit. 

## Why It's Good For The Game
K9 secborgs are overpowered compared to normal secborgs, and while normal secborgs can stunlock better, an instacrit pounce was bullshit.

## Changelog
:cl:
balance: rebalanced k9dogborgs
/:cl:

### DIFF
--- a/modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
+++ b/modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
@@ -4,7 +4,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 */
 
 /obj/item/dogborg/jaws
-	var/cellcost
+	var/cellcost = 0
 	name = "Dogborg jaws"
 	desc = "The jaws of the debug errors oh god."
 	icon = 'icons/mob/dogborg.dmi'
@@ -20,7 +20,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 	desc = "The jaws of the law. Very sharp."
 	icon_state = "jaws"
 	force = 12
-	cellcost = 1000 // Secborg can only harm with the baton, why does the K9 get a free module to do 12 damage with?
+	cellcost = 250
 	attack_verb = list("chomped", "bit", "ripped", "mauled", "enforced")
 
 
@@ -29,14 +29,17 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 	desc = "Rubberized teeth designed to protect accidental harm. Sharp enough for specialized tasks however."
 	icon_state = "smalljaws"
 	force = 6
-	cellcost = 500 // Less cell use for the smaller jaws.
 	attack_verb = list("nibbled", "bit", "gnawed", "chomped", "nommed")
 	var/status = 0
 
 /obj/item/dogborg/jaws/attack(atom/A, mob/living/silicon/robot/user)
 	..()
-	user.cell.use(cellcost)
-	user.do_attack_animation(A, ATTACK_EFFECT_BITE)
+	if(user.cell.charge <= cellcost)
+		to_chat(src,"<span class='danger'>Insufficent energy for jaws!</span>")
+		return
+	else
+		user.cell.use(cellcost)
+		user.do_attack_animation(A, ATTACK_EFFECT_BITE)
 
 /obj/item/dogborg/jaws/small/attack_self(mob/user)
 	var/mob/living/silicon/robot.R = user
@@ -352,7 +355,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 		pixel_y = 10
 		update_icons()
 		throw_at(A, MAX_K9_LEAP_DIST, 1, spin=0, diagonals_first = 1)
-		cell.use(750) //Same as stunbaton
+		cell.use(750) //Less than a stunbaton since stunbatons hit everytime.
 		weather_immunities -= "lava"
 
 /mob/living/silicon/robot/throw_impact(atom/A)
--- a/modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
+++ b/modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
@@ -4,6 +4,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 */
 
 /obj/item/dogborg/jaws
+	var/cellcost
 	name = "Dogborg jaws"
 	desc = "The jaws of the debug errors oh god."
 	icon = 'icons/mob/dogborg.dmi'
@@ -19,28 +20,23 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 	desc = "The jaws of the law. Very sharp."
 	icon_state = "jaws"
 	force = 12
+	cellcost = 1000 // Secborg can only harm with the baton, why does the K9 get a free module to do 12 damage with?
 	attack_verb = list("chomped", "bit", "ripped", "mauled", "enforced")
+
 
 /obj/item/dogborg/jaws/small
 	name = "puppy jaws"
 	desc = "Rubberized teeth designed to protect accidental harm. Sharp enough for specialized tasks however."
 	icon_state = "smalljaws"
 	force = 6
+	cellcost = 500 // Less cell use for the smaller jaws.
 	attack_verb = list("nibbled", "bit", "gnawed", "chomped", "nommed")
 	var/status = 0
 
-/obj/item/dogborg/jaws/small/attack(atom/A, mob/living/silicon/robot/user)
+/obj/item/dogborg/jaws/attack(atom/A, mob/living/silicon/robot/user)
 	..()
-	user.cell.use(500) // Less cell use for the smaller jaws.
+	user.cell.use(cellcost)
 	user.do_attack_animation(A, ATTACK_EFFECT_BITE)
-	log_combat(user, A, "bit")
-
-/obj/item/dogborg/jaws/big/attack(atom/A, mob/living/silicon/robot/user)
-	..()
-	user.cell.use(1000) // Secborg can only harm with the baton, why does the K9 get a free module to do 12 damage with?
-	user.do_attack_animation(A, ATTACK_EFFECT_BITE)
-	log_combat(user, A, "bit")
-
 
 /obj/item/dogborg/jaws/small/attack_self(mob/user)
 	var/mob/living/silicon/robot.R = user
@@ -316,7 +312,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 	var/leaping = 0
 	var/pounce_cooldown = 0
 	var/pounce_cooldown_time = 20 //Buffed to counter balance changes
-	var/pounce_spoolup = 2
+	var/pounce_spoolup = 1
 	var/leap_at
 	var/disabler
 	var/laser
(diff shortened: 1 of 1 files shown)


GrayRachnid:
This was tested, just in case.

Strazyplus:
This is not a balance change at all. every aspect has been completely negatively impacted even battery cost which is a bit extreme, as I sit on the toilet recharger enough already.

 if you are wanting to make borgs pounce comparative to a stun baton then make it a stun baton. 

The pounce is not a stun baton. it has a spool up and cool down. the dog is committing to a self-throw that if missed can stun itself. I feel it should be more stun than a stun baton because it is a dog jumping on you after all.

 Suggestion: At the start of the spoolup the borg makes a motor sound as it gets ready to pounce. would be a better addition than making ability and borg more weak.

GrayRachnid:
Your point is moot because a secborg using a baton is also risking itself to get flashed. I don't mind buffing the cool down or the missing stun if necessary but it needs to use up the same charge as a stun baton for starters. I also don't think a spoolup is necessary with these changes so we can remove it in exchange for similar knockdown of a stun baton. Also, if I wanted to replace this with a stun baton then I would've but that would be 'removing not improving.'

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
```diff
@@ -348,7 +356,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 		pixel_y = 10
 		update_icons()
 		throw_at(A, MAX_K9_LEAP_DIST, 1, spin=0, diagonals_first = 1)
-		cell.use(500) //Doubled the energy consumption
+		cell.use(1000) //Same as stunbaton
```
Weblure: I'd keep this where it's at, since your changes already require multiple pounces rather than just one. Increasing this to 1000 results in a net power consumption increase of at least 4x when downing a single target; realistically, this number is bound to be even higher, since the target is now capable of potentially dodging subsequent pounces. 

In short, I see no reason to increase the energy usage. If anything, it should be decreased. It's not like stun batons use charge every time you miss.

GrayRachnid: The secborg's stun baton uses 1000 per hit and requires 4 batons to stuncrit. This will make it similar.

Strazyplus: Why does it need to be similar when they are not the same thing? also 4x. after about 9 pounces i'd have a dead battery if I am calculating this right. regardless of most other things consuming power.

I don't like this. I would make pounce absolute not use if I wanna keep a charge in a fight.

Weblure: That's a fair point. I suppose we'll need to see how these changes compare to the stunbaton in actual gameplay.

GrayRachnid: A secborg also has a dead battery by 9 batons. It is very much the same thing. If it wasn't then it wouldn't stun the same way a baton does. It's just different in that it has a range and used to do a full stamcrit before.

Linzolle: yeah but you can't miss with a stun baton and still have it use power, it'll only use power when it successfully lands a hit

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
```diff
@@ -29,11 +29,19 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
...
-/obj/item/dogborg/jaws/attack(atom/A, mob/living/silicon/robot/user)
+/obj/item/dogborg/jaws/small/attack(atom/A, mob/living/silicon/robot/user)
 	..()
+	user.cell.use(500) // Less cell use for the smaller jaws.
 	user.do_attack_animation(A, ATTACK_EFFECT_BITE)
 	log_combat(user, A, "bit")
 
+/obj/item/dogborg/jaws/big/attack(atom/A, mob/living/silicon/robot/user)
+	..()
+	user.cell.use(1000) // Secborg can only harm with the baton, why does the K9 get a free module to do 12 damage with?
```
Ghommie: I suggest you add a `cell_cost` variable to `/obj/item/dogborg/jaws` and replace these code changes with the below one instead, less copypasting.
```
/obj/item/dogborg/jaws/attack(atom/A, mob/living/silicon/robot/user)
	..()
	user.cell.use(cell_cost)
	user.do_attack_animation(A, ATTACK_EFFECT_BITE)
```
Just remember to state the two subtypes' cell_cost values of 1000 and 500 respectively.

Also remove the `log_combat()` from here, since every item attack() should be logged anyway this pretty redunant.

YakumoChen:
Full stuns are bad, stun nerf good

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
GrayRachnid: :ok_hand:

GrayRachnid: :ok_hand:

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
GrayRachnid: That's fair, i'll adjust power usage down to 750. I'll also lower the spool up time and lower the knockout time if they miss.

Poojawa:
Why are you comparing jaws to a stunbaton? 

More over, why is it so horrifically fucking expensive?

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
```diff
@@ -19,20 +20,23 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
...
 	force = 6
+	cellcost = 500 // Less cell use for the smaller jaws.
 	attack_verb = list("nibbled", "bit", "gnawed", "chomped", "nommed")
 	var/status = 0
 
 /obj/item/dogborg/jaws/attack(atom/A, mob/living/silicon/robot/user)
 	..()
+	user.cell.use(cellcost)
 	user.do_attack_animation(A, ATTACK_EFFECT_BITE)
-	log_combat(user, A, "bit")
```
Poojawa, line 35: This needs to be readded

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
```diff
@@ -19,20 +20,23 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
...
 	desc = "Rubberized teeth designed to protect accidental harm. Sharp enough for specialized tasks however."
 	icon_state = "smalljaws"
 	force = 6
+	cellcost = 500 // Less cell use for the smaller jaws.
 	attack_verb = list("nibbled", "bit", "gnawed", "chomped", "nommed")
 	var/status = 0
 
 /obj/item/dogborg/jaws/attack(atom/A, mob/living/silicon/robot/user)
 	..()
+	user.cell.use(cellcost)
```
Poojawa: This needs a threshold safeguard so a borg can't put itself into negative cell charge. 

Why it needs to chunk 1,000 per 12 damage bite is amazingly biased bullshit of a balance change.  jumping is cheaper. Biting doesn't stun like a harm baton does either. so wtf?

Reminder that cell charge is quite literally a borg's ability to continue functioning. if it's 0'd out,  bad shit happens. It doesn't come back like stamina does.

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
```diff
@@ -338,7 +342,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 		//It's also extremely buggy visually, so it's balance+bugfix
 		return
 
-	if(cell.charge <= 500)
+	if(cell.charge <= 750)
```
Poojawa, line 341: this ought to be raised to a minimum of 1000 to avoid 0'd power draw.

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
```diff
@@ -348,7 +352,7 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 		pixel_y = 10
 		update_icons()
 		throw_at(A, MAX_K9_LEAP_DIST, 1, spin=0, diagonals_first = 1)
-		cell.use(500) //Doubled the energy consumption
+		cell.use(750) //Same as stunbaton
```
Poojawa: Batons use 1000 charge per use, from what I can see the * 0.3 multiplier is just to turn it off

It also doesn't use hard power from the borg itself, just the baton item.

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
```diff
@@ -19,20 +20,23 @@ SLEEPER CODE IS IN game/objects/items/devices/dogborg_sleeper.dm !
 	desc = "The jaws of the law. Very sharp."
 	icon_state = "jaws"
 	force = 12
+	cellcost = 1000 // Secborg can only harm with the baton, why does the K9 get a free module to do 12 damage with?
```
Poojawa: Because sec borgs also stunlock every godamn hit. biting has no stun.

Poojawa (changes requested):

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
GrayRachnid: Ghommie told me it can be removed because all of attacks are automatically logged?

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
GrayRachnid: This was lowered to 750 from 1000 in my latest commit, I forgot to change the comments.

deathride58:
>Increasing the cell cost of a 12 force attack from 18.75 to 1018.75

W H Y  though

GrayRachnid:
Addressed your comments.

>Why are you comparing jaws to a stunbaton?

Secborgs have nothing to harm with other than the stunbaton, the stunbaton does 10 damage while jaws do 12.

GrayRachnid:
I just realized i'm stupid. You can use a baton while it's off. Lowering jaws damage down to 10 instead.

modular_citadel/code/modules/mob/living/silicon/robot/dogborg_equipment.dm
Ghommie: Yea, `log_combat()` is already called on `/obj/item/attack()`

silicons (approved):

Strazyplus:
Really annoying this is got merged when it's changes were so broad... I am not sure if it's live on the server or not but last night the pounce responded like shit (especially when the server is lagging out) half of the time in combat the pounce just does not pounce. or randomly pounces incorrectly throwing me in the wrong direction.  feels like through frantic spamming of pounce button and luck gets me to pounce a moving target. it frustrating 

also why are you comparing jaws to a stunbaton? ffs

Jaws do not have a stun function, they are a borg melee weapon.  jaws with teeth not a fucking stick

PR #9411, opened 2019-10-04
Nerfs portable crew monitors - No more pocket/deep inventory usage

silicons:
## About The Pull Request

Crew monitors can no longer be used from within pockets or other forms of inventory/storage other than hands or it being in the game world adjacent or ontop of your tile, even with TK.

## Why It's Good For The Game

Both makes thematic sense (even though telekenesis doesn't obey the laws of physics yet with anything else) and a good way to rebalance the whole issue of 24/7 pocket crew monitors.

## Changelog
:cl:
balance: Portable crew monitors are no longer able to be used (see: automatically updating) from within pockets, inventory storage, or really, anything but your hands and by being next to you and in view.
/:cl:


code/modules/tgui/states/non_concealed_update.dm
```diff
@@ -0,0 +1,14 @@
+ /**
+  * tgui state: non_concealed_update
+  *
+  * Checks that the src_object is in the user's hands or adjacent, or block view updates entirely.
+ **/
+
+GLOBAL_DATUM_INIT(non_concealed_update /datum/ui_state/non_concealed_update, new)
```
Ghommie:
```suggestion
GLOBAL_DATUM_INIT(non_concealed_update, /datum/ui_state/non_concealed_update, new)
```
A missing comma that made it not compile.

Poojawa:
cool, a medical nerf. just what medbay needs

IronEleven:
> cool, a medical nerf. just what medbay needs

Blame sec powergaming with monitors I guess.

Weblure:
I'd suggest keeping the monitor visible and usable from your belt slot, since you can see/use your PDA from your ID slot. Neither of these break reality, since they're both just hanging from your waist and aren't visibly or physically obstructed. This keeps constant crew-monitoring viable for medics at the trade-off of not wearing a belt, which is a much larger trade-off for security, effectively limiting its usage as a power-gaming tool.

Also, every time you open the crew monitor window, you have to resize, reposition, and scroll to get it back to where it was. You have to do this every time you remove the monitor from your hands (currently excluding pockets/belt slot). This is incredibly annoying for EMTs, who have to juggle the monitor, tracker, and medical equipment/supplies every time they track down and heal an injured patient. I've only recently discovered the blessing of being able to use the monitor from your belt slot; I'll certainly be sad to see this feature (intentional or not) removed entirely.

Poojawa:
What it should do is stay up if you have a medical hud at the same time, tbh.

silicons:
i'll make it stay up if you have a medical hud and it's in pockets, or without one if it's in belt, inventory nesting was stupid.

nik707 (changes requested):
do the things you said you would

davidchan13:
TBH just add it as a function to the MedHUD.

PR #9413, opened 2019-10-04
Makes Medsprays align with patches

Poojawa:
Also ensures you can't spray it on ghosts, ai eyes, etc.

## About The Pull Request

Makes med sprays, which were more or less patch code when they were added, perform the same as patches do now. Yes, I double checked to make sure I was using Ghommie's fix too.

## Why It's Good For The Game

It doesn't make sense for a sealed space suit to receive medical attention when syringes, patches, cryo, and most other healing doesn't either. 

## Changelog
:cl:
balance: Med Sprays are now more aligned with patch mechanics
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/medspray.dm
+++ b/code/modules/reagents/reagent_containers/medspray.dm
@@ -19,8 +19,8 @@
 	var/can_fill_from_container = TRUE
 	var/apply_type = PATCH
 	var/apply_method = "spray"
-	var/self_delay = 30
-	var/squirt_mode = 0
+	var/self_delay = 3 SECONDS
+	var/squirt_mode = FALSE
 	var/squirt_amount = 5
 
 /obj/item/reagent_containers/medspray/attack_self(mob/user)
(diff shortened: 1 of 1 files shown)


Weblure:
I'd like to see syringes given the ability to inject through suits as a trade-off for this nerf. That'd give medics a reason to actually ever use syringes for something other than the hecking syringe gun. Hardsuits should still have soft areas around the joints where a needle would penetrate if inserted directly and deliberately, and thus it wouldn't be reality-breaking.

Poojawa:
> I'd like to see syringes given the ability to inject through suits as a trade-off for this nerf. That'd give medics a reason actually ever use syringes for something other than the hecking syringe gun. Hardsuits should still have soft areas around the joints where a needle would penetrate if inserted directly and deliberately, and thus it wouldn't be reality-breaking.

They already do if you print the armor pierce ones

Weblure:
> > I'd like to see syringes given the ability to inject through suits as a trade-off for this nerf. That'd give medics a reason actually ever use syringes for something other than the hecking syringe gun. Hardsuits should still have soft areas around the joints where a needle would penetrate if inserted directly and deliberately, and thus it wouldn't be reality-breaking.
> 
> They already do if you print the armor pierce ones

Oh yeah, I forgot those exist. This looks like a good PR, overall.

Poojawa:
its easy to forget when "better" stuff exists. :P 

As misunderstood as my balance ideas can be, with this fix I've literally brought AP syringes back into a reason for existing other than chem warfare.

silicons (approved):

PR #9419, opened 2019-10-05
[dnm this is just to remind myself to experiment with this] makes wall conduct atmospheric heat

silicons:
uhhhhhhh i don't know what i'm doing but someone reminded me this exists when they asked about heat exchange pipes


Poojawa:
Hi

I've done a few tests with this already on at least 0.4 lol

The problems I encountered were that Doors do not conduct heat. Windows do not allow heat to flow through them, and this actually causes raging fires to put themselves out despite still having ample amounts of fuel to burn, because with heat scaling requirments to continue burning.

PR #9428, opened 2019-10-07
pAI ID Hosting

Brewstone:
## About The Pull Request
Adds in a new pAI software, ID reader, which allows the pAI to host an ID card and use its access. Adds in some additional functionality to the pAI and pAI card so that the hosted ID can be dropped by the pAI (menu) or by anyone who can grab the pAI card (alt click).

To make any of this work, I had to remove the hard coded global no access for pAI. Since they're still under the silicon mob type, I changed any access checks to check their inherent ID object. Removing the ID card from the pAI, through either method, resets the inherent ID value to null.

## Why It's Good For The Game
At the moment, pAI are rarely if ever used and have almost no game impact. I think that it'd be more enjoyable for pAI to have more potential game impact and more independence from their masters. Adding the ability for pAI to have their own ID lets go out on their own (if given an ID), and adds some interesting opportunities for Antags and the station. However, needing their own ID means that some effort needs to be invested to allow the pAI to go out on their own, instead of cloning the ID which could open the door for abuse.

There are potential other additions, such as radio encryption key usage, that could be added as well. However, I don't want to overload pAI with options and abilities and would rather add them individually.

## Changelog
:cl:
Add: Adds in a new pAI software, ID Reader
Add: ID Reader functionality lets the pAI host an ID card and use that same ID card's access
Add: Alt clicking a pAICard now removes a hosted ID card
Add: pAI can drop their hosted ID via the ID Reader menu
Tweak: pAI no longer see the silicon text when examining doors ("Shift-Click to open", etc.)
Del: Removes the hardcoded no access for pAI
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -675,7 +675,8 @@
 		else
 			to_chat(user, "It looks very robust.")
 
-	if(issilicon(user) && (!stat & BROKEN))
+//No longer gives pAI the borg/AI text about manipulating doors
+	if(issilicon(user) && (!stat & BROKEN) && !ispAI(user))
 		to_chat(user, "<span class='notice'>Shift-click [src] to [ density ? "open" : "close"] it.</span>")
 		to_chat(user, "<span class='notice'>Ctrl-click [src] to [ locked ? "raise" : "drop"] its bolts.</span>")
 		to_chat(user, "<span class='notice'>Alt-click [src] to [ secondsElectrified ? "un-electrify" : "permanently electrify"] it.</span>")
--- a/code/game/objects/items/devices/paicard.dm
+++ b/code/game/objects/items/devices/paicard.dm
@@ -9,6 +9,8 @@
 	slot_flags = ITEM_SLOT_BELT
 	var/mob/living/silicon/pai/pai
 	resistance_flags = FIRE_PROOF | ACID_PROOF | INDESTRUCTIBLE
+		//Let's the pAI card be used as an ID in hand or thrown (Still sorting out undefined var error when nulling this)
+	//var/obj/item/card/id/paiCard_ID = null
 
 /obj/item/paicard/suicide_act(mob/living/user)
 	user.visible_message("<span class='suicide'>[user] is staring sadly at [src]! [user.p_they()] can't keep living without real human intimacy!</span>")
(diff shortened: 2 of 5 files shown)


Poojawa:
tbh, just make the door hacking utility have an upgrade that makes it much faster.

Brewstone:
You'd still have to have another person to plug the cable in, keeping the same problem of you being at the mercy of whoever's around. 

Plus, I don't want people rushing pAI for ghetto all access if the hack speed is actually decent.

Poojawa:
a PAI is supposed to serve its master. it'd still need a ghost player to be functional anyway

They're not supposed to be lawless drones

Brewstone:
True, but often enough pAI aren't actually bound by their summoner or are given a law which is basically "Do whatever", so having the potential for more freedom would be nice. 

And since someone still has to put the ID into the pAI, you're still reliant on someone to get you started.

silicons:
@Brewstone 
consider: pAIs are not SUPPOSED to have much of an in game impact.


is there any reason to give them more considering they are so cheaply printed, so annoying to kill unless you get it when it's in range of you, etc etc?

Brewstone:
They're cheaply printed, but still require a ghost to actually submit a personality. They're hard to kill (no stamina/damage states), but since they're Silicon, a flash still disables them. They still don't have hands and as such can't interact with most objects. They also can't pull anything or carry anything.  And, if it comes to it, with the card you can just wipe them.  So currently, as far as game impact goes, pAI can talk to people, hold open doors (by standing on the tile), and push objects, while still having all of those downsides.

With the addition of the ID software, they can also bump open doors (they can't long open it by clicking) and maybe help in a quick getaway if they pre-open doors. But before any of that can happen, they need to be given an ID. Compared to other ghost roles (lavaland vets, oldstation crew, ashwalkers, etc.), a pAI's ability to impact the game, even with this change, is minimal. 

However, when it comes to convenience for the pAI player, being able to get an ID would make life much easier. It's less likely you're trapped somewhere you can't escape from, you could explore on your own, and it could open up further options for things to do with your master. Beyond helping antags, you could have a departmental pAI moving within the department providing guidance, a medical pAI searching maint for injured people, a cargo pAI helping to move freight, etc.. This is all currently possible, but requires non-pAI input. 

And nothing about the change says that pAI can't stay with you. Your pAI takes the ID software, and wants to fuck off, but you want it with you? Don't give it an ID. It would have to find someone else to beg one from, but if you're its bound master, and you tell it not to, then its violating its laws anyway.

silicons:
i don't think we need to buff pais at this moment, they're already strong due to being a second radio, they don't need to be able to run around too.

PR #9442, opened 2019-10-08
Reverts jetpack efficiency from #9307

YakumoChen:
## About The Pull Request

Reverts the rate that jetpacks consumed air changed in #9307

## Why It's Good For The Game

I agree wholly with the nerf to the speed bonus jetpacks gave. I did it myself on the upstream TG.  It was even much harsher because space combat is pretty cancerous.

However I disagree wholeheartedly with making jetpack air last longer. 
You can move around space for an extremely long time with a jetpack with the current rate, making the need to replace or refill your air almost pointless.
Space is big, empty, and there should definitely be a resource cost aside from time to exploring it or staying in it for an extended amount of time. Furthermore, Citadel has a very small amount of z levels containing space in the first place, jamming most of the ruins in extremely close proximity. There's little need to increase the time you can use a jetpack without a refill aside from powercreep, and it means anyone with ill intent can fly around space basically forever without needing to keep a can of air to refill at.

## Changelog
:cl: YakumoChen
tweak: Jetpacks no longer last twice as long between air refills.
/:cl:


YakumoChen:
oh it doesn't merge. ok but why

silicons (approved):

PR #9446, opened 2019-10-08
abscond can no longer be cast from inside a locker

Sishen1542:
## About The Pull Request
title

## Why It's Good For The Game
bugfix

## Changelog
:cl:
fix: teleporting a locker is bad
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_scriptures/scripture_drivers.dm
+++ b/code/modules/antagonists/clockcult/clock_scriptures/scripture_drivers.dm
@@ -216,7 +216,7 @@
 	if(is_reebe(invoker.z))
 		to_chat(invoker, "<span class='danger'>You're already at Reebe.</span>")
 		return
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/clockcult/clock_scriptures/scripture_drivers.dm
```diff
@@ -216,6 +216,9 @@
 	if(is_reebe(invoker.z))
 		to_chat(invoker, "<span class='danger'>You're already at Reebe.</span>")
 		return
+	if(invoker.z == 0)
```
Ghommie:
```suggestion
	if(!isturf(invoker.loc))
```
I never recalled z, x and y coordinates being affected by containers in such way.
Regardless, this should be in line with the other hundreds of "must be on a turf" sort of checks.

silicons (approved):

PR #9453, opened 2019-10-09
Mitigates the ghost roles/mid-round antag lock out for suicide/cryo. And related fixes.

Ghommie:
## About The Pull Request
I'll start off saying that I don't share the current philosophy of cryo/suicide equalling to a forfeit of any priviledge of playing the current round, and specially when the average round tend to go well past the one hour mark (real time) duration, it tells people to heck off and take on a different activity (or server), whereby antag fishers could do so already (or just don't play here altogether, since rounds can be slow and boring).

This PR changes said design to allow people to come back to the round after a given time, currently set to 30 minutes (more if they happened to quit the game very early in the round).
Not a perfect design either, it still stops people from playing the more innocuous ghost roles and potentially disruptive alive.
It does fix most round comeback exploits and outer sources of ghostize(0) locking people out, and is generally more honest to those players who had to quit a round due to legit reasons only to come back almost a hour later and be left spectating for almost as much time.

## Why It's Good For The Game
Tweaking one aspect of this game, and fixing several of its issues.

## Changelog
:cl:
tweak: Suicide and cryo now prevents ghost/midround roles for a definite duration of 30 minutes (and more if that was done before 30 minutes in the game passed), down from the rest of the round.
fix: fixed several midround roles bypassing and nulling the aforementioned prevention measures.
/:cl:

I was indeed excessively verbose, but this could as well be a hot potato for some, so go ahead and criticize if you wish to.


deathride58 (changes requested):
Make the timer a config option, with -1 disabling the system completely.

code/modules/awaymissions/corpse.dm
```diff
@@ -43,8 +43,8 @@
 		return
 	if(isobserver(user))
 		var/mob/dead/observer/O = user
-		if(!O.can_reenter_round)
-			to_chat(user, "<span class='warning'>You are unable to reenter the round.</span>")
+		if(O.reenter_round_timeout > world.realtime)
+			to_chat(user, "<span class='warning'>You are unable to reenter the round yet. Your ghost role blacklist will expire in [round((O.reenter_round_timeout - world.realtime)/600)] minutes.</span>")
```
deathride58: If this is more than 4 hours (or if we have a config option that allows changing the total round time, whatever that is multiplied by 2), then the timer shouldn't be displayed. Also displaytimetext is a thing

deathride58 (dismissed):

silicons (approved):

PR #9466, opened 2019-10-11
Adds a config option to determine the minimum alert level for secborgs (defaulting to red alert)

deathride58:
## About The Pull Request
Title. Does exactly as it says on the tin. Sechounds are included under "secborg", and the game will give a notice if the conditions for secborgs aren't met.

## Why It's Good For The Game
**The alternative to this PR is to take TG's route of just removing secborgs entirely.** This PR allows secborgs to still exist, but only when shit is hitting the fan aboard the station. The primary issue with security borgs is that they're designed primarily for validhunting, which directly conflicts with the third party nature that silicons are supposed to have. This also makes it much easier to handle from an administrator standpoint, because we won't have to repeat "Secborgs obey their laws first, not space law" like a broken record multiple rounds in a row, and most cases where red alert is raised are cases where a borg would side with the crew regardless. This also has interesting implications for slaved borgs if the AI is malf!

## Changelog
:cl: Bhijn
tweak: Security borgs and K9s are now only available during red alert or higher.
server: Headmins or other folks with access to the server's config can choose the minimum alert level for secborgs to be chosen via the MINIMUM_SECBORG_ALERT config option. See the default game_options.txt for more info.
/:cl:

### DIFF
--- a/code/__DEFINES/robots.dm
+++ b/code/__DEFINES/robots.dm
@@ -50,3 +50,7 @@
 #define ASSEMBLY_THIRD_STEP     2
 #define ASSEMBLY_FOURTH_STEP    3
 #define ASSEMBLY_FIFTH_STEP     4
+
+
+//Checks to determine borg availability depending on the server's config. These are defines in the interest of reducing copypasta
+#define BORG_SEC_AVAILABLE (!CONFIG_GET(flag/disable_secborg) && GLOB.security_level >= CONFIG_GET(number/minimum_secborg_alert))
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -70,6 +70,9 @@
 
 /datum/config_entry/flag/disable_peaceborg
 
+/datum/config_entry/number/minimum_secborg_alert	//Minimum alert level for secborgs to be chosen.
+	config_entry_value = 3
+
 /datum/config_entry/number/traitor_scaling_coeff	//how much does the amount of players get divided by to determine traitors
 	config_entry_value = 6
 	min_val = 1
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -226,6 +226,9 @@
 		to_chat(src,"<span class='userdanger'>ERROR: Module installer reply timeout. Please check internal connections.</span>")
 		return
 
+	if(!CONFIG_GET(flag/disable_secborg) && GLOB.security_level < CONFIG_GET(number/minimum_secborg_alert))
+		to_chat(src, "<span class='notice'>NOTICE: Due to local station regulations, the security cyborg module and its variants are only available during [num2seclevel(CONFIG_GET(number/minimum_secborg_alert))] alert and greater.</span>")
+
 	var/list/modulelist = list("Standard" = /obj/item/robot_module/standard, \
 	"Engineering" = /obj/item/robot_module/engineering, \
 	"Medical" = /obj/item/robot_module/medical, \
@@ -234,7 +237,7 @@
 	"Service" = /obj/item/robot_module/butler)
 	if(!CONFIG_GET(flag/disable_peaceborg))
 		modulelist["Peacekeeper"] = /obj/item/robot_module/peacekeeper
-	if(!CONFIG_GET(flag/disable_secborg))
+	if(BORG_SEC_AVAILABLE)
 		modulelist["Security"] = /obj/item/robot_module/security
 
 	modulelist += get_cit_modules() //Citadel change - adds Citadel's borg modules.
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -267,6 +267,10 @@ ALLOW_AI_MULTICAM
 ## Uncomment to prevent the security cyborg module from being chosen
 #DISABLE_SECBORG
 
+## Determines the minimum alert level for the security cyborg model to be chosen
+## 0: Green, 1:Blue, 2:Amber, 3:Red, 4:Delta
+MINIMUM_SECBORG_ALERT 3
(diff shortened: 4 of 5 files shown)


ZeroNetAlpha:
Please yes, this was a massive pain in the ass when I was admin, imagine it still is.

Poojawa:
This essentially just removes them from the game. Hard no.

Do your job as administrators and job ban secborgs that aren't following server rules.

deathride58:
The alternative to this PR is that I instead do my job as headmin and disable secborgs entirely with no way for players to access them outside of adminbus, the same method that TG opted for back when it was actually good. Secborgs are a legitimate detriment to the quality of the gameplay experience during the average round, because as I've stated, their design being centered around validhunting directly conflicts with the role that silicons are supposed to play.

Head mains seem to have a fairly good grasp on when red alert is appropriate, and it's really not uncommon that you'll see the alert level get bumped up to red when things start going south aboard the station. By tying the availability of secborgs to what players already use to flag a round as having gone south, it both gives players an actual mechanical reason to make use of the alert level system when the station's in shambles, and resolves the majority of issues that secborgs cause by drastically reducing the likelihood of their availability during situations that attract problematic behavior from secborg players.

GrayRachnid:
yes please

Poojawa:
> Well if you don't like them being technically disabled I can just fully disable them


Sound logic, there. Why are silicon bans not a thing anymore? Did job bans stop being a method of policy enforcement?

NicoDeimos:
Strongly disagree with hard coding limitations like this for an issue that can easily be covered by admin action and actually enforcing rules. Validhunting is a playerbase issue that comes up beyond just borg players, but if it's an issue that has been endemic of secborg **players** in particular then it's something that we can readily hand out silicon bans for.
I would argue that with antags outside of crew roles and in some cases even within, a borg designed to detain threats to crew given the emphasis on having the crewsimov lawset seems perfectly reasonable. Rather than limit choices and punish players who actually follow the rules, I urge cracking down on the bad ones spoiling the bunch.

GrayRachnid:
> Strongly disagree with hard coding limitations like this

Except it's a config option that can be changed by the headmins at anytime depending on the changing environment. If Red alert is too much then it can be moved down to amber. It's a topic of discussion that can be brought up but either ways it's not a bad thing that secborgs and k9s would be disabled on default Green and Blue alert levels.

BlackMajor:
Why tho.

Borgs can set red alert by themselves, for starters. So unless you remove borgs being able to use auth devices (plz dont solo head and solo borg setting red together was nice) then there's gonna be a few roundstart red alert toggles.

Secondly uhhhh..
Gonna be honest I havn't seen a whole lot of issues of sec borgs focussing on space law first, and when they do we slap their asses backwards over it.

And in the cases where I do see borgs focussing on muh valids (i.e. antag calmly law 2's a borg for something antag related that doesn't breach their laws so they scream ;HELP SEC ANTAG) it's been on a whole swathe of variety of borgs. Secborg or not. (mostly medihounds tbh but that's just personal experience).

What I mostly see is like.. borgs just being stupid, or blissfully unaware. Y'know new players who turn around and ask "Where is telecomms?" etc. Folk who havn't come to understand the different playstyle necessary to be a good borg yet.

Ergo the common complaint folk have of "All borgs are retarded".

Removing sec borgs wont stop them being retarded.


P.s. In the scenario where this were to go through, and borgs couldn't use auth devices and secborgs were red alert only. The majority of circumstances that necessitate a red alert, a sec borg is shit outta luck for being able to influence much at that point, considering all they have going for them is a baton and disabler by default for whatever may have caused the red alert.

i.e. they'd probably have to take a buff if they were red alert only to make them worthwhile over an engiborg.

silicons:
i'm for this, it's not just about banning people who break silicon laws as much of a balance matter at this point.
i'm entirely against secborgs being in game default roundstart.

Soranayo:
I agree with Major on that point, if sec borgs get limited to red alert, they will need a buff, because that'll generally means the station is going down the shitter or is already in it, and at that point sec borgs won't be able to do much about it because the threats will either already have the weapons to deal with them easily, or they will lack any real way to stop it, disablers won't cut it when most of the antags or sec can easily disable you and make you totally useless. If this goes through without changes to sec borgs, there'll be no point in picking them in red alert situations over other modules, in my opinion.

silicons (approved):

InquisitiveEpidemic:
I fully disagree with limiting borgs to red/amber alert. 
First: and foremost it effectively removes sec borgs from anything but nuke ops due to needing to get reset but only when shit has already hit the fan.  Or in the case of Malf AI/subversion outs the AI/subversion.  (this will also change the average alert level for CWC where amber/red will be less frequent to prevent CWC sec borgs)

Second it means the only useful borgs will be medical and engineering, as janitor only helps when demons are out, and the rest are borderline or outright useless (or in the case of slime people harmful) in any non RP scenario. 

Third: it locks a player's core options for playing their role behind other player's actions, and this will be abused to meta grudge and meta game especially in modes where a Malf AI or subversion is likely.  Expect to see a huge jump in IAA rounds that never go above blue alert.  The best comparison I can come up with is locking scientists out of xenobio or circuits until say a point threshold is reached.

Fourth: This doesn't address the problem the PR focuses on fixing, validhunting, as it's a player issue and not to do with the borgs, Poojawa, Nico, and Black Major are completely right Admins have to deal with people who break the rules, not remove what was used to break the rules.

silicons:
tl;dr because i'm too tired to type out mutliple paragraphs here is that i've already talked about why i dislike secborgs, and it's not because of the validhunting stuff, many times, if anyone wants me to reiterate my reasoning for the merge i'm available in discord on dev discussion and i will paste it here if people want.

Poojawa:
joy, forced removal of sec borgs.

PR #9471, opened 2019-10-12
[s] fixes some exploits

silicons:


deathride58 (approved):
speedtestmerging because these exploits are fairly major

PR #9478, opened 2019-10-13
Makes pod people not constantly slightly fat

Putnam3145:
## About The Pull Request

Pod people would be set to NUTRITION_LEVEL_FULL if above it. Unfortunately, fatness starts *at* NUTRITION_LEVEL_FULL, not above it. This makes it so that they never quite reach fatness.

## Why It's Good For The Game

Pod people shouldn't get a negative moodlet for existing.

## Changelog
:cl:
tweak: Pod people can no longer get fat from standing in the light
/:cl:


Poojawa:
Have they tried the radical weight loss trick of complete darkness?

deathride58 (approved):

PR #9485, opened 2019-10-13
actually no better reduces vampire blooddrain hunger

silicons:


code/modules/mob/living/carbon/human/species_types/vampire.dm
```diff
@@ -45,7 +45,7 @@
 		C.adjustOxyLoss(-4)
 		C.adjustCloneLoss(-4)
 		return
-	C.blood_volume -= 0.75
+	C.blood_volume -= 0.15
```
Ghommie: Under the current drain rate, it'll take roughly 19 (excluding server lag) minutes for a fresh 100% blood volume vampire to die. With this change it'd take a whooping 1 hour and 37 minutes.

Vampires bleed from traumas or reagents etcetera just like they replenish through standard methods such as IV bags, iron or bloody mary, and that's fine.

The real deal is how normal low blood symptoms still affect vampires, which makes this a even more slighty irritating aspect of this species.

I'm cool with reducing this rate to a slighty more sustainable value up to 0.5-0.4 (drink iron and forget), but not making it so banal just because of whatever reason you did not provide in your PR description.

silicons:
yeah alright changed to 0.45 down from 0.75 and up from 0.15
reason is because vampire sucks enough with the whole bullet drain blood and bleeding and dust on losing too much blood without you dying in 20 minutes and I've noticed vampires had to feed VERY often, infact often enough that they can't top off with just one person.

i know it's part of the point to be able to feed on blood but this isn't an antagonist race, you shouldn't be forced to feed so often on people which will debuff them.

PR #9500, opened 2019-10-14
teleprods works on non-carbons too now.

Ghommie:
## About The Pull Request
Exactly what's said on the tin.

## Why It's Good For The Game
Fixing a minor issue.

## Changelog
:cl:
fix: Teleprods work on non-carbons mobs now.
/:cl:

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -1,5 +1,4 @@
 #define STUNBATON_CHARGE_LENIENCY 0.3
-#define STUNBATON_DEPLETION_RATE 0.006
 
 /obj/item/melee/baton
 	name = "stunbaton"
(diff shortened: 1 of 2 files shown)


silicons:
i don't like this because we shouldn't encourage taking another stunbaton even if it seems like people are doing it all the time now sigh

Ghommie:
The reason sec officers always carry a second stunbaton is because they start out with one in their backpack as per outfit loadout, on top of the loaded secbelts found in secoff closets. And stamcombat.
I wasn't aiming to penalize people who carry one, rather those who leave it on all time.

This is only a mild change, but stunbatons are, like, the very core of the security department and a good half of stuncombat, it's cheesy.

silicons:
ehh put the stunbaton nerf in a separate pr because you know how much people yell about it

silicons (approved):

PR #9477, opened 2019-10-12
[READY] Fixes a few organ code and lets organs freeze based on ambient temperature.

Thalpy:
## About The Pull Request

Fixes; Heart, Tongue and stomach regen.
Fixes lung damage threshholds.
Tweaks heart decay time.
Fixes graft synthtissue which does damage instead of healing it (oops), and fixes the data tracking of injected vol (tracking volume added to a mob across multiple instances is a huge pain)
Also makes synthtissue keep track of how much is grown via it's name.
Tweaks how organs are frozen, if they're on a cold tile (less than 150K) they'll freeze and won't decay. Equally, if a mob is frozen, their organs won't decay. You can still freeze mobs/organs in blood coolers, freezers, fridges and morgue units too. Held and floor organs rely on the turf, internal relies on mob temperature. (So you can bung a corpse in cyro to stop decay!)
Fixes skeletons from being too much like ghost rider, and will only become ghost rider for a short time when drinking strong acids/bases.
Fixes autosurgeons' organs decaying.
Enables slow, passive brain healing.
Prevents organ regeneration when frozen.

## Why It's Good For The Game

From recent people pointing out bugs and feedback.
Overhall coming fairly soon, but this should help things for now.
Also fixes a lot of things I assumed worked because tg had it. I can't tell if it's a difference in our codebases or oversights existent in tg too.

## Changelog
:cl: Fermis
Fix: Heart, Tongue and stomach regen.
Fix: lung damage threshholds.
Fix: Graft synthtissue
Fix: Skeleton's burning for no reason
Fix: Organ freezing handling.
/:cl:

PR #9508, opened 2019-10-15
[draft] Organ trauma framework

Thalpy:
## About The Pull Request

Adds something i'm working on; organ traumas. This is mostly so Ghommie can look at it as he said he'd like to a while ago.

As your organs get damaged enough, they gain an organ traumas (based off irl diseases and etc), which has minor negative effects on you (such as slightly lowering your stamina regen for severe heart traumas). These traumas shouldn't prevent you playing, but they reduce how "optimal" you are, and if you want to fix it, you need to go to medbay to get diagnosed (the answer isn't given, the doctor has to take a biopsy, use the microscope to determine they type, and apply the correct treatment. The treatment is like the fire emblem weapon triangle, where some treatments exacerbate other traumas (such as autoimmune treatments making bacterial infections worse)). Worst case you can always transplant in a new organ (the only exception being a bacterial trauma, which will infect everything it touches, and spreads to new organs being implanted.)) Generally, traumas are self contained, unless they get severe enough. Also genetic traumas are retained on cloning.
But the key point is, these traumas don't kill you unless the doctor is a quack and manages to max out the severity of all of them, and they're low priority, meaning if everyone is dying, you heal the dying and can know that people with traumas can function

This is all WIP so vars aren't set in stone yet.

Contains #9477

## Why It's Good For The Game

Adds depth to medbay, traumas rely on diagnosis, rather than scanners outright telling you.
Does something with the organ damage system.
Endless possibilities for fun traumas.
Sets framework for tamiorgans that aren't raised properly.

## Changelog
:cl: Fermis
add: Adds Organ Traumas
/:cl:


shellspeed1:
Hold on, will bacterial infection spread to synthetic organs?

Thalpy:
> Hold on, will bacterial infection spread to synthetic organs?

Hmm, not sure yet. I suppose it could.

shellspeed1:
But.... they are synthetic organs... Might be time to also add organs to the limb grower.

Thalpy:
> But.... they are synthetic organs...

Bacteria don't care irl, they grow everywhere. They're a big problem on prosthetic/implants and part of my thesis is on that but... It really depends if it's fun or not honestly. I'll think about it when I've written more of it.

shellspeed1:
fair point, just wondering how it would effect the functionality of an artificial organ or the like.

Thalpy:
> fair point, just wondering how it would effect the functionality of an artificial organ or the like.

I'm open to ideas! I've a bad habit of feature creep, but I could consider synthetic only traumas and stuff, after I've seen the reaction from the community to this first.

Arturlang:
I mean, isn't there a draft pull option just for this kind of situation?

Thalpy:
> 
> 
> I mean, isn't there a draft pull option just for this kind of situation?

If there is I completely missed it!

shellspeed1:
Would it be possible to make it so that if certain conditions are met that certain traumas can happen. For example if you get caught in an ash storm without internals you burn your lungs reducing how well they work?

Thalpy:
> 
> 
> Would it be possible to make it so that if certain conditions are met that certain traumas can happen. For example if you get caught in an ash storm without internals you burn your lungs reducing how well they work?

Yep! Use `gain_organ_trauma(organ, datum/organ_trauma/type/trauma)`

shellspeed1:
Time to make it so everyone blows out their lungs when they go into space without protection.

code/modules/mob/living/carbon/carbon_defines.dm
```diff
@@ -63,3 +63,6 @@
 	var/damageoverlaytemp = 0
 
 	var/drunkenness = 0 //Overall drunkenness - check handle_alcohol() in life.dm for effects
+
+	//bacterial infections
+	var/list/datum/bacterium/bacteria = list() //List of infections and where (i.e. ORGAN_SLOT_X = bacterium)
```
Poojawa, line 68: Should be an empty list

then call LAZYINITLIST later for when it's used

PR #9513, opened 2019-10-16
medibots are now TOXINLOVER (so slimes) friendly

silicons:
## About The Pull Request

Makes medibots slime friendly

## Why It's Good For The Game

Medibots murdering slimes is hilarious but slimes deserve some love and now that they can actually bleed I don't think it's an issue to let them not die to medibots like they usually would.

## Changelog
:cl:
balance: Medibots no longer kill slimes when trying to heal their toxins.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/medbot.dm
+++ b/code/modules/mob/living/simple_animal/bot/medbot.dm
@@ -51,6 +51,7 @@
 	var/treatment_fire = "kelotane"
 	var/treatment_tox_avoid = "tricordrazine"
 	var/treatment_tox = "charcoal"
+	var/treatment_tox_toxlover = "toxin"
 	var/treatment_virus_avoid = null
 	var/treatment_virus = "spaceacillin"
 	var/treat_virus = 1 //If on, the bot will attempt to treat viral infections, curing them if possible.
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/bot/medbot.dm
```diff
@@ -396,6 +397,15 @@
...
 	return FALSE
 
+/mob/living/simple_animal/bot/medbot/proc/get_avoidchem_toxin(mob/M)
+	return toxlover_check(M)? null : treatment_tox_avoid
+
+/mob/living/simple_animal/bot/medbot/proc/get_healchem_toxin(mob/M)
+	return toxlover_check(M)? treatment_tox_toxlover : treatment_tox
+
+/mob/living/simple_animal/bot/medbot/proc/toxlover_check(mob/M)
+	return HAS_TRAIT(M, TRAIT_TOXINLOVER)
```
Ghommie: These sort of functions that only return macros or simple operation without executing anything else should be avoided.

silicons: aight bet

deathride58 (approved):

Issue #9495, opened 2019-10-14
Sprite Issues (Oh no, my Privates)


Owai-Seek:
Issue reported from  Round ID: 19637 (Citadel Station 13)

Reporting client version: 512

Hand clips through Hemi and Knotted Hemi sprites on pretty much every species on the list. There's basically a line just clipping right through em. (Layering issue with hands)

I've also seen some issues with an unerect tapered, at least on lizards. Specifically while rolling on the ground. (Layering issue with Satchel)

Ghommie:
Could you provide some visual example of this layering issue on discord? Won't promise it'll be easy to fix.

PR #9528, opened 2019-10-18
Fixes some peculiarities with hands layer overlapping stuff

Ghommie:
## About The Pull Request
#9495 brought up some issues with the west dir right hand and the east dir left hand overlapping stuff it really shouldn't.
So if you are wondering why I'm not simply shifting those things mentioned in the issue a couple pixels down... 
There are some potential discordant overlaying issues concerning humans owning bodyparts of different species, currently only prevented by the fact the code for that feature hasn't been working for a long time, with transplanted limbs assuming the new owner's species while the owner var remains null. I'm not going to digress too much about it though, this is just about almost half the work to fix that niche too.

## Why It's Good For The Game
Fixing some layering issues mentioned in #9495.

## Changelog
Do not want.


Poojawa:
HANDS AREN'T REEEEEAAAAAAAAAL

silicons (approved):

PR #9532, opened 2019-10-19
Reduces stamloss of melee weapons by 20%

MrJWhit:
## About The Pull Request

This melee buff will make it so it's not so hard to put yourself in stamcrit from attacking someone. Note, this doesn't effect the stamloss of hitting an object, just against other mobs.

## Why It's Good For The Game

It helps ashwalkers actually fight mobs, as well as reducing the feel bad of attacking people who have stamina tools, and then they use them against you.

## Changelog
:cl:
balance: rebalance melee stamloss
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -82,7 +82,7 @@
 	log_combat(user, M, "attacked", src.name, "(INTENT: [uppertext(user.a_intent)]) (DAMTYPE: [uppertext(damtype)])")
 	add_fingerprint(user)
 
-	user.adjustStaminaLossBuffered(getweight())//CIT CHANGE - makes attacking things cause stamina loss
+	user.adjustStaminaLossBuffered(getweight()*0.8)//CIT CHANGE - makes attacking things cause stamina loss
 
 //the equivalent of the standard version of attack() but for object targets.
 /obj/item/proc/attack_obj(obj/O, mob/living/user)


code/_onclick/item_attack.dm
```diff
@@ -82,7 +82,7 @@
 	log_combat(user, M, "attacked", src.name, "(INTENT: [uppertext(user.a_intent)]) (DAMTYPE: [uppertext(damtype)])")
 	add_fingerprint(user)
 
-	user.adjustStaminaLossBuffered(getweight())//CIT CHANGE - makes attacking things cause stamina loss
+	user.adjustStaminaLossBuffered(getweight()*0.8)//CIT CHANGE - makes attacking things cause stamina loss
```
Ghommie, line 85: You should make it a define and add it to `code\__DEFINES\combat.dm`.

MrJWhit: I'm going off of what was used in the lower part of the code, where the stamloss to attack an object is *1.2 of the weapon weight.

Ghommie: Scratch my previous suggestion then, the stamina cost feature is so uncharming and I don't get the point of why attack_obj should be more tiresome.
What I wanted was fewer magic numbers and more coherence. But the feature is slim and rudimentary, not worth a drama.

MrJWhit: Yeah, it was either, make both of them variables that are used in one spot, or just accept that they're a modifier and go on from there.

silicons:
TALLY HO!!!!!

BlackMajor (approved):

silicons (approved):

PR #9541, opened 2019-10-20
ports (probably badly) the abductor all access gland from the tg wiki

Seris02:
## About The Pull Request

Adds the all access abductor organ from the TG wiki because why not. Also might work on the other ones in there we don't have. Also I can't test this, so it should probably get locally tested on someone's computer, or testmerged or something like that. Also the code is probably bad, tell me how to improve it, please. Also it probably removed some things because I'm on my web browser and the thing automatically removed some of the traits because it's getting all the commits from the current thing, tell me if I need to add back in those deleted lines.

## Why It's Good For The Game

Because I can't think of why not to add this. Also more abductor things. Yay.

## Changelog
:cl:
add: added the abductor all access gland
/:cl:


Linzolle:
everything that was removed in defines.dm needs to still be there.

Seris02:
ok cool it doesn't work imma just kill this PR

Seris02:
and redo it later

PR #9556, opened 2019-10-20
Fix not being able to read paper stuck to airlocks, lets AI view paper but with obfuscated text

Linzolle:
## About The Pull Request
Title. Alternative to https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9546 that isn't just a bandaid fix and fixes the underlying issue. The AI can just see paper with the stars text, I couldn't get it to check if the paper was close to a camera to be able to read it fully.

## Changelog
:cl:
fix: unable to read paper on airlocks
tweak: ai can see paper (still just shows stars)
/:cl:

### DIFF
--- a/code/modules/paperwork/paper.dm
+++ b/code/modules/paperwork/paper.dm
@@ -59,7 +59,7 @@
 	updateinfolinks()
 
 /obj/item/paper/oui_getcontent(mob/target)
-	if(!target.is_literate())
+	if(!target.is_literate() || isAI(target))
 		return "<HTML><HEAD><TITLE>[name]</TITLE></HEAD><BODY>[stars(info)]<HR>[stamps]</BODY></HTML>"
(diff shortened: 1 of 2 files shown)


code/modules/paperwork/paper.dm
```diff
@@ -59,7 +59,7 @@
 	updateinfolinks()
 
 /obj/item/paper/oui_getcontent(mob/target)
-	if(!target.is_literate())
+	if(!target.is_literate() || isAI(target))
```
Ghommie: This will also screw up the feature of holding up a paper on the camera for the AI to read it, keep in mind.
EDIT: was told the camera paper feature uses basic html windows, so this is not a problem yet.

Linzolle: This is kind of a weird way to do it to be perfectly honest though. I couldn't find a really more efficient way to do it with how oui works. Ideally I wanted a "force stars" variable in the proc but because of the way getcontent is called that's not possible without a lot of other changes.

silicons (approved):

PR #9557, opened 2019-10-21
Dynamic voting (needs testing)

Putnam3145:
## About The Pull Request

Adds an option to allow voting on threat level curve of the Dynamic game mode (or just force extended as usual), rather than simply voting between extended/secret. Keeps extended/secret vote as an option.

With dynamic voting enabled, it gives the following options:

Extended
Very calm
Somewhat calm
Balanced
Somewhat chaotic
Very chaotic

These votes determine the Dynamic game mode's threat generation curve; the more chaotic the votes, the higher the average threat level, and the more disagreement there is on how threatening it should be, the more variance there will be in the possible threat levels. Moreover, if the majority of people vote extended the mode will force extended. If the majority do *not* vote for extended, then all votes for extended are counted as "Very calm" votes.

## Why It's Good For The Game

Dynamic is good actually.

## Changelog
:cl:
add: Dynamic voting
config: Added DYNAMIC_VOTING to game_options
/:cl:


code/controllers/subsystem/vote.dm
```diff
@@ -146,6 +152,17 @@ SUBSYSTEM_DEF(vote)
...
 					else
 						GLOB.master_mode = .
+			if("dynamic")
+				if(SSticker.current_state > GAME_STATE_PREGAME)//Don't change the mode if the round already started.
+					return message_admins("A vote has tried to change the gamemode, but the game has already started. Aborting.")
+				GLOB.dynamic_forced_extended = choices["extended"]/voted.len > 0.5
+				if(GLOB.dynamic_forced_extended)
+					return message_admins("Dynamic extended has been voted for.")
+				var/mean = (choices["extended"]*-2+choices[PEACE2]*-2+choices[PEACE1]*-1+choices[CHAOS1]*1+choices[CHAOS2]*2)/voted.len
+				var/variance=(((-2-mean)**2)*choices["extended"]+((-2-mean)**2)*choices[PEACE2]+((-1-mean)**2)*choices[PEACE1]+((0-mean)**2)*choices[BALANCED]+((1-mean)**2)*choices[CHAOS1]+((2-mean)**2)*choices[CHAOS2])/voted.len //Sorry. Im sorry. Im trying to remove it
```
Ghommie: Keep in mind this will give off a division by zero error in the unlikely occurrence of no one voting.

code/controllers/subsystem/vote.dm
```diff
@@ -213,6 +230,8 @@ SUBSYSTEM_DEF(vote)
 					choices[i] = 0
 			if("roundtype") //CIT CHANGE - adds the roundstart secret/extended vote
 				choices.Add("secret", "extended")
+			if("dynamic")
+				choices.Add("extended",PEACE2,PEACE1,BALANCED,CHAOS1,CHAOS2)
```
Ghommie: You could as well add a "secret" option here too, but I doubt both non-dynamics are likely to be ever choosen since they still require over fifty percent of all votes to trigger.

code/controllers/configuration/entries/game_options.dm
```diff
@@ -364,3 +364,5 @@
 /datum/config_entry/number/auto_transfer_delay
 	config_entry_value = 72000
 	min_val = 0
+
+/datum/config_entry/flag/dynamic_voting
```
Ghommie: An admin "verb" should be added to switch on or off dynamic voting for a given round through CONFIG_SET (modified configs do not carry over successive rounds iirc). It'd be sad if this ended up backend only.

Putnam3145: I've been an admin but I'm not sure where that would go. Game panel? Secrets?

code/controllers/subsystem/vote.dm
Putnam3145: https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9557/commits/0a000f3165875ace0d8b89ef909cadb59170a5a2

code/controllers/subsystem/vote.dm
Putnam3145: Yeah, that's a worry--could make it so that a sufficiently low mean makes it do dynamic_classic_secret.

code/controllers/configuration/entries/game_options.dm
Ghommie: I think somewhere easily accessible like "server" or "fun" tabs.
Since it currently would have to be hastily toggled before gamemode voting happens to work, it shouldn't be added to the secrets or game panels.
Though I suggest you ask @kevinz000, who is an administator, unlike me.

Putnam3145: Right, "server" tab makes most sense, I'll do it

Putnam3145: This sufficient? https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9557/commits/da7d4345b5bdcaa07662de1ef5d9abe15329cb23

Putnam3145: Upon testing: it's sufficient.

code/modules/admin/admin.dm
```diff
@@ -686,6 +686,19 @@
 			log_admin("[key_name(usr)] set the pre-game delay to [DisplayTimeText(newtime)].")
 		SSblackbox.record_feedback("tally", "admin_verb", 1, "Delay Game Start") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
+/datum/admins/proc/toggledynamicvote()
+	set category = "Server"
+	set desc="Switches between secret/extended and dynamic voting"
+	set name="Toggle Dynamic Vote"
+	CONFIG_SET(flag/dynamic_voting,!CONFIG_GET(flag/dynamic_voting))
```
Ghommie: Just state a variable `dynamic_mode` (its name can be however you want) with value `!CONFIG_GET(flag/dynamic_voting)` and replace all the following `CONFIG_GET` with it. More efficient.

code/modules/admin/admin.dm
```diff
@@ -686,6 +686,19 @@
...
+/datum/admins/proc/toggledynamicvote()
+	set category = "Server"
+	set desc="Switches between secret/extended and dynamic voting"
+	set name="Toggle Dynamic Vote"
+	CONFIG_SET(flag/dynamic_voting,!CONFIG_GET(flag/dynamic_voting))
+	if (CONFIG_GET(flag/dynamic_voting))
+		to_chat(world, "<B>Vote is now between extended and dynamic chaos.</B>")
+	else
+		to_chat(world, "<B>Vote is now between extended and secret.</B>")
+	log_admin("[key_name(usr)] toggled dynamic voting.")
```
Ghommie: Replace this line with the following:
```
	log_admin("[key_name(usr)] [dynamic_mode ? "enabled" : "disabled"] dynamic voting.")
```

code/modules/admin/admin.dm
Putnam3145: https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9557/commits/296ce4fb78b45fbc6f088591420d5967068b26b2

code/modules/admin/admin.dm
Putnam3145: https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9557/commits/296ce4fb78b45fbc6f088591420d5967068b26b2

YakumoChen:
Does this count all the votes as is or does it combine into some sort of average threat level? I'm concerned in either case because youre either going to see a lot of votes be pointless due to the large spread (making extended VERY likely since extended voters will be all or nothing), or a lot of very calm on average rounds.
This especially becomes a problem when the vote is made secret since youre being pitted with 5 options against extended and no means to know what non-extended players vote.

Putnam3145:
It counts every vote. If more than 50% of voters vote extended, it will be extended, otherwise all extended votes will count as a vote for very peaceful. Extended voters aren't really all or nothing, they're given weight in the threat level rating.

Yes, I do predict a lot of rounds will be very calm on average, but the thing is that 1. even at the most peaceful possible, the mean threat level is still ~35, which is enough for traitor + a midround wizard 2. there's still a chance, albeit small, that the threat level will be high otherwise, which might actually be a failure of the system but whatever.

Of course all this is subject to tweaking and changing. This doesn't have to go unchanged forever, and if this system turns out to be too varied or random then it can be tweaked or even replaced entirely.

Too much variation is the biggest foreseeable problem here and can be dealt with without too much issue, if it does turn out to be an issue.

code/controllers/subsystem/vote.dm
Putnam3145: Upon really checking the code in the process of #9582, looks like secret's basically irrelevant as long as threat level's low, and I can't really think of a way to take those votes into account for if dynamic ends up winning anyway (-1? -2 like extended?), so I'll hold off on this until testing.

Putnam3145:
(fixed the merge conflict caused by #9582 being brought in)

code/controllers/subsystem/vote.dm
```diff
@@ -379,4 +396,10 @@ SUBSYSTEM_DEF(vote)
...
 		if(P)
-			P.player_actions -= src
\ No newline at end of file
+			P.player_actions -= src
+
+#undef PEACE2
+#undef PEACE1
+#undef BALANCED
+#undef CHAOS1
+#undef CHAOS2
```
GrayRachnid: Some of these don't exist anymore

config/game_options.txt
```diff
@@ -699,3 +557,6 @@ MONKEYCAP 64
 
 ## Uncomment to use TG-style combat
 #DISABLE_STAMBUFFER
+
+#Replaces standard extended/secret dichotomy with extended and 5 levels of chaos using dynamic.
```
GrayRachnid: comment needs to change

code/controllers/subsystem/vote.dm
Putnam3145: fixed in https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9557/commits/85fd53d61595a0bc5ce0dce467b89278b8678341

config/game_options.txt
Putnam3145: fuixed in https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9557/commits/3a48b7238090181a90e93a6affbe25efd9423f32

code/controllers/subsystem/vote.dm
```diff
@@ -124,6 +124,10 @@ SUBSYSTEM_DEF(vote)
 		message_admins(admintext)
 	return .
 
+#define EXTENDED "calm"
```
GrayRachnid: this is defined and not used, also not undefined at the end

code/game/gamemodes/dynamic/dynamic.dm
```diff
@@ -201,30 +201,35 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
...
+		if(31 to 49)
 			. += "<b>Anomalous Exogeology</b></center><BR>"
 			. += "Although your station lies within what is generally considered Nanotrasen-controlled space, the course of its orbit has caused it to cross unusually close to exogeological features with anomalous readings. Although these features offer opportunities for our research department, it is known that these little understood readings are often correlated with increased activity from competing interstellar organizations and individuals, among them the Wizard Federation and Cult of the Geometer of Blood - all known competitors for Anomaly Type B sites. Exercise elevated caution."
-		if(40 to 65)
+			set_security_level(SEC_LEVEL_GREEN)
+		if(50 to 65)
 			. += "<b>Contested System</b></center><BR>"
 			. += "Your station's orbit passes along the edge of Nanotrasen's sphere of influence. While subversive elements remain the most likely threat against your station, hostile organizations are bolder here, where our grip is weaker. Exercise increased caution against elite Syndicate strike forces, or Executives forbid, some kind of ill-conceived unionizing attempt."
-		if(66 to 79)
```
GrayRachnid: uhh don't think this was supposed to be removed.

code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
```diff
@@ -465,3 +482,244 @@
...
+/datum/dynamic_ruleset/midround/from_ghosts/ninja
+	name = "Space Ninja"
+	config_tag = "ninja"
+	antag_flag = ROLE_NINJA
+	enemy_roles = list("Security Officer","Head of Security","Captain","AI","Cyborg")
+	required_enemies = list(3,2,2,2,2,1,1,1,1,0)
+	required_candidates = 1
+	weight = 4
+	cost = 15
+	requirements = list(101,101,101,90,80,70,60,50,40,30)
```
GrayRachnid, line 681: why is ninja so expensive to run at midpop

Putnam3145: didn't get caught in my balance passes, probably, will tweak

config/dynamic_config.txt
```diff
@@ -0,0 +1,299 @@
...
+DYNAMIC_HIGH_POPULATION_REQUIREMENT PROCESSOR_OVERLOAD 5
+DYNAMIC_HIGH_POPULATION_REQUIREMENT SPACE_DUST 5
+DYNAMIC_HIGH_POPULATION_REQUIREMENT MAJOR_DUST 10
+DYNAMIC_HIGH_POPULATION_REQUIREMENT ELECTRICAL_STORM 5
+DYNAMIC_HIGH_POPULATION_REQUIREMENT HEART_ATTACK 5
+DYNAMIC_HIGH_POPULATION_REQUIREMENT RADIATION_STORM 5
+
+## Latejoin antags
+DYNAMIC_HIGH_POPULATION_REQUIREMENT LATEJOIN_TRAITOR 15
+DYNAMIC_HIGH_POPULATION_REQUIREMENT LATEJOIN_REVOLUTION 50
```
GrayRachnid, line 265: forgot to drop this down to 20

config/dynamic_config.txt
```diff
@@ -0,0 +1,299 @@
...
+DYNAMIC_HIGH_POPULATION_REQUIREMENT EXTENDED 101
+DYNAMIC_HIGH_POPULATION_REQUIREMENT CLOCKWORK_CULT 50
+DYNAMIC_HIGH_POPULATION_REQUIREMENT CLOWNOPS 101
+DYNAMIC_HIGH_POPULATION_REQUIREMENT DEVIL 101
+DYNAMIC_HIGH_POPULATION_REQUIREMENT MONKEY 101
+DYNAMIC_HIGH_POPULATION_REQUIREMENT METEOR 101
+
+## Midround antags
+DYNAMIC_HIGH_POPULATION_REQUIREMENT MIDROUND_TRAITOR 15
+DYNAMIC_HIGH_POPULATION_REQUIREMENT MIDROUND_MALF_AI 35
```
GrayRachnid, line 228: this similarly should be at 30 compared to above

config/dynamic_config.txt
```diff
@@ -0,0 +1,299 @@
...
+DYNAMIC_HIGH_POPULATION_REQUIREMENT MIDROUND_WIZARD 50
```
GrayRachnid, line 229: this too

silicons (approved):

silicons (approved):

PR #9563, opened 2019-10-21
[TESTMERGE] Goodbye, superpH, you were too wild for our world.

Thalpy:
## About The Pull Request

Some people pushed superbases too far, and well, admins messaged me with concerns, so I've added a limit to how high they can go. Considering that 2100u at pH 50+, 5k temp, purity 1, was a 13 gib radius, this should lower things by a factor of 3, to roughly 5 gib radius with a chemical payload.

Though, to note, even before this change, the other explosive chems are stronger, hence why it was kept like this for a while but, it seems chemical payloads are easier to get than I thought. Heck!

# Testmerge this please, this makes ALL containers melt from superbases. I need to know if unintentional holders are melting.
Though, testing locally didn't seem to show anything melting over the course of a round.

## Why It's Good For The Game

Booms are too big. Booms should be smaller now.

## Changelog
:cl: Fermis
tweak: tweaked how super bases/acids work but limiting them
/:cl:


code/modules/reagents/reagent_containers.dm
```diff
@@ -156,6 +156,16 @@
...
 			START_PROCESSING(SSobj, src)
+	else if((reagents.pH < -3) || (reagents.pH > 17))
+		var/list/seen = viewers(5, get_turf(src))
+		var/iconhtml = icon2html(src, seen)
+		for(var/mob/M in seen)
+			to_chat(M, "<span class='notice'>[iconhtml] \The [src]'s is damaged by the super pH and begins to deform!</span>")
+		if(reagents.pH < -3)
+			reagents.pH = -3
+		else
+			reagents.pH = 17
```
Ghommie: Replace it with:
```
	reagents.pH = CLAMP(reagents.pH, -3, 17)
```

code/modules/reagents/reagent_containers.dm
```diff
@@ -156,6 +156,16 @@
 	if(beaker_weakness_bitflag & PH_WEAK)
 		if((reagents.pH < 1.5) || (reagents.pH > 12.5))
 			START_PROCESSING(SSobj, src)
+	else if((reagents.pH < -3) || (reagents.pH > 17))
+		var/list/seen = viewers(5, get_turf(src))
+		var/iconhtml = icon2html(src, seen)
+		for(var/mob/M in seen)
+			to_chat(M, "<span class='notice'>[iconhtml] \The [src]'s is damaged by the super pH and begins to deform!</span>")
```
Ghommie: Use visible_message() instead, please. Delete these awful lines.
```
	visible_message("<span class='notice'>[icon2html(src, seen)] \The [src] is damaged by the super pH and begins to deform!</span>")
```

code/modules/reagents/reagent_containers.dm
```diff
@@ -156,6 +156,16 @@
 	if(beaker_weakness_bitflag & PH_WEAK)
 		if((reagents.pH < 1.5) || (reagents.pH > 12.5))
 			START_PROCESSING(SSobj, src)
+	else if((reagents.pH < -3) || (reagents.pH > 17))
```
Ghommie, line 159: Just to point out, this `else if` means beaker susceptible to akalyne or acid Ph are still allowed to exceed acceptable values. No great impact to the game however, cause those containers will be destroyed in seconds anyway. Your call.

code/modules/reagents/reagent_containers.dm
Thalpy: I've tried that, but [iconhtml] never works if i use visible_message(), do you know how to fix it?

Ghommie: Ok then, try and see if this works instead, icon2html is somewhat stringy.
```
	visible_message("<span class='notice'>[icon2html(src, viewers(src)] \The [src] is damaged by the super pH and begins to deform!</span>")
```

code/modules/reagents/reagent_containers.dm
Thalpy: Mm! I'm fine with it working like that because the flagged containers have melting sprites.

Thalpy:
Why does it have the tick when there's a confliiict? I'll fix this tomorrow.

Thalpy:
Fixed

Ghommie (approved):

PR #9577, opened 2019-10-22
[Balance Pass] - Supermatter Mega Delam (Tesla)

Feasel:
All I'm doing here is raising the power_penalty_thresholds to a higher number so that it takes more time and effort to force the mega delamination.

Also by leaving a wider gap between Severe and Critical stages, it gives more warning to AI and Engineers that Tesla delamination is happening soon.

As some of us know by now, it's /very/ easy to force Tesla delams and that uhh... it's aids to deal with due to all the effects happening so quickly as it powers through all the stages from just 3 simple switches.

It takes effort and time for Singulo delam to start happening due to the high amount of moles needed on. that. tile. 

Does not take much for Tesla. Fixing that now.


Putnam3145:
more like gotta reevaluate CO2's place as simultaneously the best balance of safety and power while still being the easiest to cause catastrophic delams with

Feasel:
> more like gotta reevaluate CO2's place as simultaneously the best balance of safety and power while still being the easiest to cause catastrophic delams with

I intend to see how upping the threshold will interact first before doing any alterations to gasses. Usually buying more time is all the crew needs to fix the problem and that's what I am doing right here.

silicons (approved):

PR #9578, opened 2019-10-22
Map config system improvements - Fixed and expanded map vote restrictions, map-specific announcer framework, and more

deathride58:
## About The Pull Request
This PR mostly deals with expanding the functionality of the existing backend framework, and is useful primarily to mappers that wish to be experimental with their maps. The only effects this PR has on the game as of right now is the voting system respecting map playercount restrictions, and a new 1% chance for the medibot voice to be the station announcer.

This adds 2 new variables for map entries in maps.txt, and adds 3 new variables for map config files in the individual .json files. They are as follows

- max_round_search_span [number (default: 0)] (maps.txt) - This determines the amount of previous rounds that the game will search for in order to determine if a map has been played too many times for the map to appear in the end-of-round vote. If this is 0, the map will always be available in the vote even if it's been running 24/7. This is primarily useful for gimmicky maps that can get stale fast if there's no restrictions on how often it can be played, for example, if someone were to implement a map with a gimmick that everything is severely limited in terms of budget.
- max_rounds_played [number (default: 0)] (maps.txt) - This is used in conjunction with the above. If a map has been played more than the number declared in this variable within the last few rounds dictated by the above variable, then the map will be made unavailable at the end-of-round vote.

- maptype [string (default: "station")] (map.json) - This currently has no explicit functionality. This is primarily to allow creative codermappers to specify a completely different map type from the standard station gameplay by adding hooks into the code that check for the map type. This will make it a lot easier to port FTL13 maps and have them function like ships as intended if any coders are brave enough to do so.
- announcertype [string (default: "standard")] (map.json) - This allows mappers to specify which announcer a given map to use. The default value, standard, automatically switches to the classic announcer, but has a 1% chance to instead be the new medibot announcer. In order to define a new announcer type, you can modify the get_announcer_sound() proc in priority_announce.dm
- year_offset [number (default: 540)] (map.json) - This var allows mappers to specify the offset of the ingame year from the IRL year. This is currently displayed on computers, PDAs, and various other machinery around the station. This means a mapper can make a map that takes place in a completely different time period if they so desire. You know you want to.

## Changelog
:cl:
add: There is now a 1% chance for the station's announcer to be the medibot voice instead of the classic TG announcer.
add: The map config system has been expanded to allow mappers to specify the map type, announcer voice, ingame year, and how often a given map can be voted at roundend.
fix: The map vote system now takes into account map playercount limits properly.
/:cl:


code/controllers/subsystem/mapping.dm
```diff
@@ -53,6 +53,8 @@ SUBSYSTEM_DEF(mapping)
 		if(!config || config.defaulted)
 			to_chat(world, "<span class='boldannounce'>Unable to load next or default map config, defaulting to Box Station</span>")
 			config = old_config
+	GLOB.year_integer += config.year_offset
+	GLOB.announcertype = (config.announcertype == "standard" ? (prob(1) ? "cleanbot" : "classic") : config.announcertype)
```
Ghommie:
```suggestion
	GLOB.announcertype = (config.announcertype == "Standard" ? (prob(1) ? "medibot" : "classic") : config.announcertype)
```

I'm not going to review all of this right now, but dope.

deathride58: The default value for announcertype in the map's config is `"standard"`, which is case sensitive

Ghommie: Well, it wasn't. Now it's.

silicons (approved):

PR #9595, opened 2019-10-24
Fixes examine strings not properly showing blood-stained status of items

deathride58:
Title. This is a follow-up to the RGB blood PR, and addresses a requested change from my review that was made a little over _TWO MONTHS AGO._ This took less than ten minutes to do. Eleven bloody lines to do what I was requesting, and instead of actually doing it, it was argued about every single time it was brought up. If this were actually addressed beforehand, the RGB blood PR would have taken a lot less time to actually get merged. And apparently I'm the one in the wrong for not merging it fast enough, when this change that was requested two months ago took less than ten minutes to actually do? Good. Bloody. Riddance.

## Changelog
:cl:
fix: Examining a spaceman, and other things that use get_examine_string(), will now actually properly show when an item is blood-stained.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -261,7 +261,17 @@
 		. = "[article] [src]"
 		override[EXAMINE_POSITION_ARTICLE] = article
 
+	var/should_override = FALSE
+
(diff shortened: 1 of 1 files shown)


nik707 (approved):
Sick

PR #9598, opened 2019-10-24
various runtime fixes - plasmaman tongues no longer have a maxHealth of "alien" and more

deathride58:
Title. These are very small, minor things. They don't really cause much harm to the game's functionality, but it'd be nice to not have these clog up the view runtimes panel. Might add more fixes later.

## Changelog
:cl:
fix: Plasmaman tongues no longer have a maxHealth of "alien", and no longer cause the organ's on_life to always runtime.
fix: Shooting a simplemob no longer causes runtimes prior to the blood effect being created.
fix: Removing a filter from an object that lacks filters no longer causes runtimes.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -164,7 +164,7 @@
 			if(starting)
 				splatter_dir = get_dir(starting, target_loca)
 			var/obj/item/bodypart/B = L.get_bodypart(def_zone)
-			if(B.status == BODYPART_ROBOTIC) // So if you hit a robotic, it sparks instead of bloodspatters
+			if(istype(B) && B.status == BODYPART_ROBOTIC) // So if you hit a robotic, it sparks instead of bloodspatters
 				do_sparks(2, FALSE, target.loc)
 				if(prob(25))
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -164,7 +164,7 @@
 			if(starting)
 				splatter_dir = get_dir(starting, target_loca)
 			var/obj/item/bodypart/B = L.get_bodypart(def_zone)
-			if(B.status == BODYPART_ROBOTIC) // So if you hit a robotic, it sparks instead of bloodspatters
+			if(istype(B) && B.status == BODYPART_ROBOTIC) // So if you hit a robotic, it sparks instead of bloodspatters
```
Ghommie:
```suggestion
			if(B && B.status == BODYPART_ROBOTIC) // So if you hit a robotic, it sparks instead of bloodspatters
```
In no case `get_bodypart()` should return an atom reference that is not an actual (or sort of, like chainsaw arm or synthetic armblade) bodypart.

silicons (approved):

PR #9619, opened 2019-10-26
Buffs engiborg build potential

BlackMajor:
Added the following to engiborgs.

-Roundstart basic RPED
-Atmos holofan
-A gripper that can only pick up circuit boards please god never expand on grippers
-A light replacer (Janiborgs be mad, but I'll get to reworking you later too. Engiborgs need this too)

Robotics RPED upgrade replaced with engiborg BSRPED upgrade, which does what you think.

Right now engiborgs have a lot of versatility, but can't do engineering on par with an actual engineer.
They can fix things way faster, but ultimately can't really /build/ like an engi can.
Hopefully this addresses this.

Kinda a big potential there for balance concerns, so I welcome input on this.

Part 1 of going over all the borgs and rebalancing them.


JTGSZ:
nice

silicons (approved):

PR #9631, opened 2019-10-27
Medibots now ACTUALLY heal slimes, rather than just not killing them (or how I learned binary logic the hard way)

silicons:
fixes medibots.


Putnam3145:
are you sure you want isnull instead of just !

BlackMajor (approved):

PR #9638, opened 2019-10-27
[TESTMERGE ONLY] Reverts hat tossing

silicons:
Broke lots of suit helmets due to clothing flags being overridden which erased higher up overrides.
Testmerge until that's fixed.


BlackMajor:
Fixed by
https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9640

PR #9667, opened 2019-10-30
Gee Kevinz, why does YOUR research division let you have A HUNDRED AND SIXTY fetish gas points?

ghost:
wtf

## About The Pull Request
miasma research define is now 6 instead of 120

## Why It's Good For The Game
no more 3.05e^7 research points

## Changelog
:cl:
fix: kevinz forgot to nerf miasma research and cargo value after making it produce like 100x as much lmao
/:cl:

### DIFF
--- a/code/__DEFINES/reactions.dm
+++ b/code/__DEFINES/reactions.dm
@@ -27,7 +27,7 @@
 #define NOBLIUM_RESEARCH_AMOUNT				1000
 #define BZ_RESEARCH_SCALE					4
 #define BZ_RESEARCH_MAX_AMOUNT				400
-#define MIASMA_RESEARCH_AMOUNT				160
+#define MIASMA_RESEARCH_AMOUNT				6
 #define STIMULUM_RESEARCH_AMOUNT			50
 //Plasma fusion properties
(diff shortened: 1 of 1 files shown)


ghost:
salt pr

deathride58:
good first pr

IronEleven:
A 95% nerf seems like overkill.  I'm not denying it's incredibly powerful right now but the proposed values seem like it will simply stop being worth the effort to set up at all.

ghost:
this is reverting it to before miasma generation buff, which multiplied the amount of mols made by corpses by 5x but also lowered their effectiveness. also the original 160 value makes it the BEST method of getting research points PERIOD, and can be done in similar speed to a toxins bomb with no research cap. by comparison hypernoblium fusion is worth 1000 and requires 90 minutes of setup

ShadeAware:
Will it at least still be a viable way to make points?

ShadeAware:
Still, a _95% reduction_ is a bit overkill imho. Maybe, like, 50-65%?

ghost:
50% is still 500k points in 15 minutes, aka a couple times more than you need to do every research

i dunno if anyone commenting about it has seen it in use besides the science mains that do it every round, but basically you do 15 minutes work in the roundstart map and take an rpd to it and get 1 million research points

silicons (approved):

PR #9668, opened 2019-10-30
Gives ways for various slime-killing healing viruses not to do that.

Putnam3145:
## About The Pull Request

Makes starlight condensation not hurt toxin lovers and allows the transmission thresholds of plasma fixation and radioactive resonance to do the same.

## Why It's Good For The Game

starlight condensation made every slime player immediately call for the viro's head for the pill labeled "healing disease" killing them and the other two it's nice to have more reasons to reach thresholds, plus see above

## Changelog
:cl:
balance: Made starlight condensation not kill slime people.
balance: Added not-killing-slime-people to the transmission threshold of plasma fixation and radioactive resonance.
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -368,7 +368,6 @@
 	level = 8
 	passive_message = "<span class='notice'>You feel an odd attraction to plasma.</span>"
 	var/temp_rate = 1
-	var/heals_slimes = FALSE
 	threshold_desc = "<b>Transmission 6:</b> Increases temperature adjustment rate and heals toxin lovers.<br>\
 					  <b>Stage Speed 7:</b> Increases healing speed."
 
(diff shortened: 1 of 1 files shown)


Tupinambis:
Remember when being a slimeperson had at least one single disadvantage for all of the benefits they give? Let's go back to that.

xurxomario:
> Remember when being a slimeperson had at least one single disadvantage for all of the benefits they give? Let's go back to that.

I mean, i feel like not being able to be defibbed, getting instakilled by one of the most common medicines in the station and dying to cold incredibly easy are already more than enough disadvantages, in exchange to being somewhat resistant to fire and some poisons (which in itself is kinda useless since most poisons deal other damages and/or sleep you anyway), and a bit of skipping ahead in chem (which is the most useful thing they can do, anyway). This is especially true considering that it was created as a xenobio made race. I completely agree with this change.

Putnam3145:
they still die to tricordrazine/epinephrine/omnizine/(lists 20 other medicines that heal tox damage without healing slimes), yes. i think starlight condensation's hurt ought be removed cause it's nigh impossible to avoid taking damage if you've got a virus with it

silicons (approved):

PR #9677, opened 2019-10-31
Makes clone scans lock while processing

Putnam3145:
##  About The Pull Request

As above, locks the cloner while scanning.

## Why It's Good For The Game

No self-scans by cheesing with a humanized monkey.

## Changelog
:cl:
tweak: Made the clone scanner lock while it's scanning.
/:cl:

### DIFF
--- a/code/game/machinery/computer/cloning.dm
+++ b/code/game/machinery/computer/cloning.dm
@@ -298,13 +298,14 @@
 		src.updateUsrDialog()
 		playsound(src, 'sound/machines/terminal_prompt.ogg', 50, 0)
 		say("Initiating scan...")
-
+		scanner.locked = TRUE
 		spawn(20)
 			src.scan_occupant(scanner.occupant)
 
 			loading = 0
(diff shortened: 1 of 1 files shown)


YakumoChen:
uh, isn't the delay to allow the ghost to come back to its body?

Just give the scan a sanity check or lock it when scanning

Putnam3145:
This was such an easy change that I decided proposing it alongside the code to change it would be a good idea and, yeah, that seems like a dang good reason, except that I never manage to get there in 5 seconds, much less 2, heh. There's probably a better solution, hmm.

xurxomario:
Maybe you could make the scan cancel itself if its opened while doing it? That would avoid self scans too!

ghost:
@Putnam3145 make it so the doors bolt while scanning.  scanners have a door bolting function somewhere...

Putnam3145:
good idea. done.

silicons:
can't believe i'm nerfed

silicons (approved):

PR #9681, opened 2019-11-01
buffs bluespace apprentice (targeted teleport --> blink)

silicons:
## About The Pull Request

gives bluespace apprentice blink instead of targeted area teleport

## Why It's Good For The Game

bluespace is unironically weaker than all of the other three sets, this should make them harder to steamroll and de-incentivize hiding for long periods of time. also without teleport they can blink but it's harder for them to cover a huge amount of ground quickly/also less incentivize to locker hide until it's charged.
the point of wizards is to fight, not to run forever.

## Changelog
:cl:
balance: bluespace wizard apprentice now has blink instead of targeted area teleportation
/:cl:


YakumoChen:
Bluespace wizard doesn't actually have any tools to fight though? All they do is teleport around and distract the crew.

nik707 (approved):

PR #9699, opened 2019-11-03
Bleeding Changes, Armor protects blood loss

Poojawa:
Reopening #9402 as an alternative to #9682 because this has more effort than just completely absolving bloodloss at range.

Blood ratios made a lot more forgiving, armor protects additional bleeding.

More drugs heal blood loss rate as well, though Gauze is still superior


silicons (dismissed):
why are we throwing in so many probs() and rands() we really don't need so much randomness in core combat code.

Poojawa:
because TG's code has it and it's better than 100% assured blood deletion

silicons:
ehh ill read it again in a day

silicons:
actually yeah i don't like this.
armor shouldn't be a thresholded thing in general but even if it is there's no reason for the prob(40).
old tg code was just for effects, not damage.

PR #9712, opened 2019-11-04
Rapid Light Dispenser (RLD) Nerf

KathrinBailey:
## About The Pull Request
Nerfs the RLD.

## Why It's Good For The Game
Can't go and lag the server with spamming lights, turns it into a grief weapon that lags the server into an upgrade to the light replacer.

## Changelog
:cl:
fix: RLD from 500 max matter units and loaded matter units to 200.
/:cl:


ghost:
it's still usable for grief, just takes more storage space carrying ammo

KathrinBailey:
> it's still usable for grief, just takes more storage space carrying ammo

It's still useable to grief but nothing short of making removing it will make it so you can't. Even if you reduce the capacity to 50, you can carry around a stack of plasteel and refill it. When you examine and see it can hold 500 matter units, it kinda gives you the feeling that you can place infinite lights. If it was just 200 matter units, not only is that more in line with the RCD, it kinda makes it feel like there is actually a limit on the number of lights you can place.
In effect, it doesn't actually nerf much, a proper nerf would be to change the costs of lights, but then people could just lug around more plasteel.

Nerfs the number of tube lights you can put in one go from 50 to 20.
Nerfs the number of floor lights you can put in one go from 33 to 13.

ghost:
> > it's still usable for grief, just takes more storage space carrying ammo
> 
> It's still useable to grief but nothing short of making removing it will make it so you can't. Even if you reduce the capacity to 50, you can carry around a stack of plasteel and refill it. When you examine and see it can hold 500 matter units, it kinda gives you the feeling that you can place infinite lights. If it was just 200 matter units, not only is that more in line with the RCD, it kinda makes it feel like there is actually a limit on the number of lights you can place.
> In effect, it doesn't actually nerf much, a proper nerf would be to change the costs of lights, but then people could just lug around more plasteel.
> 
> Nerfs the number of tube lights you can put in one go from 50 to 20.
> Nerfs the number of floor lights you can put in one go from 33 to 13.

no you're still thinking wrong about stopping grief the actual proper nerf would be to code a cooldown for glowsticks and make it so you can only use light placers if the user's mob can see where you're placing (no cameras), maybe some ammo changes as well

silicons:
this doesn't fix the issue just makes it more annoying

KathrinBailey:
> this doesn't fix the issue just makes it more annoying

There's nothing else I can do to nerf them because I don't know how to.
Removing RLDs from the box PR, this PR no longer has a purpose.

PR #9733, opened 2019-11-06
Makes summon (X) spells interact with dynamic's threat system, when applicable

Putnam3145:
## About The Pull Request

A wizard who uses summon events has caused far more chaos, so reflect that if dynamic mode is being played. There's options in game_options that reflect these values, which can be changed as seen fit.

Could add stuff for s/laughter demons too, but I'm unsure, since those are somewhat more localized while this is a global event. Course, that's just post hoc justification for me having forgotten about 'em.

## Why It's Good For The Game

Honestly more things should use dynamic's threat system *in general*, when dynamic is being played, and it's all customizable anyway.

## Changelog
:cl:
balance: In dynamic mode, summon events/guns/magic now count as, essentially, minor antags in their own way.
config: Added configuration that adjusts how much the summon spells adjust dynamic threat.
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -411,3 +411,27 @@
 /datum/config_entry/keyed_list/dynamic_high_population_requirement
 	key_mode = KEY_MODE_TEXT
 	value_mode = VALUE_MODE_NUM
+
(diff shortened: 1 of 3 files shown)


Putnam3145:
closed because subsumed into #9557

PR #9736, opened 2019-11-07
I want to have a word with whoever wrote the original antagonist reputation proc

silicons:
## About The Pull Request
 
Entirely reworks the antag rep proc to be easier to read, as well as support not giving people with no tickets antag rolls. Might have side effects because the old proc was so impossible to read I gave up reading it.
It's (probably?) slightly less performant unless you specify to return a list, instead of one by one though, in which case it'll be more performant than before. This probably isn't an issue because it wasn't that much better before.

## Why It's Good For The Game

Just code changes

## Changelog
:cl:
code: Antag rep proc is now easier to read and supports returning a list.
/:cl:

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -305,48 +305,73 @@
 // The odds become:
 //     Player A: 150 / 250 = 0.6 = 60%
 //     Player B: 100 / 250 = 0.4 = 40%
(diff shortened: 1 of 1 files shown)


silicons:
this actually allows config for free tickets to be set to 0 or even negatives

no one's going to be that evil though right haha

nik707 (approved):

PR #9707, opened 2019-11-04
Port Welding Hard Hats from /tg/

BonniePandora:
## About The Pull Request

Ports across the welding hard hats from /tg/ in various colours along with their corresponding action for the welding shield and addition to the Engi + Atmos 'drobes

## Why It's Good For The Game

Gives engineers and atmos-techs an early game alternative to having to swap between hard hat + welding mask or engineering scanner goggles/optical t-ray goggles + welding goggles.

As they're limited in number, a full engineering department will likely not have enough to outfit the entire team, but can be suplemented with welding gas masks once the research for them has been completed.

Also they're cool as hell who doesn't like looking like a budget Judge Dredd

## Changelog
:cl:
modules/clothing/head/Hardhat.dm:
- Added Welder Hard Hats & various colours
--They function as a combi hard hat & welder mask, but have a slightly dimmer head lamp

modules/clothing/vending/wardrobes.dm
- Added 3 Welder Hard Hats to the Engi-drobe and Atmos-drobe
- Equalized some values between the two (3 atmos winter coats now instead of 5, and 3 engineering dufflebags instead of 2)

game/objects/structures/crates_lockers/closets/secure/engineering.dm
- Replaced the welding mask in the CE's locker with a white WHH (This one has brighter buld in it! Same brightness as a regular hard hat)

datums/action.dm
- Added the item_action to lower the WHH's face shield
icons/mob/head.dmi
- Added weldervisor + welderhat, which are used by the WHH's
- Recoloured hardhat0/1_dblue to be in atmos colours as I've never seen it on the station before and this way is actually used

icons/mob/obj/hats.dmi
- Added weldervisor, again, used by the WHH's
- Repositioned the hardhat sprites up several pixels so they don't clip and look weird with the weldervisor
- Recoloured hardhat0/1_dblue to be in atmos colours as I've never seen it on the station before and this way is actually used
/:cl:

PR #9752, opened 2019-11-09
Coming out of the closet: Adds empty versions of ID-locked closets for mappers, buffs welding closet.

KathrinBailey:
## About The Pull Request
• Adds empty versions of many lockers that have ID locks to give extra freedom to mappers.
• Replaces welding closet contents with 2 welding helmets, 2 welding helmets and 4 industrial welding tools (instead of 3 helmets and 3 welding tools) as engineers spawn with industrial welders, the basic ones are just junk and probably shouldn't be held tightly under the high security of an ID lock.
• Adds a closet that's a gimmick for KatStation. Doesn't affect anything else at all, just a new closet, no one else has to use it or anything.

If people would rather the closet remain unchanged I'll undo that and keep as is.

## Why It's Good For The Game
Gives more freedom for mappers. Should be very useful for events, space ruins and unique stations.
Examples:
• Use the empty atmos locker and make a new instance, rename it "RPD Storage" and place 4 RPDs over it, now you have a locker that only atmos techs can open and it has exactly what you want inside.
• Make a more gimmick-y engineering closet for an old station space ruin, before more advanced tech like mesons and holobarriers were invented. A custom loadout can now be added without having to make a PR to add an entirely new closet.
• Have a destroyed medical ship space ruin, with some medical loot that you don't want just anyone to access easily, so you can now place it in an empty medical locker. Now only those with access can get in.
• Put AEDs in a medical closet and rename it to AED closet. Now you have your AEDs protected and you don't also have to have them stored with a bunch of other junk.

Not having empty versions of the lockers seemed like an oversight. I guess you could technically cheese some of them by just changing icon states, but this should just be cleaner. 

## Changelog
:cl:
add: Empty versions of ID-locked closets for mappers to put custom loadouts in them.
add: Empty versions of some lockers so there should be an empty version of the locker for its item state.
add: Gimmick closet for the QM for KatStation to recruit miners. No one has to use it if they don't want to. It's just a QM locker with the miner locker icon state and pre-filled with gear tailored towards turning non-miners into equipped miners, equips them better than the existing conscription kit.
tweak: Adds maid uniform to the janitor closet.
balance: Replaces the current welding closet contents with two welding helmets, two goggles, and four industrial welding tools (red ones). It is ID-locked to engineering access. Normal welding tools in here a pointless since they can be obtained by public vendors.
/:cl:


dapnee:
QM can already order a crate that does the same thing as your locker, plus miners can buy kits out of the mining vendor which is basically the same thing.

KathrinBailey:
> QM can already order a crate that does the same thing as your locker, plus miners can buy kits out of the mining vendor which is basically the same thing.

I know, the locker isn't something to be put on Box/Meta/Delta/Pubby, it was an edit I made ages ago for my map that I just copy and pasted to keep. You don't have to place it or use it on your own map, it's tailored to a very specific gimmick. If it existing bothers people I can just remove it until its time for the map to be tested.

Weblure:
Options are good, I see no reason not to merge this.

KathrinBailey:
> Options are good, I see no reason not to merge this.

Thanks!
Now that [welding hard hats](https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9707) have been merged, I could add them to the closet if anyone wants, but I won't change anything now unless requested.

silicons (changes requested):
remove all the duplicated code please

example:

/obj/structure/closet/secure_closet/miner
name = miners equipment
icon state = mining
req access = list(seomthing something)

and then you did 
/obj/structure/closet/secure_closet/miner/empty
and all that again

please remove duplicate/already done definitions, it'll be inherited when you do the subtype

so you'd only need to define /obj/structure/closet/secure_closet/miner/empty and then override PopulateContents to not do anything

also you didn't actually make said closets empty because populatecontents was never overridden, therefore it'll inherit it from its parent which is the full one.

PR #9756, opened 2019-11-09
turns on the chat subsystem (#47653)

Ghommie:
## About The Pull Request
Holy crap it was not initialized and was using the fallback old system as a result, that's why buckshot still lagged so much.

## Why It's Good For The Game
Fixing bad ports.

## Changelog
Nothing.


GrayRachnid:
amazing

silicons (approved):

PR #9833, opened 2019-11-17
Ports "Ban system and interface update" and its following updates

Ghommie:
## About The Pull Request
Porting tgstation PR #41176 by Jordie0608 (and post merge fixes by the same man). #42861 by MrStonedOne. Plus bits from successive PRs by kingofkosmos that dealt with missing message spans etcetera.

DNM tag atm because: 
- It requires the host (or anyone with such database access) to update the Schema.
- While I'm pretty sure sql database code stuff in the repository was left untouched since note/ban severity levels were added, I'm more than quite unaccustomed to it.
- This is the sort of thing that you do NOT want to go through a testmerge.

## Why It's Good For The Game
The new ban system interface is :ok_hand:, also better stay updated as much as possible, knowing the sync is dead and many contributors prefer to bother themselves with the small stuff.

## Changelog
:cl: Jordie0608, MrStonedOne.
admin: Ban interface rework. The banning and unbanning panels have received a new design which is easier to use and allows multiple role bans to be made at once.
prefix: Ban search and unbanning moved to unbanning panel, which is now a separate panel to the old banning panel.
admin: Some tweaks to the ban interface:
add: Bans apply to computer_ids by default
add: Bans only apply to IPs by default if its a permaban.
admin: Note: The behavior of the "Use IP and CID from last connection of key" box has been changed. It now only fills in checked values, where as before it would error if CID or IP was check.
/:cl:


Ghommie:
This goes on my wall of shame.

PR #9856, opened 2019-11-18
Refactoring code to allow all living mobs to use shields.

Ghommie:
## About The Pull Request
Moved down most of check_shields() and check_reflect() from human to living. Had to go through all attack types, bullect_act and hitby to make sure they all checked the parent return value when possible.
Cleaned some old code and oversights. A few istances hacky checks here and there to make sure the feature works for all living mobs capable of holding items.
I did test this locally and it works. But you still may want to review it cause I preferred to preserve functionality over code quality.

I also refactored a chunk of the embedding code down to carbon, for now it's still humans only but I plan to move it all down to carbon level later.

## Why It's Good For The Game
Allows all living mobs of benefiting from shields.

## Changelog
:cl:
add: Refactored code to allow all living mobs to use shields and not only humans.
tweak: Monkys will now retaliate against aliens attacking them (as if they even posed a threat to start with).
/:cl:


silicons (dismissed):

silicons (commented):
@deathride58

Linzolle (approved):

PR #9863, opened 2019-11-19
No adrenals for borgs please

BlackMajor:
Simply disables borg sprinting while vtec is installed (and fixes a bug where vtec could be installed multiple times).

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -74,7 +74,7 @@
 /obj/item/borg/upgrade/vtec/action(mob/living/silicon/robot/R, user = usr)
 	. = ..()
 	if(.)
-		if(R.speed < 0)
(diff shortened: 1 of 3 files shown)


Tename:
Ya, Vtec doesn't even really seem noticeable without sprint and I don't really want to be spamming the Vtec change speed button at the top left instead of just hitting shift.

BlackMajor:
@Tename 
The alternative is I just make sprinting while vtec is installed cost 0 charge.

Which now that I think about it I might do.

Tename:
> @Tename
> The alternative is I just make sprinting while vtec is installed cost 0 charge.
> 
> Which now that I think about it I might do.

that would probably be better, unless you add a easy hotkey for changing vtec speed :/

ghost:
@BlackMajor daily reminder to fix your code

nik707 (approved):

PR #9864, opened 2019-11-19
Able to use Legion Cores in-hand

Knouli:
## About The Pull Request
Hacks in a new proc for if the player uses a legion core in their hand rather than clicking on themselves. This is equivalent to clicking on the self, with nearly all the same checks as the original, skipping out on the user's death and distance - as the player would not be able to activate the legion core in-hand in the first place if they were dead, and if self healing, the distance would be null anyhow.

This is also my first hack for SS13. Yell at me and tell me how I messed it up.

Edit: Logging is now incorporated in the scope of this PR. With unique logs for all three use cases of legion core healing to be saved via LOG_ATTACK.

## Why It's Good For The Game
The lavalands megafuana sprites are huge, big-huge, enough to swallow up and block the character if they are just above them. This creates a frustrating scenario where you want to heal with a core but cannot access your own sprite as it is overlapped by the mega. This circumvents that issue as you can now instead use the legion core in hand for the same effect as if clicking on yourself.

Edit: Logging will improve round reconstruction and provide better insight for administrators performing after-action reviews.

## Changelog
:cl:
add: attack_self proc for the legion core which triggers a self-heal al la the previous 'afterattack' proc, as if clicking on the character's own sprite to self-heal
add: admin logging for all three use cases of legion core healing - afterattack, attack_self, and implanted ui_action_click
/:cl:

### DIFF
--- a/code/modules/mining/equipment/regenerative_core.dm
+++ b/code/modules/mining/equipment/regenerative_core.dm
@@ -59,6 +59,7 @@
 		to_chat(owner, "<span class='notice'>[src] breaks down as it tries to activate.</span>")
 	else
 		owner.revive(full_heal = 1)
+		owner.log_message("[owner] used an implanted [src] to heal themselves! Keep fighting, it's just a flesh wound!", LOG_ATTACK, color="green") //Logging for implanted legion core use
 	qdel(src)
 
 /obj/item/organ/regenerative_core/on_life()
@@ -85,6 +86,7 @@
 				SSblackbox.record_feedback("nested tally", "hivelord_core", 1, list("[type]", "used", "self"))
 			H.revive(full_heal = 1)
 			qdel(src)
+			user.log_message("[user] used [src] to heal [H]! Wake the fuck up, Samurai!", LOG_ATTACK, color="green") //Logging for 'old' style legion core use, when clicking on a sprite of yourself or another.
 
 /obj/item/organ/regenerative_core/attack_self(mob/user) //Knouli's first hack! Allows for the use of the core in hand rather than needing to click on the target, yourself, to selfheal. Its a rip of the proc just above - but skips on distance check and only uses 'user' rather than 'target'
 	if(ishuman(user)) //Check if user is human, no need for distance check as it's self heal
@@ -97,6 +99,8 @@
 			SSblackbox.record_feedback("nested tally", "hivelord_core", 1, list("[type]", "used", "self"))
 		H.revive(full_heal = 1)
 		qdel(src)
+		H.log_message("[H] used [src] to heal themselves! Making use of Knouli's sexy and intelligent use-in-hand proc!", LOG_ATTACK, color="green") //Logging for 'new' style legion core use, when using the core in-hand.
+
 
 /obj/item/organ/regenerative_core/Insert(mob/living/carbon/M, special = 0, drop_if_replaced = TRUE)
 	. = ..()
--- a/code/modules/mining/equipment/regenerative_core.dm
+++ b/code/modules/mining/equipment/regenerative_core.dm
@@ -86,6 +86,18 @@
 			H.revive(full_heal = 1)
 			qdel(src)
 
+/obj/item/organ/regenerative_core/attack_self(mob/user) //Knouli's first hack! Allows for the use of the core in hand rather than needing to click on the target, yourself, to selfheal. Its a rip of the proc just above - but skips on distance check and only uses 'user' rather than 'target'
+	if(ishuman(user)) //Check if user is human, no need for distance check as it's self heal
+		var/mob/living/carbon/human/H = user //Set H to user rather than target
+		if(inert) //Inert cores are useless
+			to_chat(user, "<span class='notice'>[src] has decayed and can no longer be used to heal.</span>")
+			return
+		else //Skip on check if the target to be healed is dead as, if you are dead, you're not going to be able to use it on yourself!
+			to_chat(user, "<span class='notice'>You start to smear [src] on yourself. It feels and smells disgusting, but you feel amazingly refreshed in mere moments.</span>")
+			SSblackbox.record_feedback("nested tally", "hivelord_core", 1, list("[type]", "used", "self"))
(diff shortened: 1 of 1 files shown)


DeltaFire15:
Can't you still use things inhand if stamcrit?

r4d6:
This make using the Legion Core in-hand the same as using it on yourself.
If you were able to use it in stamcrit before, why would that have changed?

DeltaFire15:
Because attack_self is using the core inhand which afaik is possible if in stamcrit whilr attack_mob (or whatever the click proc is called) only works outside of stamcrit
iirc

r4d6:
But, isn't the core just healing physical damage? and not doing anything to Stamina?

DeltaFire15:
It uses H.revive(full_heal = 1)
Which is pretty much a full heal / adminheal
Which means uhh stamdamage gets healed too afaik

r4d6:
I think that you can't take items out of your backpack in crit, so unless you took the Core in your hand before falling in crit, but not using it until after being in crit.
Also, if he has time to get the core out before falling in crit, why not use it right now? I really doubt the user will fall in crit _at that exact moment_.

Arturlang:
You might want to add an check if they are stunned.

DeltaFire15:
Yah

code/modules/mining/equipment/regenerative_core.dm
```diff
@@ -86,6 +86,18 @@
 			H.revive(full_heal = 1)
 			qdel(src)
 
+/obj/item/organ/regenerative_core/attack_self(mob/user) //Knouli's first hack! Allows for the use of the core in hand rather than needing to click on the target, yourself, to selfheal. Its a rip of the proc just above - but skips on distance check and only uses 'user' rather than 'target'
+	if(ishuman(user)) //Check if user is human, no need for distance check as it's self heal
+		var/mob/living/carbon/human/H = user //Set H to user rather than target
+		if(inert) //Inert cores are useless
```
Trilbyspaceclone, line 94: Make sure to check if they are dead or not other wise it can work on dead players.

DeltaFire15: They literally left a comment in the code and in the PR why the check is gone, people dead can't use inhands anyways, can they?

silicons: how is a dead player going to use that

silicons:
sorta out of scope but add logging for legion cores if you're able

not a requirement, not going to make you do it but if it's easy for you please go right ahead

Knouli:
Without change you are unable to activate the legion core in hand via this new proc call, nor the previously existing proc call while in stamina crit. Similarly, you are able to utilize the preexisting click method to apply the legion core while stunned. (At least I was able immediately after eating a taser bolt. If this is different, I'll need a nudge in another direction.)
image

I skipped out the death check feeling it was redundant, believing that a character would not be able to activate the core in the state of death in the first place. If there's something that would say otherwise where the person activating the core on themselves could be dead while doing so, then the check would be required, and I would add it.

Logging should be achieved through this line.

SSblackbox.record_feedback("nested tally", "hivelord_core", 1, list("[type]", "used", "self"))

At least, that is my understanding of what this line represents, and is a direct carbon copy from the self-heal clause from the standard click method proc just above.

code/modules/mining/equipment/regenerative_core.dm
Thalpy: What if they're dead and it's in their hand, and they're still in their body and press z?

Thalpy:
> Logging should be achieved through this line.
> 
> SSblackbox.record_feedback("nested tally", "hivelord_core", 1, list("[type]", "used", "self"))
> 
> At least, that is my understanding of what this line represents, and is a direct carbon copy from the self-heal clause from the standard click method proc just above.

That's usually for statbus, I presume Kev wants it added to admin logs. Here's an example:

`log_game("FERMICHEM: [M]'s possesser has been booted out into a astral ghost!")`

silicons:
oh i mean it should be in attack logs

silicons:
log_attack() i believe or something

no worries if you're too busy it's not really within scope of pr it's just a nice thing to have

code/modules/mining/equipment/regenerative_core.dm
Knouli: This change only influences cores held in the hand, not implanted cores. Implanted cores will retain preexisting functionality - which is to activate automatically without player activation upon threshold of entering into crit, if I'm not mistaken. This pull utilizes a widespread proc, attack_self, which is used by near any item I was able to view in the code already - which ought to have handling for dead characters attempting to use items.

Which is all to say, upon death, you drop the core you held in hand.
image


In the case of the use of an anti-drop implant, the standing proc, attack_self has handling in-built already to prevent use of the core, even in death.
image


There presently stands no logical situation where a dead character will have a core in-hand, and able to use it. There exists code already to prevent the dead from using the core and a death check is unnecessary outside of redundancy.

Knouli:
At Kevinz's request logging as been added to the scope of this PR. I am not certain of the conventions that go into logs, their colors, and their significance, but I got some simple attack logs that will denote between different use cases. Be it from clicking sprites via the 'old' proc, making use of the PR's use in hand via the 'new' proc, or using an implanted legion core to heal.

image


Additionally, to debunk all notions of this PR changing how legion cores work with the dead using it on their selves, here is a picture of a very dead kitsune with both a legion core implanted into their chest, and a legion core stuck in their hand via an anti-drop implant. Take note, they are very dead and not coming back.

image


Finally, through the course of testing this PR, I came across the realization that the procedure for implanting a legion core is bugged - where the surgeon will attempt to implant the legion core into the chest via the organ manipulation surgery. Upon reaching the final step, implantation, the progress bar will appear, but upon completion, instead of implanting the legion core, it will instead use it immediately upon the patient, as if applied externally, instead of implanting the core itself. Testing for this particular use case was performed via VV edits to directly add the legion core organ.

silicons (approved):
the log procs are silly

PR #9882, opened 2019-11-20
[READY] Allows addition of reagents to sleepers for injection and detection of failed organs (minor GUI tweak.)

Thalpy:
## About The Pull Request

This allows any medicine reagent to be added to the sleepers, with the traditional chems being synthesizeable within the machine, instead of a button injection. This is done this way to avoid conflicts on upgrading and purity. In addition to normal upgrade function, increasing the matter bin increases the capacity of the sleeper, and the manipulator allows the injection of 10, 5, 1 unit respectively, with a fully upgraded unit displaying purity of the chem too.

Generally, only medicines can be added by slapping the sleeper with a beaker/reagent holder, except in a few cases: If the machine is emmaged or if you have CMO access, it will accept any chem.

EMPs cause a reagent from the synthesize list to randomise. Upgrading fixes this.

Finally, only the CMO can purge chems from the sleeper too.

Deconstructing a sleeper will spit out it's reagents into a special reagent holder that can be used to transfer them back (so they're not lost on deconstruct).

Thanks to Persi for the neat idea, I kinda ran off with it cause I wanted to add purity effects to them later too.

(by the way the sleeper is NO_REACT for obvious reasons)

image

Additionally failing organs are highlighted (except the brain)

## Why It's Good For The Game

Adds more features to the sleeper, and hopefully lets chemists give more helpful chems in new and usual places!

## Changelog
:cl: Fermis
refactor: refactored sleepers!
/:cl:


MrJWhit:
What if there isn't a CMO to purge the system? Is the only solution deconstruct and reconstruct the sleeper?

Thalpy:
> 
> 
> What if there isn't a CMO to purge the system? Is the only solution deconstruct and reconstruct the sleeper?

Yeah that works too. I was mostly worried of meanies purging everything and making it magically disappear. I was toying with the idea of purging transferring them to a bottle instead but then you can use it as a dispenser which I don't want.

Arturlang:
Thats a lotta commits, might be worth squashing some of them for easier reading in the future.

Thalpy:
I don't know how you can squash them, also I tend to use them as waypoints incase I edit something but don't like it and want to go back.

Poojawa:
there's nothing bad about too many commits. maintainers can squash on merge anyway.

silicons (approved):

PR #9861, opened 2019-11-19
Blackpowder bugfix. Alternative to #9884

Useroth:
## About The Pull Request

Alternative to https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9884

Fixes the problem of blackpowder exploding where the explosion reaction _happened_, instead of exploding where it is after the explosion delay happens.

## Changelog
:cl:
fix: blackpowder now explodes where its actual location is, instead of where the reaction happened
tweak: containers that had blackpowder reacted in them have their colour changed (if that's not enough, I can make them blink like the hot potato)
/:cl:

PR #9884, opened 2019-11-20
Black powder fix. Alternative to #9861

Useroth:
## About The Pull Request

Alternative to https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9861

Makes blackpowder explode instantly, instead of after a delay.
Imo, it's not imbalanced. The explosion is comparable to a meth explosion, which is instantaneous.

## Changelog
:cl:
fix: resolves the issues revolving around blackpowder exploding where the reaction happened, instead of where it actually is through making it explode instantly
tweak: the explosion delay moved from blackpowder directly into bomb cherries, to keep them functioning as intended
/:cl:


silicons:
the silly part is where you consider methsplosion had to be nerfed when people made one hit kill syringes with it
but i guess this can be testmerged for now

BlackMajor (approved):

PR #9899, opened 2019-11-21
brings back old insect sprites for anthro insects.

kiwedespars:
title.

also gives them their old diet.

and removes their weakness to flashes that allow them to be blinded through a welding mask.


Ghommie:
Be careful about totally replacing the current .dmi with an older one. You aren't only reverting the insect sprites there...

Actually, you really should be careful about indiscriminately reverting code changes by uploading older files, dude.

kiwedespars:
oh yeah.

PR #9911, opened 2019-11-22
Explosions now throw mobs around like ragdolls.

Useroth:
## About The Pull Request

Accidentially came up with this idea and thought it's funny enough to implement.

## Changelog
:cl:
tweak: explosions now throw mobs around
/:cl:


Putnam3145:
Actually, devastation-level explosions already throw people around in /mob/living/carbon/human/ex_act, and I think if you wanna implement this you're gonna wanna do it in ex_act to begin with?

Useroth:
I re-used the item throwing code tbh.

Putnam3145:
Yeah, same code as the already existing code that throws mobs around when they're exploded.

If the item throwing code's in the explosion proc rather than ex_act, then yeah, it should probably go there, otherwise probably in ex_act?

MrJWhit:
Any chance you could make a gif of it in action, so I can watch things get flung around by autistic monkies?

silicons (changes requested):
do a static typecache, explosion performance is really important

Ghommie (approved):
Ok. Even though I don't like the extra loop.. You should do what Kev said.

Ghommie:
Staled.

PR #9914, opened 2019-11-22
Showers Redux: Remastered: II: Son of Shower

lolman360:
## About The Pull Request

showers now use water and splash water on their tile

slimepeeps are now cooled down by showers a LOT
## Why It's Good For The Game

preparation for plumbing showers

## Changelog
:cl:
tweak: showers now use water really
balance: jellypeople get cold in water
/:cl:


code/game/objects/structures/watercloset.dm
```diff
@@ -322,8 +331,8 @@
 	O.remove_atom_colour(WASHABLE_COLOUR_PRIORITY)
 	if(isitem(O))
 		var/obj/item/I = O
-		I.acid_level = 0
-		I.extinguish()
+
+
```
Linzolle, line 335: If you're going to remove this, then also remove the first two lines. They don't do anything if the latter two lines are gone. There's also no need to keep the whitespace here.

Putnam3145:
i'm marking this as stale but should also note that it has literally never compiled

PR #9924, opened 2019-11-23
Chameleon Nanopaste (closed for further testing)

Seris02:
## About The Pull Request

Adds chameleon nanopaste to the game. It's an item that can hide items by spreading nanite paste over them. I did this because it seemed like a neat idea.
The disguised items flicker sometimes though.
It's available under devices for 7 TC.

## Why It's Good For The Game

Because it seems like it would be fun, and it's another way for traitor's to hide their items in plain sight.

## Changelog
:cl:
add: added chameleon nanopaste
imageadd: added icons for chameleon nanopaste
/:cl:


Ghommie:
This feature has several issues and doesn't ever take in account items that can update their icons.
You could use alternate appearances or add a dummy item container instead, but without appropriately modifying the code to fit such concept, except them to be just as flawed.

Seris02:
That has an easy fix
or not

Seris02:
nah it's fine. If it's transforming, then it would take a bit for the nanites to re-cover it, and it calls flicker every now and then which will reset it again.

Seris02:
closed for now.

PR #9932, opened 2019-11-24
Make organs no longer radiaoctively contaminable.

Putnam3145:
## About The Pull Request

People are claiming genitals are causing permacontamination. The fix is a single line, so I've made the fix. I don't know if it's the actual problem, but testing it is trivial, so here you go.

## Why It's Good For The Game

Permacontamination is bad.

## Changelog
:cl:
fix: Organs can no longer be radioactively contaminated.
/:cl:

### DIFF
--- a/code/modules/surgery/organs/organ_internal.dm
+++ b/code/modules/surgery/organs/organ_internal.dm
@@ -28,6 +28,7 @@
 	var/now_fixed
 	var/high_threshold_cleared
 	var/low_threshold_cleared
+	rad_flags = RAD_NO_CONTAMINATE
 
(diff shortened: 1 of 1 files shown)


Poojawa:
Nerfing Big Dick Energy, smh

Linzolle (approved):

Ghommie:
Yet organs are actually moved into nullspace when inserted inside mobs, in order to prevent them from being affected the same way other items are.
Whoever claimed that probably has little to no knowledge about this castle of cards, nor gave organs code a scrutiny to back their points.

Though, it may indeed be another sort of object found inside the mob contents but not in their inventory, such as certain implants' radios/storage/gps (latter should be componentized like tgstation did months ago). Or cavity implants- FUCK I WAS TOO SLOW, IT GOT MERGED ANYWAY!

PR #9941, opened 2019-11-25
Personal closets can use PDAs/tablets now

### LINKED ISSUE #9929: Personal Closets don't recognize IDs inside PDAs, only IDs.
## Round ID:
N/A

## Testmerges:
N/A

## Reproduction:
Just put your ID in a pda and try accessing your own personal closet. Don't try this with all access.

Putnam3145:
## About The Pull Request

Fixes #9929

## Why It's Good For The Game

A bit of consistency can't hurt. It's not the only thing that can't use PDAs, only ID cards, and there's *lots* of stuff that can't use tablets, but whatever, one step at a time.

## Changelog
:cl: Putnam
fix: Personal closets can use anything that holds an ID card now.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/personal.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/personal.dm
@@ -9,11 +9,22 @@
 	if(registered_name)
 		to_chat(user, "<span class='notice'>The display reads, \"Owned by [registered_name]\".</span>")
 
-/obj/structure/closet/secure_closet/personal/check_access(obj/item/card/id/I)
+/obj/structure/closet/secure_closet/personal/check_access(obj/item/I)
 	. = ..()
 	if(!I || !istype(I))
 		return
-	if(registered_name == I.registered_name)
+	if(istype(I,/obj/item/modular_computer/tablet))
+		var/obj/item/modular_computer/tablet/ourTablet = I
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/crates_lockers/closets/secure/personal.dm
```diff
@@ -9,11 +9,22 @@
...
+		if(card_slot)
+			return registered_name == card_slot.stored_card.registered_name || registered_name == card_slot.stored_card2.registered_name
+	var/obj/item/card/id/ID
+	if(istype(I,/obj/item/pda))
+		var/obj/item/pda/ourPDA = I
+		ID = ourPDA.id
+	else if(istype(I,/obj/item/card/id))
+		ID = I
+	if(registered_name == ID.registered_name)
 		return TRUE
```
Ghommie: Just replace all this code with the following, dude.
```
	var/obj/item/card/id/ID = I.getID()
	if(!ID || registered_name != ID.registered_name)
		return FALSE
	return TRUE
```

GrayRachnid:
can you add wallets too

code/game/objects/structures/crates_lockers/closets/secure/personal.dm
Putnam3145: doesn't actually work for tablets as such, since it only checks the one ID for that, but yeah, apparently so

Linzolle (approved):

PR #9950, opened 2019-11-26
Nerfs the Clockwork Marauder meta

silicons:
## About The Pull Request

Marauders are now on a 3 minute cooldown per summon unless summoned on Reebe.
Marauders now take 6 damage per second in space or 4 while next to visible space (out of their 120 health)

Marauders also glow a bit brighter because why not.

## Why It's Good For The Game

Station clockwork marauder spam is getting a too much. Frankly I want to just revert clockcult back to station clockcult rather than Reebe Greytiding Simulator but until we have better plans for how clockcult should be working we might as well do this.
10 tile per second mobs running around in space smashing every window they can is ridiculous. If you want to do that, have a cultist steal a space suit and fire axe. 
Marauders are still a powerful thing to unleash on station, just not spammable waves of death.

## Changelog
:cl:
balance: Clockwork marauders are now on a configured summon cooldown if being summoned on station. They also rapidly bleed health while in or next to space. And they glow brighter.
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -384,5 +384,5 @@
 	config_entry_value = 1800
 	min_val = 0
 
-/datum/config_entry/flag/allow_clockwork_marauder
+/datum/config_entry/flag/allow_station_clockwork_marauder
 	config_entry_value = TRUE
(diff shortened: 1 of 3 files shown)


silicons:
CLARIFICATION: By next to visible space, I mean oview(1). That means if they're directly next to a space tile they can see. E.G., they breach a window, and run onto the spot the window was at.

silicons:
added a config option for disabling station marauders on request.

Putnam3145 (approved):

Ghommie:
Not funny, but I'll let it pass.

PR #9983, opened 2019-11-29
refactoring altclick interaction to allow parent calls without forcing the turf contents stat menu open.

### LINKED ISSUE #9758: You actually cannot alt+click a tile again to close the floor menu list
## Round ID:
N/A

## Testmerges:
N/A

## Reproduction:
Alt click a tile, it shows you a list of items. Alt click it again, it doesn't close. In the past, you could alt click again to close it. Now, this isn't an issue for your normal everyday player because all you have to do is move away from the tile and it'll auto-close, but. You can't do that as AI. If an AI alt clicks a tile it permanently keeps that tile open and in vision, which is somewhat intended, but you can't close it after which is annoying because sometimes I don't want a tile permanently open due to the fact that it takes an extra tab space.

Ghommie:
## About The Pull Request
Moving the "open turf contents stat menu" code away from Altclick() and into its own specific proc.
Giving most Altclick() procs parent calls and return values to prevent interactable objects from opening said stat menu...
Otherwise Altclickon() (and its variants) will proceed to call the aforementioned proc, which now also contains additional location checks (inspired from tgstation #47597 by Skogol) to prevent objects inside storages from opening the menu, plus the option to close the tab by altclicking on the same turf again.

## Why It's Good For The Game
This will close #9758, and fix an annoying aspect of the game that has been lingering for a lot of time now.

## Changelog
:cl: Ghommie, Skogol
refactor: refactored altclick interaction to allow alt-click interactable objects to parent call without forcing the turf contents stat menu open.
tweak: Alt clicking will no longer show turf contents for items inside bags etc.
tweak: Alt clicking the source of your turf contents stat menu will now close said menu.
/:cl:


Ghommie:
Tested, looks ready.

code/game/objects/items/storage/belt.dm
```diff
@@ -757,23 +757,13 @@
 	STR.rustle_sound = FALSE
 	STR.max_w_class = WEIGHT_CLASS_BULKY
 	STR.can_hold = typecacheof(fitting_swords)
+	STR.quickdraw = TRUE
 
 /obj/item/storage/belt/sabre/examine(mob/user)
 	. = ..()
 	if(length(contents))
 		. += "<span class='notice'>Alt-click it to quickly draw the blade.</span>"
```
Linzolle, line 765: remove this line if you're going to remove the AltClick quickdraw.

Linzolle:
Looks ready to merge after that one fix with the sabre sheath. edit nevermind

code/game/objects/items/storage/belt.dm
Linzolle: wait a second, never mind. I misunderstood how the quickdraw var works.

Linzolle (approved):

PR #9995, opened 2019-11-29
[Testmerge] Comments out Tinfoil hat until it's fixed

YakumoChen:
## About The Pull Request
Makes you unable to print Tinfoil Hats from the autolathe

## Why It's Good For The Game

The new holymelon, Tinfoil hats are broken in such a way they give you a random brain disability instead of the intended one (Phobia of command) in exchange for anti-magic, something several people are abusing to become immune to cults and wizards (and abductors)

### DIFF
--- a/code/modules/research/designs/autolathe_desings/autolathe_designs_sec_and_hacked.dm
+++ b/code/modules/research/designs/autolathe_desings/autolathe_designs_sec_and_hacked.dm
@@ -177,10 +177,10 @@
 	build_path = /obj/item/kitchen/knife/butcher
 	category = list("hacked", "Dinnerware")
 
-/datum/design/foilhat
-	name = "Tinfoil Hat"
-	id = "tinfoil_hat"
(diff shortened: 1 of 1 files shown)


GrayRachnid:
HAHAHHAHAHAHAHAHA

no

Xantholne:
Cult have an absurd winrate as of late, especially clock cult having a hat that you cannot remove that also gives you a phobia of certain things and related things is the only solid counter to cults that rely heavily on magic.

That, and how hard cult snowballs makes them a trade off worth taking against a 30 second long stun/mute that you cannot counter.

ghost:
>multiple short comments

on other note i'm ok with this i guess but we need a countermethod against cults if we're removing this

Linzolle:
uhh that's not even true. it doesn't work like that. foilhats do NOT protect you from magic.

YakumoChen:
it's holy water, not this, but I'll be looking into that next

PR #10001, opened 2019-11-29
im late to the quads or 10k party but i can still make a bad pr

Hatterhat:
makes every gun autofire by default if you hold down lmb


hollyhocc:
that's a wonderful idea

GrayRachnid:
madlad

PR #10008, opened 2019-11-29
adds sleeping carp wrist wrench/violent grab hallucination

Seris02:
## About The Pull Request

Adds a sleeping carp hallucination which picks from the mobs around and then 50/50 wrist wrench / violent grab.

## Why It's Good For The Game

Because I think it would be fun to see people going CARP CARP and then realising it was a hallucination because they weren't hurt (or just not realising).

## Changelog
:cl:
add: added sleeping carp hallucination
/:cl:


Arturlang:
Why not make it stun for the proper time and with the same effects without the actual damage?

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,29 @@ GLOBAL_LIST_INIT(hallucination_list, list(
...
+	var/list/mobsyup = list()
+	for (var/mob/living/carbon/A in orange(C,2))
+		if (get_dist(C,A) < 2 && C != A)
+			mobsyup += list(A)
+	if (mobsyup.len == 0)
+		qdel(src)
+		return
+	var/mob/living/carbon/G = pick(mobsyup)
+	if (rand(0,1))
+		C.visible_message("<span class='warning'>[C] falls to the ground screaming and clutching [G.p_their()] wrist!</span>", \
```
Ghommie:
```suggestion
		C.visible_message("<span class='warning'>[C] falls to the ground screaming and clutching [C.p_their()] wrist!</span>", \
```
It's an hallucination, right?

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,29 @@ GLOBAL_LIST_INIT(hallucination_list, list(
...
+						  "<span class='userdanger'>[G] grabs your wrist and violently wrenches it to the side!</span>")
+		C.emote("scream")
+		C.dropItemToGround(C.get_active_held_item())
+		C.Knockdown(60)
+		qdel(src)
+		return
+	else
+		to_chat(C,"<span class='userdanger'>[G] violently grabs you!</span>")
+		qdel(src)
+		return
```
Ghommie: Remove these two qdel(src) and return, and add a single qdel(src) at the end of the proc, without return. Also new line endings (cause I plan to port travis new line endings linter at some point later), will ya?

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,29 @@ GLOBAL_LIST_INIT(hallucination_list, list(
...
+
+/datum/hallucination/sleeping_carp
+
+/datum/hallucination/sleeping_carp/New(mob/living/carbon/C, forced = TRUE)
+	set waitfor = FALSE
+	..()
+	var/list/mobsyup = list()
+	for (var/mob/living/carbon/A in orange(C,2))
+		if (get_dist(C,A) < 2 && C != A)
+			mobsyup += list(A)
```
Ghommie:
```suggestion
			LAZYADD(mobsyup, A)
```
No need to wrap the atom in a list toh.

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,29 @@ GLOBAL_LIST_INIT(hallucination_list, list(
 	H.preparePixelProjectile(target, start)
 	H.fire()
 	qdel(src)
+
+/datum/hallucination/sleeping_carp
+
+/datum/hallucination/sleeping_carp/New(mob/living/carbon/C, forced = TRUE)
+	set waitfor = FALSE
+	..()
+	var/list/mobsyup = list()
```
Ghommie:
```suggestion
	var/list/mobsyup
```

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,29 @@ GLOBAL_LIST_INIT(hallucination_list, list(
...
+	for (var/mob/living/carbon/A in orange(C,2))
+		if (get_dist(C,A) < 2 && C != A)
+			mobsyup += list(A)
+	if (mobsyup.len == 0)
```
Ghommie:
```suggestion
	if (!LAZYLEN(mobsyup))
```
Even though the macro is just a placeholder for future possibly more reliable alternatives to length(A).

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,29 @@ GLOBAL_LIST_INIT(hallucination_list, list(
...
 	qdel(src)
+
+/datum/hallucination/sleeping_carp
+
+/datum/hallucination/sleeping_carp/New(mob/living/carbon/C, forced = TRUE)
+	set waitfor = FALSE
+	..()
+	var/list/mobsyup = list()
+	for (var/mob/living/carbon/A in orange(C,2))
+		if (get_dist(C,A) < 2 && C != A)
```
Ghommie:
```suggestion
		if (get_dist(C,A) < 2)
```
should be impossible since orange doesn't check the epicenter turf.

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,29 @@ GLOBAL_LIST_INIT(hallucination_list, list(
...
 	H.fire()
 	qdel(src)
+
+/datum/hallucination/sleeping_carp
+
+/datum/hallucination/sleeping_carp/New(mob/living/carbon/C, forced = TRUE)
+	set waitfor = FALSE
+	..()
+	var/list/mobsyup = list()
+	for (var/mob/living/carbon/A in orange(C,2))
```
Ghommie: Why two tiles orange for a thing that should only happen at melee range? Also edge cases for people behind walls.

Seris02:
> Why not make it stun for the proper time and with the same effects without the actual damage?

It does

code/modules/flufftext/Hallucination.dm
Seris02: Because I wasn't sure if orange(C,1) would only get the the things on the same tile, but also you can see in the next line I make sure it's in melee range.

code/modules/flufftext/Hallucination.dm
Seris02: It does
I know it does because while I was coding blood red slimes I filled a space with walls using orange and it filled the center.

code/modules/flufftext/Hallucination.dm
Seris02: oh woops, that was a typo, thanks

code/modules/flufftext/Hallucination.dm
```diff
@@ -1294,3 +1295,26 @@ GLOBAL_LIST_INIT(hallucination_list, list(
...
+	..()
+	var/list/mobsyup
+	for (var/mob/living/carbon/A in orange(C,1))
+		if (get_dist(C,A) < 2)
+			LAZYADD(mobsyup,A)
+	if (!LAZYLEN(mobsyup))
+		qdel(src)
+		return
+	var/mob/living/carbon/G = pick(mobsyup)
+	if (rand(0,1))
```
Putnam3145: this is unclear, should be prob(50)

Putnam3145: should probably be prob(50) for clarity

Putnam3145 (changes requested):

code/modules/flufftext/Hallucination.dm
Seris02: I mean ok. But also. Isn't it clear enough?

Putnam3145 (approved):

PR #10017, opened 2019-11-30
Makes filters volume-based

Putnam3145:
## About The Pull Request

Makes filters pump like volume pumps instead of like gas pumps. Also makes gas filters push to each output independently: filtered gas ALWAYS goes to filtered output and unfiltered gas ALWAYS goes to unfiltered output, if either of those can't be pushed to it just leaves it in the input.

## Why It's Good For The Game

It'll cause a few less supermatter disasters, for one. For another, it's kinda weird that filters were *literally pessimal* as far as throughput goes (the only thing worse than a gas pump is a passive gate, and those are equal in the case of much higher input pressure).

## Changelog
:cl:
balance: Gas filters now push gas the same way volume pumps do.
balance: Gas filters now won't clog if only one output is clogged.
/:cl:


Arturlang:
When are we going to replace filters with scrubber filterinng

Linzolle (approved):

PR #10022, opened 2019-12-01
techfabs are now easier to investigate

Sishen1542:
~title~

ok so basically now you can investigate research to find any printed items, due to admin message spam problem


Putnam3145:
this is gonna get spammy as hell at times, probably, but the code checks out

Sishen1542:
fixd

Putnam3145 (approved):

PR #10035, opened 2019-12-03
Commenting out pAI circuitry

kiwedespars:
## About The Pull Request

circuitry is pretty op. removes the pAI part of it so you have to still manually activate your circuits.

## Why It's Good For The Game

prevents people from being able to have someone else think for their door bolting/flameshooting/item grabbing bullshit.

## Changelog
:cl:
del: commented out pAI circuitry, probably.
/:cl:

### DIFF
--- a/code/modules/integrated_electronics/subtypes/smart.dm
+++ b/code/modules/integrated_electronics/subtypes/smart.dm
@@ -249,7 +249,7 @@
 		idc.access = Pl
 
 // - pAI connector circuit - //
-/obj/item/integrated_circuit/input/pAI_connector
+/* /obj/item/integrated_circuit/input/pAI_connector
 	name = "pAI connector circuit"
 	desc = "This circuit lets you fit in a personal artificial intelligence to give it some form of control over the bot."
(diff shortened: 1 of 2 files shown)


ShadeAware:
An unfortunate hit to a lot of roleplaying circuit devices, but I suppose it is a far better solution than removing cloning or the system entirely

ShadeAware:
On that note, prehaps a different circuit that simply allows you to put a pAI IN the device without giving it any control would be a good addition as a backup

EDIT: Now that I think about it more, would it be possible to simply comment out the activators of the circuit? That way, you preserve currently printable circuits (mostly) allowing a pAI to be put in while still neutering non-roleplay oriented circuits. Since, you know, they can't _activate_ things anymore.

silicons:
i'm sure glad i'm ""on break"" and not touching this one lmfao

Ghommie:
I'd rather hinder pAI circuits salt potentials (I already halved down the locomotion circuit speed) than let these PRs stay open forevermore.

PR #10062, opened 2019-12-06
[TESTMERGE ONLY]Revert "Ports "Slightly improves visible and audible messages."" - the pAI nerf we needed but didn't deserve

silicons:
Reverts Citadel-Station-13/Citadel-Station-13#10056


Putnam3145:
my reign of terror already claims a casualty. truly there is a lesson to be learned here

PR #10064, opened 2019-12-06
Fixes silicon hearing - "Wow I do good work when I'm not literally spamming reverts without looking, I didn't even make the reverts for the right PRs"

silicons:
fixes silicons not being able to hear :^)


Ghommie:
Either ways, this is my fault cause I was the one that touched their flags_1.
SPEEDMERGE IT NOW!!

Linzolle (approved):

PR #10091, opened 2019-12-07
Reduces the blue, amber and red alert sounds volume.

Ghommie:
## About The Pull Request
These alert sounds are just too fucking loud I can hear them from the opposite side of this room, coming out of my headphones, with the mixer volume set to less than 15%.
Did not actually touch the delta alert sound, it's already quiet compared to these wankers.

## Why It's Good For The Game
Quieting these annoying sounds down.

## Changelog
:cl:
soundadd: Blue, Amber and Red security alert sounds should be half as loud now.
/:cl:


Putnam3145:
honestly yeah i can go for this
also fun fact half as loud is only -3 decibels
also fun fact NAME YOUR PRS IN A WAY THAT ACTUALLY DESCRIBES THEM SO THE HISTORY ISN'T ALL SHIT UP please

silicons (approved):

PR #10095, opened 2019-12-07
Buffs clown ops by removing their clumsiness and adding a new trait to be used in place of several clown role checks.

Ghommie:
## About The Pull Request
What's said on the tin. Not removing other antags clown role checks for the clumsiness mutation cause that could screw up with clown ops admin shenanigeans dynamic whatever.

Oh, yea, this also means clown ops effectively can make use of more clowns features (such as the still backend clown car), without having to be clumsy.
Oh yea, their mood also suffer from having no clown shoes on.

## Why It's Good For The Game
The future is now, clownman.

## Changelog
:cl:
balance: Buffed clown ops by removing their clumsiness and adding a new trait to be used in place of several clown role checks.
tweak: Clown ops too also suffer from not holding or wearing clown shoes now.
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -163,6 +163,7 @@
 #define TRAIT_FRIENDLY			"friendly"
 #define TRAIT_ASSBLASTUSA       "assblastusa"
 #define TRAIT_CULT_EYES 		"cult_eyes"
+#define TRAIT_CLOWN_MENTALITY	"clown_mentality" // The future is now, clownman.
 #define TRAIT_FREESPRINT		"free_sprinting"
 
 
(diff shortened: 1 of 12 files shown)


YakumoChen:
This isn't actually the problem with clown ops. They're meant to be non-lethal, in fact. The only problem is all of their weapons were not designed for our stamina system and most of them don't even work at all. All of their toys were from a bygone era when slips were hard stuns.

Also clown ops don't have access to any of the lethal weapons explicitly because they're clumsy and the weapons wouldn't work. Make sure you're not actually fucking them over by making what gear they have unusable.

Ghommie:
Ok, I'll buff their toys in a separate PRs, meanwhile I think this should still go through cause the impairments of clumsiness go well over being unable to fire most guns.
Saying they are meant to be non-lethal, though, is such a bogus statement. They got access to many of the "lethal" items normal ops and traitors have access to, while their specific items were designed with an obvious clown theme in mind rather than a stamina combat community clichè that's growing dumb and repetitive as fuck over time.

silicons (approved):

PR #10100, opened 2019-12-07
[READY] Tweaks botany related reagent pHes so they don't negatively affect botany.

Thalpy:
## About The Pull Request

Someone who was playing botany today pointed out that I have their reagents within the water + acid range, and it's causing problems for them. I've reduced the acidity of the reagents, and set the vapor to only occur at less than 2, instead of less and equal to.

If I missed any reagents, let me know and I can look over them and/or edit them too.

## Why It's Good For The Game

I don't want to upset botanists and they shouldn't have to deal with this when it's outside of their control.

## Changelog
:cl: Fermis
tweak: tweaks botany reagent pHes
/:cl:


Putnam3145 (approved):

Thalpy:
Woah, speedrun

PR #10120, opened 2019-12-09
Night vision trait but it just boosts barely-lit visibility

silicons:
## About The Pull Request

Readds night vision as a powerful night vision effect in general, assuming you can see a tile. Also gives you a 5x5 rather than 3x3 darksight range.

## Why It's Good For The Game

Old night vision was too long ranged. This is still pretty strong, just not "I never need lights and no one can hide from me" levels of strong.

## Changelog
:cl:
add: Night vision readded as a darkness dampening effect rather than darksight.
/:cl:

### DIFF
--- a/code/__DEFINES/lighting.dm
+++ b/code/__DEFINES/lighting.dm
@@ -59,7 +59,7 @@
 #define LIGHT_RANGE_FIRE		3 //How many tiles standard fires glow.
 
 #define LIGHTING_PLANE_ALPHA_VISIBLE 255
-#define LIGHTING_PLANE_ALPHA_NV_TRAIT 192
+#define LIGHTING_PLANE_ALPHA_NV_TRAIT 160
 #define LIGHTING_PLANE_ALPHA_MOSTLY_VISIBLE 192
 #define LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE 128 //For lighting alpha, small amounts lead to big changes. even at 128 its hard to figure out what is dark and what is light, at 64 you almost can't even tell.
 #define LIGHTING_PLANE_ALPHA_INVISIBLE 0
--- a/code/__DEFINES/lighting.dm
+++ b/code/__DEFINES/lighting.dm
@@ -59,10 +59,12 @@
 #define LIGHT_RANGE_FIRE		3 //How many tiles standard fires glow.
 
 #define LIGHTING_PLANE_ALPHA_VISIBLE 255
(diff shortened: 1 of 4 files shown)


Putnam3145 (approved):

code/datums/traits/good.dm
```diff
@@ -188,3 +188,18 @@
...
+	mob_trait = TRAIT_NIGHT_VISION
+	gain_text = "<span class='notice'>The shadows seem a little less dark.</span>"
+	lose_text = "<span class='danger'>Everything seems a little darker.</span>"
+
+/datum/quirk/night_vision/on_spawn()
+	var/mob/living/carbon/human/H = quirk_holder
+	var/obj/item/organ/eyes/eyes = H.getorgan(/obj/item/organ/eyes)
+	if(!eyes || eyes.lighting_alpha)
+		return
+	eyes.Insert(H) //refresh their eyesight and vision
```
Ghommie: What the hell? Just modify the eyes' `lighting_alpha` and `see_in_dark` right here and call `H.update_sight()` instead.

code/modules/surgery/organs/eyes.dm
```diff
@@ -1,3 +1,5 @@
+#define NIGHT_VISION_DARKSIGHT_RANGE 2
```
Ghommie: This is the damn same value of standard eyes' `see_in_dark`, kevinz.

code/datums/traits/good.dm
silicons: yeah i just copied the old code :joy: didn't know that was a thing not my code lmao one sec

code/modules/surgery/organs/eyes.dm
silicons: bruh what

Arturlang:
It seems it causes meson level nightvision in a 3 tile or so radius

code/__DEFINES/lighting.dm
```diff
@@ -59,11 +59,13 @@
 #define LIGHT_RANGE_FIRE		3 //How many tiles standard fires glow.
 
 #define LIGHTING_PLANE_ALPHA_VISIBLE 255
-#define LIGHTING_PLANE_ALPHA_NV_TRAIT 250
+#define LIGHTING_PLANE_ALPHA_NV_TRAIT 175
```
Ghommie: considering `LIGHTING_PLANE_ALPHA_MOSTLY_VISIBLE` sits at 192, this should be 215-220 at the least (quirks should never give the user significant advantages). Otherwise, it's all cool for merge.

PR #9770, opened 2019-11-11
Revert "Makes the temperature notification more realistic by making the notification look for the difference between body temperature and ambient temperature instead of only looking at body temp (#8025)"

Ghommie:
## About The Pull Request
That PR actually made temperature alerts fucked up quirky, throwing the hot alert when cold and viceversa, if the enviroment temperature was more or less nominal. Credits to Kepler PR # 982 by CoolCatIsFeelingGroovy for reminding me about this issue.

## Why It's Good For The Game
Reverting an old PR of results opposite to the intended ones.

## Changelog
:cl:
fix: fixed human hud temperature alerts.
/:cl:

Issue #9927, opened 2019-11-24
HUD Temperature icon is flipped


GrayRachnid:
## Round ID:
N/A

## Testmerges:
N/A

## Reproduction:
Shows too cold for hot and too hot for cold.

Putnam3145:
see https://github.com/Citadel-Station-13/Citadel-Station-13/pull/9770, which fixes this

MarinaGryphon:
hey it's just a logic flip, `loc_temp-H.bodytemperature` should be`H.bodytemperature-loc_temp`

MarinaGryphon:
> hey it's just a logic flip, `loc_temp-H.bodytemperature` should be`H.bodytemperature-loc_temp`

actually im dumbass

PR #10133, opened 2019-12-09
Fixes temperature alerts being backwards

Putnam3145:
## About The Pull Request

Temperature alerts were based on current body temperature, now they're based on baseline.

Closes #9770. Why was this PR even made when the actual change is ONE LINE

## Why It's Good For The Game

It doesn't say you're scorching hot when you're actually hypothermic and just moved to regular temperature anymore

## Changelog
:cl:
fix: Temperature alerts work properly now
/:cl:


GrayRachnid:
If this works it also closes #9927

deathride58:
This makes it so that things that directly modify body temperature (for example, watcher beams) don't actually affect the notification.
The proposed alternative of adding separate indicators for how your spaceman feels vs what their body temperature is would be very much worth visiting, and would be fairly easy to do

Ghommie:
> 
> 
> This makes it so that things that directly modify body temperature (for example, watcher beams) don't actually affect the notification.

This. It doesn't really fix anything.

Thalpy:
Why not just have the function consider body temp only, maybe?

Ghommie:
> 
> 
> Why not just have the function consider body temp only, maybe?

That's what it originally did before deathride modified it.

ghost:
> Why not just have the function consider body temp only, maybe?

pr for that got closed

Ghommie:
This only trades in one issue with another, and both are equally bad. Merge #9770 if you want to instead.

PR #10162, opened 2019-12-13
Removes medihounds (actually just makes them a medical aux sprite)

BlackMajor:
Same treatment that janihounds got, being compressed into their base module.

This PR does the following.
-Removes medihound as a module type
-Adds medihound sprite options for medical borgs, the same treatment Engihounds and Janihounds currently have
-Makes medical borgs have a defib roundstart
-Removes borg defibs from exosuit fabs, since it's default

-Also fixes a small bug where ejecting a patient in a voresleeper via unequipping the module meant you couldn't eat anyone again.


Arturlang:
Good change, less snoflakey mediborgos the better

silicons (approved):

PR #10170, opened 2019-12-13
kneecaps tasers - makes taser knockdown require stamloss, nerfs taser stamloss, and removes the status effect's combat mode disabling

deathride58:
## About The Pull Request
Title. Does pretty much exactly as it says on the tin, nerfs tasers to fill their specific niche. My original taser rework PR intended to make tasers fill a specific niche of pacifying targets to make it easier to incapacitate them, instead of their prior state where tasers were the end-all-be-all solution to all questions of "how do i incapacitate a spaceman". However, even with my taser rework, tasers were still way too overtuned, as they dealt way too much staminaloss and continued to remain the end-all-be-all method of incapacitation, albeit in a much stronger way due to the debuff. So this PR tackles the problem of tasers being overtuned by directly affecting all of their qualities that make them superior to disablers, while still leaving the rest of their debuff intact, allowing tasers to act as a way to make a target approachable, but stopping them from being the easiest way to completely incapacitate a spaceman.

## Changelog
:cl: Bhijn
del: Tasers no longer block combat mode
balance: Tasers now only deal 5 direct staminaloss on hit, but no longer have a hard cap
balance: Taser knockdowns now require the target to have at least 36 staminaloss. The knockdown does no extra staminaloss over the taser's base staminaloss
balance: Tasers now only add a +2 to your movement delay, down from the previous +8. This means you'll still be unable to run, but you won't be completely immobile, either.
/:cl:


GrayRachnid:
👎 After experiencing it. Still the same issue as previous iterations. The slowdown is so ridiculous that even on meth I was only able to move 2 tiles before the tase effect wore off. Anybody with two braincells can immediately break distance and disable while moving away from you.

Granted, my circumstance only applies to melee but buffing ranged weapons and discouraging melee is the last thing we should be doing. Tasers are a cockblock to ranged combat and now they cockblock melee combat too.

Edit: Not to mention, now that they don't knock you down they can just fire in your general direction instead of clicking on your spaceman.

code/datums/status_effects/debuffs.dm
```diff
@@ -105,30 +105,29 @@
...
-/datum/status_effect/no_combat_mode/electrode/on_creation(mob/living/new_owner, set_duration)
+/datum/status_effect/electrode/on_creation(mob/living/new_owner, set_duration)
 	if(isnum(set_duration)) //TODO, figure out how to grab from subtype
 		duration = set_duration
 	. = ..()
 	if(iscarbon(owner))
 		var/mob/living/carbon/C = owner
-		if(C.combatmode)
-			C.toggle_combat_mode(TRUE)
 		C.add_movespeed_modifier(MOVESPEED_ID_TASED_STATUS, TRUE, override = TRUE, multiplicative_slowdown = 8)
```
ghost: slowdown 8

deathride58:
This is now obsolete as of #10581

Ghommie:
Stale and seems to have been remade into newer PRs.

PR #10182, opened 2019-12-14
OHOHOHO, NAR'SIE? YOU DARE APPROACH ME?

KeRSedChaplain:
## About The Pull Request
Marauders will now act more akin to guardians/holoparasites than a blood cult construct. The difference between holoparas/guardians and these Marauders is that only the host can manifest them, unless said host is considerably damaged. When manifested, marauders will take damage based on how far away from their host they are, and when not manifested they will heal damage taken. When the Marauder isn't recovering and is un-manifested, they heal their host up to 65% health.

## Why It's Good For The Game
People complain about marauder spam on the station, so this should hopefully help lessen it.

## Changelog
:cl:
add: Adds in a HUD specific to Marauders.
tweak: Changes Marauders from being their own entity to being part of the cultist that summons them.
config: Comments out old Marauder defines in case someone wants to bring them back easily in the future.
/:cl:

### DIFF
--- a/code/__DEFINES/clockcult.dm
+++ b/code/__DEFINES/clockcult.dm
@@ -83,10 +83,12 @@ GLOBAL_LIST_EMPTY(all_scripture) //a list containing scripture instances; not us
 
 #define RATVARIAN_SPEAR_COOLDOWN 300 //The cooldown period between summoning another Ratvarian spear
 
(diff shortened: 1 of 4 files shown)


Xantholne:
Alternative was to just apply z level blocker, so this still has some good use for them.

KeRSedChaplain:
neat, Travis broke

ghost:
pog

KeRSedChaplain:
fucked up with making this PR so I'm gonna redo it
why do I keep making these at night

PR #10217, opened 2019-12-17
Nerfing radioactive contamination

Putnam3145:
## About The Pull Request

This nerfs radioactive contamination by making it STRICTLY reducing over time--no more back-and-forth contamination infinitely increasing forever despite conservation of energy and radiation NOT WORKING THAT WAY. It does this by doing the following:

1. Radiation waves now keep track of how much contamination they've laid down on things and will stop contaminating as soon as they run out.
2. Rather than contaminating every single atom on a tile with full strength, it contaminates each of them proportionally to how many atoms there are on the tile--for example, if there are 5 atoms on a tile, each atom will get 1/5 as much contamination as if there were only 1.

Also makes contamination no longer random--a rad pulse of sufficient strength will always contaminate a certain amount, no RNG element to it whatsoever. 

## Why It's Good For The Game

Radioactive contamination is almost always simply a bad meme. It's a complete pain in the ass to fix, is too easy to cause and leads to a bunch of really dumb stuff like supermatters making 10x as much power as they should because the chamber has been filled with monkeys. It should be noted that this doesn't *completely* destroy those strats; monkeys hold onto radioactive contamination while turfs don't, so they allow pulses to "linger" for longer, but it should no longer be the be-all-end-all of supermatter power.

## Changelog
:cl: Putnam
balance: Contamination is no longer an infinitely spreading deadly contagion causing mass panic
/:cl:

### DIFF
--- a/code/datums/radiation_wave.dm
+++ b/code/datums/radiation_wave.dm
@@ -3,6 +3,7 @@
 	var/turf/master_turf //The center of the wave
 	var/steps=0 //How far we've moved
 	var/intensity //How strong it was originaly
+	var/remaining_contam //How much contaminated material it still has
 	var/range_modifier //Higher than 1 makes it drop off faster, 0.5 makes it drop off half etc
 	var/move_dir //The direction of movement
 	var/list/__dirs //The directions to the side of the wave, stored for easy looping
@@ -18,6 +19,7 @@
 	__dirs+=turn(dir, -90)
 
 	intensity = _intensity
+	remaining_contam = intensity
 	range_modifier = _range_modifier
 	can_contaminate = _can_contaminate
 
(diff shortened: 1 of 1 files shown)


Ghommie (approved):
Sorry for being late at approving this. I did not carefully read through it at first and assumed it was just salt, but it actually makes sense.

Perhaps radiation should be buffed a later considering how it should be proportionally split across multiple atoms now. Also testmerge first, I do not want unwarranted disasters.

B1gb3ast:
So when are you actually going to differentiate between contaminating radiation and just radiation.
Because if I'm standing behind a glass wall I don't see how UV rays turn me radioactive.

Putnam3145:
@B1gb3ast i WANTED to but it would be slow as hell

Ghommie:
TODO: buff radiation effects on mobs.

PR #10224, opened 2019-12-17
Makes trait blacklisting remove random positives instead of removing every quirk (new and improved)

Seris02:
## About The Pull Request

Makes it so that if a job has a trait blacklisted, it removes random positive traits to the value of the blacklisted trait instead of removing every positive and negative trait.

## Why It's Good For The Game

Because removing everything so you don't have to code removing randoms is just lazy. Either that or they didn't think of this, and should have.

## Changelog
:cl:
tweak: made it so trait blacklisting removes random positives instead of removing everything
/:cl:


Ghommie:
I'm not going to bother to review it completely, sorry, the more I look at it the more flaws I spot, starting with the lack of specific negative and positive quirk lists, neutral quirks being potentially added to the pickquirks list, while also having an unnecessarily long code.
I may take over and write down an actual not-so-bad method after the quirk code update port is merged. No promises.

Also, did you actually test it?

Seris02:
Obviously I tested it, and I've thought of some ways to improve it, and if you look at it, you'll notice that the points for positive is positive and same for negative quirks, ie, we only need to add them on and select only the ones that are the opposite, which I've done. Neutral quirks are added to the list, but can't be selected later on. If you're going to look at the code, read it properly instead of making judgements on first sight please.
As a sidenote, you ask me if I've "actually tested it" when you don't test your own stuff. Your new quirks update killed all the traits which is a REALLY easy thing to spot if you had actually tested it. You PR stuff without testing it quite often. Don't ask me if I've tested it when you haven't tested your own shit.
BTW, I still value your reviews, just not the ones that say "I'm only going to skim through this" and then list things which aren't actually problems.

Seris02:
It compiles locally, don't know why it's erroring

Ghommie:
I did actually read it throughfull and, not skim around the lines, but I was a bit unwilling to write down a review at the time. 
Also good to know it works, sorry for crudely doubting that in my previous comment.

code/controllers/subsystem/processing/quirks.dm
```diff
@@ -75,7 +76,17 @@ PROCESSING_SUBSYSTEM_DEF(quirks)
...
+			pointscut += quirk_points_by_name(i)
+	if (pointscut != 0)
+		for (var/i in shuffle(our_quirks))
+			if (quirk_points_by_name(i) < pointscut || (pointscut < 0) ? quirk_points_by_name(i) <= 0 : quirk_points_by_name(i) >= 0)
+				continue
+			else
+				our_quirks -= i
+				cut += i
+				pointscut += quirk_points_by_name(i)
+			if (pointscut == 0)
```
silicons: <=

Ghommie (approved):

PR #10230, opened 2019-12-18
[DNM] Improves random DNA features generation and cleaning up the snowflakes.

Ghommie:
## About The Pull Request
See title.
Killed several snowflak global lists, fixed ckey restricted mutant bodyparts code from being hacky and not functional  and turned all those mutant bodyparts related DNA features text strings into defines to make things easier to handle.

## Why It's Good For The Game
Getting race-swapped into a slime person is currently a bit yucky.
DNM tag cause I gotta test it but I'm busy.

## Changelog
:cl:
refactor: Thousand lines random DNA features generation and snowflakes clean-up.
/:cl:


IronEleven:
Does this just take all the extra bits off of our default slime appearance or will we lose the ability to swap species to slime, make changes, save, then swap back to human and save again to have a go-to slime look?

Ghommie:
No, it currently only stops newly spawned humans from having some of their DNA's features (such as ears, tails etc.) from being randomized, and it shouldn't erase your preferences.
Technically, this doesn't even stop preferences from overriding these DNA features, and previously existing non-antropomorph characters will still look like weird mutants when their species is changed to another, but it should stop new characters from this madness as of now.

I'll may as well add a filter later so non-compatible dna features (such as mam_tails for humans) are not transferred from preferences to the spawned mob to stop previously existing non-anthros from also looking like garbage when race-swapped, but I'll probably restrict it to a savefile conversion to aknowledge theorical future snowflake players that want to set DNA features to stop their raceswapped snowflake anthro counterpart from looking too generic. Because changing species in the character setup doesn't reset unused DNA features, it's the species datum that dictates which features should be displayed or not.

Poojawa:
Actually, I tried to mitigate this by forcing the other bodyparts to 'none' if changed in the preference window.

This is also why I rolled back the lazy change that made humans have mam_tails listings, specifically so not every human was a demi when random spawned. If this works as you intend, good. Otherwise, welp.

Also you really ought to chill with the furry hate my dude. It's just unbecoming at this point.

Ghommie:
Ehhh, I'll get back to this soon (tm)

PR #10243, opened 2019-12-19
Lewd chem sanity pass/"ERP preferences"

Putnam3145:
## About The Pull Request

1. Puts arousal, vore into their own tab on the pref screen
2. In said tab, also adds options for: penis enlarging, breast enlarging, forced masculinization, forced feminization, hypno
3. Removes acute hepatic pharamacokinesis because its effects are fully replaced by four of the above toggles
4. Removes "lewdchem" toggle because its effects are fully replaced by "hypno" (and yes, "lewdchem" was a hypno opt-in)
5. Makes it so that both the user *and* the victim of MKUltra have to opt in to hypno for *either* to get the lewd messaging

## Why It's Good For The Game

MKUltra's caused, uh, a lot of drama recently, primarily due to the fact that it does not signal well whether it's being used for fetish or antag purposes, to the point that admins ruled that it can *only* be used for fetish purposes. This should alleviate that if not fix it.

Neither "lewdchem" nor "acute hepatic pharmakocinesis" actually told you what you were opting into/out of. I opted into "lewdchem" thinking "ah, yes, lewd doesn't sound that bad" but it's actually an opt-in to hypno, which I'm not particularly comfortable with. This is a highly inoptimal situation that our preferences should not allow. Similar is true for AHP: no description of what these chemicals actually *do*, so you don't know what you're opting out of unless explicitly told; no conveyance that the overdose has a *completely separate* fetish effect, so you don't know that you're also opting into *that*; and so on. This fixes that too.

## Changelog
:cl:
add: New tab in preferences screen: "ERP preferences"
add: New opt-outs for individual effects of incubus draught, succubus milk
del: Acute hepatic pharmacokinesis has been removed, replaced with above
tweak: Renamed "Toggle Lewdchem" to "Toggle Lewd MKUltra", since that's what it actually means, and made it toggle the "hypno" setting (rename it again if more hypno mechanics are added).
tweak: Made MKUltra's lewd messages require both people involved to have hypno opted-in.
/:cl:

### DIFF
--- a/code/__DEFINES/citadel_defines.dm
+++ b/code/__DEFINES/citadel_defines.dm
@@ -109,6 +109,11 @@
 #define MEDIHOUND_SLEEPER	(1<<0)
 #define EATING_NOISES		(1<<1)
 #define DIGESTION_NOISES	(1<<2)
+#define BREAST_ENLARGEMENT	(1<<3)
+#define PENIS_ENLARGEMENT	(1<<4)
+#define FORCED_FEM			(1<<5)
+#define FORCED_MASC			(1<<6)
+#define HYPNO				(1<<7)
 
 #define TOGGLES_CITADEL (EATING_NOISES|DIGESTION_NOISES)
 
--- a/code/__DEFINES/citadel_defines.dm
+++ b/code/__DEFINES/citadel_defines.dm
@@ -115,7 +115,7 @@
 #define FORCED_MASC			(1<<6)
 #define HYPNO				(1<<7)
 
-#define TOGGLES_CITADEL (EATING_NOISES|DIGESTION_NOISES)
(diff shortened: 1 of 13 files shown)


deathride58:
Can we get a screenshot of how the new tab looks?

deathride58 (dismissed):
It might be a good idea to add a little disclaimer at the top of the menu that these toggles are only for features explicitly designed to be erotic content. Additionally, it might be a good idea to have two prefs for hypno, one for hypno in general (affecting whether or not mkultra and gaseous mkultra even have an effect on you), and another for lewd hypno (affecting mkultra flavor text)

Putnam3145:
image

silicons (approved):

PR #10246, opened 2019-12-20
Makes mob/Stat on a 1 second timer rather than fast-as-possible to tone down highpop lag

silicons:
## About The Pull Request

title

## Why It's Good For The Game

Admins on MC/SDQL2 panel get an exception because with those you need updates as fast as possible.
Updating stat for 80 players at once is untenable. Yes, we have worse lag issues, but honestly this isn't bad enough that we need to keep this at an as-fast-as-possible rate.

## Changelog
:cl:
tweak: Stat() slowed down for anti-lag measures.
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -568,7 +568,7 @@ GLOBAL_VAR_INIT(exploit_warn_spam_prevention, 0)
 
 /mob/Stat(delayoverride)
 	. = ..()
-	var/statdelay = delayoverride || 20
+	var/statdelay = delayoverride || 10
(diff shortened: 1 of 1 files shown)


Ghommie:
Confused it with update_stat() for a moment and was about to call you a lunatic. Glad I was wrong.
May cause some awful delays on stuff such as altclicking for turf contents, but considering the server struggle it's worth a shot.

Ghommie (approved):

Ghommie:
I can only wonder how many servers Jay is running on the same server machine.

You should also run a performance benchmark to see what's really bogging the server once in a while.

PR #10247, opened 2019-12-20
ATMOS Firelocks and more

Seris02:
## About The Pull Request

Changes ATMOS Holofan's sprite to holo_fan and the name of the barrier to "holo fan". Adds Atmos Holofirelock with the holo_firelock sprite. Makes a change to the way temperature spreads so things can block temperatures if you want now.

### I tested this locally, so this is something that should definitely be testmerged first if it's even considered, as I don't know how this will affect the time dilation since I'm playing around with how temp works.

The time dilation wasn't affected during local testing.

## Why It's Good For The Game

Because having NOTHING be able to fully block temperatures is real bad.

## Changelog
:cl:
add: added atmos holofirelocks and temperature blocking
tweak: tweaked how atmos holofan looks
/:cl:


Thalpy:
My concern about this is that it will trivialise the frost punk creeping freezing temperatures in the upcoming Streetstation map.

Seris02:
> My concern about this is that it will trivialise the frost punk creeping freezing temperatures in the upcoming Streetstation map.

nope
definitely not, unless you guys changed the same files I did, in which case it will conflict, and that's it
they're about as easy to get as atmos holofans, which is to say not very easy, and only two spawn roundstart

MrJWhit:
I'm worried about someone remotely releasing captured fusion in the hallway, while the person is in another z-level and there's nothing that can be done to stop it while the station cooks itself.

Thalpy:
> 
> 
> > My concern about this is that it will trivialise the frost punk creeping freezing temperatures in the upcoming Streetstation map.
> 
> nope
> definitely not, unless you guys changed the same files I did, in which case it will conflict, and that's it
> they're about as easy to get as atmos holofans, which is to say not very easy, and only two spawn roundstart

Ah, if they're only obtainable from roundstart that should be fine - I just won't add any to the map then.

Seris02:
They spawn in an atmos tech's locker... sorry. And are printable from engie lathe. Which is what I meant by the same availability as atmos holofans

MrJWhit:
It's still worrying that a competent atmos tech can just completely cook a station from three z-levels over from a quality fusion can.

silicons:
i'd swear there was a better way to do this, rather than adding yet another loop to the laggeist thing in the game.

Seris02:
> i'd swear there was a better way to do this, rather than adding yet another loop to the laggeist thing in the game.

I can tell you right now.
There isn't.
I spent about two hours combing through all the code I could find and the only thing that could even possibly REMOTELY stop temperature is "blocks_air" as a turf var, and I'm still not sure if that does either.
So I made something that anything could use, but added "TESTMERGE THIS FIRST"

Ghommie (approved):
Ok.

Ghommie:
If the server somehow and somewhat dunk itself thanks to this silly change, you have my blessing to revert it.

Seris02:
sure

PR #10248, opened 2019-12-20
Restores function to the hound modules' tongues and noses

CameronWoof:
## About The Pull Request

Dogborgs now get their functional noses and tongues back. Bellies are still scene-only.

## Why It's Good For The Game

The unique flavor of the hound modules is important to preserve. Destroying interesting and appropriately flavorful mechanics from hound modules in order to standardize silicons is misguided, and only creates a situation where the role loses some of its draw in order to satisfy no-one. 

These features are limited in their use, but satisfying and fun to bring into play when the opportunity arises, and removing them was a mistake. 

## Changelog
:cl:
tweak: Dogborg tongues and noses are now functional again
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -137,11 +137,11 @@
 		rebuild_modules()
 	return I
 
-//Adds flavoursome dogborg items to dogborg variants without mechanical benefits
+//Sleepers are purely flavor, the tongue and nose are functional
 /obj/item/robot_module/proc/dogborg_equip()
-	var/obj/item/I = new /obj/item/analyzer/nose/flavour(src)
+	var/obj/item/I = new /obj/item/analyzer/nose(src)
 	basic_modules += I
-	I = new /obj/item/soap/tongue/flavour(src)
+	I = new /obj/item/soap/tongue(src)
 	basic_modules += I
 	I = new /obj/item/dogborg/sleeper/K9/flavour(src)
 	if(istype(src, /obj/item/robot_module/engineering))


YakumoChen:
no

Sishen1542:
fetish content pr, downvoted

CameronWoof:
If you disagree, please explain in your comment how the addition of these features will negatively affect your game, other than the knowledge that people will derive joy from them and you won't.

chefferz:
This literally doesn't affect anyone. Calling this fetish content is a big fat ???.

Sishen1542:
:thinking:

ghost:
we're on a fetish content freeze

GrayRachnid:
Powercreep.

If you want to analyze the air then play an engi or a standard. If you want to clean then play a service. Other modules don't need these items to function.

silicons:
This is unnecessary.

CameronWoof:
It is not strictly necessary; the modules will obviously survive without. But I think it would be a mistake to deny hound modules what made them unique and interesting in order to squash everything down to perfect, bland parity. What's the mechanically ideal game design is not necessarily the most fun, and I think that we can afford small differences like this in order to preserve part of the server's culture, and the form of its unique hound borgs.

Ghommie (approved):
I can agree this goes against the current design direction of having dogborgs as little more than glorified reskins of base vanilla counterparts without much to fancy about.
On the other hand, the mostly insignificant advantage this PR provides to dogborgs is not worth the drama and I can also see how the current nose and tongue items are nothing more than slobbering roleplay garbage.

EDIT: Screw it, Gray is right, you should play engi or service instead.

Ghommie:
Ah, screw it. As Gray said,

Ghommie:
Reopening because of grace period. Remember this is not a meme PR and shouldn't be treated as such.

Soranayo:
Technically speaking both of these have gameplay affecting usage. Tongue can extinguish burning people and be used as a low efficiency stunbaton while emagged, and the nose is technically an analyzer (Meaning you can scan anomalies with it.), if you remove those functions, i don't have any issue with it, but the current way dogborgs are going currently (Aka, being just skins, which i agree on, honestly.), i can't say i like this PR, as minor as it is.

PR #10255, opened 2019-12-21
Hopefully fixes the broken Christmas clothes that no one knows how they broke?

Xantholne:
## About The Pull Request

Fixes red winter clothes, santa boot icons, etc etc.
Literally none of us currently know how they broke, but they worked this time after pretty much rewriting it and re importing the somehow missing obj icons for the hoodies in a local version

## Changelog
:cl:
fix: Christmas clothes that where missing stuff should work again
/:cl:


Ghommie:
It seems that the uniform sprites are already present in uniform.dmi. 
If they still don't display on mob somewhat, you may want to make sure they have the .dmi state is the same as the typepath's `icon_state`, that they have no item_color (variable to be deprecated) set to a text string different than their `icon_state` and that they have no `alternate_worn_icon`.

Ghommie:
Looks like digitigrade versions of these coats are missing too.

Xantholne:
As an object icon? Never made any.

Ghommie:
Yea, but since the current christmas coats settings allow digitigrades, so it should be wise to copypaste those sprites to suit_digi (plus a few extra pixels for that leg gap).
Otherwise add a `mutantrace_variation = NO_MUTANTRACE_VARIATION` line to each coat type.

silicons (approved):

PR #10256, opened 2019-12-21
Adds an alcohol intolerance trait (NOT light drinker)

Putnam3145:
## About The Pull Request

Adds an alcohol intolerance trait that makes you incapable of handling alcohol. *Completely* incapable. This isn't like light drinker, which makes you drunk faster. You are no longer able to get drunk, and you vomit the instant you ingest any alcohol whatsoever. No 

## Why It's Good For The Game

I mean, it's sort of a pref thing a la #10243, except it's all IC stuff and not really kink related so I figure it should go in its own way.

Also I wanted an IC-enforced no-drinking trait that isn't the maybe-overpowered for-some-reason-negative light drinker, so there you go.

## Changelog
:cl: Putnam
add: Alcohol intolerance trait, which forces vomiting on any alcohol ingestion
/:cl:


Thalpy:
You could consider making this trait give bonus metabolic stress if you wanted it to be worth negative points too.

Putnam3145:
prooobably won't. could make alcohol cause metabolic stress by default, really, instead of damaging the liver directly, at least at low boozpwr

Thalpy:
Alcohol does in the pr, but it accumulates it per tick, and I assume this purges it quickly.

code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
```diff
@@ -77,6 +77,14 @@ All effects don't start immediately, but rather get worse over time; the rate is
 				// +10% success propability on each step, useful while operating in less-than-perfect conditions
 	return ..()
 
+/datum/reagent/consumable/ethanol/on_mob_add(mob/living/L,amount)
+	if(HAS_TRAIT(L, TRAIT_NO_ALCOHOL) && boozepwr > 0)
+		if(iscarbon(L))
+			var/mob/living/carbon/C = L
+			C.vomit()
+		L.reagents.remove_all_type(/datum/reagent/consumable/ethanol, amount*4, 0, 1)
+	return ..()
```
Ghommie: Ever played a flyperson in the past just to annoy the heck out of everyone by vomiting all around the station? Well, this feels quite the same, except it stuns the owner.

What about dealing toxin damage on species/handle_chemicals() with an odd chance of vomiting (it clears the mob's reagents), diminished "metabolization rate" and return value TRUE?

Putnam3145: Vomiting makes you dry heave if you're low on nutrition, though? Flies can eat vomit so you can sorta just do that in a loop.

Also, vomiting didn't *seem* to remove reagents, I thought I checked pretty thoroughly.

Also, OOC main reason I want this is to give immediate, heavy feedback that my character doesn't want alcohol and not to give it, so making it more subtle isn't really the intent?

Ghommie: Umh, a little extreme but gotcha.
After all, tgstation has the vomi/tg/oose, and on goon you can make cheese with puke, milk and a fire extinguisher, why should I bother?

silicons:
this seems like a net benefit, why do we need this?

Putnam3145:
i mean, if it's a benefit it can be a positive trait, and this is me counting alcohol in my "chemistry prefbreak salt PR" list.

code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
```diff
@@ -77,6 +77,14 @@ All effects don't start immediately, but rather get worse over time; the rate is
 				// +10% success propability on each step, useful while operating in less-than-perfect conditions
 	return ..()
 
+/datum/reagent/consumable/ethanol/on_mob_add(mob/living/L,amount)
```
Ghommie: Do you think this could be `on_mob_metabolize()` instead?

Krysonism:
> this seems like a net benefit

Seems like it would purge weapons grade alcohols like Bacchus blessing, neurotoxin and hippie's delight.
 
Somewhat situational sure but not a net negative I think.

code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
Putnam3145: went with on_mob_add so i could use amount there, though it is fully possible it's unnecessary for some reason or another

Ghommie: You could use the `volume` var instead, but whatever.

Putnam3145: yeah i rethought it and made it less of a "meme effect", maybe recheck it

Putnam3145:
I updated the effects, for the record, so this should probably get another look at

Ghommie (approved):
All good now.

PR #10274, opened 2019-12-22
no more 40% time dilation from broodmother

ghost:
## About The Pull Request
poojblood means blood and gibs stack on the same tile and each one of these gibs makes like 10 decals or something along those lines, and each use of the ability summons anywhere between 3 to 8 of these critters, not a good combination

## Why It's Good For The Game
lag

## Changelog
:cl:
fix: broodmother baby lag
/:cl:


ghost:
@ me if this is broken somehow but i didn't test it because one line change 😂

Ghommie (approved):
Gory is good but considering there are going to be a lot of them, better this way.

P.S. gibspawners could use some improvements instead of initializing dummy humans and species for dna data at some point later, damnit. :rage1:

...since when Github had non-unicode emojis?

ghost:
yeah unfortunately the gib is too powerful, i've seen one broodmother fight on box 20 minutes in send tidi from 10% to 38%

PR #10295, opened 2019-12-26
Brain surgery can now fix an abductee's mindsnap objectives

silicons:
## About The Pull Request

title

## Why It's Good For The Game

"Kev, why are abductors hyperlethal"
Because they permanently fuck up crewmembers in ways that can't be fixed.
This fixes that and makes it so you need to put some effort in to fix it, rather than just mindshielding. 
## Changelog
:cl:
balance: Abductor mindsnapping (aka abductee objectives) can now be "cured" with brain surgery.
/:cl:


silicons:
also before anyone says "brain trauma", those are cured by cloning :)

YakumoChen:
I don't agree with this, abductees will just seek the easiest way to have their objective removed instead of actually doing it.

That or it will lead to a meta of "get your organ removed ASAP" which also completely goes against the spirit of the antag.

Ghommie (changes requested):
Definitely what Yakumo said.
Not that I dislike the idea of abductee "de-antagging", but I do not want to encourage people to have their glands removed furthermore.
But if it has to come down to surgery, then I could say **brain surgery should fix it instead**.

Also cloning is lazy and too much of a more efficient alternative to lobotomy, mind restoration virus etc., If anything the brain traumas being transferred on cloning should also be ported, because that's pretty cheesy.

necromanceranne:
Traumas carried by cloning is one fix and it is a good one too. Gutting the power of the cloner is another potential fix. Cloning upgrades makes cloners far too effective at fixing crew when it should be a last resort. Either there needs to be a reasonable resource cost for cloning (biomaterial of some kind, something for genetics to actually do besides find mutations maybe), or cloning should never be perfect and always effective.

ghost:
@kevinz000 make a persistent-through-clones phobia of conspiracies, like permanent tinfoil

Ghommie (approved):

PR #10296, opened 2019-12-26
Added a ghost role with its own spawner for the purposes of chillling in the ninja holding facility.

Useroth:
## Why It's Good For The Game

People who have been taken out of the round can now choose to continue their RP or whatever.

## Changelog
:cl:
add: Ghost cafe spawner. For letting people spawn as their own character in the ninja holding facility. It bypasses the usual check, so people who have suicided/ghosted/cryod may use it.
add: Dorms in the ninja holding facility.
/:cl:

### DIFF
--- a/code/game/objects/structures/ghost_role_spawners.dm
+++ b/code/game/objects/structures/ghost_role_spawners.dm
@@ -624,4 +624,19 @@
 	uniform = /obj/item/clothing/under/color/random
 	shoes = /obj/item/clothing/shoes/sneakers/black
 	id = /obj/item/card/id
+
+
+/datum/outfit/ghostcafe/pre_equip(mob/living/carbon/human/H, visualsOnly = FALSE, client/preference_source)
+	..()
+	var/suited = !preference_source || preference_source.prefs.jumpsuit_style == PREF_SUIT
+	if (CONFIG_GET(flag/grey_assistants))
+		if(suited)
+			uniform = /obj/item/clothing/under/color/grey
+		else
+			uniform = /obj/item/clothing/under/skirt/color/grey
+	else
+		if(suited)
+			uniform = /obj/item/clothing/under/color/random
+		else
+			uniform = /obj/item/clothing/under/skirt/color/random
 		
--- a/code/game/objects/structures/ghost_role_spawners.dm
+++ b/code/game/objects/structures/ghost_role_spawners.dm
@@ -607,33 +607,21 @@
 	density = FALSE
 	death = FALSE
 	assignedrole = "Ghost Cafe Visitor"
+	flavour_text = "Is this what life after death is like?"
+	skip_reentry_check = TRUE
+	banType = "ghostcafe"
 
 /obj/effect/mob_spawn/human/ghostcafe/special(mob/living/carbon/human/new_spawn)
 	if(new_spawn.client)
-		new_spawn.fully_replace_character_name(newname = new_spawn.client.prefs.real_name)
-		new_spawn.nameless = new_spawn.client.prefs.nameless
-		new_spawn.gender = new_spawn.client.prefs.gender
-		new_spawn.dna.features = new_spawn.client.prefs.features.Copy()
-		new_spawn.flavor_text = new_spawn.dna.features["flavor_text"]
-		new_spawn.dna.custom_species = new_spawn.client.prefs.custom_species
-		new_spawn.eye_color = new_spawn.client.prefs.eye_color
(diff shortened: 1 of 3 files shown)


code/game/objects/structures/ghost_role_spawners.dm
```diff
@@ -595,3 +595,45 @@
...
+		new_spawn.socks = new_spawn.client.prefs.socks
+		new_spawn.socks_color = new_spawn.client.prefs.socks_color
+		new_spawn.saved_undershirt = new_spawn.client.prefs.undershirt
+		new_spawn.undershirt = new_spawn.client.prefs.undershirt
+		new_spawn.shirt_color = new_spawn.client.prefs.shirt_color
+		new_spawn.saved_underwear = new_spawn.client.prefs.underwear
+		new_spawn.underwear = new_spawn.client.prefs.underwear
+		new_spawn.undie_color = new_spawn.client.prefs.undie_color
+		new_spawn.set_species(new_spawn.client.prefs.pref_species, icon_update=1)
+		new_spawn.give_genitals(TRUE)
```
Ghommie: Any reason for doing all of this when
https://github.com/Citadel-Station-13/Citadel-Station-13/blob/765832373d490468e349d96ec5e2dbdfbc593e9d/code/modules/client/preferences.dm#L2252
could be used instead? (there is also a separate modular_citadel bit for the extra bits)

code/game/objects/structures/ghost_role_spawners.dm
```diff
@@ -595,3 +595,45 @@
 
 /obj/effect/mob_spawn/human/pirate/gunner
 	rank = "Gunner"
+
+/obj/effect/mob_spawn/human/ghostcafe
+	name = "ghost cafe sleeper"
+	uses = -1
```
Ghommie, line 601:
```suggestion
	permanent = TRUE
```

Ghommie (approved):
You could make the ninja holding facility's cryopods not "punishable" like how #9862 tried to do with perma's while you are at it. (Just make sure the ghost verb prompt message is not mistleading here this time)

code/game/objects/structures/ghost_role_spawners.dm
Useroth: That's a proc I wasn't even aware of.
I've looked at the virtual reality code (which copies shit from pre-existing mob) and how turning a ghost into human worked in the admin panel, and neither of those fit my purposes.
I'll test it now.

code/game/objects/structures/ghost_role_spawners.dm
Useroth: the "permanent" var only affects whether the spawner will be qdel'd after running out of uses
here it has no effect, because the spawner will never run out of uses
```
	if(!permanent && !uses)
		qdel(src)
```

Ghommie: ok.

Useroth:
Also, to be honest, it might be tricky to add non-punishable cryopods that can't be abused after having a spawner that bypasses the "punishment".

code/game/objects/structures/ghost_role_spawners.dm
```diff
@@ -595,3 +595,33 @@
...
+		var/datum/outfit/O = new /datum/outfit/ghostcafe()
+		O.equip(new_spawn)
+		SSjob.equip_loadout(null, new_spawn, FALSE)
+		SSquirks.AssignQuirks(new_spawn, new_spawn.client, TRUE, TRUE, null, FALSE, new_spawn)
+
+/datum/outfit/ghostcafe
+	name = "ID, jumpsuit and shoes"
+	uniform = /obj/item/clothing/under/color/random
+	shoes = /obj/item/clothing/shoes/sneakers/black
+	id = /obj/item/card/id
```
Ghommie, line 626: Do you think you could add some special equip() (probably with parent call at the end) for jumpsuit/jumpskirt preferences in a similar manner to: https://github.com/Citadel-Station-13/Citadel-Station-13/blob/05e3fc1a016f9121c19d6fcb888475881124525e/code/modules/jobs/job_types/assistant.dm#L31
?

Ghommie (approved):
I'm not going to take this PR hostage because of some oversight with the ghost punishment thingy that'll be fixed in a near future.

By the by I don't like Citadel's tendency to slap in dorm rooms everywhere.

code/game/objects/structures/ghost_role_spawners.dm
Useroth: That's a pretty good idea tbh. Done.

PR #10304, opened 2019-12-27
Overhauling antags to be more freeform/not-require-greentext

Putnam3145:
## About The Pull Request

Just removing greentext is kinda low effort and leads to weird shittery behavior, so here's something easier to keep track of for everyone involved, probs.

## Why It's Good For The Game

All the stuff people say about greentext applies, plus I think giving traitors more freedom to do antaggy things is good.

## Changelog
:cl:
add: Traitors now have more freeform goals
del: Removed greentext
del: Removed most traitor goals
/:cl:


YakumoChen:
It's nice but can we have the "ruin everything" objective not contain the latest memes?

MrJWhit:
The only way to actually kill people is with the highest possible chaos level, which doesn't seem in keeping with the even the medium chaos elements at _all_.

EDIT: Especially because you go from "almost remove them" to "hijack the fucking shuttle".

Putnam3145:
Second to last is totally allowed (even encouraged) to kill, just encouraged not to cremate/nutcrack and so on. Obviously the wordings of all these I'm not attached to, though.

MrJWhit:
I think there could be another option that is "remove a few people from the round". It's just that without a source of body disposal, it's more or less a joke to survive in the round for more 20 minutes. People act violently and heard against the idea of a crew member dying and being on suit sensors, and body disposal is easily half the battle for antagging.

Ghommie:
I have seen similar formulas on abductee/sinned antags before and it just doesn't work.
Leaving objectives up to interpretation with no incentive nor reward may leave players with a higher tendency about not following them.
This is all a caprice about the competitive nature of most antagonists being unwanted and it doesn't consider greentext as a mean to encourage antags to play their role.

You (the staff) hop on the phrase "antagonists are supposed to make rounds interesting" without admitting they already do so by stopping the game from being 24/7 extended.

GrayRachnid:
'Teach the people from the engineering department a lesson by killing them and leaving their bodies in the open.'

YakumoChen:
More specific goals could be added but I think the current flavor ones I wrote work pretty well. Something like "kill engineers" would have to be paired with a relevant objective (Donk Corp in this example), would need checks to see if there are actual engineers to murder, and besides, "leave them in the open" is too close-ended and wouldn't be rules-enforced, all of this would add a lot of complexity to what the point is to make objectives more open-ended.

I'd personally like to change AI objectives as well but looking at objective code gives me a headache.

EmeraldSundisk:
Nitpicky, but I thought it was Donk Co. and WaffleCo. The idea of different factions of the Syndicate is interesting, but I dunno if it should simply be based on threat level alone. You could certainly make some interesting stories out of it though, especially considering different groups within the Syndicate don't exactly see eye-to-eye at all times. In fact, some fucking hate each other.

Some ideas off the top of my head:
- Get the shuttle called and end the shift as quickly as possible.
- (Probably an Animal Rights Consortium objective if we're using them) Kill the Cook and serve their meat to the public to see how they like it. (They could be resurrected, they'd just need to be gibbed for the meat.)
- Kill a given target and resurrect them as a podperson.
- Ensure the AI/a cyborg is delawed and remains active until the end of a round (they would NOT be inherently loyal to the person that freed them).

I could probably come up with more.

YakumoChen:
Co. is literally short for Corporation.

The point of the objectives is to give you a *lot* of freedom in objectives without removing them entirely. A vague goal to aim your gimmick around. Retooling them back into very specific objectives would ruin the point of the PR again.

silicons:
assuming this one is dead reopen if im wrong

PR #10306, opened 2019-12-27
Removes most of vampire armor and removes stun scaling, and associates the buffs with physiology

Arturlang:
Finally, after people yelling at me that vampires were extremely weak, i was told vampires were too powerfull. This hopefully fixes excessive armor, somewhat, though im unsure how to deal with vampires who wear armor, as it might possibly cause an 100 percent armor protection which might not be something we want.
## Changelog
:cl:
tweak: Vampires are no longer as tanky 
/:cl:

### DIFF
--- a/code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
+++ b/code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
@@ -183,7 +183,7 @@
 	BuyPower(new /datum/action/bloodsucker/masquerade)
 	BuyPower(new /datum/action/bloodsucker/veil)
 	// Traits
-	for (var/T in defaultTraits)
+	for(var/T in defaultTraits)
 		ADD_TRAIT(owner.current, T, "bloodsucker")
 	if(HAS_TRAIT(owner.current, TRAIT_TOXINLOVER)) //No slime bonuses here, no thank you
 		had_toxlover = TRUE
@@ -200,10 +200,10 @@
 		var/mob/living/carbon/human/H = owner.current
 		var/datum/species/S = H.dna.species
 		// Make Changes
-		S.brutemod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
-		S.coldmod = 0
-		S.stunmod *= 0.35
-		S.siemens_coeff *= 0.75 	//base electrocution coefficient  1
+		physiology.brutemod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
+		physiology.coldmod = 0
+		physiology.stunmod *= 0.35
+		physiology.siemens_coeff *= 0.75 	//base electrocution coefficient  1
 		//S.heatmod += 0.5 			// Heat shouldn't affect. Only Fire.
 		//S.punchstunthreshold = 8	//damage at which punches from this race will stun  9
 		S.punchdamagelow += 1       //lowest possible punch damage   0
--- a/code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
+++ b/code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
@@ -200,7 +200,7 @@
 		var/mob/living/carbon/human/H = owner.current
 		var/datum/species/S = H.dna.species
 		// Make Changes
-		S.brutemod *= 0.2											//  <--------------------  Start small, but burn mod increases based on rank!
+		S.brutemod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
 		S.coldmod = 0
 		S.stunmod *= 0.35
 		S.siemens_coeff *= 0.75 	//base electrocution coefficient  1
--- a/code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
+++ b/code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
@@ -201,9 +201,9 @@
 		var/datum/species/S = H.dna.species
 		var/datum/physiology/physiology
 		// Make Changes
-		physiology.brutemod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
-		physiology.coldmod = 0
-		physiology.stunmod *= 0.35
+		physiology.brute_mod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
+		physiology.cold_mod = 0
+		physiology.stun_mod *= 0.35
 		physiology.siemens_coeff *= 0.75 	//base electrocution coefficient  1
 		//S.heatmod += 0.5 			// Heat shouldn't affect. Only Fire.
 		//S.punchstunthreshold = 8	//damage at which punches from this race will stun  9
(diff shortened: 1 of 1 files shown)


necromanceranne:
The core of the issue here is that it's a race damage modifier, which I think because of how the damage reduction calculations work, will always allow you to bring damage to 0 if your armour would reduce the damage you take to below what your damage mod would reduce it too.

Basically, the core issue is more with armor and damage reduction calculations, and not actually with the ability. Typically, you should only ever have a hardcap of 90% damage reduction from armor even if your armor would bring you to 100%. This gets really fuck with damage modifications however.

What happened in a recent round that I saw was that because those attacking the vampire didn't have any armor penetration weapons whatsoever (which makes sense given it was all baton smashes and magrifle shots), the armor provided by their wizard hardsuit (40% against everything) was enough to protect them from the damage to a degree that they could just simply outheal everything.

Armor penetration also does nothing for damage modifiers, which means even IF they had a high armor pen weapon (there was actually a syndicate sniper rifle floating around, which I got out of the christmas tree and gave to security), then it would STILL not be enough as the damage reduction from damage modifiers ignores armor penetration.

This needs a much deeper rework than simple stopgap fixes, and is one of the biggest and most glaring examples of how fucked up armor really is. I've talked a lot about this, and if stun combat wasn't completely bypassing armor entirely, armor would be busted overpowered as would armor stacking. Vampires, because they nearly disregard stamina combat for the most part, demonstrate this most readily.

Arturlang:
This should help somewhat and make it harder to do so, god knows how long itll take for a rework

Arturlang:
Ive ditched their burn and most of their burn armor in return giving then 5 more health per level, allowing them to get 10 extra maxhealth per level, i plan on also making leveling cost blood later, and fortitude should still be usefull as it blocks stakes, and knockdowns, while also giving decent amount of armor.

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
```diff
@@ -319,12 +319,10 @@ datum/antagonist/bloodsucker/proc/SpendRank()
 	if(ishuman(owner.current))
 		var/mob/living/carbon/human/H = owner.current
 		var/datum/species/S = H.dna.species
-		S.burnmod *= 0.025 			// Slightly more burn damage
```
Ghommie, line 322: >burn modifier dunked down to 1/40th

>slightly more

:joy:

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
```diff
@@ -200,9 +200,9 @@
...
 		var/datum/species/S = H.dna.species
 		// Make Changes
-		S.brutemod *= 0.5											//  <--------------------  Start small, but burn mod increases based on rank!
+		S.brutemod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
 		S.coldmod = 0
-		S.stunmod *= 0.25
+		S.stunmod *= 0.35
 		S.siemens_coeff *= 0.75 	//base electrocution coefficient  1
 		//S.heatmod += 0.5 			// Heat shouldn't affect. Only Fire.
 		//S.punchstunthreshold = 8	//damage at which punches from this race will stun  9
```
Ghommie: Just noticed how the buffs affect the current species' coefficients instead of the physiology datum's and how, with upmost certainty, it can and will be fucked up by species change.
Please use the physiology equivalents where possible (cause there is no variable for punches min / max threshold yet) instead.
Here is the [physiology datum file](https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/code/modules/mob/living/carbon/human/physiology.dm)
Also the var name:
https://github.com/Citadel-Station-13/Citadel-Station-13/blob/7e5e241bd2c72dd3e451344ee5e6aa9dca9bed41/code/modules/mob/living/carbon/human/human_defines.dm#L63

Also the fact these buffs are not undone on antag removal worries me a little but that's outside the scope of this PR.

Ghommie (changes requested):

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
Arturlang: uhhh it should be undone on antag removal

Arturlang: `	if(ishuman(owner.current))
		var/mob/living/carbon/human/H = owner.current
		H.set_species(H.dna.species.type)`

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
Arturlang: I forgot to update that comment...
Not that it matters anymore

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
```diff
@@ -199,11 +199,12 @@
 	if(ishuman(owner.current))
 		var/mob/living/carbon/human/H = owner.current
 		var/datum/species/S = H.dna.species
+		var/datum/physiology/physiology
```
Ghommie: This is the proof I never were or will be good enough at explaining how things should be done. 
Remove this line, it will cause null runtimes. :pensive:

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
```diff
@@ -199,11 +199,12 @@
...
 		// Make Changes
-		S.brutemod *= 0.5											//  <--------------------  Start small, but burn mod increases based on rank!
-		S.coldmod = 0
-		S.stunmod *= 0.25
-		S.siemens_coeff *= 0.75 	//base electrocution coefficient  1
+		physiology.brute_mod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
```
Ghommie:
```suggestion
		H.physiology.brute_mod *= 0.8										//  <--------------------  Start small, but burn mod increases based on rank!
```

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
```diff
@@ -199,11 +199,12 @@
...
+		physiology.cold_mod = 0
```
Ghommie:
```suggestion
		H.physiology.cold_mod = 0
```

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
```diff
@@ -199,11 +199,12 @@
...
+		physiology.stun_mod *= 0.35
```
Ghommie:
```suggestion
		H.physiology.stun_mod *= 0.35
```

code/modules/antagonists/bloodsucker/datum_bloodsucker.dm
```diff
@@ -199,11 +199,12 @@
...
+		physiology.siemens_coeff *= 0.75 	//base electrocution coefficient  1
```
Ghommie:
```suggestion
		H.physiology.siemens_coeff *= 0.75 	//base electrocution coefficient  1
```

Ghommie (changes requested):

Ghommie (approved):

PR #10310, opened 2019-12-27
[Needs discussion]Dead xenomorph hosts no longer burst.

Arturlang:
This gives xenomorphs a reason to keep alive hosts, until they burst, so aliens wont be no longer encouraged to kill their host once infected, which gives the host no chance to escape, why not give them a chance, maybe? Xenomorphs already are horribly strong.
🆑
balance: Xenomorph embryos no longer grow in dead hosts.
/🆑

### DIFF
--- a/code/game/objects/items/body_egg.dm
+++ b/code/game/objects/items/body_egg.dm
@@ -27,12 +27,6 @@
 		INVOKE_ASYNC(src, .proc/RemoveInfectionImages, owner)
 	..()
 
-/obj/item/organ/body_egg/on_death()
-	. = ..()
-	if(!owner)
-		return
-	egg_process()
-
 /obj/item/organ/body_egg/on_life()
 	. = ..()
 	egg_process()
--- a/code/modules/mob/living/simple_animal/hostile/headcrab.dm
+++ b/code/modules/mob/living/simple_animal/hostile/headcrab.dm
@@ -56,6 +56,12 @@
 	var/datum/mind/origin
 	var/time
 
+/obj/item/organ/body_egg/changeling_egg/on_death()
+	. = ..()
+	if(!owner)
+		return
+	egg_process()
+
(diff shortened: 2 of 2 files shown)


ghost:
people killing hosts to stop xenos
xenos being unable to reproduce in dangerous environments (lavaland/spaced areas, people unsuited will slowly die)

Arturlang:
Hmm. What do you suggest to make this better, @Commandersand ?

Owai-Seek:
As far as I know, incubation is already heavily nerfed when someone dies. Even if someone were kept alive for longer, the facehugger keeps them knocked out for at least like, half, of the incubation cycle, unless someone takes the time to take it off their face. If someone takes that time, they're probably making themselves a target for a xeno attack or another facehugger toss.

Maybe make it so if a xeno starts slashing the host, it also damages the egg "organ", making incubation take longer. You could also make it so disembowelment makes the body completely unable to hold an embryo. (I'm not sure if this is the case or not, or if Xeno's can disembowel, I don't actually see them that often?) Just pingponging ideas honestly.

Arturlang:
I dont think xenos can disembowel, not 100 percent sure @Owai-Seek

ghost:
the whole point was to give people a chance to escape right?
maybe have being nested increase stamina gain or a smaller wait time to resist out so xenos have to check up on you more in the early stage

Ghommie (changes requested):
Incubation time is not actually nerfed on dead people as of now.

But, considering most benos make nests near the vacuum of space for easier escape, salty players succumbing as a "fuck you xenos", and similar hardships with keeping hosts alive, I think this change is a bit too drastic.

If anything, keeping hosts alive should be awarded before the opposite is penalized.

silicons:
what ghom said, it kinda fucks xenos over a bit hard.

Arturlang:
Alright, ill close the PR since its a no, and will need a bit more work.

PR #10326, opened 2019-12-28
Significantly lowers the forcefield projector health

chefferz:
Does what it says on the tin. 

Having a 250 health portable shield that you can deploy on up to three tiles, has shielding resistant to everything but melee, and can be reset and redeployed quickly has been nothing but abused lately.

This drops the health down to 100 from 250, keeping shield integrity. This also shortens the distance you can place them to four tiles, so no more reaching across the map with this.

### DIFF
--- a/code/game/objects/items/devices/forcefieldprojector.dm
+++ b/code/game/objects/items/devices/forcefieldprojector.dm
@@ -10,11 +10,11 @@
 	lefthand_file = 'icons/mob/inhands/misc/devices_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/misc/devices_righthand.dmi'
 	materials = list(MAT_METAL=250, MAT_GLASS=500)
-	var/max_shield_integrity = 250
-	var/shield_integrity = 250
+	var/max_shield_integrity = 100
+	var/shield_integrity = 100
 	var/max_fields = 3
 	var/list/current_fields
-	var/field_distance_limit = 7
+	var/field_distance_limit = 4
 
 /obj/item/forcefield_projector/afterattack(atom/target, mob/user, proximity_flag)
 	. = ..()


silicons:
(doesn't mean it's a bad pr tho)

Hatterhat:
at least have the courtesy to branch properly 4head

Ghommie (approved):
The field distance limit nerfs seems a bit unrelated to the issue. 5 or 6 would be also a better value if you ask me. All good otherwise.

silicons (changes requested):
make it sustain a Beam() to any shields projected, and not work from inside backpack (or be severely weakened by being inside a container rather than being on the ground). field distance being 7 is fine.
don't make this item completely useless because someone used it effectively a few rounds.

Ghommie:
> not work from inside backpack (or be severely weakened by being inside a container rather than being on the ground)

What do you exactly mean? Aren't they supposed to be held in some way to do their job like most tools?

silicons:
@Ghommie 
haahhahaha god I wish
Right now you basically do the following:
backpack
14 forcefield projectors in 2 boxes
take out, project forcefields, put back in box
repeat ad nauseum until you have an obscene amount of forcefields up to lock down the area

taking out that alone would make these far more niche in its usage instead of a free area lockdown.

Detective-Google:
> @Ghommie
> haahhahaha god I wish
> Right now you basically do the following:
> backpack
> 14 forcefield projectors in 2 boxes
> take out, project forcefields, put back in box
> repeat ad nauseum until you have an obscene amount of forcefields up to lock down the area
> 
> taking out that alone would make these far more niche in its usage instead of a free area lockdown.

I have used these to great memey effect due to the fact that they block tasers (and projectiles)
there's a reason these only allow adjacent placement on tg iirc.

it's to the point where it takes more time to remove one of these fields than it would to remove an actual wall and they can be placed instantly at range.

GrayRachnid:
they're funny and deserve a nerf.

PR #10327, opened 2019-12-28
ports tg genetics

Seris02:
## About The Pull Request

tg genetics duh

Thank Azarak for porting TG genetics in the first place.

## Why It's Good For The Game

tg genetics good

## Changelog
:cl:
add: tg genetics
/:cl:


Arturlang:
One question, does it bring back hulk?

Seris02:
I don't think so

silicons (dismissed):
needs a complete list of changes from base of new tg genetics.
i'm not a fan of porting over hyper's rebalances to it blindly, as someone from /tg/ who feels strongly about balance done by servers that aren't /tg/ or citadel (because it rarely works right)

needs a more gradual death if you're too unstable rather than the one-minute-and-then-you-dust thing if that isn't changed already.

ghost:
gonna dump a bunch of balance suggestions here since they'll get drowned out in something like discord:

separate powers into categories:
biotic for things like invisibility or hulk
psionic for things like mindread or tk
supernatural for anything that doesn't fit in?

make each category of power have a "counter", which is just stacking debuffs the more powers you have inside a category.
the category counters would just be replacement for current instadeath but without the ability to ignore it completely.

for example:

too many biotic powers randomly changes your species/appearance constantly and causes your limbs to randomly lose control, paralysis/hitting yourself, etc

too many psionic powers gives you hallucinations and other vivid effects like the drug effect, spinny screen, and also slowly regenerating brain damage

too many supernatural powers... i dunno supernatural in my mindscape right now is just a catchall for anything that doesn't fit so you could just give them more impact individual effects


each power should have its own level of negative effects based on how powerful it is; a power like telepathy not really having an impactful downside (although it would still apply to the psionic category downside counter) besides maybe auditory hallucinations or something like self-regenerating minor brain damage, while telekinesis might have non-regenerating brain damage on use or make it harder for you to use your hands. Something like spacewalking might make you weaker to heat and crushing force (burn/melee) because of changes in biology.

this method of downsides should of course account for if an ability is meant for fun/cosmetic purposes or beating the shit out of people or emergent gameplay; a power like being able to make snow should just make you occasionally chilly where a power like goon's cryokinesis would go far enough to damage and actually slow.

some powers should be introduced/modified to COMPLETELY change your gameplay; and idea i've seen so far is mind read passively saying things to you but you're mechanically deaf, and actively using it just going back a few lines. I don't really have many ideas or thoughts on this besides the fact that gameplay changers add more replayability and therefore some level of fun.


similar powers within the same power category should have "synergies", lowering the cost of having them both by a small amount, combining them into a single(?) power, and modifying some of their aspects - either completely for something like goon's gamma ray exposure+strength for hulk or slightly for something like mixing the ability to make snow and the ability to make ice, and now it just lets you do either and makes your ice traps last slightly longer/snow sheets produced more at a time.

Ghommie:
Merge conflicts.

code/game/machinery/computer/dna_console.dm
```diff
@@ -147,28 +205,56 @@
...
+			var/max_line_len = 7*DNA_BLOCK_SIZE
+			if(viable_occupant)
+				temp_html += "<div class='dnaBlockNumber'>1</div>"
+				var/len = length(viable_occupant.dna.uni_identity)
+				for(var/i=1, i<=len, i++)
+					temp_html += "<a class='dnaBlock' href='?src=[REF(src)];task=pulseui;num=[i];'>[copytext(viable_occupant.dna.uni_identity,i,i+1)]</a>"
+					if ((i % max_line_len) == 0)
+						temp_html += "</div><div class='clearBoth'>"
+					if((i % DNA_BLOCK_SIZE) == 0 && i < len)
+						temp_html += "<div class='dnaBlockNumber'>[(i / DNA_BLOCK_SIZE) + 1]</div>"
```
Ghommie: Remember to readd unicode support changes here.

Seris02: again
wut

code/game/machinery/computer/dna_console.dm
```diff
@@ -58,12 +104,26 @@
...
+	jokerready = world.time + JOKER_TIMEOUT
+
+	stored_research = SSresearch.science_tech
+
+/obj/machinery/computer/scan_consolenew/examine(mob/user)
+	..()
+	if(jokerready < world.time)
+		to_chat(user, "<span class='notice'>JOKER algorithm available.</span>")
+	else
+		to_chat(user, "<span class='notice'>JOKER algorithm available in about [round(0.00166666667 * (jokerready - world.time))] minutes.")
```
Ghommie: Remember examine was refactored a while ago to use a list instead of multiple chat messages.

code/game/machinery/computer/cloning.dm
```diff
@@ -485,7 +485,7 @@
 	R.fields["id"] = copytext_char(md5(mob_occupant.real_name), 2, 6)
 	R.fields["UE"] = dna.unique_enzymes
 	R.fields["UI"] = dna.uni_identity
-	R.fields["SE"] = dna.struc_enzymes
+	R.fields["SE"] = dna.mutation_index
```
Ghommie, line 488: By the by, this will be displayed as "\list" on the UI since this is a list, please fix that one way or another.

Ghommie (changes requested):
I also heard auto-cloning is busted but I don't know if and how this PR is involved with that issue.

Ghommie (approved):
lgfm. A little more testmerging just to be sure and it's in.

PR #10343, opened 2019-12-29
[s] Fixes anti-ghost-role vars a little.

Ghommie:
## About The Pull Request
Stops ghost roles time locks (or whatever we can call them) from resetting due respawns and roles that don't check it (such ghost cafe, round-ending things such as nar-sie / ratvar / true devils, CTF) by moving the variable to the client, leaving some remnants for clientless ghosts (tied to the player on login)

## Why It's Good For The Game
Stops some inconsistencies and exploits with this bad feature.

## Changelog
None.


Putnam3145 (approved):

Putnam3145:
this is *exactly* what i needed for some stuff I need to do actually, nice

PR #10346, opened 2019-12-30
Nerfs limb stamcrit by doubling the base limb stam regen rate and making disabled limbs double the regen rate further

deathride58:
## About The Pull Request

This PR more or less does exactly as it says on the tin. Increases the limb stamina regen rate to 4 per process cycle, and makes it so that limbs regenerate at double that rate when disabled. This means that instead of it taking 25 seconds (!!!!) for a limb to regenerate itself out of staminacrit, it now only takes 6.25 (Roughly.) seconds for a limb to regen out of stamcrit. This PR also further reduces the cheese involved with limb stamcrit by giving limbs a very similar mechanic to what's featured for the normal spaceman stamcrit, where you take less staminaloss the more staminaloss you take while in stamcrit. This multiplier resets instantly once the limb is no longer disabled.

## Why It's Good For The Game

While I do have a complete overhaul of stamina mechanics as a whole in the works, the current limb stamina mechanics are in a state where it's simply unreasonable to leave it as-is in favor of making everyone wait for me to actually have the time to finish that overhaul. I also have a fetish for shotguns, and I want energy shotguns to stay. Don't judge.

## Changelog
:cl: Bhijn
add: Limbs now regenerate their stamina faster while disabled
add: Limbs now have the same incoming stamina damage multiplier mechanics as spacemen, where the more staminaloss they take while disabled, the less staminaloss they'll take.
balance: Limbs have had their base stamina regen rate doubled to match the doubled stamina regen rate of standard spacemen.
/:cl:

### DIFF
--- a/code/modules/surgery/bodyparts/bodyparts.dm
+++ b/code/modules/surgery/bodyparts/bodyparts.dm
@@ -29,6 +29,7 @@
 	var/burn_dam = 0
 	var/stamina_dam = 0
 	var/max_stamina_damage = 0
+	var/incoming_stam_mult = 1 //Multiplier for incoming staminaloss, decreases when taking staminaloss when the limb is disabled, resets back to 1 when limb is no longer disabled.
 	var/max_damage = 0
 	var/stam_heal_tick = 0		//per Life(). Defaults to 0 due to citadel changes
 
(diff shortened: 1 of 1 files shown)


silicons:
yeah ok

silicons (approved):

PR #10352, opened 2019-12-30
Changes to the way SM handles power related to integrity and max integrity damage

Seris02:
## About The Pull Request

SM power is now multiplied by the formula of:

Screen Shot 2019-12-30 at 6 56 17 pm

x in the second formula is also integrity, woops
first formula is applied from 100% integrity to 25% integrity, and second one is applied from 25% to 0%.

SM integrity's max is now damaged by the SM delamming, and needs a spray that can be ordered from cargo to fix it.

## Why It's Good For The Game

I dunno. I just wanted to code it and do maths, and it was fun. There'll probably need to be some improvements to this.

## Changelog
:cl:
tweak: tweaked the way the SM works
/:cl:


Ghommie:
You are the expert here, Putnam.

deathride58:
What if power output fluctuated depending on its integrity, so that max integrity has a mostly stable power output, a slightly lower integrity has a power output that slowly wobbles a little, while an integrity of 1 makes the power output fluctuate rapidly between extreme amounts.

Strazyplus:
> What if power output fluctuated depending on its integrity, so that max integrity has a mostly stable power output, a slightly lower integrity has a power output that slowly wobbles a little, while an integrity of 1 makes the power output fluctuate rapidly between extreme amounts.

That's an interesting idea, the orginal idea was with the crystal damaged its output would be lower depending on its Integrity. 

Fluctuation with the risk of severe high spikes that can cause a random APC to turn off like a circuit breaker and require a manual restart. If the integrity gets low enough why not have a random chance for the APC to explode. Since the MALF AI can cause APCS to explode I would use that code snippet.

[snipped my station-wide apc shut off ideas in this post. let's avoid mass light updates. haha]

Tupinambis:
I like this change, SM should be more dangerous than it is.

silicons:
I don't like permanent damage.

Putnam3145:
permanent damage causes engineering radio spam, so either remove it or fix that, or possibly make it have a "stability" thing that changes permanently as above

Ghommie:
Putnam's issue resolved?

Seris02:
yep

Ghommie (approved):

PR #10356, opened 2019-12-30
Adds space to sprint and toggle sprint as preferences

deathride58:
## About The Pull Request

Does exactly as it says on the tin. Adds two new preferences to the game, a toggle between having sprint on shift or spacebar, and toggle sprint.

## Why It's Good For The Game

More preferences for people who don't like the current setup is always good.

## Changelog
:cl: Bhijn
add: Added a preference to make the sprint hotkey be a toggle instead of a hold bind
add: Added a preference to bind the sprint hotkey to space instead of shift.
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -195,6 +195,9 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	var/auto_fit_viewport = TRUE
 
 	var/uplink_spawn_loc = UPLINK_PDA
+	
+	var/sprint_spacebar = FALSE
+	var/sprint_toggle = FALSE
 
 	var/list/exp = list()
 	var/list/menuoptions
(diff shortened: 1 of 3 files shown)


Ghommie:
I don't like this change, it will hinder a possible custom keybinding port (the TGMC one that tg ported a month or so ago, not Yog's).
Just move the sprint binding to the space key, it's better for everyone (shift button is kinda fucky and conflicts with examine).

Ghommie (changes requested):
See above.

deathride58:
This won't actually conflict at all with a port of keybindings (unless you're just blindly copying and pasting everything). It'll be very easy to bump the save version up and add a few lines to the version conversion to automatically set the keybinding to what the pref is. If adding conversion code is too difficult, asking people to reset their bindings isn't asking much at all. Making space be the default sprint key with no way to change it goes against what most people expect the sprint key to be, the pref is so that people have a choice of tradition vs convenience in edge cases (specifically, attempting combat outside of combat mode).

Ghommie:
Of course it won't be hard to "fix" in that case (simply a moderate waste of time). It just bitters me having merging changes that'll be likely reverted or scratched away at some foreseeable point.

Yet I still don't see any reason with shift key sprint should linger around other than muscle memory and people unable to use the space bar for whichever reason.

Ghommie (approved):

PR #10361, opened 2019-12-30
Ghost Cafe patrons now spawn with chameleon kits

CameronWoof:
## About The Pull Request

Adds the "/ghostcafe" variant of the chameleon box, which does not contain a PDA, and adds it to the ghost cafe outfit.

## Why It's Good For The Game

Fashion is important.

## Changelog
:cl:
add: Ghost Cafe patrons now spawn with chameleon kits. Dress up! Be fancy!
/:cl:

### DIFF
--- a/code/game/objects/structures/ghost_role_spawners.dm
+++ b/code/game/objects/structures/ghost_role_spawners.dm
@@ -663,6 +663,4 @@
 	new /obj/item/clothing/head/chameleon(src)
 	new /obj/item/clothing/mask/chameleon(src)
 	new /obj/item/storage/backpack/chameleon(src)
-	new /obj/item/radio/headset/chameleon(src)
-	new /obj/item/stamp/chameleon(src)
 	new /obj/item/clothing/neck/cloak/chameleon(src)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/ghost_role_spawners.dm
```diff
@@ -649,3 +650,19 @@
...
+	new /obj/item/clothing/under/chameleon(src)
+	new /obj/item/clothing/suit/chameleon(src)
+	new /obj/item/clothing/gloves/chameleon(src)
+	new /obj/item/clothing/shoes/chameleon(src)
+	new /obj/item/clothing/glasses/chameleon(src)
+	new /obj/item/clothing/head/chameleon(src)
+	new /obj/item/clothing/mask/chameleon(src)
+	new /obj/item/storage/backpack/chameleon(src)
+	new /obj/item/radio/headset/chameleon(src)
+	new /obj/item/stamp/chameleon(src)
```
Ghommie: I think these two could and should be removed. Especially the headset.

CameronWoof: The stamp is harmless, and the lathe in the ghost bar can already make radio headsets (that don't work on the z-level anyway).

Ghommie: Eeeeeeeh, just suppose we may want to use this oufit for similar spawners at some point, that may work against it.

Ghommie (approved):

PR #10371, opened 2019-12-31
call of duty loot boxes

ghost:
## About The Pull Request
uncomments abandoned crates from cargo packs

## Why It's Good For The Game
it's not but they're funny

## Changelog
:cl:
add: loot crates in cargo contraband
/:cl:

### DIFF
--- a/code/modules/cargo/packs/misc.dm
+++ b/code/modules/cargo/packs/misc.dm
@@ -17,7 +17,7 @@
 	name = "Loot Box"
 	desc = "Try your luck with these highly secure loot boxes! Solve the lock, win great prizes! WARNING: EXPLOSIVE FAILURE."
 	contraband = TRUE
-	cost = 10000
+	cost = 15000
 	contains = list(/obj/structure/closet/crate/secure/loot)
 	crate_name = "abandoned crate"
 	crate_type = /obj/structure/closet/crate/large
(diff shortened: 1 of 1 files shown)


YakumoChen:
you can reliably solve these puzzles with an online tool though

ghost:
not wrong (but if you do you're stinky)

code/modules/cargo/packs/misc.dm
```diff
@@ -13,17 +13,16 @@
 //////////////////// Paperwork and Writing Supplies //////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-/* I did it Kevin
 /datum/supply_pack/misc/abandonedcrate
```
Trilbyspaceclone, line 16: I always wanted to add in 10mm guns into cargo, just thought it would never get merged
```	
	if(97)
			new /obj/item/clothing/mask/balaclava
			new /obj/item/gun/ballistic/automatic/pistol(src)
                        new /obj/item/ammo_box/magazine/m10mm(src)

Ghommie (approved):
I reliably solve these puzzles already 9 times out of 10 on my own. Also it's one huge fucking RNG roll. Most of this stuff is icky useless and I have been thinking of reworking it since forever, but there is always other things to do.

lolman360:
can we add goon code locks instead

Ghommie:
Hangman code wouldn't require much work actually: the vocables must be stored in a server config though, to avoid even easier cheating (along hangman solvers).

PR #10375, opened 2019-12-31
Add a Radiation Hardsuit

r4d6:
## About The Pull Request
CentCom decided to send its "latest" technology in radiation protection suits to the Station with a radiation hardsuit.

This add a space-proof and radiation-proof hardsuit, even though it is only resistant to rads and bio, everything else being set to 0.

This also add a crate to buy one of those hardsuit at 3500$ (Between the Engie Hardsuit and the Atmos Hardsuit) With a gasmask and an internal tank.

## Why It's Good For The Game

I wanted to do a radiation proof hardsuit, and either do it as a Traitor Item or a Cargo Item, and chose for a Cargo Crate instead. This could be also useful in case of a Delaming SM, to get a shard as a traitor, or to do change to a Singulo Engine without spending the rest of the shift in the shower.

Keep in mind it doesn't protect against melee, bullets, lasers, energy weaponry, bombs, fire or acid. **at all**. So I feel it is balanced, as the normal radsuit has more protection in general than this one, but isn't space-proof.

## Changelog
:cl:
add: Added a Radiation Hardsuit
add: Added a Radiation Hardsuit crate
/:cl:


CameronWoof:
Why is this a hardsuit instead of a softsuit/voidsuit? Wouldn't that make more sense with its armor values?

r4d6:
Do we even have soft suits ?

GrayRachnid:
There's already a hardsuit that cargo can order which has 100% radiation protection. The name is escaping me but I don't think this is necessary when it already exists.

code/modules/cargo/packs/engineering.dm
```diff
@@ -73,6 +73,17 @@
 	crate_name = "atmospherics hardsuit"
 	crate_type = /obj/structure/closet/crate/secure/engineering
 
+/datum/supply_pack/engineering/radhardsuit
+	name = "Radiation Hardsuit"
+	desc = "The Singulo is loose? Do you need to do a few changes to its containment and don't want to spent the rest of the shift under the shower? Get this Radiation Hardsuit! It protect from radiations and space only. Ask the CE to open."
+	cost = 5000
```
Trilbyspaceclone: This is not 5000"$" this is 10k in dollars. Could prob be 3500 do to being a passive hardsuit.

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -204,6 +204,31 @@
...
+	desc = "A special suit that protects against radiation and space. Not much else unfortunately."
+	icon_state = "hardsuit-rad"
+	item_state = "nothing"
+	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 100, "fire" = 0, "acid" = 0)
+	helmettype = /obj/item/clothing/head/helmet/space/hardsuit/engine/rad
+	resistance_flags = FIRE_PROOF
+	rad_flags = RAD_PROTECT_CONTENTS | RAD_NO_CONTAMINATE
+	tauric = FALSE		//Citadel Add for tauric hardsuits
+
+/obj/item/clothing/head/helmet/space/hardsuit/engine/rad/attack_self()
```
Trilbyspaceclone, line 229: Is this ment to use engi hardsuit icon?

r4d6: It is using its own suit icon, a recolor of old suit, and I didn't want to do a tauric/naga version nor an in-hand version, to the tauric flag is set to false and the icon set to nothing mean it will not show up on the player when holding it in hands.

code/modules/cargo/packs/engineering.dm
r4d6: You know crate prices better than me. Also I think it sell for the same price than the Engie hardsuit due to me not setting a price and it being a child of the hardsuit.

GrayRachnid:
Found it, the MASON Rig has 100 rad protection. Not sure why this is necessary.

r4d6:
The MASON RIG has a _lot_ of general protection, and I am not sure if it was removed, but at one point you were able to touch the SM with it. It also cost more.

The MASON Rig is basically a straight upgrade to the Engie hardsuit and maybe the Atmos one. Saying that because it exist there is no need for a radiation hardsuit is like saying "Hey, the CE suit exist, why bother with the normals one?". At least this is the way it sound to me.

The Radiation Hardsuit only protect from Radiation, but cost less as a result. The MASON protect against lots of thing, and is costly.

GrayRachnid:
I mean technically if you want to protect yourself from radiation then radiation suits are the answer. It's just that cargo bloat has been happening a lot often and I don't see a point in adding more to it when alternatives already exist.

If you want to protect yourself from both radiation and pressure then get a MASON.
That's how I feel about it at least.

r4d6:
The Radsuit cost 7000$, and only protect from space and Rads
The MASON cost 30000$, and protect from :
- Bombs
- Rads
- Acid
- Being on fire
- Being in a fire
As a side note, it can hold a RCD or a RPD and has a Diagnostic HUD.

GrayRachnid:
image

MASON is 15,000, not 30,000.

If you want to be rad protected, wear a rad suit.
If you want pressure protection, wear a hardsuit.
If you want both, give cargo a reason to spare you some money or ask the CE for it.

Rad and Pressure defenses in one are expensive for a reason, the reason being fusion and supermatter handling are the only reason you'd need the both together. Both of these should not be made easier to handle.

Tupinambis:
Soft suit makes a lot more sense, and honestly they are underutilized on TG code. I'd say make it a soft suit and maybe add it as a rare drop in rad lockers. Otherwise this is pretty useful for working with the SM and Singularity alike so I am down with it.

Trilbyspaceclone:
Fucken why are mason rigs still in cargo?

Putnam3145 (approved):
code's good, though i'm still questioning a bit what niche the standard engi hardsuit has with this in mind

PR #10380, opened 2020-01-01
powerfists throw people faster but also leak their gas into the air

Hatterhat:
## About The Pull Request
port of tgstation/tgstation#43000, see title and changelog
## Why It's Good For The Game
fist
## Changelog
:cl: Unit2E & JMoldy, ported by Hatterhat
balance: the powerfist now punches people away at high velocity depending on setting.
add: the powerfist now leaks the gas it uses onto the tile you're on.
tweak: Adds weak punch variations for empty and near-empty punching.
/:cl:

### DIFF
--- a/code/game/objects/items/powerfist.dm
+++ b/code/game/objects/items/powerfist.dm
@@ -50,7 +50,6 @@
 		if(tank)
 			updateTank(tank, 1, user)
 
-
 /obj/item/melee/powerfist/proc/updateTank(obj/item/tank/internals/thetank, removing = 0, mob/living/carbon/human/user)
 	if(removing)
 		if(!tank)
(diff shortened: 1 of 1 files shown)


necromanceranne:
This is probably one of the more brutal changes the power fist ever got, and made it obscenely powerful because of the stun as a result of setting 3 flinging people at max velocity into walls. 

For cit, given being out of combat mode is 1.5x melee damage, setting 3 on an unprepared combatant with no armor would likely be 90 brute, and wall collision is 10. That's straight crit. I have no idea what the effects of max velocity collision would do here other than knockdown but it is savagely powerful on the unsuspecting.

Is this a bad thing? Not really because the power fist requires aiming and frankly melee weapons are a joke. It does let you use bodies as ranged weapons too.

(In addition to this change on tg, the power fist received a 2 tc discount from the original price from 8 tc. Which meant that it was probably better than an esword for the esword+ebow combo at the time. When ebows also lost 2 tc on their price, it became a combo you could mix with all sorts of weapons.)

necromanceranne:
Actually uuuh does this weapon even interact with combat mode it seems to be apply.damage now that I'm looking at it.

Hatterhat:
i got no idea how that shit plays out

Arturlang:
It does actually kill in 3 hits, critting in 2 on unarmored people.

nik707:
This one I'm gonna leave to others because to me it sounds obscenely OP but IDRK

Ghommie (approved):
Its damage output is not influenced by the silly stam-combat-mode-thing.
Let's see how this plays out.

PR #10393, opened 2020-01-02
Significantly lowers the forcefield projector health, lowers armor.

chefferz:
_**Redid this because I am a dumb and did not branch properly.**_

Having a 250 health portable shield that you can deploy on up to three tiles, has shielding resistant to everything but melee, and can be reset and redeployed quickly has been nothing but abused lately.

This drops the health down to 100 from 250, keeping shield integrity. Armor is halved on bullets and lasers.


chefferz:
I would like to note that Kevinz suggested making it only work when held in hand, but I do not know how to implement that.

silicons (approved):

silicons:
yeah acceptable
that said i don't even see why it needs armor but eh yeah dab

PR #10396, opened 2020-01-02
reverts that one TEG nerf that forced it to require an arbitrary burnmix

Hatterhat:
reverts #8056 because i dunno it seems like a TEG would be clogged to shit after it burns enough to replace all(!!!) the fuel with co2 and then you don't have an engine you have a very large brick with pipes


Hatterhat:
also thermodynamics compliance is for cowards

Detective-Google:
would be nice for TEGs to be literally useful at all for anything because even non-freezer cheese TEGs are useless atm.

Ghommie (approved):

PR #10399, opened 2020-01-02
Knocks down cargo passive gen 750 -> 500

silicons:
## About The Pull Request

title

## Why It's Good For The Game

People keep saying that cargo has nothing to do because of this, and with Trilby's exports I feel like this is augmenting easy exports too much. This is supposed to be an emergency reserve thing letting you trickle order if you have nowhere else to go but it's a little strong right now.

## Changelog
:cl:
balance: Cargo passive gen is now 500 down from 750.
/:cl:


Ghommie (approved):
ok

PR #10400, opened 2020-01-02
Gives ability to buckle yourself to a toilet

chefferz:
I don't think this needs more explanation.

### DIFF
--- a/code/game/objects/structures/watercloset.dm
+++ b/code/game/objects/structures/watercloset.dm
@@ -5,6 +5,8 @@
 	icon_state = "toilet00"
 	density = FALSE
 	anchored = TRUE
+	can_buckle = 1
+	buckle_lying = 0
 	var/open = FALSE			//if the lid is up
 	var/cistern = 0			//if the cistern bit is open
 	var/w_items = 0			//the combined w_class of all the items in the cistern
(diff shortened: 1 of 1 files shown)


ghost:
this is very uncanny

chefferz:
> 
> 
> this is very uncanny

if you look at the branch name you might find a easter egg to why this is being done preemptively!!?

code/game/objects/structures/watercloset.dm
```diff
@@ -5,6 +5,8 @@
 	icon_state = "toilet00"
 	density = FALSE
 	anchored = TRUE
+	can_buckle = 1
```
silicons: 1 --> TRUE

necromanceranne:
What, you need a seatbelt so you don't fall off and shit yourself?

code/game/objects/structures/watercloset.dm
chefferz: I tested this locally. Can_buckle has to equal 1.

chefferz: Reference > https://github.com/Citadel-Station-13/Citadel-Station-13/blob/f2fc50119d33b36d1daccf33d2fc103f44c95ab9/code/game/objects/structures/beds_chairs/chair.dm

Ghommie (approved):
I doubt the branch name is a reference to space vacuum toilets... maybe Spaceball?

PR #10401, opened 2020-01-02
Add three more Supermatter Engine variation for Box

r4d6:
## About The Pull Request
This add three more SM Variations, the 3x, the 1x3 and the 5x5, all visible with the original 3x3 as a reference point [here](https://imgur.com/a/3LaQPIi)

## Why It's Good For The Game
This allow for a little more variations concerning the Engines without actually adding any new Engines. Because for as much as I love the Antimatter Engine, it is boring and leave no place for experimentation, and the TEG is currently broken so there is no point to adding it. Plus I had those variations in mind for a while now.

This _**does**_ require the headmins to update the configs though. Or else they will never see the light of day.

## Changelog
:cl:
add: Added 3 SM Engine variations.
/:cl:


Putnam3145:
good lord 3x is a meme. 1x3 would be fun, 5x5 is almost easy mode, but 3x would be, uh, "most engineers should probably just do the solars at this point" unless the room is expanded a little

FloranOtten:
The smaller 1x3 room is a problem because it cannot contain as much gas. A properly upgraded SME with straightpipes, extra cooling, etc. Could handle this, but a single scrubber and vent would struggle to drain and refill the room in a delamination (possibly even normally but that's speculation). Additionally, the outermost rad collectors would do nothing, and a delaminating engine would heat up the room significantly due to the no shutters.


The 3x engine is a meme and i can see nothing going wrong beyond possible server crashes with delams and/or a nasty phenomenon where the engine can produce a plas/ox mix, which is instantly processed by the other crystal causing it to produce a more dangerous mix, which the first crystal processes to create a more dangerous mix, etc.


The 5x5 could work, but I'd imagine the room fills with more gas in the center due to our Atmos system, making delam more likely and recovery harder.

r4d6:
Nothing prevent the admins from simply not adding the 1x3 engine to the rotation

Putnam3145:
> the engine can produce a plas/ox mix

correction: the engine does nothing but produce a rather hot plas/ox mix and radiation, so this is a guaranteed thing that will happen

Putnam3145 (approved):

PR #10420, opened 2020-01-03
Increase radiation plasma usage in radiation collector

MrJWhit:
## About The Pull Request

Makes the unfilled plasma tanks run out in the middle of the shift, _not_ the plasma tanks refilled with the plasma tank. The increase is +170% the normal rate, it looks like a lot, but I can show the math and trials that I did if anyone's curious.

EDIT: Also, please testmerge this before merging, the math could be off for a full shift, some of the timing is off on my testing.

## Why It's Good For The Game

A few reasons:
1. It teaches engineers to fill plasma tanks, giving more of a reason to learn how to properly mass manage plasma tanks.
2. Gives players who are good at the SM another avenue to follow up on that.
3. Teaches players how the radiation collectors work ("Why did the SM stop producing energy") from purely in-game mechanics.
4. Gives more of a source of tritium for general usage.

## Changelog
:cl:
tweak: Increases plasma usage in radiation collectors
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/encryptionkey.dm
+++ b/code/game/objects/items/devices/radio/encryptionkey.dm
@@ -38,7 +38,7 @@
 	name = "robotics radio encryption key"
 	desc = "An encryption key for a radio headset.  To access the engineering channel, use :e. For research, use :n."
 	icon_state = "rob_cypherkey"
-	channels = list("Science" = 1, "Engineering" = 1)
+	channels = list("Science" = 1)
 
 /obj/item/encryptionkey/headset_med
 	name = "medical radio encryption key"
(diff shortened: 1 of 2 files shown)


dapnee:
Would this actually effect anything after the SMES are filled by that point? I feel APCs should eat more power but also be able to restore their power from the wire networks as well if you want this to be pronounced enough. Other than that, I’m for this.

MrJWhit:
Yeah, this doesn't actively drain power on the SMES, and they likely can supply the station for the time that it takes to leave, but, it still promotes the actual act of messing with the radiation collectors. I'm not going to lie, messing with the APC drain is honestly pretty scary because that's not a minor element, which I'm more comfortable with as a coder.

r4d6:
Wasn't there a reason for that to be removed?

davidchan13:
Could this at least be coupled with part upgrades to the collectors doing something? (Less gas usage, better power/point generation, ect...)

Ghommie (approved):

PR #10423, opened 2020-01-03
Remove the engie-borg ability to RCD R-Walls

r4d6:
## About The Pull Request
This remove the ability to deconstruct r-walls from the EngieBorg's RCD

## Why It's Good For The Game
All the others actual RCDs that can remove R-walls are Admin-Only. And the Engie Borgs were recently buffed since they can now place circuits into machine frames, make APCs, Air Alarms and all that, plus they don't need to upgrade their RCD and don't need to refill them. So it is only fair to remove something that didn't really make sense given that they are not _that_ weak to need it and they have AA anyway.

I don't consider Mech's RCD to be an actual RCD since it is both not a subtype of a RCD like all the others, but also because it don't do the same checks as every other RCDs.
And ERTs are admin-only.

## Changelog
:cl:
del: Removed the ability to RCD R-walls for borgs
/:cl:

### DIFF
--- a/code/game/objects/items/RCD.dm
+++ b/code/game/objects/items/RCD.dm
@@ -556,7 +556,6 @@ RLD
 /obj/item/construction/rcd/borg
 	no_ammo_message = "<span class='warning'>Insufficient charge.</span>"
 	desc = "A device used to rapidly build walls and floors."
-	canRturf = TRUE
 	upgrade = TRUE
 	var/energyfactor = 72
 


silicons:
Entirely against this. Nothing wrong with the special engineering cyborg being able to build/deconstruct better than others. You get almost nothing better late game than regular engineers, there's nothing wrong with them having this and a few other additional better perks. I don't care about consistency when it comes to a better balance decision that allows for more interesting scenarios. 
Cyborgs are limited. Regular RCDs are not. This isn't also a "they don't need it" thing, you don't remove anything that isn't needed on a day by day basis. Don't balance on presumptions after reading code when you yourself admittedly didn't even realize this was a thing for the entire time you've played.

r4d6:
Fine

PR #10447, opened 2020-01-05
[READY] Bloodsucker tweaks

Arturlang:
This tweaks bloodsuckers to have DNA, due to people deciding that blackmailing people tonget clone scans is a good idea.
Also adds a new check for vampire antag datum for simplicity, and adds a no clone check to cloning, instead of the TRAIT_NOCLONE that DNA scanners use.
Masquarade now also allows for the vampire to be infected by viruses and will undo their inherent flash resistance, as per these are common methods to find out who is a vampire. Stamina regen and clone damage was also nerfed somewhat, as they were really high previously. The cloak ability is no longer a completele invisibility, and takes some more time to reach max invisibility.Lastly, vampires unconcious will now regen faster, otherwise torpor lasts a really long time.
Also buffs staking by allowing neckgrabs and unconcious people to be staked
## Changelog
:cl:

tweak: Bloodsuckers are now vulnerable to their DNA being messed with
tweak: Masquarade ability for bloodsuckers now makes them vulnerable to viruses and disables their flash protection
tweak: Cloak of Darkness for bloodsuckers is no longer as good.
tweak: You can now stake people who you have neckgrabbed
tweak: Legion cores no longer heal vampires
This will make the very bad meta against bloodsuckers not as bad, and let them be stealthy without being cheesed.


code/modules/antagonists/bloodsucker/objects/bloodsucker_crypt.dm
```diff
@@ -486,9 +484,7 @@
 	if(lit)
 		for(var/mob/living/carbon/human/H in viewers(7, src))
 			var/datum/antagonist/vassal/T = H.mind.has_antag_datum(ANTAG_DATUM_VASSAL)
-			var/datum/antagonist/bloodsucker/V = H.mind.has_antag_datum(ANTAG_DATUM_BLOODSUCKER)
-			if(V || T) //We dont want vassals or vampires affected by this
-				return
+			if(!isvamp(H) || !T) //We dont want vassals or vampires affected by this
```
ghost: and not or? this is where travis failed

Arturlang: Whoops, will fix tommorow

necromanceranne:
Maybe give vampires the ability to remove viruses at will, just so they don't have to suffer from a virus if they so choose (aka usually when it's malicious), while also circumventing antag tests.

Arturlang:
I cant really do that easily, though i could attach the virus weakness to massquarade to further incentivize using it

code/modules/antagonists/bloodsucker/objects/bloodsucker_crypt.dm
Arturlang: It compiles locally, goddamnit it seems i lost some of the changes

Arturlang:
@necromanceranne Currently, in this PR, they can keep viruses as long as they have masquarade on, i think, and while masquarade is not on, they cannot be infected by viruses.

code/modules/antagonists/bloodsucker/items/bloodsucker_stake.dm
```diff
@@ -78,7 +78,7 @@
 	var/mob/living/carbon/C = target
 	// Needs to be Down/Slipped in some way to Stake.
 	if(!C.can_be_staked() || target == user)
-		to_chat(user, "<span class='danger'>You cant stake [target] when they are moving moving about! They have to be laying down!</span>")
+		to_chat(user, "<span class='danger'>You cant stake [target] when they are moving moving about! They have to be laying down or grabbed by the neck!</span>")
```
Linzolle:
```suggestion
		to_chat(user, "<span class='danger'>You can't stake [target] when they are moving about! They have to be laying down or grabbed by the neck!</span>")
```

Linzolle (approved):
looks good aside from that line.

PR #10458, opened 2020-01-06
Adds the ability to knock down sprinting spacemen with heavy objects

deathride58:
## About The Pull Request

This PR adds the ability to knock down a sprinting spaceman by simply smacking them with a heavy object. For items using the default weight calculation, this means bulky items and larger. This can behavior can be forced on  on a per-item basis with the `SPRINT_KNOCKDOWN_FORCED` value for `sprint_knockdown` or disabled entirely via `SPRINT_KNOCKDOWN_DISABLED`

## Why It's Good For The Game

Sprinting is still ridiculously strong in combat, and the sprint buffer exacerbated the issues quite a bit. This means the meta in melee combat is still to just yakkety-sax your way around the opponent, which simply is not fun to play against since you have to rely on high resolution displays with DPI scaling disabled and low mouse sensitivities for a reasonable accuracy rate in combat (And I'm someone that unashamedly exploits that fact). There's also the additional current issue that bulky items and heavier/larger are fairly useless in combat due to their high stamina cost for little to no gain in way of utility or damage, as lighter/smaller items more or less do exactly the same things they can do but more efficiently.

So this PR hits both birds with one stone. This allows you to use any bulky item to knock down spacemen that're mid-sprint by simply attacking them once. This means that sprinting in melee combat suddenly becomes a much riskier thing to do against most targets, as getting hit one single time while you're sprinting instantly puts you on the ground, which is a fairly inefficient place to fight due to the incredibly reduced movement speed and reduced melee damage. This also adds actual utility to bulky objects in the context of combat, as suddenly you have a choice between a light stamina-efficient object that deals 15 force or a heavy stamina-inefficient object that deals 15 force and is capable of knocking down sprinting spacemen, creating a much better variety in terms of viable options for melee combat, and giving more stationary combat styles a way to hard counter the current yakkety-sax meta.

## Changelog
:cl:
add: You can now knock down sprinting spacemen by simply smacking them with a sufficiently heavy object. Heavier objects are capable of disarming on top of the knockdown. Yakkety-sax meta begone.
/:cl:

### DIFF
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -196,3 +196,11 @@ GLOBAL_LIST_INIT(shove_disarming_types, typecacheof(list(
 #define TOTAL_MASS_HAND_REPLACEMENT	5 //standard punching stamina cost. most hand replacements are huge items anyway.
 #define TOTAL_MASS_MEDIEVAL_WEAPON	3.6 //very, very generic average sword/warpick/etc. weight in pounds.
 #define TOTAL_MASS_TOY_SWORD 1.5
+
+//Defines that control the sprinting knockdown on hit.
+#define SPRINT_KNOCKDOWN_DISABLED	0
+#define SPRINT_KNOCKDOWN_DEFAULT	1
+#define SPRINT_KNOCKDOWN_FORCED		2
+#define MIN_SPRINT_KNOCKDOWN_MASS	TOTAL_MASS_BULKY_ITEM
+#define SPRINT_KNOCKDOWN_STRENGTH	13 //Strong enough for huge items to do a disarm, too weak for bulky items to disarm. Based off total mass
+#define CAN_SPRINT_KNOCKDOWN(I)		(I?.sprint_knockdown && (I?.sprint_knockdown == SPRINT_KNOCKDOWN_FORCED || I?.getweight() >= MIN_SPRINT_KNOCKDOWN_MASS))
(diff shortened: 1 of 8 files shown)


BlackMajor:
YOU BETTER MAKE THIS WORK WITH THROWING THEM TOO SO I CAN YEET A HAMMER AT THE BACK OF A GUYS HEAD WHILE HE RUNS AWAY

deathride58:
OKAY, DAD, **GOD**

ghost:
can you make them say OOF when they're hit like that?

necromanceranne:
Doesn't this mean eswords knock people down? Very gamer.

deathride58:
Eswords are too light to proc the knockdown effect, as are a handful of other lighter bulky objects.

necromanceranne:
Ah I see. Also, worth considering whether you also want Proto-Kinetic Crushers also knocking people down, as, much like spears, they have an embed chance and are bulky.

ghost:
not really related to this pr but right now there's no reason to use blunt weapons because sharp and blunt weapons have the same effects in combat (bleeding/brain damage) but sharp weapons generally do more damage and can dismember corpses and sharp weapons usually cost less stamina

necromanceranne:
That's not really a necessary contrast, since melee weapons in spess aren't categorized so neatly. Most heavy objects tend to be blunt ones, like toolboxes, and make for good throwing weapons. Toolbox throwing is kind of a classic example. Many sharp weapons range from terrible throwing weapons to some of the best ones, but often you lose your weapon due to embedding. They have their perks and downsides.

Nobody wants to make stun batons or baseball bats any stronger in terms of lethal forcr than they are since they carry additional effects from being hit by them. Often it's more important what an individual weapons does to someone than the distinction of whether it is cutting or bludgeoning or burning and whether that has to necessarily be the determinating factor in whether or not its a good weapon or what effects that should have on you.

Also power fist says hello.

Ghommie:
Conflicts.

PR #10463, opened 2020-01-07
Allowed random engines to be weighted

Putnam3145:
## About The Pull Request

Allows adding a weight to the random engines for box station. Also renames the random engine ruins to make more clear what they are.

## Why It's Good For The Game

A bit more variety in these things is good. More variety in maps is *very* good.

## Changelog
:cl:
config: Random engines are now weighted.
/:cl:


KathrinBailey:
This is great, but it'd be cool if you lowered the singulo chances until it gets a revamp to not irradiate engineering and upped SM chances, since having engines that can destroy the entire station for nearly no effort whatsoever is kinda bad.

Ghommie (approved):

PR #10409, opened 2020-01-02
IT WORKED ON LOCAL

silicons:

PR #10464, opened 2020-01-07
adds key_cache[] resets to the returns added by #10443 (obsolete now)

deathride58:
over on rp, this not being present meant that everyone got kicked out when they tried to join due to the isbanned check when combined with tg's isbanned antispam. this ports that fix that we were missing
(relies on #10409 )


deathride58:
obsolete due to new commit

PR #10474, opened 2020-01-08
[s] Fixes traitors

Putnam3145:
## About The Pull Request

Traitors were checking for storytellers in non-dynamic modes, sometimes. Yikes.

Also allowed removing of objectives, which #9716 silently commented out a single line of while keeping the rest of the proc exactly the same, making it seem as if objectives were being deleted when they weren't. Yikes.

## Why It's Good For The Game

Assassination bad? But it's not working cause of a bug? So let's bring it back.

## Changelog
:cl:
fix: Traitor objectives don't take a dump when they try to add an assassinate anymore
admin: Objective removal with antag panel no longer commented out silently while still being an option that gives useful feedback on stuff it's not doing in any respect
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -525,7 +525,7 @@
 		if(!objective)
 			to_chat(usr,"Invalid objective.")
 			return
-		//qdel(objective) Needs cleaning objective destroys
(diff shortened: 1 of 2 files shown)


Ghommie (approved):
I was merely following whichever changes were done on the original PR, that's it

PR #10475, opened 2020-01-08
Voracious quirk update

KokiSnake:
## About The Pull Request
My first code contribution.  Small change to the voracious quirk so that it also gives you a happy moodlet for overeating.  

## Why It's Good For The Game
I think there ought to be a quirk which removes the punishment for overeating, especially for citadel and its proclivities.  Being fat shouldn't always be a bad thing.

## Changelog
:cl:
add: New mood event, "fathappy"
add: Added extra text to voracious quirk so that people know about new behavior
code: added check for voracious trait when overweight
code: I don't know why but it pushed a removal and re-adding of wuv.dm? Don't think it's necessary to push but don't know how to remove it
/:cl:


Putnam3145:
(that's not what the code changelog is for, for the record, it's for code-only things that don't have an effect on the game)

ghost:
someone has a foodie fetish

code/datums/components/mood.dm
```diff
@@ -258,7 +258,10 @@
 /datum/component/mood/proc/HandleNutrition(mob/living/L)
 	switch(L.nutrition)
 		if(NUTRITION_LEVEL_FULL to INFINITY)
-			add_event(null, "nutrition", /datum/mood_event/fat)
+			if(HAS_TRAIT(L, TRAIT_VORACIOUS))
+				add_event(null, "nutrition", /datum/mood_event/fathappy)
+			else
+				add_event(null, "nutrition", /datum/mood_event/fat)
```
deathride58, line 264: You can oneline this check by making use of ternaries, like so
```
			add_event(null, "nutrition", (HAS_TRAIT(L, TRAIT_VORACIOUS) ? /datum/mood_event/fathappy : /datum/mood_event/fat))
```

code/datums/mood_events/needs_events.dm
```diff
@@ -3,6 +3,10 @@
 	description = "<span class='warning'><B>I'm so fat...</B></span>\n" //muh fatshaming
 	mood_change = -4
 
+/datum/mood_event/fathappy
+	description = "<span class='nicegreen'>I'm so fat...just how I should be!</span>\n"
```
deathride58, line 7: could probably just go with a nod to the original wellfed flavortext with less grammatical errors
```suggestion
	description = "<span class='nicegreen'>My belly feels round and full, just as it should be!</span>\n"
```

deathride58 (approved):
If anything else puts wuv.dm in its proper place in the DME then it'll be fairly easy to fix conflicts in there

deathride58:
You don't need to be incredibly verbose in the changelog section! A simple `add: Makes the voracious trait give you a positive moodlet for being fat instead of a negative.` would suffice

code/datums/mood_events/needs_events.dm
KokiSnake: Oh okay!  I figured it was more legible if it was broken up but i'll definitely make use of that in future.
Fair enough, I just went with the text that made the most sense to me at the time.
You aren't wrong about the foodie fetish...
And yeah, I just figured more commentary is better, i'll be less explicit if i do prs in the future.
Alright putnam, good to know :)

Ghommie: > My belly feels round and full

Please don't spread your enthusiasm for vore around the sane portion of the codebase like that, deathride.

Ghommie:
Just port this PR: https://github.com/tgstation/tgstation/pull/41541, thank you.

PR #10489, opened 2020-01-09
Fixes and improvements for chameleon clothes

Seris02:
## About The Pull Request

Makes it so the code actually gets run where it didn't before and:
Clothes now actually cover the relevant bits when you use chameleon.
ie before, you changed it to a hardsuit, your tail would stick out
now the tail won't
same with hats and etc for all the chameleon clothes

## Why It's Good For The Game

Because chameleon clothes are supposed to be a disguise, but they aren't very good if your tail is sticking out of the hardsuit which normally doesn't allow tails to stick out.

## Changelog
:cl:
tweak: the way chameleon clothes update on change
fix: chameleon cloak icon
/:cl:


Linzolle (approved):
thank god

PR #10494, opened 2020-01-09
allows you to do normal talking with a chameleon mask [closed pending recoding]

Seris02:
## About The Pull Request

Alt click a chameleon mask to stop hissing, rattling, lisping, etc.

Also.
If you can find a way to do this with comsig and whatever else, go for it, I tried and failed because it did extremely weird things.

## Why It's Good For The Game

If you're going to change your voice, at least make it be able to sound real.

## Changelog
:cl:
add: chameleon mask voice regulator
/:cl:


silicons (dismissed):

Seris02:
I really couldn't find another way to do it.
actually wait. hm. nevermind I can do that.

Ghommie (changes requested):
This PR has some issues: it adds a silly forcednormalspeech variable to a mob and requires snowflake checks to work half-right. It's pretty janky considering that with a few adjustments this feature could reliably use signals.

PR #10510, opened 2020-01-10
reverts additional sms

silicons:
not tested and not working


r4d6:
You can just ask an headmin to change the config.

PR #10512, opened 2020-01-10
Halves research points obtained through bombs. Makes 25k cap.

chefferz:
Does as it says on the tin.

Getting a bit exhausting seeing research make a 50k point bomb sub 10 minutes into the shift and have a fully decked out research tree that can counter whatever has been announced. This makes it less stronger, because it seems almost every antagonist now has to hard counter science far too early in the shift.

### DIFF
--- a/code/game/machinery/doppler_array.dm
+++ b/code/game/machinery/doppler_array.dm
@@ -121,7 +121,7 @@ GLOBAL_LIST_EMPTY(doppler_arrays)
 		say("Explosion not large enough for research calculations.")
 		return
 	else if(orig_light < 4500)
-		point_gain = (83300 * orig_light) / (orig_light + 3000)
+		point_gain = (41650 * orig_light) / (orig_light + 3000)
 	else
 		point_gain = TECHWEB_BOMB_POINTCAP
 
--- a/code/__DEFINES/research.dm
+++ b/code/__DEFINES/research.dm
@@ -72,4 +72,4 @@
 	TECHWEB_POINT_TYPE_GENERIC = "General Research"\
 	)
 
-#define TECHWEB_BOMB_POINTCAP		50000 //Adjust as needed; Stops toxins from nullifying RND progression mechanics. Current Value Cap Radius: 100
+#define TECHWEB_BOMB_POINTCAP		25000 //Adjust as needed; Stops toxins from nullifying RND progression mechanics. Current Value Cap Radius: 100


ghost:
just a note that deleting toxins will just tell people to find out the other far far more broken methods of gaining points, and toxins is a 1 and done flat gain that requires knowledge of how atmos works (or a guide) where-as the other methods of gaining points don't have a cap and can gain points faster without having a guide

Hatterhat:
:salt:

GrayRachnid:
Time to bring forth the meta of harvesting corpses.

code/game/machinery/doppler_array.dm
```diff
@@ -121,7 +121,7 @@ GLOBAL_LIST_EMPTY(doppler_arrays)
 		say("Explosion not large enough for research calculations.")
 		return
 	else if(orig_light < 4500)
-		point_gain = (83300 * orig_light) / (orig_light + 3000)
+		point_gain = (41650 * orig_light) / (orig_light + 3000)
```
GrayRachnid, line 124: This shouldn't have been touched.

GrayRachnid (changes requested):

silicons (changes requested):
Bad nerf, there's ways to make toxins more interesting or atleast more hands on/less blindfire that doesn't make it useless/near useless and still the same boring run (average scientists run it in 10 minutes).

I'd rather something else be done like requiring a far lower specific range of detonation (which forces people to figure it out although I'm sure someone will make a calculator or some shit), which will make toxins more interactive.
There's in general not enough ways to get points. There's no reason to hammer down one of the only ones.

chefferz:
> 
> 
> just a note that deleting toxins will just tell people to find out the other far far more broken methods of gaining points, and toxins is a 1 and done flat gain that requires knowledge of how atmos works (or a guide) where-as the other methods of gaining points don't have a cap and can gain points faster without having a guide

If you know of ways to cheese points, then bring them up.

chefferz:
> 
> 
> Bad nerf, there's ways to make toxins more interesting or atleast more hands on/less blindfire that doesn't make it useless/near useless and still the same boring run (average scientists run it in 10 minutes).
> 
> I'd rather something else be done like requiring a far lower specific range of detonation (which forces people to figure it out although I'm sure someone will make a calculator or some shit), which will make toxins more interactive.
> There's in general not enough ways to get points. There's no reason to hammer down one of the only ones.

Gonna be honest and would like a bit more input from maintainers where this isn't their bread and butter. Getting "gamered" because research has enough points 10 minutes in to neuter you as an antag is ridiculous. All this does is delete 10 minutes worth of points.

GrayRachnid:
>If you know of ways to cheese points, then bring them up.

>Time to bring forth the meta of harvesting corpses.

davidchan13:
There is not enough reliable point generation as is and toxins isn't repeatable, and now you want to nerf that even when science rarely maxes out on research anyways? Why.

Putnam3145:
Hyper-noblium production gets me 250000 if I did really, really badly, and 1.5 million on average. BZ production lets you make some points without much issue by just pumping a bit of plasma and N2O into the scrubbers pipes. You can deconstruct slime crossbreeds in the destructive analyzer for 2000 points *at least*.

Personally? Cap it at 50000, but you can't do it in one bomb.

keronshb:
If anything I think the PR should be increasing costs of weapon-based tech trees, rather than the bomb.  Research nanites were also introduced to make points come by faster as well, so rushing nanites after parts has a use now too.

Most of the time the bomb will be rushed to get the parts needed for the station + BoH.  About halfway through making the bomb the miners will already have Mining + Mining Satchels.  So by the time the bomb is finished you should be able to comfortably get all tiers of parts + whatever you want.  So if the points for weapons were raised, this would dissuade weapon rushing if this is the issue and it could be put into more research related trees like nanites or dissection surgery.

Also science is usually a high priority target just because they can research in general, toxin bomb or not, and that the bombs CAN be that destructive vs most antagonists (wiping out cults summoning, wiping out the nukie ship, wiping out the pirate ship, etc).

Ghommie:
Seems like most players and staff is not really in favor of this change.. I honestly don't care too much but there could always be a few other ways to shake off bombastic research, it's not top priority.

PR #10546, opened 2020-01-12
Prevent Playback Devices from triggering Voice Analyzer

r4d6:
## About The Pull Request
This prevent the Voice Analyzer from recognizing Playback Devices as a valid voice input, thus preventing Playback devices from triggering them.

## Why It's Good For The Game
I was told you could attach one of each together and make them trigger each other, thus creating an endless loop and lots of lag. This fix it.

## Changelog
:cl:
fix: fixed being able to use a playback device and a voice analyzer to activate each others
/:cl:

### DIFF
--- a/code/modules/assembly/playback.dm
+++ b/code/modules/assembly/playback.dm
@@ -24,7 +24,7 @@
 	recorded = raw_message
 	listening = FALSE
 	languages = message_language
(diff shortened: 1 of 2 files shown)


Ghommie (approved):
welpsies

PR #10551, opened 2020-01-13
[s]Fixes bloodsucker mesmerize mute

Arturlang:
Whoops, used wrong procs for status effects


Arturlang:
Isnt extremely high priority but is still really important

Ghommie (approved):

PR #10565, opened 2020-01-15
Revert "shotgun pointblanking people actually throws them"

silicons:
Reverts Citadel-Station-13/Citadel-Station-13#10379

sorry
makes pointblanking useless because the throw tends to throw them out of your pellet's way. this also isn't even the way to do this.


necromanceranne:
That's kinda odd since it works fine on tg. The pellets should hit before the throw. The only time it gets weird is stuff like cryodarts and bomb mixes in them because unlike Frag-12, the explosion wasn't following the person. 

Frag-12 would always work fine with it.

There is clearly something else happening here that makes it non-functional.

Hatterhat:
probably guncode being hot garbage and not having the nice /tg/ guncode refactor

Ghommie (approved):

PR #10581, opened 2020-01-17
[WIP] Taser re-rework: Turns tasers into a right-click action of hybrid tasers, nerfs electrodes similarly to taser kneecapping

deathride58:
## About The Pull Request
This PR has the same goals as #10170. It has the intended goal of turning tasers into the utility item they were meant to be as of the combat rework. However, it does so in ways that slightly differ from #10170 in a way that addresses feedback about both the original taser rework and the taser kneecapping.

**Notice:** This PR will be receiving active adjustments in accordance to live observations and player feedback. Things marked as (experimental) are things I am going to be playing around with very frequently. Expected date that the PR will be considered complete is one week from now, January 24th.

Currently unimplemented changelog bits as of last PR desc update include:
- tweak: (todo) Renamed hybrid tasers to something that makes more sense (Lorenerds, hmu. Give me a name that makes sense for a dual-function sidearm.)

## Changelog
:cl: Bhijn
add: Added the ability for projectile weapons to use an alternate fire mode. The taser part of hybrid tasers are now an alternate fire feature.
del: Electrodes no longer block combat mode
balance: Electrodes no longer deal any direct staminaloss. All of their stamloss now comes from the status effect. The knockdown is unaffected and only procs if the target is standing.
balance: Electrodes now only add +2 to your movement delay instead of +8
balance: (experimental) Electrodes now only have half the energy cost. This is primarily to alleviate concerns regarding electrodes being balanced around their old energy cost, where they were THE meta, and had to use up a significant amount of the cell in order to even somewhat compete with disabler bolts.
balance: (experimental) Added a cooldown of 3 seconds to hybrid taser alt-fire. This is primarily just to stop you from being dumb and magdumping your entire cell's worth in electrodes, as this no longer does anything meaningful. Though I suppose it could also help people avoid blindly spamming electrodes to get someone down
/:cl:


lolman360:
why change it to a rightclick? using attackself to switch modes inhand is literally how every other multi-mode energy weapon works, why does the taser need to be snowflakey?

PR #10585, opened 2020-01-17
Rebalances virology a little

Arturlang:
Removes virology reagent blood duplication by making blood a required ingredient instead of a catalyst, also doubles cooldowns for pandemic, making blood and vaccines, as it can make enough to not require conservation.

:cl:
tweak: Virology reagents will no longer magically duplicate blood
tweak: The PanD.E.M.I.C 2200. can no longer produce as much blood and virus vaccines.
/:cl:
Should hopefully making cures for viruses less brainless, and require doctors to not have to rely on making blood out of thin air for tranfusion.


code/modules/reagents/chemistry/recipes/others.dm
```diff
@@ -222,9 +222,8 @@
 /datum/chemical_reaction/mix_virus
 	name = "Mix Virus"
 	id = "mixvirus"
-	results = list(/datum/reagent/blood = 1)
-	required_reagents = list(/datum/reagent/consumable/virus_food = 1)
-	required_catalysts = list(/datum/reagent/blood = 1)
```
Ghommie, line 227: Keep the catalyst, just remove the results. I suspect that doing otherwise may delete the original blood and its data when mixed with amount of virus food equal to or greater than the blood reagent's.

html/changelogs/viro_tweaks.yml
```diff
@@ -0,0 +1,37 @@
...
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, this gets changed to [] after reading.  Just remove the brackets when you add new shit.
+# Please surround your changes in  double quotes ("). It works without them, but if you use certain characters it screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - tweak: "Virology reagents will no longer magically duplicate blood" 
+  - tweak: "The PanD.E.M.I.C 2200. can no longer produce as much blood and virus vaccines."
+
```
Ghommie, line 37: Even if the autochangelog is broken, I don't think is worth the time versus running the changelog compiler and PR'ing it later.

Ghommie (changes requested):

code/modules/reagents/chemistry/recipes/others.dm
Arturlang: Thought it was the other way around, alright.

html/changelogs/viro_tweaks.yml
Arturlang: Someone told me to make a changelog so i started doing so again, but alright.

Arturlang:
Closed because ke:b: did it better

PR #10594, opened 2020-01-18
Removes shrink rays from abductor console.

chefferz:
Does as says on tin.

Removes purchase option of shrink rays from console as an abductor.

This was discussed in depth with multiple maintainers and determined to be something that was forked over from TG but a bit too chaotic for Citadel. This only writes out the purchase option but keeps the physical gun for other uses.


Ghommie (approved):
> This was discussed in depth with multiple maintainers [...]

Well, guess I missed the discussion then.
Sad to see such a fun item vaulted, but some people can't really digest the current state of abductors. Crybabies.

PR #10595, opened 2020-01-18
Zombo

ZomboTheSkunk:
## About The Pull Request

A new away mission able to be selected., SyndicateAssault

## Why It's Good For The Game

More variety for away missions able to be chosen during extended rounds, so the crew has something more to do.

## Changelog
:cl:
Adds the map SyndicateAssault.dmm
/:cl:


Ghommie:
We already have plenty old lotsa-guns-enemies-n-loot centered away missions. This is nothing new. 

You forgot to convert the map to the tgm format and added the escape shuttle area to it (That might cause some issues, you should use a different area). Also the ice spiders and bears will trample and trigger the snare traps anyway.

PR #10605, opened 2020-01-19
Port TG Layers updates and Freezer update

r4d6:
## About The Pull Request
This port https://github.com/tgstation/tgstation/pull/41664 & https://github.com/tgstation/tgstation/pull/48330

#48330 allow freezers to operate on every layers, by using a multitool on the board to change layers
#41664 make pipes better and do some cleanup of the Atmos Code and resprite the layer manifold

### List of changes :

- Make Freezers work on all three 3
- Allow Freezers to switch layers using a multitool on the board
- Scrubbers and vents now stay in the center no matter what their layer is
- Layer Manifold resprited to be more of an adaptor
- Layer Manifold renamed to Layer Adaptor
- Some unused icons were removed
- Add a lot of pipe icons, so that they look better
- Pipes should no longer leave their tile when on layer 1 or 3

## Why It's Good For The Game
1st, freezers working on multiple layers is just a straight plus.
2nd, Those pipes are just straight up better and don't change the mapping in any way. Also code is better too

## Changelog
:cl:
- Make Freezers work on all three 3
- Allow Freezers to switch layers using a multitool on the board
- Scrubbers and vents now stay in the center no matter what their layer is
- Layer Manifold resprited to be more of an adaptor
- Layer Manifold renamed to Layer Adaptor
- Some unused icons were removed
- Add a lot of pipe icons, so that they look better
- Pipes should no longer leave their tile when on layer 1 or 3
/:cl:


r4d6:
Right, so due to outdated code, it seem as if the connections between freezers and pipes on layer 1 & 3 is visually broken. It should still work though, but it will look ugly

silicons (approved):

PR #10616, opened 2020-01-19
Renames created volume in pyrotechnics to multiplier to alleviate confusion

silicons:
Created volume is a misnomer. It uses MULTIPLIER, aka the most "units" of reactions it can make.

With a 1:1:1:1 reaction, you'd expect it to be 600 created for 150/150/150/150 because that's default right?
No, it's 150, because that's the most amount of "times" it can react, assuming one reaction is a single unit (which it isn't but it's more simplified this way.)
With nitroglycerin which ACTUALLY SAYS IT CREATES 2 UNITS FOR 1:1, you'd expect 300/300 to be 600 CREATED VOLUME right?
No.
300. Because that's the most amount of unit reactions it can produce, regardless of result. 

The procs use multiplier, not created volume. This was just misleading from the start.

### DIFF
--- a/code/modules/food_and_drinks/recipes/food_mixtures.dm
+++ b/code/modules/food_and_drinks/recipes/food_mixtures.dm
@@ -15,9 +15,9 @@
 	required_catalysts = list(/datum/reagent/consumable/enzyme = 5)
 	mob_react = FALSE
 
-/datum/chemical_reaction/tofu/on_reaction(datum/reagents/holder, created_volume)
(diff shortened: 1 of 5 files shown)


Ghommie (changes requested):
Then all on_reaction procedures should have their created_volume arg renamed to multiplier for code consistency and fewer issues with possible calls using named arguments.
Don't get me wrong, I'm all in for the clarity that this change brings, but it hinders the code a little in the current state.

Ghommie (approved):

PR #10630, opened 2020-01-20
Nerfs cloning

silicons:
## About The Pull Request

60 clone damage at tier 1 --> 80 clone damage at tier 1
No clone damage at full tier 4 --> 10 clone damage at tier 4
No mutation at tier 4/50% chance of mutation at tier 1 (rng sucks too) --> always bad SE mutation
400% speed at tier 4 --> 200% speed at tier 4

All scaling is linear.

## Why It's Good For The Game

Cloning makes death too worthless at times.
I don't want to take the nuclear route of removing cloning, I also don't want to remove prescanning (at this time at the very least), autocloning is whatever and I feel like removing it will just prompt doctors to camp the cloner. (I don't agree with autocloning though, it's just that with prescanning and without autocloning you're just going to have these issues, and I haven't identified a method to solve said issues yet.)

It may be worthwhile in the future to have autocloning notify medical and be able to be postponed for cases where medical is trying to defib someone instead of clone them.

Cloning was buffed way too much by the introduction of tier 4 parts a few years back anyways. While this does knock down tier 1 effectiveness, it also makes tier 4 not an end-all entirely to cloning being imperfect in usual circumstances. 

The goal here is to make cloning not almost effortless to recover from. I will never agree with stuff like "Add permanent modifiers to max health or something to make cloning suffering" or anything like that which can't be fixed through medical even if it takes effort, but since we don't have a middle ground yet I feel this would make medbay have to still treat people who are cloned yet not make cloning permanently gank you from any effectiveness in the round.

## Changelog
:cl:
balance: Cloning has been nerfed.
/:cl:

### Feedback/discussion is very accepted and welcome.

Side note, experimental cloners are untouched, so your duplicate army memes are safe.

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -4,7 +4,7 @@
 //Potential replacement for genetics revives or something I dunno (?)
 
 #define CLONE_INITIAL_DAMAGE     150    //Clones in clonepods start with 150 cloneloss damage and 150 brainloss damage, thats just logical
-#define MINIMUM_HEAL_LEVEL 40
+#define MINIMUM_HEAL_LEVEL 20
 
 #define SPEAK(message) radio.talk_into(src, message, radio_channel)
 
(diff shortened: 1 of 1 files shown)


Arturlang:
Maybe make the cloner require meat and add a machine that can proccess bodies for it?

ghost:
i like still having clone damage even at tier 4 because at tier 4 cryo tubes are useless
biomass seems like a good idea because medbay can make infinite of it
do you think you could add speed cloning to tier 4? basically eject early and they only get more clone damage instead of missing organs/limbs

EmeraldSundisk:
"It may be worthwhile in the future to have autocloning notify medical and be able to be postponed for cases where medical is trying to defib someone instead of clone them."

I could see this being useful, currently the only way to do this is back someone's data up on a disk and wipe it from the computer, reinserting the disk to clone them.

Tupinambis:
The main problem with cloning nerfs that I see right now, is that with bloodloss being as ridiculous as it is, and organ damage being a thing despite kinda being confusing (but also clearly fucked by bloodloss) AND the changes to medical not being fully documented, cloning tends to be the best option once someone has died in nearly every circumstance because it is much harder and more resource consuming to revive someone that probably has low blood, probably has organ failure, and probably has a fuck ton of damage in general that you can literally only fix through surgery or synthflesh.

It's good that stuff has been added to make medical have more depth, but the way that it has been done is incredibly obtuse (as in, baymed deadass makes more sense to me) and I think that needs to be solved before cloning is touched.

GrayRachnid:
> AND the changes to medical not being fully documented

They are very well documented, there's 3 books about them spread out in medbay and chemistry.

Tupinambis:
Ah my mistake, I never noticed them.

Poojawa:
> 
> 
> Maybe make the cloner require meat and add a machine that can proccess bodies for it?

It's called the Gibber, it's roundstart in the Kitchen Cold room.

shellspeed1:
Replace prescanning with empty cloning.

Ghommie (approved):

PR #10646, opened 2020-01-22
Resolves Minor issues with vore code

### LINKED ISSUE #10611: Vore can make you go permanently blind.
Issue reported from  Round ID: 21047 (Citadel Station 13)

Reporting client version: 512

I've actually seen people complain about this more than once, they get blinded in the vore room but after being ejected, they're still blinded.

Poojawa:
## About The Pull Request

fixes #10611

removes redundant hear checks. It's now only performed once towards the end of a belly cycle. Preferences still apply. 

Tweaked the sound output to respect falloff. sound should be quieter at range, rather than full blast in one's ear.

## Why It's Good For The Game

something something fixing stuff

### DIFF
--- a/code/modules/vore/eating/bellymodes.dm
+++ b/code/modules/vore/eating/bellymodes.dm
@@ -18,7 +18,8 @@
 			return SSBELLIES_PROCESSED
 
 	next_process = times_fired + (6 SECONDS/wait) //Set up our next process time.
-	var/to_update = FALSE
+	var/play_sound //Potential sound to play at the end to avoid code duplication.
+	var/to_update = FALSE //Did anything update worthy happen?
 
 /////////////////////////// Auto-Emotes ///////////////////////////
 	if(contents.len && next_emote <= times_fired)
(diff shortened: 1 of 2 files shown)


code/modules/vore/eating/bellymodes.dm
```diff
@@ -71,18 +72,9 @@
...
-						if(!H.client || !(H.client.prefs.cit_toggles & DIGESTION_NOISES))
-							continue
-						LAZYADD(hearing_mobs, H)
-					last_hearcheck = world.time
-				for(var/mob/living/H in hearing_mobs)
-					if(H && H.client && (isturf(H.loc) || (H.loc != src.contents)))
-						SEND_SOUND(H,pred_digest)
-					else if(H && H in contents && H.client)
-						SEND_SOUND(H,prey_digest)
+				if(M && M.client && M.client.prefs.cit_toggles & DIGESTION_NOISES)
```
Ghommie, line 75: So the world.time hearcheck removal for "preys" is intended, right?

Ghommie (changes requested):
I was thinking... it'd be far safer if `Exited` and `Entered` were used for mob blindness.
Example:
```
/obj/belly/Entered(atom/movable/AM, atom/oldloc)
	. = ..()
	if(isliving(AM))
		var/mob/living/L = AM
		M.become_blind("belly_[REF(src)]"
```
And `cure_blind()` for `Exited()`.

code/modules/vore/eating/bellymodes.dm
Poojawa: Yes because it was simply tied into the predator's noises

Poojawa:
Sure hope you test it :^)

Ghommie (approved):

Ghommie:
That'll be live server testing. :clown_face:

PR #10649, opened 2020-01-22
Completely reworks the hijack objective, and shuttle hijacking in general

silicons:
## About The Pull Request

Instead of needing everyone dead or "captured" on the shuttle, the hijack objective now doesn't care about that at all.
Shuttle hijacking is now based on a telegraphed multi-stage action on the emergency shuttle console that anyone with hijack_speed set on their mind (only traitors at this point in time) are able to hijack the shuttle.
All traitors by default are able to hijack, although at half speed (10 seconds to increase stage) rather than full speed.
Hijack-objective traitors are able to hijack shuttles at full speed (5 seconds to increase stage)

User must be standing still to hack a stage of the process.
Each stage will increase the time left by 30 seconds if the shuttle is in flight, to provide more time to fend off any attackers and continue the hijack.
Reaching the final stage while the shuttle is in flight will force the shuttle to dock in 10 seconds.
Neither of the above will happen if the shuttle is hacked while docked with the station, however.

## Why It's Good For The Game

Hijack almost always results in a giant murderbone a la keep as many off the shuttle as you can, followed by an immediate emag or just straight up maxcapping the shuttle

This will not solve any of that entirely, because it is still, by far, the easiest way to ensure a mechanical victory.

Instead, the reason I am making this is to make hijack potentially more interesting. Now you can choose to fortify/barricade in the bridge and keep loyalist crew out of it. This is obviously harder than just purging the station but I have faith that some people could pull it off. You are no longer mechanically required to ensure everyone on the shuttle dies. Sure, yes, you can just do the old method and hack the console while everyone's dead/on fire, but hey, why not allow for more should the antagonist choose to use this.

This also buffs cases of "I nearly hijacked but someone or another managed to survive the rampage and stowed away." Now you just hack the console, and you still win.

Potential balance concerns may include the fact that many of our shuttles are cramped as fuck and I don't think it'd be a fun time to have to hijack a cramped shuttle vs one of the bigger ones, so this might require some tweaking. Maybe shuttles should be bigger overall, or maybe the smaller ones can have lower hijack stage time complexity and cooldowns. Admittedly, something like Cere shuttle would probably be the least brutal of any shuttle for antagonists to accomplish this, due to the way larger size which allows for better defensive measures.

## Changelog
:cl:
add: Shuttle hijacking has been completely reworked. Alt-click the shuttle as a hijack-capable mind (so traitors, and especially traitors with hijack) to begin or continue a hacking process.
/:cl:


code/datums/mind.dm
```diff
@@ -62,6 +62,8 @@
 
 	var/force_escaped = FALSE  // Set by Into The Sunset command of the shuttle manipulator
 
+	var/hijack_speed = 0			//whether or not we can hackerman the shuttle and hijack it. 0 means you can't, 1 is normal speed hack, 2 is double, 0.5 is half, etc.
```
Putnam3145: good lord just use a component. hell, use an element. hell, you're already adding it to the damn antag datum, just USE THE ANTAG DATUM, there's not much good reason to have this duplicated

Putnam3145: i'm saying this cause i want to reduce the mind/mob bloat at some point and this adds to it and DAMN

ghost:
can you make hijacked shuttles CRASH into centcomm?

Arturlang:
That would be a neat idea to happen on perhaps attempted hijack but failing it in the course of it

ShadeAware:
Failed but attempted hijacks should crash into CC, successfully hijacked shuttles should change course for a Syndicate base 👀

necromanceranne:
The only concern I have is metagamers destroying the console to prevent hijacks. You know people will do this even if it's nonsensical

code/modules/shuttle/emergency.dm
```diff
@@ -3,13 +3,38 @@
...
+#define STAGE_3 3
+#define STAGE_4 4
+#define HIJACKED 5
+
 /obj/machinery/computer/emergency_shuttle
 	name = "emergency shuttle console"
 	desc = "For shuttle control."
 	icon_screen = "shuttle"
 	icon_keyboard = "tech_key"
+	obj_flags = INDESTRUCTIBLE
```
silicons: ANTI POWERGAMER DEVICE DEPLOYED

code/modules/shuttle/emergency.dm
```diff
@@ -133,6 +158,71 @@
...
+
+/obj/machinery/computer/emergency_shuttle/proc/announce_hijack_stage()
+	var/msg
+	switch(SSshuttle.emergency.hijack_status)
+		if(NOT_BEGUN)
+			return
+		if(STAGE_1)
+			var/datum/species/S = new
+			msg = "AUTHENTICATING - FAIL. AUTHENTICATING - FAIL. AUTHENTICATING - FAI###### Welcome, technician [S.random_name(pick(MALE, FEMALE))]."
+			qdel(S)
```
Ghommie: Remove the species clutter.
```
			msg = "AUTHENTICATING - FAIL. AUTHENTICATING - FAIL. AUTHENTICATING - FAI###### Welcome, technician [prob(50) pick(GLOB.first_names_male) : pick(GLOB.first_names_female)] [pick(GLOB.last_names)]."
```

Ghommie (approved):

PR #10353, opened 2019-12-30
no longer can you wave at things you don't know are there

Seris02:
## About The Pull Request

makes a viewers call in waving to check if you can actually see the thing

## Why It's Good For The Game

you can no longer wave at walls and floors from behind things

## Changelog
:cl:
fix: waving at walls from behind a wall
/:cl:

PR #10667, opened 2020-01-23
no longer can you wave at unseen things, nor bystanders will be informed of what you are waving at if they can't see it themselves.

Ghommie:
## About The Pull Request
#10353 but improved. You may not know what exactly someone is waving his hand to if you can't see it yourself. Also adding a 0.4 secs cooldown, it's pretty awful when people click around repeatedly and spam it, and stopping turfs from being targettable to prevent the aforementioned cooldown from making targetting movables no more viable.

## Why It's Good For The Game
I don't know, do you want the feature removed? Yes, it's a little snowflake, but I prefer to avoid clunky overhead. This will close #10353. 

## Changelog
:cl:
fix: Something about empty hand combat mode right click waving hands defying common sense and being spammy.
/:cl:


silicons (approved):
i'm hurt but ok

PR #10675, opened 2020-01-23
adds in a 'no eorg' preference that teleports you to the ghost cafe in round end

Seris02:
## About The Pull Request

see title

## Why It's Good For The Game

Because people get salty over eorg and this is one way to stop it, at least for a while until I have to code in some edge case to stop them from eorging in the ghost cafe.

## Changelog
:cl:
add: Ability to be transported to the ghost cafe on roundend instead of eorg
/:cl:

### DIFF
--- a/code/__HELPERS/roundend.dm
+++ b/code/__HELPERS/roundend.dm
@@ -191,7 +191,8 @@
 			H.add_hud_to(M)
 
 	CHECK_TICK
-	for(var/mob/m in GLOB.mob_list)
+	for(var/M in GLOB.player_list)
+		var/mob/m = M
 		if (!m.client)
 			continue
 		m.client.verbs += /client/proc/tele_to_ghost_cafe
--- a/code/__HELPERS/roundend.dm
+++ b/code/__HELPERS/roundend.dm
@@ -192,18 +192,30 @@
 
 	CHECK_TICK
 	for(var/mob/m in GLOB.mob_list)
-		if (m.client)
-			m.client.verbs += /client/proc/tele_to_ghost_cafe
+		if (!m.client)
+			continue
+		m.client.verbs += /client/proc/tele_to_ghost_cafe
 		if (m.client && m.client.prefs && m.client.prefs.eorg_tele)
 			ADD_TRAIT(m,TRAIT_PACIFISM,ROUNDSTART_TRAIT)
-			if (ishuman(m))
-				var/mob/living/carbon/human/H = m
-				H.revive(full_heal=TRUE,admin_revive=TRUE)
-				m.status_flags |= GODMODE
-				for (var/obj/item/W in H)
-					if (W == H.w_uniform || W == H.shoes)
-						continue
-					H.dropItemToGround(W)
+			m.status_flags |= GODMODE
+			if (!ishuman(m))
+				continue
+			var/mob/living/carbon/human/H = m
+			H.revive(full_heal=TRUE,admin_revive=TRUE)
+			for (var/I in H)
+				var/obj/item/W = I
+				if (!istype(W))
+					continue
+				if (W == H.w_uniform || W == H.shoes)
+					continue
+				H.dropItemToGround(W)
+			if (istype(H.dna.species,/datum/species/plasmaman))
+				qdel(H.w_uniform)
+				H.equip_to_slot_or_del(new /obj/item/clothing/under/plasmaman(H),SLOT_W_UNIFORM)
+				H.equip_to_slot_or_del(new /obj/item/clothing/head/helmet/space/plasmaman(H),SLOT_HEAD)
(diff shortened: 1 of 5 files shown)


code/__HELPERS/roundend.dm
```diff
@@ -191,7 +191,20 @@
 			H.add_hud_to(M)
 
 	CHECK_TICK
-
+	for(var/mob/m in GLOB.mob_list)
```
Ghommie: One, all atoms in mob_list should be mobs, so the typecheck is not necessary versus a new line `var/mob/M = m`.
Two, it'll tele people left stranded on the station to the ghost cafe too.
Third, I think it should be better if you used the player_list.

code/__HELPERS/roundend.dm
```diff
@@ -191,7 +191,20 @@
...
+		if (m.client)
```
Ghommie: Early return if they don't have a client.

code/__HELPERS/roundend.dm
```diff
@@ -191,7 +191,20 @@
...
+			m.client.verbs += /client/proc/tele_to_ghost_cafe
+		if (m.client && m.client.prefs && m.client.prefs.eorg_tele)
+			ADD_TRAIT(m,TRAIT_PACIFISM,ROUNDSTART_TRAIT)
+			if (ishuman(m))
+				var/mob/living/carbon/human/H = m
+				H.revive(full_heal=TRUE,admin_revive=TRUE)
+				m.status_flags |= GODMODE
+				for (var/obj/item/W in H)
```
Ghommie:
```
				for (var/i in H.get_equipped_items(TRUE))
					var/obj/item/I = i
```

code/__HELPERS/roundend.dm
```diff
@@ -191,7 +191,20 @@
 			H.add_hud_to(M)
 
 	CHECK_TICK
-
+	for(var/mob/m in GLOB.mob_list)
+		if (m.client)
+			m.client.verbs += /client/proc/tele_to_ghost_cafe
+		if (m.client && m.client.prefs && m.client.prefs.eorg_tele)
+			ADD_TRAIT(m,TRAIT_PACIFISM,ROUNDSTART_TRAIT)
+			if (ishuman(m))
```
Ghommie: the human check is only necessary for the uniform and shoes. bring it down there.

code/__HELPERS/roundend.dm
```diff
@@ -191,7 +191,20 @@
...
+				var/mob/living/carbon/human/H = m
+				H.revive(full_heal=TRUE,admin_revive=TRUE)
+				m.status_flags |= GODMODE
+				for (var/obj/item/W in H)
+					if (W == H.w_uniform || W == H.shoes)
+						continue
+					H.dropItemToGround(W)
```
Ghommie: Oh yea, btw this will light plasmamemes on fire.

Ghommie (changes requested):
I have a certain perplexity on this feature.
Under normal conditions, the game will restart in 1 minute anyway, so roleplay is irrelevant.
On delayed restarts, people may assume the cafe as a no EORG zone and ahelp the grievers while the admins are busy dealing with other tickets.

code/__HELPERS/roundend.dm
Putnam3145: For reference: the second line is faster than the typecheck. You should prefer it when it's possible to do.

code/__HELPERS/roundend.dm
Putnam3145: Also, this shouldn't be ROUNDSTART_TRAIT added, though it's kind of hilariously irrelevant

code/__HELPERS/roundend.dm
Seris02: why though

code/__HELPERS/roundend.dm
Seris02: and the revival but ok

code/__HELPERS/roundend.dm
Seris02: true.

code/__HELPERS/roundend.dm
Ghommie: cavity implants, and any other item that's not equipped but found inside the mob anyway.

Seris02: it works for the ninja net though but alright.

silicons:
this is just hugbox

Seris02:
basically yeah

code/__HELPERS/roundend.dm
Seris02: I want it to tele people on the station with this on, and it does it after it collects the roundend report, so it shouldn't affect anything.

Ghommie:
This is irrelevant, you can still use the ghost cafe in case of delayed round ends.

PR #10688, opened 2020-01-24
Fixes SOME of the problems with hilbert hotel

silicons:
## About The Pull Request

hilbert hotel now checks canreach while checking in
hilbert hotel now has safeties to prevent checking in while already checked in which covers the fact that the reservation load is blocking and sleeps which makes it so you can check in twice before it's done loading
hilbert hotel user container check should be slighter faster and is far more thorough, so no more abusing it for teleport

## Why It's Good For The Game

please don't abuse this !fun! and !interesting! and !useful! item all the time to saturate transit space and use it as a counterless hand teleporter thank you very much for your cooperation

## Changelog
:cl:
fix: hilbert hotel is now less of an exploity mess
/:cl:


Ghommie:
Travis is screaming at you to stop making web edits like this.

Ghommie (approved):

Ghommie (approved):

PR #10694, opened 2020-01-24
[s] more reasonable reagent logging

silicons:
might be a bad idea because of so many usr-less transfers, ideas are appreciated


silicons:
nvm horrible idea

silicons:
nvm not a horrible idea if i don't do it horribly

Ghommie (approved):

PR #10698, opened 2020-01-24
Made ghost cafe mobs dust upon leaving, allowed ghost cafe bans

Putnam3145:
## About The Pull Request

See title. 

## Why It's Good For The Game

Less exploits are always good. This limits exploit potential massively.

## Changelog
:cl:
tweak: Ghost cafe mobs are super duper attached to the ghost cafe.
/:cl:

### DIFF
--- a/code/datums/elements/dusts_on_leaving_area.dm
+++ b/code/datums/elements/dusts_on_leaving_area.dm
@@ -1,6 +1,6 @@
 /datum/element/dusts_on_leaving_area
 	element_flags = ELEMENT_DETACH | ELEMENT_BESPOKE
-	id_arg_index = 1
+	id_arg_index = 2
 	var/list/attached_mobs = list()
 	var/list/area_types = list()
 
--- a/code/datums/elements/dusts_on_leaving_area.dm
+++ b/code/datums/elements/dusts_on_leaving_area.dm
@@ -1,0 +1,28 @@
+/datum/element/dusts_on_leaving_area
+	element_flags = ELEMENT_DETACH | ELEMENT_BESPOKE
+	id_arg_index = 1
+	var/list/attached_mobs = list()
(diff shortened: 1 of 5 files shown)


Ghommie (approved):

Ghommie:
This makes the ghost cafe's hilbertshotel quite useless though.

code/datums/elements/dusts_on_leaving_area.dm
```diff
@@ -0,0 +1,28 @@
...
+	if(!attached_mobs.len)
+		STOP_PROCESSING(SSprocessing,src)
+
+/datum/element/dusts_on_leaving_area/process()
+	for(var/m in attached_mobs)
+		var/mob/M = m
+		var/area/A = get_area(M)
+		if(!(A.type in area_types))
+			M.dust(force = TRUE)
+			Detach(M)
```
silicons, line 28: shouldn't this go before dust as i'm assuming elements auto detach on qdel?

also is it really worth elementing everything when you can just hook ghost cafe area to dust people with the ghost cafe roll or
feels really expensive for a niche  thing =|

Ghommie: While, it should indeed detach automatically on Destoy(), keep in mind that elements are singletons assignable to multiple datums, and are only created if there is no other existing element that meets the criterias (mostly just the type, in certain cases arguments too). TL;DR lightweight components.

Putnam3145: Yeah, this is a "just-in-case", and it won't cause any runtimes or issues AFAIK.

Also, I explicitly wrote this to be non-snowflakey for e.g. weird magical ghost roles in the future.

Also, it's not that expensive.

Also, that wouldn't let ghost cafe mobs enter the hilbert hotel.

PR #10707, opened 2020-01-25
Raises damage precision to two decimal places

silicons:
## About The Pull Request

Title

## Why It's Good For The Game

Nanite regeneration breaking due to low damage amount alerted me to this but I don't think it'll hurt us to have higher precision here. can't be that expensive right??

## Changelog
:cl:
experimental: Damage precision is now two decimal places instead of one.
/:cl:


Arturlang:
There is no way this can end horribly.

Ghommie (approved):
You might want to round up the reports from health analyzers & co later then.

PR #10712, opened 2020-01-25
Shoving someone who is on the ground now always disarms their active held item. Punch stamina cost reduced. 

silicons:
## About The Pull Request

Also makes shove-up cost constant rather than stambuffer

## Why It's Good For The Game

Made shove-up cost constant rather than staminabuffer as it's probably better that way if someone's staminabuffer is, say, edited for whatever reason, and without this the new cost would be impredictable. 

Anyways for the actual balance changes:
Shoving's horribly underpowered on Citadel, this makes shoving someone down more of a punishment/consequence than the current disarm which only applies if they have a gun and are standing up. And I guess you can argue that being down slows them but honestly that isn't too relevant for what you'd usually be using shoves for.
Also punch, with its average of 5 damage, is just outright terrible at a 1:1 average damage to stamina cost. Rage cages are nearly untenable and stamcritting from trying to fistfight one person is a load of why. This won't make it vastly more superior or anything, just makes it slightly better in a fight.

## Changelog
:cl:
balance: Shoving now always disarms if someone is down. Punching now takes 3.5 stam instead of 5 to do.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1475,10 +1475,6 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
 		target.grabbedby(user)
 		return 1
 
-
-
-
-
 /datum/species/proc/harm(mob/living/carbon/human/user, mob/living/carbon/human/target, datum/martial_art/attacker_style)
 	if(!attacker_style && HAS_TRAIT(user, TRAIT_PACIFISM))
 		to_chat(user, "<span class='warning'>You don't want to harm [target]!</span>")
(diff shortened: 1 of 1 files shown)


Tupinambis:
Punch changes good, but I feel like a guaranteed disarm on downed opponents is a terrible idea with how jenky combat mode can be, and more importantly with how antags tend to have less opportunities to put people on the ground when compared to security. I am already making a PR to remove stunbaton disarm for a reason, this just nullifies that decision.

I'd rather shoves just do more stamina damage, and act as essentially a nonlethal version of a punch and be balanced as such.

PR #10714, opened 2020-01-25
is_ctf_target now checks for innate faction var rather than clothing

silicons:
fixes mushrooms being near-invulnerable


Ghommie (approved):

code/modules/awaymissions/capture_the_flag.dm
```diff
@@ -440,7 +437,7 @@
 		. = TRUE
 	if(ishuman(target))
 		var/mob/living/carbon/human/H = target
```
Ghommie: Change this to a mob typecheck since it now checks the faction and not the worn suit.

Ghommie (changes requested):

Ghommie (approved):

PR #10727, opened 2020-01-26
Makes meteors require higher station population to roll more destructive waves.

chefferz:
Does as it says on the tin

Also shifts the weight up.
Also limits 1 of each meteor wave type.
Also pushes the time for them to start out a bit more so engineering actually has time to make shields.

### DIFF
--- a/code/modules/events/meteor_wave.dm
+++ b/code/modules/events/meteor_wave.dm
@@ -6,7 +6,7 @@
 /datum/round_event_control/meteor_wave
 	name = "Meteor Wave: Normal"
 	typepath = /datum/round_event/meteor_wave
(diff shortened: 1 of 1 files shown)


silicons:
A bump in numbers was needed but these are way too high and I don't see why you need 1 max occurence for the weaker waves.

chefferz:
> 
> 
> A bump in numbers was needed but these are way too high and I don't see why you need 1 max occurence for the weaker waves.

What do you propose then.

GrayRachnid:
😂

Ghommie:
Kevinz is already handling meteor waves in a more acceptable way. Closing this without second thought.

PR #10728, opened 2020-01-26
Meteor waves are now directional (and tell you the direction on the ""1 minute""+ warning)

silicons:
## About The Pull Request

title

## Why It's Good For The Game

i'm stubborn about the meteor timer and its purpose
now you can have a better time shielding if you are very limited on time, and this will also make it easier to have a vague idea of where meteors are going to come from

## Changelog
:cl:
balance: meteor waves are now directional and announces the direction on the command report.
/:cl:


Iroquois-Pliskin:
I proposed this in the original time change PR and I'm happy to see this being done, even if it's later on.

KathrinBailey:
This is much nicer and a better solution than expecting engineers to drain the ore silo every single round as a requirement, which isn't enough half the time because expecting gamer miners to do something like getting glass is a bit too extreme.

code/modules/events/meteor_wave.dm
```diff
@@ -59,11 +61,21 @@
...
+	var/directionstring
+	switch(direction)
+		if(NORTH)
+			directionstring = " from the north."
+		if(SOUTH)
+			directionstring = " from the south."
+		if(EAST)
+			directionstring = " from the east."
+		if(WEST)
+			directionstring = " from the west."
```
Ghommie: Nautical terms (such as fore, aft, startboard, port) please, it is pretty wrong to use cardinal directions on automated announcements and the such when there is no north star to orientate around in the game. (Yes, we'll still gergally use cardinals, but think of this as an incentive to know these basic terms)

Also should replace "from" with "toward" since the terms refer to sides of the ship and not the direction.

Ghommie (changes requested):

code/modules/events/meteor_wave.dm
```diff
@@ -64,13 +64,13 @@
...
+			directionstring = " towards the fore"
 		if(SOUTH)
-			directionstring = " from the south."
+			directionstring = " towards the aft"
 		if(EAST)
-			directionstring = " from the east."
+			directionstring = " towards starboard"
 		if(WEST)
-			directionstring = " from the west."
+			directionstring = " towards port"
```
YakumoChen, line 73: High level commit, excellent change

Ghommie (approved):

PR #10729, opened 2020-01-26
The meteor wave report time to impact is now accurate (tip you were getting 2-3 minute warnings all along)

silicons:
title because HAHA PLEASE STOP THINKING STARTWHEN IS SECONDS WHY DOES EVERY CODER DO-
oh because events is a bad system.
yeah.


Putnam3145 (approved):
what fun.

PR #10739, opened 2020-01-26
Add a Dwarf Language

r4d6:
## About The Pull Request
This add a dwarf language to the game. For now only dwarves and curators are able to use though.

## Why It's Good For The Game
More languages is good, and Dwarf Fortress has thousands of words to use that I took for this language.

Also if we plan to add dwarves as a roundstart race instead of an Halloween-only thing, I think they should have their own language.

## Changelog
:cl:
add: Added a dwarf language
/:cl:

Thanks to Autism for dealing with the special characters.


Ghommie (commented):
Neat, Extra points if you make it so speaking in their own language doesn't apply their super-heavy-scottish dialect world replacement.

Probably easiest way to do it would be uncommenting this line:
https://github.com/Citadel-Station-13/Citadel-Station-13/blob/9cbe0b5bc19c7d3e42fef18ab46d1766ba0b64ef/code/__DEFINES/components.dm#L189
And then add this at the start of the species' handle_speech()
```
	if(ispath(speech_args[SPEECH_LANGUAGE], /datum/language/dwarf)
		return
```

Ghommie (changes requested):
Oh yea, forgot to say: There are quite a few syllabes containing non-alphabetical characters such as %, £, < and ^, I assume that might have something to do with certain ASCII characters such as accented vowels not being recognized by the program or your locale. Is that intended?

ItsTisme:
> 
> 
> Oh yea, forgot to say: There are quite a few syllabes containing non-alphabetical characters such as %, £, < and ^, I assume that might have something to do with certain ASCII characters such as accented vowels not being recognized by the program or your locale. Is that intended?

Yeah bassicly, I tryed to replace them with there closest look alike (‰ = %) but it still looks very odd, cant do much about that though

r4d6:
> Neat, Extra points if you make it so speaking in their own language doesn't apply their super-heavy-scottish dialect world replacement.

I mean some people asked for them keeping their accent. so I am not sure about that, at least for now.

Ghommie (approved):

PR #10741, opened 2020-01-26
The much needed stunbaton nerf.

Tupinambis:
## About The Pull Request

Who would win? A heavily armed series of antags with various amounts of stun resistance and rapid stamina regen, or one zappy stick.

It's always the zappy stick, this nerfs stunbaton stamina damage from 49 (3 hits to hardcrit) to 38.5 (4 hits to hard crit) with the stunprod taking an extra hit to stamcrit. Additionally, stunbatons no longer disarm.

Zappy stick now does limb based damage and disables limbs in two hits.

## Why It's Good For The Game

Getting hit a single time with a stunbaton should not consistently game over people. As of now if you get stunbatonned once you have pretty much no chance to fight back. 

The removal of the disarm should make energy swords ( and melee weapons in general) significantly more useful as an antag than they currently are, and the stamina damage nerf should hopefully give people on the recieving end of a stunbaton more time to react and fight back.

Limb based damage adds strategy and such.

## Changelog
:cl:
balance: stunbatons now take 5 hits to inflict hard crit, up from 3
balance: stunbatons no longer disarm targets
/:cl:


kiwedespars:
too hard of a nerf.

if stunbatons could do stam damage to a specific limb, disabling limbs in 2 hits while needing 5 to hard stamcrit, that'd be epic.

edit:
regardless i think it's a good change to an extent because this means we can also nerf other instant wins, like blood cult stun and the sort. isn't it funny how the only way you can beat an instant win cult stun is through using a baton that wins in 3 hits, 2 if they were running and already lack stamina?

ghost:
haha yes

ShadeAware:
**GOOD** 👏**CHANGE** 👏

Linzolle:
Interesting change, will have to see how it works in practice though.

silicons:
I don't like overreliance on limb damage in general but I'm kind of biased as someone who doesn't have a numpad so obviously I don't want things that heavily benefit and give massive advantages to those with numpads.

Regardless, this will need.. a lot of testing. An alternative if this doesn't work would be disarm but no knockdown, but yeah, we'll see how this goes.

However I'd like for this to be handled after we resolve the taser question as I'd rather not shake up two things at once (without them being related in the same PR/meant to be applied at once).

Tupinambis:
I updated this to do 4 hits instead of 5. With natural stunbaton power drain and the fact I already removed disarm, 5 was probably excessive.

Tupinambis:
Oh fuck I broke it

PR #10744, opened 2020-01-27
Moar engines!

r4d6:
## About The Pull Request
This add 4 new engines to Boxstation.
1st, the Thermo-Electric Generator, recently un-nerfed, courtesy of MrJWhit. It has the same weight as the other default engines
2nd, the Antimatter Engine, a little boring, but has the lowest weight for it to happen
3rd, the Budget Engine, due to a budget cut by NT, the Engine of the station has been replaced with a serie of PACMANS and SMES, Thanks to Feasel for giving the idea
4th and final engine, the Empty Engine! NT has decided to be even more lazy and left a big empty space where the engine is normally supposed to be. Has a weight of 0. Probably would make it load it if the config is somehow empty at a latter date.

They will still require an Headmin to update the configs.

## Why It's Good For The Game
More diversity is good, it will also reduce the chances of getting a singulo engine if it is in the rotation.
Also keep in mind that, while a few of them could be considered meme engines, that the headmins can always remove them from the rotation.

## Changelog
:cl:
add: Added more engines
/:cl:

I would like for a test-merge first if possible though, I could have messed something up.


Putnam3145:
`Probably would make it load it if the config is somehow empty at a latter date.`
i'd prefer the supermatter be the default tbh but that's not really here

r4d6:
It would just require a check when it pick a map from the config, then an overwrite if the config is empty, not really a problem or a priority right now.

config/game_options.txt
```diff
@@ -591,6 +591,10 @@ BOX_RANDOM_ENGINE Box Singulo,3
 BOX_RANDOM_ENGINE Box SM 1x3,1
 BOX_RANDOM_ENGINE Box SM 5x5,1
 BOX_RANDOM_ENGINE Box SM 3x,0
+BOX_RANDOM_ENGINE Box TEG,3
+BOX_RANDOM_ENGINE Box Empty,0
+BOX_RANDOM_ENGINE Box Antimatter,1
```
Ghommie, line 596: The antimatter engine is old, not sure if functional, and tgstation removed it already.

config/game_options.txt
```diff
@@ -591,6 +591,10 @@ BOX_RANDOM_ENGINE Box Singulo,3
...
+BOX_RANDOM_ENGINE Box P.A.C.M.A.N,1
```
Ghommie, line 597: I'm not sure this is a great idea.

Ghommie (approved):
All good but perhaps you should move on and make some other maps' engine randomizable too.

config/game_options.txt
r4d6: I know it work, We've used it multiple times on the server.

config/game_options.txt
r4d6: A Headmin can always change those values if it is needed.

r4d6:
I should, but I fear the spriting.
Pubby look easy enough
Delta don't really need it
Meta is going to be hell, since you have to include the space loop, with it the maint hallway, without taking the security post, and also take the freezers without taking the Protolathe as well. So much pixel measurement is going to be needed for that one.

PR #10747, opened 2020-01-27
shoving someone into a wall on the second shove now also disarms their weapon

silicons:
## About The Pull Request

title

## Why It's Good For The Game

bit silly to make shoving into a wall be more useless than a normal shove if they have a weapon

## Changelog
:cl:
balance: shoves buffed, now shoving into a wall twice rapidly will also disarm their weapon.
/:cl:


Ghommie (changes requested):
Just noticed there is a lack of NODROP checks for chat messages and combat logs here. Could you please add some since you are on that code already?

Ghommie (approved):

PR #10755, opened 2020-01-27
small refactor to melee_attack_chain

silicons:
## About The Pull Request

Updates melee_attack_chain to tg's varient and makes pre_attackby return values match the pattern of everything else

## Why It's Good For The Game

Less cluttered code, tool_attack_chain is useless, etc etc

## Changelog

No real in game changes


Ghommie (approved):
All nice but there is that ctf claymore you added whose pre_attack still hasn't its boolean return values switched.

PR #10777, opened 2020-01-29
Gangs can now only tag with GANG spraycans, rather than lathed spraycans

silicons:
## About The Pull Request

Title

## Why It's Good For The Game

I think it's better for gangs as a gamemode to have more of am emphasis on leaders (we also get an extra LT from /tg/ implementation back in the day btw) to have to pass out gear. If you want your influence points, hand out spraycans. The meta should not be to pen someone and leave them to lathe spraycans at public lathe and then go blasting with a sniper rifle.

## Changelog
:cl:
balance: Gangs can now only tag with a gang uplink bought spraycan.
/:cl:

### DIFF
--- a/code/game/objects/items/crayons.dm
+++ b/code/game/objects/items/crayons.dm
@@ -619,8 +619,7 @@
 	is_capped = TRUE
 	self_contained = FALSE // Don't disappear when they're empty
 	can_change_colour = TRUE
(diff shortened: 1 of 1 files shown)


Detective-Google:
<--- this player doesn't even know cans came from places other than the lathe.

Ghommie (approved):

lolman360:
^ this player is low iq

PR #10788, opened 2020-01-30
crushing magboots for tators

Seris02:
## About The Pull Request

Crushing magboots for traitors, only engineers + atmos techs, crushes anyone you walk over while the magpulse is enabled.

## Why It's Good For The Game

more traitor items + fun times.

## Changelog
:cl:
add: crushing magboots
/:cl:

### DIFF
--- a/code/modules/clothing/shoes/magboots.dm
+++ b/code/modules/clothing/shoes/magboots.dm
@@ -80,7 +80,7 @@
 
 /obj/item/clothing/shoes/magboots/crushing/equipped(mob/user,slot)
 	. = ..()
-	if (slot == SLOT_SHOES)
+	if (slot == SLOT_SHOES && magpulse)
 		RegisterSignal(user, COMSIG_MOVABLE_MOVED,.proc/crush)
 
 /obj/item/clothing/shoes/magboots/crushing/dropped(mob/user)
--- a/code/modules/clothing/shoes/magboots.dm
+++ b/code/modules/clothing/shoes/magboots.dm
@@ -71,12 +71,17 @@
 			to_chat(A,"<span class='userdanger'>[user]'s magboots press down on you, crushing you!</span>")
 			A.emote("scream")
 
+/obj/item/clothing/shoes/magboots/crushing/attack_self(mob/user)
+	. = ..()
+	if (magpulse)
+		RegisterSignal(user, COMSIG_MOVABLE_MOVED,.proc/crush)
+	else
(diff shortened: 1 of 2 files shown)


necromanceranne:
With a flamethrower full of water vapour this might be fun.

code/modules/clothing/shoes/magboots.dm
```diff
@@ -57,3 +57,27 @@
...
+	var/turf/T = user.loc
+	for (var/mob/living/A in T)
+		if (A != user && A.lying)
+			A.adjustBruteLoss(rand(10,13))
+			to_chat(A,"<span class='userdanger'>[user]'s magboots press down on you, crushing you!</span>")
+			A.emote("scream")
+
+/obj/item/clothing/shoes/magboots/crushing/equipped(mob/user,slot)
+	. = ..()
+	if (slot == SLOT_SHOES)
```
Ghommie: For efficiency, I think these signals should be registered/unregistered on attack_self() (magpulse activation/disactivation), instead of early returning crush() if magpulse is not TRUE.

So make sure to add a magpulse check here once that's is done.

Ghommie (changes requested):
See above.

code/modules/clothing/shoes/magboots.dm
Seris02: true

Seris02: actually no, because then if they turn it on and then unequip it it'll still call it, if I check if it's actually on them in crush() and unregister if not, it'll not work if they put it back on while it's still activated, so they'll have to turn it back on, and it'll only work for that specific person.

code/modules/clothing/shoes/magboots.dm
```diff
@@ -71,13 +71,18 @@
...
 	. = ..()
-	if (slot == SLOT_SHOES)
+	if (magpulse)
 		RegisterSignal(user, COMSIG_MOVABLE_MOVED,.proc/crush)
 	else
 		UnregisterSignal(user,COMSIG_MOVABLE_MOVED)
 
+/obj/item/clothing/shoes/magboots/crushing/equipped(mob/user,slot)
+	. = ..()
+	if (slot == SLOT_SHOES)
```
Ghommie:
```suggestion
	if (slot == SLOT_SHOES && magpulse)
```

Ghommie (approved):

PR #10791, opened 2020-01-30
[READY] Further tweaks to bloodsuckers

Arturlang:
Bloodsucker levels up now take blood to do, and they dont get a free level up at the start, while also every night is a bit longer to smooth out the leveling late game. Vassals now also require a lot more blood and the blood cost scales with the amount of thralls. Hopefully. Needs testing.
## Changelog
:cl:
add: Each night for bloodsuckers will last longer due to the sun dimming.
tweak: You cant level up for free, now you have to pay blood for power.
balance: Bloodsuckers are no longer as resistant to hard stuns and regenerate less stamina. and a lot of their cooldowns are a lot higher.
/:cl:
Theres a large amount of tweaks so i might have missed something, but please testmerge this so we can see how it turns out.

### DIFF
--- a/code/datums/mood_events/generic_negative_events.dm
+++ b/code/datums/mood_events/generic_negative_events.dm
@@ -201,51 +201,52 @@
 
 /datum/mood_event/vampcandle
 	description = "<span class='umbra'>Something is making your mind feel... loose...</span>\n"
-	mood_change = -10
+	mood_change = -15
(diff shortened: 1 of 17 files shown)


Arturlang:
Ready to be fullmerged, seems good so far, will do a rework of a lot of stuff later on maybe in atomized PRs

Ghommie (approved):

PR #10802, opened 2020-01-31
Well Deserved Pepperspray Buffs. Now you can actually use it!

Tupinambis:
## About The Pull Request

Pepperspray has been buffed by a large amount. It now ONLY checks for gas protection (which only gas masks, bio suits, rad suits?, and space suits have.)

As it stands, this weapon has a mix of both Flash and Stunbaton traits, and most importantly has a hard knockdown (It disarms!) which if #10741 gets merged, means pepperspray has a leg up, especially against conversion antags and groups of enemies. 

This should result in conversion antags just being a tad easier to deal with, as not everyone will have a mask.

*I swear to god if you can print gas masks at the lathe*

## Why It's Good For The Game

Pepperspray is quite literally useless currently, and I think it's an awfully pretty fun and unique option to just never be used. These change should hopefully result in pepperspray getting regular and standard use.

## Changelog
:cl: Tupinambis
balance: Pepperspray now only checks gas protection.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/food_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/food_reagents.dm
@@ -285,58 +285,24 @@
 	var/mob/living/carbon/victim = M
 	if(method == TOUCH || method == VAPOR)
 		//check for protection
-		var/mouth_covered = 0
-		var/eyes_covered = 0
-		var/obj/item/safe_thing = null
+		var/face_covered = 0
 
-		//monkeys and humans can have masks
+		//gas masks are one example of protection. standard issue to security and operatives.
 		if( victim.wear_mask )
-			if ( victim.wear_mask.flags_cover & MASKCOVERSEYES )
-				eyes_covered = 1
-				safe_thing = victim.wear_mask
-			if ( victim.wear_mask.flags_cover & MASKCOVERSMOUTH )
-				mouth_covered = 1
-				safe_thing = victim.wear_mask
-
-		//only humans can have helmets and glasses
-		if(ishuman(victim))
-			var/mob/living/carbon/human/H = victim
-			if( H.head )
-				if ( H.head.flags_cover & HEADCOVERSEYES )
-					eyes_covered = 1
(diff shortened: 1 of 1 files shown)


KrabSpider:
Shouldn't there be a minimum on the amount of capsaicin you need to induce the effects...?

I have a bit of concern about the possibility of people breaking into security checkpoints, filling up a spray bottle of *any* kind with pepperspray from the refill stations to have a 2-tile range stunbaton (pre-nerf) that ignores shields and projectile immunity without much concern for running out of charge or the like.

silicons:
if we remove hardstuns from a staple of melee for security why would we increase the power of a resprited and speedier spray bottle with how downright stupid reagent spraying code is?

ghost:
reagent spray code is funny, it INJECTS it into you instead of spraying it
why?

Tupinambis:
I undid the stamina damage buffs, probably dumb move to begin with

code/modules/reagents/chemistry/reagents/food_reagents.dm
```diff
@@ -285,49 +285,15 @@
...
-			victim.blur_eyes(3)
-			victim.blind_eyes(2)
-			victim.confused = max(M.confused, 3)
-			victim.damageoverlaytemp = 60
-			victim.Knockdown(80, override_hardstun = 0.1, override_stamdmg = min(reac_volume * 3, 15))
-			return
-		else if ( eyes_covered ) // Eye cover is better than mouth cover
-			victim.blur_eyes(3)
-			victim.damageoverlaytemp = 30
+		if (face_covered)
```
Ghommie, line 296: At this point, just replace lines 291-297 with this:
```
		if (victim.wear_mask && victim.wear_mask.clothing_flags & BLOCK_GAS_SMOKE_EFFECT)
				return
```
Also remove the `face_covered` var.

Ghommie (changes requested):

Ghommie:
Hello?

PR #10805, opened 2020-01-31
guns/ammo/projectiles now properly clean up while being deleted

silicons:
Title, they should be deleting contents properly using QDEL_NULL instead of relying on the default atom/object behaviors for wiping contents.


Putnam3145 (approved):

Putnam3145:
i think this is fine but i've had bad experiences with spooky things like modifying all projectiles

PR #10806, opened 2020-01-31
[Balance Patch - Tend Wounds Edition]

Feasel:
Seriously why the fuck is this doing such little healing and taking so much time to do?

Like, it's literally useless with our synthflesh+silver sulf... meta

Fuckin hell, even bicardine/kelotane heals faster than this.

Not anymore.

Today, I'm dramatically improving the healing and giving a decent time reduction to doing this surgery.

I haven't touched the "nerf to healing" on dead people... maybe should if it's not enough... but that'll be for later after community tries it out.

--------

PATCH NOTES

- Doubled healing on all Tend Wounds

- Roughly around 40% reduction to time needed to do the step that actually heals.

===

Marked it so maintainers can edit if I have fucked up any code and I didn't see it.

### DIFF
--- a/code/modules/surgery/healing.dm
+++ b/code/modules/surgery/healing.dm
@@ -25,7 +25,7 @@
 	name = "repair body"
 	implements = list(TOOL_HEMOSTAT = 100, TOOL_SCREWDRIVER = 65, /obj/item/pen = 55)
 	repeatable = TRUE
-	time = 25
+	time = 15
 	var/brutehealing = 0
 	var/burnhealing = 0
 	var/missinghpbonus = 0 //heals an extra point of damager per X missing damage of type (burn damage for burn healing, brute for brute). Smaller Number = More Healing!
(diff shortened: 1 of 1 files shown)


Linzolle:
I think the reason is this is meant to be performed on dead bodies to prepare for defibrillation, but yes it does take a long time.

Feasel:
@Linzolle Well... the original intent was that this is supposed to be a chem-less way to heal people.

Hence why the bonus heal is reduced on dead bodies... but the healing is so low that this thing is negligible at any point before this change.

It takes 50 seconds to do 100 healing before patch. 

50... seconds. Who is going to wait 50 seconds with our medical system?

It's also designed in the way as like a "revival"-lite surgery... but once again. 50 seconds for 100 healing. That's big issue with cloning being around and organ damage the way it is.

Ghommie (approved):

PR #10816, opened 2020-01-31
pH strip feedback fix/tweak

ghost:
## About The Pull Request

Fixes the feedback message for trying to use a used pH strip by changing it from "`user` has already been used" to "`src` has already been used". Also adds "used" to their name when used as additional feedback to make it more clear.

## Why It's Good For The Game

it's a bugfix

## Changelog
:cl:
fix: Fixes an error in pH strips' feedback message.
/:cl:

### DIFF
--- a/modular_citadel/code/modules/reagents/objects/items.dm
+++ b/modular_citadel/code/modules/reagents/objects/items.dm
@@ -107,7 +107,7 @@
 		if(-INFINITY to 1)
 			color = "#c6040c"
 	desc += " The paper looks to be around a pH of [round(cont.reagents.pH, 1)]"
-	name = "used " + name
+	name = "used [name]"
 	used = TRUE
 
 /obj/item/fermichem/pHmeter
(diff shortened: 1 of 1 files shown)


modular_citadel/code/modules/reagents/objects/items.dm
```diff
@@ -107,6 +107,7 @@
 		if(-INFINITY to 1)
 			color = "#c6040c"
 	desc += " The paper looks to be around a pH of [round(cont.reagents.pH, 1)]"
+	name = "used " + name
```
Ghommie:
```suggestion
	name = "used [name]"
```
Little peeve, generally safer to use, even though names should always be text.

Ghommie (changes requested):

modular_citadel/code/modules/reagents/objects/items.dm
ghost: gotcha, fixed

Ghommie (approved):

PR #10823, opened 2020-02-01
nodrop electric gloves for traitors

Seris02:
## About The Pull Request

nodrop gloves that stop you from picking anything up/dropping anything while toggled on
costs 6 TC (basically infinite glue but can be toggled on/off which is why it's so high but is also very obvious with a name of 'electric gloves)

## Why It's Good For The Game

more traitor items

## Changelog
:cl:
add: nodrop traitor gloves
/:cl:


ghost:
maybe make them vulnerable to emp? perhaps dropping whatever you're holding and disabling nodrop or something

Ghommie (changes requested):
The feature is janky all around. There are lot of edge cases that can remove either the gloves or the held objects from the mob and cause the no drop trait to stick around. You'll most likely have to use component signals to overcome these issues.

Ghommie:
We are kinda beating a dead horse here. Adding more and more uplink items is not going to add quality. I'm considering to keep the wall running boots around but you'll have to reework them a little as I said.

PR #10830, opened 2020-02-01
[s] fixes a glitch with the PKA

Seris02:
## About The Pull Request

stops people from installing the same modkit over and over I can recreate the glitch but not gonna tell how obviously

## Why It's Good For The Game

fixes

## Changelog
:cl:
fix: glitch with PKA
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/kinetic_accelerator.dm
+++ b/code/modules/projectiles/guns/energy/kinetic_accelerator.dm
@@ -86,10 +86,9 @@
 		M.uninstall(src, FALSE)
 
 /obj/item/gun/energy/kinetic_accelerator/attackby(obj/item/I, mob/user)
-	if(istype(I, /obj/item/borg/upgrade/modkit) && yeetdelay < world.time)
+	if(istype(I, /obj/item/borg/upgrade/modkit))
 		var/obj/item/borg/upgrade/modkit/MK = I
 		MK.install(src, user)
-		yeetdelay = world.time + 20
 	else
 		..()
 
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/energy/kinetic_accelerator.dm
```diff
@@ -24,6 +24,7 @@
 	var/list/modkits = list()
 
 	var/recharge_timerid
+	var/yeetdelay = 0 //exploit fix
```
Ghommie: this should be a modkit variable.

code/modules/projectiles/guns/energy/kinetic_accelerator.dm
```diff
@@ -296,7 +297,7 @@
 
 /obj/item/borg/upgrade/modkit/proc/install(obj/item/gun/energy/kinetic_accelerator/KA, mob/user)
 	. = TRUE
-	if(src in KA.modkits) // Sanity check to prevent installing the same modkit twice thanks to occasional click/lag delays.
+	if(src in KA.modkits && KA.yeetdelay < world.time) // Sanity check to prevent installing the same modkit twice thanks to occasional click/lag delays.
```
Ghommie:
```suggestion
	if((src in KA.modkits) || KA.yeetdelay >= world.time) // Sanity check to prevent installing the same modkit twice thanks to occasional click/lag delays.
```

Ghommie (changes requested):
I'm pretty unsure how this fixes anything, there is already a check against this sort of issue... eeeh, whatever

Seris02:
found the real culprit

Ghommie (approved):

PR #10840, opened 2020-02-01
Save panic bunker between rounds

Putnam3145:
## About The Pull Request

See title. Makes it so admins don't have to be 100% on 100% of the time.

## Why It's Good For The Game

Better that admins don't have to be on at the same time as people trying to enter.

## Changelog
:cl:
admin: Panic bunker is now round-to-round persistent
/:cl:


code/datums/world_topic.dm
```diff
@@ -187,3 +187,6 @@
 
 	if(!key_valid)
 		GLOB.topic_status_cache = .
+
+/datum/world_topic/panic_bunker
```
deathride58: There's already the whitelist TGS command

deathride58 (dismissed):
Make sure you keep track of round IDs and autocull old entries. Be sure to add config entries so that headmins can configure this

code/datums/world_topic.dm
Putnam3145: wow i did NOT know i'd saved that, whoops

code/controllers/subsystem/persistence.dm
```diff
@@ -259,6 +260,16 @@ SUBSYSTEM_DEF(persistence)
...
 	SaveRandomizedRecipes()
+	SavePanicBunker()
+
+/datum/controller/subsystem/persistence/proc/LoadPanicBunker()
+	var/bunker_path = file("data/bunker_passthrough.json")
+	if(fexists(bunker_path))
+		GLOB.bunker_passthrough = json_decode(file2text(bunker_path))
+		var/round_id = GLOB.round_id
+		for(var/ckey in GLOB.bunker_passthrough)
+			if(round_id - GLOB.bunker_passthrough[ckey] > 100)
```
Ghommie: The panic bunker duration should be in real time for consistency and being easier to manage. And perhaps a configuration instead of a hardcoded number.

Ghommie (changes requested):

code/controllers/subsystem/persistence.dm
Putnam3145: done

Ghommie (approved):

code/controllers/subsystem/persistence.dm
```diff
@@ -259,6 +260,15 @@ SUBSYSTEM_DEF(persistence)
...
 		CollectAntagReputation()
 	SaveRandomizedRecipes()
+	SavePanicBunker()
+
+/datum/controller/subsystem/persistence/proc/LoadPanicBunker()
+	var/bunker_path = file("data/bunker_passthrough.json")
+	if(fexists(bunker_path))
+		GLOB.bunker_passthrough = json_decode(file2text(bunker_path))
+		for(var/ckey in GLOB.bunker_passthrough)
+			if(daysSince(GLOB.bunker_passthrough[ckey]) >= CONFIG_GET(number/max_bunker_days))
```
Ghommie, line 270: Uh, actually, you should use a variable with CONFIG_GET(number/max_bunker_days) as value instead of running the CONFIG_GET every loop.

Putnam3145: this is run exactly once, when the subsystem initializes

PR #10904, opened 2020-02-05
granting golems GPSes that are off by default

Hatterhat:
## About The Pull Request
title
## Why It's Good For The Game
allows free golems to find their way, makes it easier for them (or miners) to find their way back and/or to them. they all start off so you don't have grugshits rushing their lathes
## Changelog
:cl:
add: The ships often crashed by Free Golems on Lavaland now have GPSes. They're off, by default, but an awakening Golem could easily turn one on.
/:cl:


necromanceranne:
The reason this was axed was because they had a protolathe that printed guns. So you could game it and get free guns. Like, for instance, those antags with their emags and test firing pins.

silicons:
do i really want unrestricted lathes easier to find?

silicons (approved):

PR #10913, opened 2020-02-05
[Balance Pass - Dissection Early Game Buff]

Feasel:
Doing this so science has some real ways to do science point generation.

Normal bodies are low enough in value that by the time you do a body normally, passive gen has already beaten too far.
So I just guaranteed success with better tools, but left a tiny bit off just for memes/ayylmao tools must be best.

Scapel - 90% success from 45% 

Advanced - 99% success from 65% success

Ayylmao - 100% Because ayylmao. 

Knife and shard adopt old-ish values because ghetto man should have a fun option (45% and 25%)

Now go gather dem bodies for SCIENCE!!!

## Changelog
:cl:
balance: Buffed dissection success chances.
/:cl:

### DIFF
--- a/code/modules/surgery/experimental_dissection.dm
+++ b/code/modules/surgery/experimental_dissection.dm
@@ -24,8 +24,8 @@
 
 /datum/surgery_step/dissection
 	name = "dissection"
-	implements = list(/obj/item/scalpel/augment = 75, /obj/item/scalpel/advanced = 60, /obj/item/scalpel = 45, /obj/item/kitchen/knife = 20, /obj/item/shard = 10)// special tools not only cut down time but also improve probability, doesn't use TOOL_SCALPEL because different scalpels have different probs
-	time = 125
+	implements = list(/obj/item/scalpel/augment = 100, /obj/item/scalpel/advanced = 99, /obj/item/scalpel = 90, /obj/item/kitchen/knife = 45, /obj/item/shard = 25)// special tools not only cut down time but also improve probability, doesn't use TOOL_SCALPEL because different scalpels have different probs
+	time = 100
 	silicons_obey_prob = TRUE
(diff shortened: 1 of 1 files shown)


code/modules/surgery/experimental_dissection.dm
```diff
@@ -24,8 +24,8 @@
 
 /datum/surgery_step/dissection
 	name = "dissection"
-	implements = list(/obj/item/scalpel/augment = 75, /obj/item/scalpel/advanced = 60, /obj/item/scalpel = 45, /obj/item/kitchen/knife = 20, /obj/item/shard = 10)// special tools not only cut down time but also improve probability, doesn't use TOOL_SCALPEL because different scalpels have different probs
-	time = 125
+	implements = list(/obj/item/scalpel/augment = 100, /obj/item/scalpel/advanced = 99, /obj/item/scalpel = 90, /obj/item/kitchen/knife = 45, /obj/item/shard = 25)// special tools not only cut down time but also improve probability, doesn't use TOOL_SCALPEL because different scalpels have different probs
```
Linzolle: /obj/item/scalpel/augment refers to the toolset implant, not the alien scalpel. /obj/item/scalpel/alien is what you are looking for.

Feasel: Ahh... so I guess alien is automatic 100?

Ehh. I'll allow augmented scalpel to be a thing... people should be encouraged to get implants and shit anyways.

Linzolle: currently, the alien scalpel will be 90 with how you have, because the game will assume that since the alien subtype wasn't defined to just use the normal scalpel's values

Ghommie (approved):

Ghommie (approved):

PR #10914, opened 2020-02-05
Stunbatons now start with a 15k cell instead of 10k (50% capacity increase)

silicons:
## About The Pull Request

title

## Why It's Good For The Game

Baton nerf effectively cut their stun capacity in half. I could propose decreasing stun cost, but there was already an existing issue with bluespace cells allowing you to basically use it.. forever. This gives it the necessary buff while not making bluespace cell even more prevalent than it was before.

## Changelog
:cl:
balance: Batons now start with a 15k rather than 10k cell.
/:cl:

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -48,7 +48,7 @@
 		baton_stun(hit_atom)
 
 /obj/item/melee/baton/loaded //this one starts with a cell pre-installed.
-	preload_cell_type = /obj/item/stock_parts/cell/high
+	preload_cell_type = /obj/item/stock_parts/cell/high/plus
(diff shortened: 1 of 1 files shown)


Tupinambis:
I think it's a bit early to jump to this balance change. I don't think it will be necessary. However, giving sec high-cap+ cells on spawn (to put in ONE) of their stunbatons is a good compromise, and will teach newer sec that you can in fact switch out stunbaton cells.

And then we will probably have to nerf bluespace powercells once every sec officer is using them every round but, ya know.

silicons:
I was under the impression batons did 25 stam instead of 39 (which.. is 3 hit stamcrit I guess)
Nvm

PR #10917, opened 2020-02-05
Removes the hypospray verb for toggling spray/inject mode and makes it into an Alt click proc.

Arturlang:
Its really hard to use especially in emergencies, and makes the hypospray feel cumbsy for changing from modes
## Changelog
:cl:
tweak: The MK2 hypospray now
/:cl:
Also adds a neat little examine text so you actually know its a feature.

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -423,18 +423,19 @@
 		else
 			unload_hypo(vial,user)
 
-/obj/item/hypospray/mkii/verb/modes()
-	set name = "Toggle Application Mode"
-	set category = "Object"
-	set src in usr
-	var/mob/M = usr
-	switch(mode)
-		if(HYPO_SPRAY)
-			mode = HYPO_INJECT
-			to_chat(M, "[src] is now set to inject contents on application.")
-		if(HYPO_INJECT)
-			mode = HYPO_SPRAY
-			to_chat(M, "[src] is now set to spray contents on application.")
+/obj/item/hypospray/mkii/AltClick(mob/living/user)
+	if(user.canUseTopic(src, BE_CLOSE, FALSE,))
+		switch(mode)
+			if(HYPO_SPRAY)
+				mode = HYPO_INJECT
+				to_chat(user, "[src] is now set to inject contents on application.")
+			if(HYPO_INJECT)
+				mode = HYPO_SPRAY
+				to_chat(user, "[src] is now set to spray contents on application.")
+
+/obj/item/hypospray/mkii/examine(mob/user)
+	. = ..()
+	. += "<span class='notice'>You can click on it while holding <b>Alt</b> to toggle its mode from spraying to injecting.</span>"
 
 #undef HYPO_SPRAY
 #undef HYPO_INJECT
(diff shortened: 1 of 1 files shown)


code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -423,18 +423,19 @@
...
-	var/mob/M = usr
-	switch(mode)
-		if(HYPO_SPRAY)
-			mode = HYPO_INJECT
-			to_chat(M, "[src] is now set to inject contents on application.")
-		if(HYPO_INJECT)
-			mode = HYPO_SPRAY
-			to_chat(M, "[src] is now set to spray contents on application.")
+/obj/item/hypospray/mkii/AltClick(mob/living/user)
+	if(user.canUseTopic(src, BE_CLOSE, FALSE,))
```
Ghommie: Parent call, please. Also remove the second arg and the last comma.

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -423,18 +423,19 @@
...
+		switch(mode)
+			if(HYPO_SPRAY)
+				mode = HYPO_INJECT
+				to_chat(user, "[src] is now set to inject contents on application.")
+			if(HYPO_INJECT)
+				mode = HYPO_SPRAY
+				to_chat(user, "[src] is now set to spray contents on application.")
```
Ghommie, line 435: Ensure the return value is TRUE if the user can use the src.

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -423,18 +423,19 @@
...
+
+/obj/item/hypospray/mkii/examine(mob/user)
+	. = ..()
+	. += "<span class='notice'>You can click on it while holding <b>Alt</b> to toggle its mode from spraying to injecting.</span>"
```
Ghommie:
```suggestion
	. += "<span class='notice'><b>Alt-Click</b> it to toggle its mode from spraying to injecting and viceversa.</span>"
```

Ghommie (changes requested):

code/modules/reagents/reagent_containers/hypospray.dm
Arturlang: ..() just after /obj/item/hypospray/mkii/AltClick(mob/living/user) ?

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -423,18 +423,21 @@
...
-	var/mob/M = usr
-	switch(mode)
-		if(HYPO_SPRAY)
-			mode = HYPO_INJECT
-			to_chat(M, "[src] is now set to inject contents on application.")
-		if(HYPO_INJECT)
-			mode = HYPO_SPRAY
-			to_chat(M, "[src] is now set to spray contents on application.")
+/obj/item/hypospray/mkii/AltClick(mob/living/user)
+	..()
```
Ghommie:
```suggestion
	. = ..()
```
I wasn't explicit on the "parent call" suggestion before. My bad.

Ghommie (approved):

PR #10993, opened 2020-02-10
adds the bow

Linzolle:
## About The Pull Request
It's a bow. Currently admin-spawn only, planning on making it craftable in the future depending on what it'll be used for. Does not include the hardlight bow, that might come in the future.

## Why It's Good For The Game
Either as a ranged weapon for Ashwalkers, or for something something planetstation.

## Changelog
None.


ghost:
poggers

code/modules/projectiles/guns/ballistic/bow.dm
```diff
@@ -0,0 +1,53 @@
...
+	return
+
+/obj/item/gun/ballistic/bow/chamber_round()
+	chambered = magazine.get_round(1)
+
+/obj/item/gun/ballistic/bow/afterattack()
+	. = ..()
+	if (chambered)
+		chambered = null
+		magazine.get_round(0)
```
silicons, line 29: is this supposed to be TRUE/FALSE or 0/1?

Linzolle: technically it could be FALSE since the 0 is for a variable called "keep" which refers to whether it keeps the round in the magazine but everywhere else its used is with 0/1. this could also technically be magazine.stored_ammo[1] = null I think.

Ghommie (changes requested):
You need to rework firing pin code more to stop pin-less ranged arms from having the pin spawned or accepting pins on attackby. Bonus points if you make them unviable for the pin removal crafting recipe. Otherwise all good even though the fact it's considered a gun by the code triggers me a little.

Ghommie (approved):

PR #11013, opened 2020-02-11
Add two more pickaxes, Titanium & Plasteel

r4d6:
## About The Pull Request
In the recent surge of the dwarves, two more type of pickaxes have been introduced to promote the old, traditional way of mining : The Plasteel and Titanium Pickaxe.

The Plasteel pickaxe break rock walls a bit slower than the titanium pickaxe, but is better at breaking skulls.
Meanwhile,
The Titanium Pickaxe break rock walls faster, but don't break open skulls as easy as the plasteel pickaxe.

In more gameplay term, the Plasteel Pickaxe has the speed of the silver pickaxe and the damage of the diamond, while the Titanium Pickaxe has the speed of the diamond and the damage of the silver.

## Why It's Good For The Game
Pickaxes are normally overrated and only for those who don't have a plasma cutter or PKA.
Hopefully those new pickaxes will be enough to tip the balance a little bit, to make them stand side by side with the Advanced Plasma Cutter, or at least a bit closer.

**_Strike The Earth !_**

## Changelog
:cl:
add: Added Plasteel Pickaxe
add: Added Titanium Pickaxe
/:cl:


Ghommie (approved):
Mixed feelings but we don't have the updated material datums yet.

PR #11015, opened 2020-02-11
Temporary(?) revert for "ENDS THE PLUSHIE DRAMA"

Ghommie:
Despite not touching or having any preprocessor function, every time it was attempted to testmerge #10277 there were some pretty shady undefined token errors with the watchdog's server compile. Trying to resolve this issue may take some good dose of trial and error.


Ghommie:
Ok, turns out the real issue was Kevinz giving Izzy a misnamed config dmi file, causing things to go bad at compile time thanks to missing rsc files.

PR #11016, opened 2020-02-11
Some sprite fixes. 

Microvolnovka19:
## About The Pull Request

So. Size 4 tentacle was broken, also 1-4 sizes for tapered,hemi and knotted hemi was the same. I fixed it.


Ghommie:
This should already being handled on #10987.

Ghommie:
EDIT: Misread the title.
Good change but you got to do it over again since I have cleaned up the file quite a bit.

PR #11024, opened 2020-02-11
Fixes brig cell timers, alternatively "WHO DECIDED TO USE NUMBER OF SECONDS FROM 2000 INSTEAD OF FROM THE CURRENT DAY"

silicons:
## About The Pull Request

title
realtime apparently doesn't update as fast as we think.

## Why It's Good For The Game

brig cells update too slowly and are inaccurate.

## Changelog
:cl:
fix: Brig cells now are more accurate by using timeofday instead of realtime
/:cl:

### DIFF
--- a/code/game/machinery/doors/brigdoors.dm
+++ b/code/game/machinery/doors/brigdoors.dm
@@ -67,7 +67,7 @@
 		return
 
 	if(timing)
-		if(world.timeofday - activation_time >= timer_duration)
+		if(REALTIMEOFDAY - activation_time >= timer_duration)
 			timer_end() // open doors, reset timer, clear status screen
 		update_icon()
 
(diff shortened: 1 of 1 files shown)


code/game/machinery/doors/brigdoors.dm
```diff
@@ -43,8 +41,6 @@
 	Radio = new/obj/item/radio(src)
 	Radio.listening = 0
 
-/obj/machinery/door_timer/Initialize()
```
silicons, line 46: why the fuck was it overriding initialize twice i wonder.

Linzolle:
is THAT why oh my god

Tupinambis:
this explains why it wouldn't FUCKIN WORK NO MATTER WHAT I CHANGED.

Ghommie (approved):

PR #11034, opened 2020-02-12
_new wrapper now works

silicons:
Well this technically did work (but I didn't test if it does so I guess it's "technically should work") but calling it with a list in SDQL2 is basically impossible so :^)


silicons:
nevermind
[] for list in sdql2

PR #11055, opened 2020-02-13
probably a better way of doing assassinate-once

Putnam3145:
## About The Pull Request

Makes once-assassinations based on a signal instead of processing all the damn time

## Why It's Good For The Game

an optimization that might also make it more consistent, who knows

## Changelog
probably not needed


Ghommie (changes requested):
There are going to be some issues with the feature if the target is deleted without dying (monkeyizing, polymorph etcetera). Maybe it could use some mind transfer hooks.

Ghommie:
Hello?

PR #11080, opened 2020-02-13
hey someone gave me this idea for stunbatons (and I sort of expanded it) and it's worth a test

silicons:
## About The Pull Request

Baton stamina damage reduced to 25 down from 35 when stunning and knocking down. 
**This means a 5 hit stamcrit, up from 3.** (100/25 round up by 1 for regen vs 100/35 round up) 
**Guns are being made unable to be used while in soft stamcrit in another PR. Same with cult stuns and whatever else gets brought up as Bhijn has said that they were infact unintentional.**
Batons by default now get a high capacity + cell and 750 hitcost.
For the sake of math, this is a total stamina damage of ~500 (up from 350 before this PR) with default cells. This also restores (see below) the 2:1 stunbaton:stunprod efficiency. Only stunprods lost knockdown soooo :^) don't know if that's going to turn out better or worse.
Batons are now able to force disarm again but only when used in right click, and **using it in right click does not knock down or do stamina damage.**
clowning around (clumsy hitting yourself) now hits a bit harder again. clown nerf clown nerf 🤡 

Stunprods are the same in everything including damage but 1000 hitcost instead of 750 without knockdown. Open to making it knockdown with less damage (but seriously how less can you get than 25 lmao)
Mathematically: stunprods now get from a 10k cell (because where are assistants getting anything else (^:) 250 stam damage total vs the old 150. Open to buffing this a bit too because shame on you tupin for nerfing the greytide.  😞 

Limb specific damage is gone. Requiring limb targeting is stupid.

## Why It's Good For The Game

Gives security their disarms back with some method of outplaying it (no stunning + disarming at the same time).
I can already see this being either too powerful due to sec players having twitch reflexes or the low delay meaning you can immediately knockdown afterwards/pick their weapon up, or too weak due to the other side having the same reflexes and killing you before you can knock them down again but it's worth a try.
I can also see complaints that the stunforce is a bit low (it is somewhat of a reduction, you can now only hard stamcrit 3 people or soft stamcrit 4) and it can be raised if this turns out to work like shit.

Raising capacity and balancing around 15k also reduces the balance issue of "40k bluespace cell". That said, it'll also make charged slime cores a slight bit weaker so.. oof, but the charge rate on those is pretty powerful so who knows how that'll play out.

Also y'all nerfed stunprods too hard, yikes. Maybe I'm nerfing it harder with no knockdown though, so give comments below.

## Changelog
:cl:
balance: Stunbatons changed yet again.
/:cl:

### DIFF
--- a/code/game/objects/items/balls.dm
+++ b/code/game/objects/items/balls.dm
@@ -20,7 +20,7 @@
 	throw_range = 14
 	w_class = WEIGHT_CLASS_SMALL
 
-/obj/item/toy/tennis/pre_altattackby(atom/A, mob/living/user, params)	//checks if it can do right click memes
+/obj/item/toy/tennis/alt_pre_attack(atom/A, mob/living/user, params)	//checks if it can do right click memes
 	altafterattack(A, user, TRUE, params)
 	return TRUE
 
--- a/code/game/objects/items/melee/energy.dm
+++ b/code/game/objects/items/melee/energy.dm
@@ -259,7 +259,7 @@
 	light_color = "#37FFF7"
 	actions_types = list()
 
-/obj/item/melee/transforming/energy/sword/cx/pre_altattackby(atom/A, mob/living/user, params)	//checks if it can do right click memes
+/obj/item/melee/transforming/energy/sword/cx/alt_pre_attack(atom/A, mob/living/user, params)	//checks if it can do right click memes
 	altafterattack(A, user, TRUE, params)
 	return TRUE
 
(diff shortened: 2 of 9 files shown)


YakumoChen:
Is there even a point to stunprods at this point? They're bulky items that are pretty wildly ineffective at stun, for what's one of the only good common ghetto stuns.

Tupinambis:
I really do not like right click combat mode mechanics at all. They clunky. It is interesting that you effectively nerfed batons further and just ignored soft stamcrit, but considering it is what I originally wanted to do, I approve.

Let stunprods disarm on right click too.

silicons:
@Tupinambis it soft stamcrits in ~5 hits now.
stunprods can disarm on right click

Ghommie (approved):

PR #11084, opened 2020-02-14
!tgs valentine

silicons:
Adds a chat command for random flirt lines

and makes valentine cards Initialize instead of New.


code/modules/events/holiday/vday.dm
```diff
@@ -64,9 +64,11 @@
...
 	w_class = WEIGHT_CLASS_TINY
 
-/obj/item/valentine/New()
-	..()
-	message = pick("Roses are red / Violets are good / One day while Andy...",
+/obj/item/valentine/Initialize(mapload)
+	message = pick(GLOB.flirts)
+	return ..()
+
+GLOBAL_LIST_INIT(flirts, "Roses are red / Violets are good / One day while Andy...",
```
Ghommie: You need to move the global list initi before any reference. In this case, since it's mentioned in multiple files, move it to _globalvars/list/flavor_misc

Ghommie (changes requested):

Ghommie (approved):

PR #11110, opened 2020-02-16
you kids like extend votes?

Hatterhat:
## About The Pull Request
extend votes. stolen from quotefox/Hyper-Station-13#96
## Why It's Good For The Game
green shifts are bad but you can make them exponentially worse by voting on this
## Changelog
:cl:
add: Extend votes! Ported from Hyper, ported from AUstation.
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -84,12 +84,14 @@
 
 /datum/config_entry/number/vote_autotransfer_initial //length of time before the first autotransfer vote is called (deciseconds, default 2 hours)
 	config_entry_value = 72000
-	integer = FALSE
 	min_val = 0
 
 /datum/config_entry/number/vote_autotransfer_interval //length of time to wait before subsequent autotransfer votes (deciseconds, default 30 minutes)
 	config_entry_value = 18000
-	integer = FALSE
+	min_val = 0
+
+/datum/config_entry/number/vote_autotransfer_maximum // maximum extensions until the round autoends
+	config_entry_value = 4
 	min_val = 0
 
 /datum/config_entry/flag/default_no_vote	// vote does not default to nochange/norestart
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -82,6 +82,16 @@
 	config_entry_value = 600
 	min_val = 0
 
+/datum/config_entry/number/vote_autotransfer_initial //length of time before the first autotransfer vote is called (deciseconds, default 2 hours)
+	config_entry_value = 72000
+	integer = FALSE
+	min_val = 0
+
+/datum/config_entry/number/vote_autotransfer_interval //length of time to wait before subsequent autotransfer votes (deciseconds, default 30 minutes)
+	config_entry_value = 18000
+	integer = FALSE
+	min_val = 0
+
 /datum/config_entry/flag/default_no_vote	// vote does not default to nochange/norestart
 
 /datum/config_entry/flag/no_dead_vote	// dead people can't vote
--- a/code/controllers/subsystem/shuttle.dm
+++ b/code/controllers/subsystem/shuttle.dm
@@ -55,7 +55,6 @@ SUBSYSTEM_DEF(shuttle)
 
 	var/lockdown = FALSE	//disallow transit after nuke goes off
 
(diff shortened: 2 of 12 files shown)


silicons (dismissed):

Ghommie:
Make it compile.

Arturlang:
Why not port the add antag option from bay for !!FUN!!

code/controllers/configuration/entries/general.dm
```diff
@@ -82,6 +82,16 @@
 	config_entry_value = 600
 	min_val = 0
 
+/datum/config_entry/number/vote_autotransfer_initial //length of time before the first autotransfer vote is called (deciseconds, default 2 hours)
+	config_entry_value = 72000
+	integer = FALSE
```
Ghommie: The value is in deciseconds and byond ticks are 1 ds minimum. I don't understand why it should be a float.

code/controllers/configuration/entries/general.dm
```diff
@@ -82,6 +82,16 @@
...
+	min_val = 0
+
+/datum/config_entry/number/vote_autotransfer_interval //length of time to wait before subsequent autotransfer votes (deciseconds, default 30 minutes)
+	config_entry_value = 18000
+	integer = FALSE
```
Ghommie: Same.

Ghommie (changes requested):

keronshb:
Honestly I do see the good in this but also the bad.

Some rounds are suppose to be short fun things (Wiz, Nukies, Gangs, maybe Cult if they're speedrunning)

Most should be a good medium run (Cult, Traitors, Bloodsuckers, etc), but I feel this might turn those into big shitter safari rounds.  These are the 1:30 - 2:30 hr rounds.

I know this would be left up to the players to vote in game, but rounds end up going past 2:30 (even then it feels too long), people might get round fatigued a lot more.  Especially if people start rallying behind votes like they use to when they weren't hidden for dynamic.

This feels like a big blow also for the tots or cult that was caught too early + killed and then it's "forced extended" for the rest of the shift and if it gets voted... even longer?  I know ghost roles do exist, but just like the station you can only do so much with them in the current allotted time that going past that feels like an eternity.

I feel like if this does stay, midrounds should be 2x or 3x the appearance rate when the round is voted to be extended.

EDIT:

Yes inb4 the 4head response "There's OTHER servers and you can just go do something else!"  Imagine waiting THAT long though?  Coming back 2+ hours later and the round is still going?

Ghommie (changes requested):
Remember to remove the shuttle subsystem's `auto_call` variable and modify its `autoEnd` procedure accordingly.

Ghommie (approved):

code/controllers/subsystem/shuttle.dm
```diff
@@ -643,10 +643,11 @@ SUBSYSTEM_DEF(shuttle)
...
-	if((world.realtime - SSshuttle.realtimeofstart) > auto_call && EMERGENCY_IDLE_OR_RECALLED) //2 hours
+/datum/controller/subsystem/shuttle/proc/autoEnd() //CIT CHANGE - allows shift to end without being a proper shuttle call?
+	if(EMERGENCY_IDLE_OR_RECALLED)
 		SSshuttle.emergency.request(silent = TRUE)
 		priority_announce("The shift has come to an end and the shuttle called. [seclevel2num(get_security_level()) == SEC_LEVEL_RED ? "Red Alert state confirmed: Dispatching priority shuttle. " : "" ]It will arrive in [emergency.timeLeft(600)] minutes.", null, "shuttlecalled", "Priority")
-		log_game("Round time limit reached. Shuttle has been auto-called.")
-		message_admins("Round time limit reached. Shuttle called.")
+		log_game("Round end vote passed. Shuttle has been auto-called.")
+		message_admins("Round end vote passed. Shuttle has been auto-called.")
 		emergencyNoRecall = TRUE
```
Ghommie:
```suggestion
	emergencyNoRecall = TRUE
```

code/controllers/subsystem/shuttle.dm
```diff
@@ -643,10 +643,11 @@ SUBSYSTEM_DEF(shuttle)
...
+		endvote_passed = TRUE
```
Ghommie:
```suggestion
	endvote_passed = TRUE
```

Ghommie (changes requested):

Ghommie (approved):

PR #11113, opened 2020-02-16
Adds diethylamine and saltpetre to the biogenerator

CameronWoof:
## About The Pull Request

- Ammonia can be printed from the biogenerator at 25 biomass for 10 units.
- Saltpetre can be printed from the biogenerator at 75 biomass for 10 units.

## Why It's Good For The Game

**Who does this benefit?** Botanists and public garden users who either cannot or do not want to construct chem dispensers in their department, but still want to participate in the stat manipulation part of the department to grow better produce for the kitchen. 

**Who does this not benefit?** Serial powergamers will still want to get their chem dispensers, as it can make unstable mutagen to fuel their projects, and with chem macros, can produce diethylamine and saltpetre in greater quantities faster.

**Why not just give them a chem dispenser?** If we're still making any attempt at roleplay standards, chemical synthesis _should_ be a skill reserved for chemists, not gardeners. I don't like the non-interactivity of chem dispensers, or the behavior they encourage: breaking into technical storage, stealing any public-access cell charger you can get your hands on, and generally availing yourself to uncontested access to one of the strongest machines on the station, all just for fertilizer.

With this PR, botanists can create these important chemicals by _working their department_. It will take a few harvests of the initial, low-yield and low-potency plants to get their hands on these bottles, but once they're established, they can focus more on experimenting and filling orders than trying to source an infinite supply of fertilizer chems.

They will **still** need to source mutagen, but I think that is a reasonable opportunity for coordination with chemistry that doesn't feel overbearing or annoying for either side.

## Changelog
:cl:
add: Ammonia and saltpetre can now be made at the biogenerator.
/:cl:

### DIFF
--- a/code/modules/hydroponics/hydroponics.dm
+++ b/code/modules/hydroponics/hydroponics.dm
@@ -664,7 +664,7 @@
 		adjustHealth(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 1))
 		adjustNutri(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 2))
 		if(myseed)
-			myseed.adjust_yield(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 0.02))
+			myseed.adjust_yield(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 0.034))
 		adjustPests(-rand(1,2))
 
 	// Nutriment Compost, effectively
--- a/code/modules/hydroponics/hydroponics.dm
+++ b/code/modules/hydroponics/hydroponics.dm
@@ -664,7 +664,7 @@
 		adjustHealth(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 1))
 		adjustNutri(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 2))
 		if(myseed)
-			myseed.adjust_yield(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 0.034))
+			myseed.adjust_yield(round(S.get_reagent_amount(/datum/reagent/diethylamine) * 0.02))
 		adjustPests(-rand(1,2))
 
 	// Nutriment Compost, effectively
--- a/code/modules/reagents/reagent_containers/bottle.dm
+++ b/code/modules/reagents/reagent_containers/bottle.dm
@@ -112,6 +112,11 @@
 	desc = "A small bottle of diethylamine."
 	list_reagents = list(/datum/reagent/diethylamine = 30)
 
+/obj/item/reagent_containers/glass/bottle/saltpetre
+	name = "saltpetre bottle"
+	desc = "A small bottle of saltpetre."
+	list_reagents = list(/datum/reagent/saltpetre = 30)
+
 /obj/item/reagent_containers/glass/bottle/facid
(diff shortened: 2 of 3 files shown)


silicons:
=|

Derkuleen:
I was requested to specifically review and place my input on this pull. Frankly i think its a pretty good idea. This way botany can allow those that are really good at it to not feel obligated to grey tide for chem dispensers. those that are only kind of good at it can still improve plant life for the station so less food can go farther which is certainly relevant.

Ghommie:
FYI, we don't have to make the game any easier because of some players' intolerance toward certain behaviours.

However make them reagent designs (check the first two ones from the start of the file for reference) and raise their cost ratio a little.

Ghommie (changes requested):
FYI, we don't have to make the game any easier because of some players' intolerance toward certain behaviours.

However turn them reagent designs (check the first two ones from the start of the file for reference) and raise their cost ratio a little.

davidchan13:
Imo it should be ammonia with t1, deith unlocked at t2, saltpetre at t3 and mutagen at t4. By that point it's easy to get a chem dispenser anyways might as well allow botanist to use their biomass for something other than synthmeat and leather.

Can't say I approve of changing the chem strength of dieth, but I could support adding a bottle or two of mutagen to the nutrimax vendor when hacked.

Owai-Seek:
I do like the idea of the biogen getting access to more stuff on upgrade. AFAIK the only thing it currently does is make things much cheaper, which is kinda lame.

As for the bottles, you could keep the ratio the same while just making the bottles 30u, other bottles in the biogen already do this, at least for robust harvest.

CameronWoof:
Gating them behind upgrades would ruin the point, as it would be faster to get a chem dispenser than wait on the upgraded parts. The purpose of this is to help botany's _early game_.

davidchan13:
I get what you're going at but imo saltpetre is too strong for botany to have mass produceable round start. 5u translates into +2 potency, 100u at once reduced production by 1 for faster harvest. Compound this on itself and dieth to Max yield and botany will have max stat plants before most science players even consider doing upgrades.

I suggest ammonia as round start because it functions as a weaker dieth, it can improve yield but takes 100u, and is a decent fertilizer too, (janitors and cook would also have more reason to work with botanist for some space cleaner)

Nano and pico manipulators can be researched and applied in the first 15 minutes of the round, so long as botany is diligent enough to ask or flag down the nerd as they move to do medical.

I agree botany isn't in a great state in no small part due to outside departments having little interest helping them, but I don't think allowing them to side step a hurdle entirely is the best way to do this, we should instead be finding the interesting or useful to trade or encourage people to upgrade their shit.

Ghommie (changes requested):
Umh, what David said, saltpetre is a tad too strong (though I understand the only plant that provides sulfur right now is the gatfruit), and the diethylamine is made with ammonia plus ethanol anyway.

CameronWoof:
I can reduce the diethylamine to ammonia, but the saltpetre is required to make this a viable alternative to chem dispenser rushing. I **really** believe it is important that it stays, or the impact of this PR on the health of the department will be minimized. Please reconsider.

Ghommie (approved):

silicons:
ugh fine i hate botany

PR #10389, opened 2020-01-01
[TESTMERGE ONLY] dumb hotfix

Putnam3145:
## About The Pull Request

Panic!

## Why It's Good For The Game

server stop crash please

## Changelog
:cl:
fix: Vendor icons are gone
/:cl:

PR #11122, opened 2020-02-16
[wip] vendors use spritesheets instead of icon2base64

Linzolle:
## About The Pull Request
basically porting https://github.com/tgstation/tgstation/pull/43914 and some further improvements to it
right now this is broken and I can't figure out why so I'm putting it up to have it looked at. Some icons just display as maid dresses in the vendors and I cannot find any discernible reason why. The items that are displays as dresses seem to all be completely unrelated.

This absolutely needs a testmerge whenever the issue is fixed.

Edit: I found out why this happens, it's because any item that is not inside a vendor that is present on the map round-start will have a null icon. I need to find a good way to update the spritesheet if the item does not exist, but I'm not sure how...

## Why It's Good For The Game
fucking closes #10389.

## Changelog
None.
/:cl:


Arturlang:
Was going to do this when i got around to tgui next vending machines

Linzolle:
i'm going to just close this for now since I still have no idea what the issue is

Linzolle:
merged into #11751

PR #11124, opened 2020-02-16
Bandaids a projectile runtime, removes an unused variable

silicons:
hitscan_last was never used
moves qdeleted cheks to after movements, because in theory everything before CANNOT cause a projectile to be prematurely deleted.

this can reveal some edge cases in which projectiles don't draw hitscan tracers properly when forcemoving in certain ways but, I'm going to have to rewrite this entire thing sooner or later anyways so whatever, bandaid it is.


Ghommie (approved):
Testmerge.

PR #11135, opened 2020-02-17
Bloodsucker Ability Tweaks

jack-avery:
##  About The Pull Request

Bloodsucker Mesmerize was an instant hardstun from within 1 tile; now is a slow before winding up to a full hardstun for longer. Both user and target are slowed so as to prevent the user from just sprinting and breathing down their neck leaving no chance for escape. The target only has to be out of view / further than 3 tiles of the user when the bar reaches full in order to prevent the hardstun.

Bloodsucker Lunge was a (near) instant aggrograb from within 3 tiles; now is a 20 frame (@30fps) windup before the aggrograb (within 4 tiles) with audio telegraphing played ONLY to the target with directional audio depending on user's position.

## Why It's Good For The Game

Mesmerize hardstun is avoidable, if difficult. Adds slight counterplay to an otherwise immediate and punishing ability; for vampire players, balances the windup with the stun lasting ~50% longer.

Lunge is now dodgeable by those who are quick to respond to audio telegraphing, which comes with directional audio in order to help.

## Changelog
:cl:
tweak: bloodsucker mesmerize now w/ windup
tweak: bloodsucker lunge now w/ windup
balance: mesmerize hardstun now lasts longer and range increased as a result
balance: lunge telegraphed, range increased as a result
add: bloodsucker lunge flavor text
add: bloodsucker lunge telegraphing audio
/:cl:


Arturlang:
Seems a lot better than the previous implementation, yeah

jack-avery:
unusable -anon

PR #11142, opened 2020-02-18
makes vampire haste more reliable and slightly less of an example of how not to code 

silicons:
## About The Pull Request

vampire haste now ticks every move rather than using a sleep loop with walk to
it has its own movement loop that automatically aborts if it can't move successfully 3 times in a row instead of using byond walk_to

it now disarms like telebatons do using 0.1 second hardstun rather than direct proccall.
if you go down from incapacitation and are resting you do not slip a few tiles anymore. just unnecessary.

you can no longer hit the same person multiple times during the move.

## Why It's Good For The Game

old code was shit, my code is presumably less shit.

also renames vampire powers to not be appended with bs_ because why the fuck is that necessary.

seriously i hate vampire code.

## Changelog
:cl:
refactor: vampire "immortal haste" has been reworked to be more reliable and consistent.
/:cl:


Arturlang:
Probably appended with bs_ to show its bloodsucker related, because vassal powers were in the same folder.

silicons:
should probably move them to different folders then

Ghommie (approved):

PR #11144, opened 2020-02-18
subtle looc

Seris02:
## About The Pull Request

looc but subtle

## Why It's Good For The Game

because it's annoying to be an antag and then having to mentorhelp or drag your fellow antag away to ask them something loocly/doing something else subtly and having to ask it in full looc

## Changelog
:cl:
add: subtle looc
/:cl:

### DIFF
--- a/code/modules/client/verbs/slooc.dm
+++ b/code/modules/client/verbs/slooc.dm
@@ -1,0 +1,73 @@
+/client/verb/slooc(msg as text)
+	set name = "SLOOC"
+	set desc = "Subtle LOOC, seen only by those within close distance."
+	set category = "OOC"
+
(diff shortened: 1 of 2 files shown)


Ghommie:
I can't really approve this. Most subtle(r) commands are enhanced copypasta. 

If something "out of character" must be said through a different channel than (A/L)OOC (like whispers, pda messages etc.), I suggest to wrap the message in double parenthesis and hope the local admin is not hitler, that's how it's done on other servers, whitin common sense.

That said people would still cautiously step aside to avoid possible eavesdropping, so there is little gain from this feature.

Ghommie:
Whops, didn't mean to close it so fast.

PR #11150, opened 2020-02-19
Bloodsucker Abilities Tweaks/Rebalances

jack-avery:
## About The Pull Request

Take 2 it's a big one folks

Bloodsucker Ability Tweaks/Rebalances
Note that apart from CLOAK which is getting a buff, these are meant to be SIDEGRADES,
reducing the "insta-stun" potential while replacing that potential with more power in other areas.

MESMERIZE
- Range increased to 3 tiles.
- Initial hard-stun removed.
- Both target and user slowed for 3 seconds while the hard-stun charges.
- Hard-stun base length increased 53%, to a minimum of 15 seconds.
- Original initial casting requirement for both the user and target to be looking at each other remains.

To prevent confusion, during the 3 second wind-up slow:
- Both the user (bloodsucker) and the target (victim) will move slowly for the FULL 3 seconds.
- Once the slow has begun, the user is able look away without breaking Mesmerize.
- The only way to prevent the hard-stun is to lack line of sight with the bloodsucker at the moment that the 3-second slow is complete. Only the bloodsucker will see the "progress bar" for this.

LUNGE
- Range increased to 5 tiles.
- 0.6 second windup time.
- Target receives an audio cue with directional audio to telegraph the attack.

To prevent confusion:
- ONLY the target can hear this audio cue.

CLOAK
- Now usable in brightly lit areas.
- Walking forced in bright areas; running/sprinting allowed in darker areas.
- User alpha (opacity) increased to 35/255 (13%, a 6% increase).
- Can only be toggled on when no other players can directly see the bloodsucker.

## Why It's Good For The Game

Complaints about some abilities being woefully underpowered or not worth taking, other abilities being instant hard-stuns with little to no counterplay (anti-fun) even if they were hard to pull off (in the case of mesmerize, which the main issue was having to be in melee range). This PR aims to promote diversity in abilities chosen as well as having some abilities synergise. In this case, Mesmerize and Lunge. Cloak buffed to where people might want to actually take it.

-- PLEASE NOTE --
This PR is a WORK IN PROGRESS balance. The way the abilities function in this PR, I believe will, be CLOSER to how I see the abilities functioning. **There will absolutely be more PRs as an active balancing effort.**

## Changelog
:cl:
add: some flavor text for lunge, cloak
tweak: mesmerize, cloak functionality/requirements
balance: mesmerize, lunge, cloak
soundadd: lunge telegraph sound
spellcheck: mesmerized status icon flavor text
/:cl:


Arturlang:
Worth a testmerge

jack-avery:
oops filenames time 2 fix

jack-avery:
ok but why

jack-avery:
sweating bullets wondering if this would make it look like im trying to commit the 200 something commits between when this post was made and now but it fixed it thank god

code/modules/antagonists/bloodsucker/powers/lunge.dm
```diff
@@ -7,9 +7,9 @@
 	button_icon_state = "power_lunge"
 	bloodcost = 10
 	cooldown = 120
-	target_range = 3
+	target_range = 5
```
Arturlang, line 10: Iffy on nearly doubling its range. Not a big worry but just iffy.

jack-avery: they don't move any faster, and the telegraphing in combination with more travel time for larger ranges is plenty of warning
edit: wording

Ghommie (approved):

PR #11151, opened 2020-02-19
Fixed map vote type config

Putnam3145:
## About The Pull Request

wow

## Why It's Good For The Game

i would like this config to work.

## Changelog
please no

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -480,18 +480,15 @@ SUBSYSTEM_DEF(ticker)
 		switch(vote_type)
 			if("PLURALITY")
 				SSvote.initiate_vote("map","server",hideresults=TRUE)
-				return
 			if("APPROVAL")
 				SSvote.initiate_vote("map","server",hideresults=TRUE,votesystem = APPROVAL_VOTING)
-				return
 			if("IRV")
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/ticker.dm
```diff
@@ -480,12 +480,16 @@ SUBSYSTEM_DEF(ticker)
...
 				SSvote.initiate_vote("map","server",hideresults=TRUE,votesystem = APPROVAL_VOTING)
+				return
 			if("IRV")
 				SSvote.initiate_vote("map","server",hideresults=TRUE,votesystem = INSTANT_RUNOFF_VOTING)
+				return
 			if("SCORE")
 				SSvote.initiate_vote("map","server",hideresults=TRUE,votesystem = MAJORITY_JUDGEMENT_VOTING)
+				return
 		// fallback
 		SSvote.initiate_vote("map","server",hideresults=TRUE)
```
Ghommie: Wouldn't it be better if you just used an else statement and indentated this row under it instead of all these return lines?

Putnam3145: i forgot there's an "else" for switches in byond. switch statements are my enemy

Ghommie (approved):

PR #11167, opened 2020-02-20
hand teles uses a (slightly) less atrocious sprite for the beam

silicons:
## Why It's Good For The Game

[s]nerfed it too hard (still think it's powerful) and people complained[/s] fine

and the default beam is fucking atrocious

oh and the beam is longer for Those Who Care.

## Changelog
:cl:
tweak: hand teles use a less atrocious beam
/:cl:


Ghommie (changes requested):
The lowered delay is unnecessary, these damn things shouldn't be used in combats and the such as much as they do.

Ghommie (approved):

PR #11181, opened 2020-02-21
measurement changes

Detective-Google:


code/__DEFINES/citadel_defines.dm
```diff
@@ -38,9 +38,9 @@
 #define COCK_SIZE_MIN		1
 #define COCK_SIZE_MAX		20
 
-#define COCK_GIRTH_RATIO_MAX		1.25
-#define COCK_GIRTH_RATIO_DEF		0.75
-#define COCK_GIRTH_RATIO_MIN		0.5
+#define COCK_GIRTH_RATIO_MAX		.42
```
Putnam3145: it needs to be 0.42 and so on

Putnam3145 (dismissed):

code/__DEFINES/citadel_defines.dm
Detective-Google: did it not fucking commit the changes GOD I LOVE WEBEDITS.

"commit directly to branch" my ass 😂

Detective-Google: oh no I just left out the zeros whoops

dapnee:
.... girth is circumference. Thanks for making me google this cause I ain’t got one myself.

silicons (approved):

Ghommie (changes requested):
Replace "girth" with "diameter" in the defines and so on.

Detective-Google:
I wish github notiffs didnt suck

Linzolle:
i'm not sure about you but i don't think the average 8-inch dick is 2 inches in diameter that seems kind of. dumb? what do I know though.

Detective-Google:
> i'm not sure about you but i don't think the average 8-inch dick is 2 inches in diameter that seems kind of. dumb? what do I know though.

Most players assume width when they hear girth, meaning that they see "they have a 20in length and 15in in girth". believe it or not having a sausage over a foot long in diameter is a bit quirky, so to convert from the circumference to the diameter you divide by pi or in this case 3 which makes the measurement more in line with what people expect (though the ratio itself might not be ideal in the first place, now it is at least measured in a way that doesn't confuse players.)

Ghommie (approved):
cursed.

PR #11190, opened 2020-02-21
Virology works again

Psody-Mordheim:
## About The Pull Request
Fixes runtime error
Makes polyvitigo just give colorful reagent until it hits 100u

## Why It's Good For The Game
Makes virology actually work, instead of being broken which it has been for the last 8 days.

It's almost like someone didn't test their changes to the mobtype defines after making them, hmmmst

## Changelog
:cl:

fix: /mob/living/proc/CanContractDisease(datum/disease/D) doesn't look for mobtypes in the disease anymore.

fix: Polyvitigo actually makes you colorful or whatever, honk
/:cl:

### DIFF
--- a/code/datums/diseases/_MobProcs.dm
+++ b/code/datums/diseases/_MobProcs.dm
@@ -17,9 +17,8 @@
 	if(HasDisease(D))
 		return FALSE
 
-	if(!(D.infectable_biotypes & mob_biotypes))
+	if(!(D.infectable_biotypes))
 		return FALSE
-
 
 	if(!(type in D.viable_mobtypes))
 		return FALSE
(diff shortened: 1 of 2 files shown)


Arturlang:
Oh shit, i thought it was the fucking synthblood, not the biotypes port, which is really weird, since i ported it from TG.

Psody-Mordheim:
> Oh shit, i thought it was the fucking synthblood, not the biotypes port, which is really weird, since i ported it from TG.

It was both but you had to fix synthblood first to even notice this.

code/datums/diseases/_MobProcs.dm
```diff
@@ -17,10 +17,9 @@
 	if(HasDisease(D))
 		return FALSE
 
-	if(!(D.infectable_biotypes & mob_biotypes))
```
Ghommie, line 20: The real issue is there are still a dozen critters that use mob biotypes lists instead of bitflags.
Removing this will just alter and break some features.

Ghommie (changes requested):

Ghommie:
Properly fixed on #11244.

PR #11217, opened 2020-02-24
plushies now use SAFEPICK

silicons:
runtime fix


code/game/objects/items/plushes.dm
```diff
@@ -37,7 +37,9 @@
 /obj/item/toy/plush/random_snowflake/Initialize(mapload, set_snowflake_id)
 	. = ..()
 	var/list/configlist = CONFIG_GET(keyed_list/snowflake_plushies)
-	var/id = pick(configlist)
+	var/id = safepick(configlist)
```
Ghommie, line 40: While I like not having this runtime happen every single round, you should make sure the configs actually work if you haven't done so already and then ping headmins with all your might until it's fixed server side.

Ghommie (approved):

PR #11221, opened 2020-02-24
Reverts #10744 (More random engines)

YakumoChen:
## About The Pull Request

Title

## Why It's Good For The Game

All of these engine choices are objectively terrible.
-TEG: Hyper-safe, very easy to set up, goes against design philosophy for the main engine, can already be built by engineering anywhere, in the space of a pod. Produces 10MW of power with low effort.
-AM: Ugly. God-awful. Boring. Takes literally a single minute of setup. Produces more power than the most autistic SM setups. Only makes a shitty explosion when sabotaged. This should not be a main engine choice.
-PACMAN: No. There's already one of these in the SMES room anyway.
-Empty: No.

The purpose of the main engine on station is twofold: To give the engineers something to set up, maintain, and tinker with, and to give antagonists something highly threatening to sabotage. None of these engines provide EITHER of these. The TEG comes the closest to one criteria but critically lacks any sort of risk to build.

If the engineers want to build a TEG, they can do so while still having a supermatter or singularity to babysit, and very easily do so.

## Changelog
:cl:
del: Engineering may now only choose from dangerous engines again.
/:cl:


Arturlang:
Ehh, why not give the TEG some love, its pretty fun.

YakumoChen:
You don't even need cargo to build one, I believe. You can just build it anywhere, there's no need to remove the dangerous, sabotagable engine from rotation while doing so.

r4d6:
Just bother the headmins to set the engines weight to 0.

MrJWhit:
So, I disagree with this PR on two levels, 

First is that this is directly against the whole idea of modularity, because this is the same as forcing the chance of spawn to zero, but without just... asking a head admin to reduce the weight to zero. Talk to them rather than making a PR to straight up remove them.

Secondly, the removal of the TEG. This has been talked about for a few hours on the Discord, and we've come to the following conclusions:
1. The TEG is susceptible to sabotage really easily, there's a distro and waste pipe in the system. 
2. The TEG being able to be produced by engineering to produce a lot of power in a small area is an issue, yes, but that's not what this pr is doing. The argument literally doesn't make sense. It's a giant engine that fills up the same space as SM, possibly even more. 
3. The TEG is arguably more complicated than the SM because it requires both a heating angle and a cold angle, rather than just a cold angle.

This is a bad PR and should be closed.

MrJWhit:
For the record: A headmin has been pinged about reducing it from their side.
The empty engine is already set at zero, and wasn't changed.
The AM machine was reduced to zero.

silicons:
config issue, please only code remove non functional engines.
the others are stupid jokes but they're functional.

PR #11225, opened 2020-02-24
Make the mining base's placement random

r4d6:
## About The Pull Request
Port https://github.com/yogstation13/Yogstation/pull/4612

## Why It's Good For The Game
Now the mining base won't be in the same spot next to the gulag anymore. So miners that want to go there first need to locate where on the map the base is to orient themselves.

## Changelog
:cl:
change: Mining base is now randomly placed on lavaland
/:cl:


Ghommie (approved):
You should give some curve to the border line below the bridge that connects the public mining base/break room to the rest of the station, but all good.

silicons:
i feel like there's a better way to handle this than a goddamn ruin

r4d6:
Well, that's how Yog does it, and it does make sense, the mining station need to be randomly placed on lavaland, there is a whole system to place ruins on lavaland, so it is a ruin.

Ghommie:
Kevinz salty because the solution was fairly simple.

CharAznable117:
Please change the spawning back to how it was, with a fixed mining base spawn, as megafauna spawn most often than not right next to the mining base with this new system and have destroyed it without being lured at round start. Its a major hassle to even deal with this and is a massive issue for resource gathering. I also don't believe this solves the gulag issue, as I think moving the gulag someplace else, via the random spawning perhaps may be a better solution.

PR #11239, opened 2020-02-25
new item test

Detective-Google:


CameronWoof:
Security buff. I like it.

dapnee:
xenobio can do this, but it needs crossbreeding. Dislike giving crew more ways of pacifying other players just cause, I don't see pacification surgery or the flower being used and adding this would make them invalid cause this is way easier.

Owai-Seek:
I think it should probably require research to be created, and/or be locked to the security lathe so not everyone can get it. Either that, or require something only security would have in its crafting (IE: An actual full-blown taser)

If someone without a mindshield tries to remove it, maybe an alert is sent to security, or it zaps the person trying to touch it?

necromanceranne:
Please don't tell me you're getting PR ideas from Girly Station, Google.

silicons:
alright first of all why pacification isn't uh, supposed to be this easy to get

second

> If someone without a mindshield tries to remove it, maybe an alert is sent to security, or it zaps the person trying to touch it?

no.

Detective-Google:
> Please don't tell me you're getting PR ideas from Girly Station, Google.

i had the idea and I felt like making it so it exists that's basically it.

Detective-Google:
ehh overall I did this more as a fun project and threw it up on git since I figured people might be interested, they're not so I'm not gonna worry about it.

PR #11290, opened 2020-03-01
Traitor classes: a new paradigm.

Putnam3145:
## About The Pull Request

Adds traitor classes. There are now five types of traitors:

1. MI13 operatives. They steal, sabotage and assassinate/maroon people (once-only for assassination; permanent assassination has a place, but I'm not sure where).
2. Waffle Co operatives. They're given free reign to terrorize, mock or otherwise antagonize the crew, as long as they escape with their lives.
3. Gorlex Marauders. They can do whatever they want, as long as they hijack the shuttle.
4. Tiger Cooperators. They can do whatever they want, so long as they die a glorious death.
5. Syndicate Operatives. Actually a fallback, just in case something odd happens; is assigned objectives just like the old way.

## Why It's Good For The Game

Flavor objectives suck?? As it turns out?? And I think this system could be an interesting flexible thing. Obviously people should review this, not just maintainers, to tell me if it sucks, cause lemme tell you, flavor objectives were not looked at hard enough 😂

## Changelog
:cl:
add: Traitor classes for traitors: a new way for traitors to have objectives assigned.
/:cl:


Theolorex:
Regarding the concept, I personally like it but I do think there needs to be _some_ level of _encouraging_ (not forcing, thats solved through server rules) traitors to do something that isn't mass killing. Where original objectives would tell you to kill _specific_ people or to do objectives that would be hard if you had sec after you like getting the SM crystal, there isnt much encouragement from 2-3 of the classes (Donk Co, Tiger, _maaaybe_ Gorlex)

==============

I do also have a concept for _another_ team, could have it as mimes or something rad

(im def not the greatest at understanding how far you can get with code, so ~~everything~~ something here'll probably sound impossible to do/add too much flags or somethin)

there should be class specifically for terrorizing/masterminding the station
1. Create a hull breach in 5 places (or in specific places) on the station
2. Gib (specific person or random person) and take a picture of their remains (Optional: add a clause to drop it in a specific place like medbay or bar) OR scatter it across (specific place)
3. Revive a traitor/hostile enemy
4. Start a fire in (place with high traffic, like medbay)
5. Poison 10 people through food
6. Murder and steal the identity of (high value role) and pose as them until the end of the shift (gives you something for this objective to look like them)

This kind of stuff fits a bit into the Tiger's concept with die a glorious death or MI13's with the sabotage but I dont think they got the encouragement enough or the aesthetic

ghost:
Very interesting approach to freeform jecties.  However, a mandatory targeted assassination role feels disproportionately restrictive compared to the other classes.

Putnam3145:
Yeah, the assassination class is definitely the most questionable out of these. I kinda like the idea, but there's no real discouragement to prevent rampages, which should be done. The problem is partially that stealth traitoring is... well, people don't like it much. Too many ways to perfect-crime, and if the crime isn't perfect you're probably gonna get caught in a couple hours, and so on.

code/modules/antagonists/traitor/classes/subterfuge.dm
```diff
@@ -0,0 +1,40 @@
...
+		maroon_objective.owner = T.owner
+		maroon_objective.find_target()
+		T.add_objective(maroon_objective)
+	else
+		if(prob(15) && !(locate(/datum/objective/download) in T.objectives) && !(T.owner.assigned_role in list("Research Director", "Scientist", "Roboticist")))
+			var/datum/objective/download/download_objective = new
+			download_objective.owner = T.owner
+			download_objective.gen_amount_goal()
+			T.add_objective(download_objective)
+		else if(prob(70)) // cum. not counting download: 40%.
```
ghost, line 31: putnam horny

Putnam3145: i thought i fixed that.

keronshb:
I'm going to have to agree RE: stealth traitoring.  ESPECIALLY stealth killing.  There are a lot of ways to give that away (how loud sound wise that almost every item is to kill someone with, people not respecting/skirting around the blackout rule, PAIs, blood, stamina if it's melee, !! going through friggin z levels and last minute headset calls, suit sensors, and even fingerprint scanners).

Also I suppose the definition of "murder spree" is different from person to person.  I've seen someone kill anyone that tried to stop them, and then they complain that the traitor is murderboning when it's in self defense.

EmeraldSundisk:
While I get there's a lot to it, something like killing a cook and serving their gibbed corpse's meat as food seems right up the Animal Rights Consortium's alley. Maybe have it that their meat simply needs to be used in the creation of a food item, not necessarily consumed, and there's nothing saying the person can't be resurrected.

(Yeah I know there'd have to be more for them, maybe have it be one option among a bigger objective pool although protecting non-sapient creatures is a given.)

Putnam3145:
I was trying to think of how to integrate the ARC into it and, yeah, most of my ideas came up short. An entire traitor subclass that targets chefs seems off.

EmeraldSundisk:
It does, although they'd probably see them as more of a priority than others. I suppose one idea is them having a goal to not only protect a station pet but make sure it escapes on a shuttle out of custody just the same as a person.

Hatterhat:
let the gorlex gamers buy the contractor kit tho

Putnam3145:
wait, i forgot, did we merge the new hijack, cause if so, i probably should tbh

Hatterhat:
is this breaking uplinks it might be

silicons (approved):

PR #11305, opened 2020-03-02
you can no longer stamcrit a goddamn xenomorph come on citadel why is this a thing 

silicons:
this is just completely unnecessary now that i fixed the infinite hardstun bug

why was this a thing in the first place? it's a xenomorph, you lethal them and use face protection, not spam disablers/rubbers until they're ""stamcrit"" but not yet stamcrit what the fuck?


Ghommie:
Linter, missing keyword arguments from `mob/living/carbon` and `mob/living`'s  `adjustStaminaLoss`.

Ghommie (approved):

PR #11307, opened 2020-03-02
gets rid of fly species requirement from hulk

ghost:
## About The Pull Request
also fixes (i assume this isn't intentional) chunky man not being removed on mutation loss.

probably a relatively bad idea unless we commit to tg-oon genetics

## Why It's Good For The Game
honestly the racial requirement of flyperson doesn't restrict this at all to anyone actually looking to gamer gear up, no work needed to become a flyperson and now you have the benefits of hulk - so the "restriction" just produces more bugs than it's worth to 'prevent people from using hulk'. furthermore, with it being soft removed by tg genetics, i'm sure someone somewhere can find a better balance for hulk than "run into teleporter", "be certain race", or "you can take less niche superpowers if you take hulk but still can take the really important ones" before it gets fully introduced again.

also hulk isn't actually super overpowered here because it wasn't updated with stamcombat so you can actually take out hulks nonlethally just fine on citcode. kinda funky.

## Changelog
:cl:
balance: removed (mostly aesthetic) requirement to become hulk
fix: changed add to remove on crit
/:cl:


Ghommie:
First half pending #10327. Thanks for the fix though.

ghost:
honestly i'm almost tempted to make it so only golems can have hulk 😆 
don't think about it too hard

Ghommie:
It's clobberin' time.

silicons (approved):

ghost:
travis isn't my fault :(

PR #11322, opened 2020-03-04
Typing indicators (without baycode lag edition)

silicons:
## About The Pull Request

Title
30 seconds timeout if the user doesn't speak. Support (but not implemented) for different typing indicators per mob/species.

ONLY SUPPORTS SAY CHATBOX. Due to the mechanics of how this (as in the part where I make it still clientside in essence) works (abusing the ANY + UP/DOWN macros to basically async by "binding" two procs to one key), this will ONLY support say box. If someone wants to code a lagless implementation (and a damn good implementation that is as stable and workable in times of crash/lag) of a JS based input bar, be my guest but this is how it is for now because for a gameplay server I'm absolutely not trading performance for this whereever possible. This is good enough for its intended purpose.

## Why It's Good For The Game

Roleplay, and before anyone says "but what if typebaiting": Hey it'll happen either way, you can use say bar to opt out (or if there's enough demand someone can make a preference for opting out but I honestly think it's unnecessary and this won't be that big of an issue).

## Changelog
:cl:
add: Typing indicators! Only works if you use say/me popups though.
/:cl:

### NOTE: IF WE EVER GET REBINDABLE KEYS, THIS WILL NEED TO BE MODIFIED TO FOLLOW SUIT. I'm already damn picky about rebindable keys and I'd want to require that it have support for clientside-based macros (aka using winset to bind clientside verbs to not introduce baycode server based sayverbs which incur a round trip latency for the popup box appearing). Just something to note for anyone looking into that.


code/modules/mob/living/carbon/human/species.dm
```diff
@@ -2218,3 +2221,8 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
 /datum/species/proc/start_wagging_tail(mob/living/carbon/human/H)
 
 /datum/species/proc/stop_wagging_tail(mob/living/carbon/human/H)
+
+
+/////// TYPING INDICATORS ///////
+/datum/species/proc/get_typing_indicator_state()
```
Ghommie: proc overhead.

code/modules/mob/living/carbon/human/typing_indicator.dm
```diff
@@ -0,0 +1,2 @@
+/mob/living/carbon/human/get_typing_indicator_icon_state()
+	return dna?.species?.get_typing_indicator_state() || ..()
```
Ghommie:
```suggestion
	return dna?.species?.typing_indicator_state || ..()
```

code/modules/mob/mob_defines.dm
```diff
@@ -128,3 +128,13 @@
...
+
+	///////TYPING INDICATORS///////
+	/// Set to true if we want to show typing indicators.
+	var/typing_indicator_enabled = FALSE
+	/// Default icon_state of our typing indicator. Currently only supports paths (because anything else is, as of time of typing this, unnecesary.
+	var/typing_indicator_state = /obj/effect/overlay/typing_indicator
+	/// The timer that will remove our indicator for early aborts (like when an user finishes their message)
+	var/typing_indicator_timerid
+	/// Default typing indicator timeout
+	var/typing_indicator_timeout = 30 SECONDS
```
Ghommie: That's too much time, 6-8 seconds would suffice, the chatbox is not really suitable for long messages. Also any reason this shouldn't be a define instead?

EDIT: oh wait, does this things actually check keybindings instead of the chatbox?

code/modules/mob/typing_indicator.dm
```diff
@@ -0,0 +1,44 @@
...
+/*!
+ * Displays typing indicator.
+ * @param timeout_override - Sets how long until this will disappear on its own without the user finishing their message or logging out. Defaults to src.typing_indicator_timeout
+ * @param state_override - Sets the state that we will fetch. Defaults to src.get_typing_indicator_icon_state()
+ * @param force - shows even if src.typing_indcator_enabled is FALSE.
+ */
+/mob/proc/display_typing_indicator(timeout_override = typing_indicator_timeout, state_override = get_typing_indicator_icon_state(), force = FALSE)
+	if(!typing_indicator_enabled || force)
+		return
+	add_overlay(get_indicator_overlay(state_override))
```
Ghommie: This overlay should also be stored in a variable...

code/modules/mob/typing_indicator.dm
```diff
@@ -0,0 +1,44 @@
...
+	addtimer(CALLBACK(src, .proc/clear_typing_indicator, state_override), timeout_override, TIMER_STOPPABLE)
+
+/*!
+ * Removes typing indicator.
+ * @param state_override Sets the state that we will remove. Defaults to src.get_typing_indicator_icon_state()
+ */
+/mob/proc/clear_typing_indicator(state_override = get_typing_indicator_icon_state())
+	deltimer(typing_indicator_timerid)
+	cut_overlay(get_indicator_overlay(state_override))
```
Ghommie: Of course the overlay should be deleted or the variable made null for the typing indicator to keep working.

code/modules/mob/typing_indicator.dm
```diff
@@ -0,0 +1,44 @@
...
+	// doesn't exist, make it and cache it
+	if(ispath(state))
+		. = GLOB.typing_indicator_overlays[state] = state
+	// We only support paths for now because anything else isn't necessary yet.
+
+/// Gets the state we will use for typing indicators. Defaults to src.typing_indicator_state
+/mob/proc/get_typing_indicator_icon_state()
+	return typing_indicator_state
+
+/*!
```
Ghommie: IIRC, the proper doc format uses two asterisks and no exclamation point, as well as the following lines  of the commentary having two spaces at the start instead of one, but w/e.

code/modules/keybindings/bindings_mob.dm
```diff
@@ -3,6 +3,17 @@
...
+	if(SSinput.typing_indicator_binds[_key])
+		var/macroset = winget(user, "mainwindow", "macro")
+		var/list/L = SSinput.macro_set_reverse_lookups[macroset]
+		var/valid = TRUE
+		if(length(L) > 1)
+			for(var/i in 2 to length(L))
+				if(!client.keys_held[L[i]])
+					valid = FALSE
+					break
+		if(valid)
```
Ghommie, line 19: And the below proc shouldn't run if said overlay variable exists to avoid typing indicators from potentially stacking up.

Ghommie (changes requested):

Ghommie:
It is going to be pretty unpractical and lame since it relies on the keybinding instead of the skin chatbox.

code/modules/mob/mob_defines.dm
silicons: yes

Linzolle:
Personally I think having it in the preferences is a must-have, I can foresee many people complaining about type-baiting and not wanting to use the verb bar.

silicons:
Patched to work if: 1. you use a keybind
2. You use say/me in command bar

DOES NOT intercept command bar direct inputs for say/me though.

code/modules/mob/typing_indicator.dm
silicons: It's not because say, you start talking in clockcult/whatnot, and it wants to apply a different overlay
Not implemented but I made it ensured that it will support multiple typing states.

code/modules/mob/typing_indicator.dm
silicons: Overlay is just a path but I need a way to prevent stacking hmm.

silicons:
ehh if anyone has interest in this feel free to pick it up but unfortunately it is at the current moment not worth it, as I am forced to chosee between lag and typing indicators, and as someone who's here for the game I am not choosing to create input lag.

The other option is javascript input but I don't have time to do that.
Sorry folks. I wish there was a way.

PR #11325, opened 2020-03-04
Removed some bad flavor objectives.

Putnam3145:
## About The Pull Request

Traitor classes ought to make this obsolete, but can't make the problem better in the meantime. They're just strings, you can read the code, I don't know how to make shorthand for this.

## Why It's Good For The Game

These flavor objectives caused nothing but shittery and/or confusion.

## Changelog
:cl:
del: Removed some particularly bad flavor objectives.
/:cl:


Linzolle (approved):

necromanceranne:
Not to be a debbie downer but even at the risk of a bomb detonation I've seen some real apeheads in sec just rush into rooms with a hostage. The results are about as predictable as you would think. I genuinely don't think that objective ever works, even if fun in theory, and it is primarily something you shouldn't try and force antags to do given the playerbase itself isn't going to give a toss to even humor antagonists in this way.

PR #11327, opened 2020-03-04
Fixes reagent containers transfer cycling

Arturlang:
Fixes reagent container inhand use transfer, all it really was lacking is a proper return at the end, previously it simply cycled through all of them.
## Changelog
:cl:
fix: Fixed reagent container transfer amount cycling.
/:cl:


Arturlang:
A extremely simple fix. It is also tested and does indeed fix the problem.

Ghommie (approved):

PR #11329, opened 2020-03-04
Made laser carbines and magrifles bulkier.

Ghommie:
## About The Pull Request 
Title. Both have a bulky weight class now, the former also made unsuitable for gun akimbo.
On the good side, they can be worn on the back now, and the former's accuracy was rised.
Also made the (unused) hyperburst huge, the magpistol is still normal sized.

## Why It's Good For The Game
The laser carbine and mag rifle having the same size of the laser gun and mag pistol is pretty much stupid design, and the fact both the former are easily printable only amplifies the issue.

## Changelog
:cl:
balance: Made laser carbines and magnetic rifles bulkier. They can be worn on the back though.
imageadd: slightly different inhands sprites for the laser carbine.
/:cl:


code/modules/projectiles/guns/energy/laser.dm
```diff
@@ -56,12 +57,14 @@
 /obj/item/gun/energy/laser/carbine
 	name = "laser carbine"
 	desc = "A ruggedized laser carbine featuring much higher capacity and improved handling when compared to a normal laser gun."
```
Trilbyspaceclone, line 59: Improved handling?

Trilbyspaceclone:
also
image
Did you mean to delet this?

code/modules/projectiles/guns/energy/laser.dm
Ghommie: Not my description of the item but that's not a typo.

Ghommie:
It was a duplicate, Some icondiffbot entries that end up with `-X` where X is a number may be duplicates anyway, and as of Byond version 513 it's also easier to spot duplicates in the editor.

silicons:
huge is a bit bigger than bulky i feel like you're making magrifles kinda garbage even if they're a printable energy-ballistic

Ghommie:
They work just as well right now (save gun akimbo), they just require the user to manage their inventories more carefully now (read: armor suit storage), but you and others were accostumed to cancerously spam boxes of magazines and maybe carry a reserve magrifle before so I understand if it feels shitty for you now.
Also the only huge item here is the adminspawn hyperburst and its toy version.

silicons:
magazine box spam is already solved by energy fam, at this point it's literally just autorifles but worse in every single way, there's no reason at all to take it over those (because honestly if you have enough to print a lot of magrifles you can sell some for a crate of smgs).
it's not about carrying reserve magrifles, it was already only good because of magazine spam post burst fire nerf, and then made to use energy. that's fine, magazine spam was cancer and I say this as someone who uses it, but at this point what even is the point of using this? 


.. on the other hand I really don't like NT having ballistics
:pensive:

Ghommie:
You should have thought well that when you buffed wt550s you made other guns worse by comparision, so let's not develop around that "gamer gear" mindgame while the server administration enforces idiotic rules, shall we?

silicons (approved):

Ghommie:
Maaaaan, look at all those typos I have done in these comments, surely must have been too distracted to formulate sentences right.

PR #11335, opened 2020-03-05
Made flavor texts into an element datum, appliable to every atom.

Ghommie:
## About The Pull Request
Title. This is only backend support for now.

## Why It's Good For The Game
Code support for various shenanigeans, such as dumping the entire bee movie script into a bikehorn without delivering a huge wall of text on the chat screen upon examine.

## Changelog
None, backend.


code/datums/elements/flavor_text.dm
```diff
@@ -0,0 +1,112 @@
...
+		addendum = _addendum
+	always_show = _always_show
+
+	RegisterSignal(target, COMSIG_PARENT_EXAMINE, .proc/show_flavor)
+
+	if(can_edit && ismob(target)) //but only mobs receive the proc/verb for the time being
+		var/mob/M = target
+		if(!verb_instance)
+			verb_instance = new /datum/element/flavor_text/proc/set_flavor (src, "Set [_name]", _desc)
+		M.verbs += verb_instance
```
Ghommie: I'm not even sure this piece of byond procpath magic actually works yet.

code/datums/elements/flavor_text.dm
```diff
@@ -0,0 +1,140 @@
+GLOBAL_LIST_EMPTY(mobs_with_editable_flavor_text) //et tu, hacky code
```
Putnam3145, line 1: does this *need* to be a global list? It's not used in any other files.

Ghommie: That's the least awful way to allow `mob/proc/manage_flavor_texts` to call `flavor_text/set_flavor` proc that I had in mind, and probably better than scrolling through every associated value of `SSdcs.elements_by_type` or a new mob level variable to store flavor text elements in when we have plenty of bad ones already.

silicons (approved):

PR #11349, opened 2020-03-06
Pixel shifting for ERP

monster860:
This code is completely 100% untested.

## About The Pull Request

What this does is make it so that if you are holding ctrl+shift and then do a direction it causes you to pixel shift

## Why It's Good For The Game

well you see I was doing a threesome ERP and I was on top of someone with someone else and we couldn't position ourselves correctly so I open this PR so that no one else has to deal with the limitations of tile based movement when trying to ERP.

## Changelog
:cl:
add: Use Ctrl-Shift-direction key to shift your characters position. Use for ERP.
/:cl:

### DIFF
--- a/code/modules/mob/mob_defines.dm
+++ b/code/modules/mob/mob_defines.dm
@@ -38,7 +38,7 @@
 	var/resting = 0			//Carbon
 	var/lying = 0
 	var/lying_prev = 0
-	var/is_shifted = 0
+	var/is_shifted = FALSE
 
 	//MOVEMENT SPEED
 	var/list/movespeed_modification				//Lazy list, see mob_movespeed.dm
--- a/code/modules/mob/mob_defines.dm
+++ b/code/modules/mob/mob_defines.dm
@@ -38,6 +38,7 @@
 	var/resting = 0			//Carbon
 	var/lying = 0
 	var/lying_prev = 0
+	var/is_shifted = 0
 
 	//MOVEMENT SPEED
 	var/list/movespeed_modification				//Lazy list, see mob_movespeed.dm
--- a/code/modules/keybindings/bindings_mob.dm
+++ b/code/modules/keybindings/bindings_mob.dm
@@ -61,16 +61,28 @@
 	if(client.keys_held["Ctrl"])
 		switch(SSinput.movement_keys[_key])
 			if(NORTH)
-				northface()
+				if(client.keys_held["Shift"])
+					northshift()
+				else
+					northface()
 				return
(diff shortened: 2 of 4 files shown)


monster860:
Against my better judgement I tested the PR. Here's a picture

image

monster860:
image

BlackMajor:
PR of the year.
Make sure it works for other mobs too. Like ash drakes, borgs and xenos.

monster860:
it should work on all mobs

BlackMajor:
Bless.

dapnee:
absolute madlad. 
now for this to be used for combat /s

Ragolution:
> absolute madlad.
> now for this to be used for combat /s

If it doesn't already do so on tile transition, this should have a line specifying a recenter on movement, to prevent abuse.

Detective-Google:
> > absolute madlad.
> > now for this to be used for combat /s
> 
> If it doesn't already do so on tile transition, this should have a line specifying a recenter on movement, to prevent abuse.

Should also be a limit to prevent shifting out of the tile they're currently in(don't want people shifting their sprites into walls and whatnot.)

monster860:
you can't shift more than 16 pixels in a direction to prevent aboos

Ragolution (approved):
I think it's funny I can even approve things.

Ghommie:
The pixel shift doesn't reset on move, perhaps add two variables to store the x and y shifted pixels amount and a comsig hook to reset them on move.

Ghommie (changes requested):
The pixel shifting doesn't reset on move, perhaps add two variables to store the x and y shifted pixels amount and a comsig hook to do the job.

monster860:
there I did the thing

code/modules/mob/mob_defines.dm
```diff
@@ -38,6 +38,7 @@
 	var/resting = 0			//Carbon
 	var/lying = 0
 	var/lying_prev = 0
+	var/is_shifted = 0
```
silicons:
```suggestion
	var/is_shifted = FALSE
```

Detective-Google:
~~@monster860 as of commit c8535275ce being testmerged on the server, using the "set custom holoform" pAI option and then swapping your holo to the new one whilst pixel shifted will result in your "default" position to being the position you were shifted in when you set the holoform.~~

Apparently it's an issue with tauric bodies(specifically the naga) instead? unsure from my testing.

Ghommie (approved):

PR #11351, opened 2020-03-06
tasers are now as fast as disablers and 7 --> 14 tile range  but do not have impact damage

silicons:
## About The Pull Request

changes projectile speed to match disablers + 14 tiles range vs 12 for hybrids and 7 for the other ones
Removes impact damage from taser

## Why It's Good For The Game

maybe i should have just killed it instead of making it a weird thing but uh ok

Tasers were meant to be an alternative fire mode to slowdown. Not a better or worse disabler.

## Changelog
:cl:
balance: disabler taser alt fire shots are faster and have a 14 tile range now
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -132,7 +132,7 @@
 	var/slowdown_priority = 50		//to make sure the stronger effect overrides
 	var/affect_crawl = FALSE
 	var/nextmove_modifier = 1
-	var/stamdmg_per_ds = 1		//a 20 duration would do 20 stamdmg, disablers do 24 or something
+	var/stamdmg_per_ds = 0		//a 20 duration would do 20 stamdmg, disablers do 24 or something
 	var/last_tick = 0			//fastprocess processing speed is a goddamn sham, don't trust it.
 
 /datum/status_effect/electrode/on_creation(mob/living/new_owner, set_duration)
@@ -167,6 +167,7 @@
 	slowdown_priority = 100
 	nextmove_modifier = 2
 	blocks_combatmode = TRUE
+	stamdmg_per_ds = 1
 
 /datum/status_effect/electrode/no_combat_mode/on_creation(mob/living/new_owner, set_duration)
 	. = ..()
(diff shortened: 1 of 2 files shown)


silicons:
oh yeah not just those have the taser shots ugh

Ragolution:
why tho

Ghommie (changes requested):
No, nerf disablers cost instead, a good chunk of the disabler buffs were only made to discourage taser use when the latter mode was still dominant.

silicons:
Updated to entirely remove the DoT and just be a slow at the cost of 2.5 disabler shots.
Tasers were meant to be a slowdown alt fire for niche/situational cases, not just a better disabler or a worse one (which is what it is as of right now, there's no reason to use it when you can just use disablers).

And re: disabler nerfs, I don't think it's necessary with security already having been nerfed a lot, the current sprint changes, and the fact that just about nothing else has been patched/redone to match the new numbers and such. It's important to realize that while security shouldn't AUTOMATICALLY WIN 1v1s, we're already long, long past that point with new weapons. Antags shouldn't take on security without a good reason to and it shouldn't be incentivized by security being made out of paper mache (is that how you spell it lmao). We're a long ways away from the old stun & done security combat, you now have a decent-ish chance with just mediocre preparations if you are skilled, and for an antagonist, your usual equipment makes you pretty good already (even if baton disarm will always be a fucking meme but it's as of right now kind of a bandaid for how terrible every other part of the system is).

I don't want to reward validhunting/mindless charging the antags but I also don't want antagonists to be downright unstoppable half the time. In a perfect world, yes, the crew will band together and security will have teamwork. But consider that solo antagonists should rely on stealth and surprise ambushes instead of going and charging down security default vs default gear, you can already do this if you well, buy loud weapons, but they shouldn't automatically win just like how sec shouldn't automatically win. Anyways, regarding teamwork and requiring groups of people to take down one SOLO ANTAGONIST:

This is citadel.
Even on /tg/, that doesn't work out. This game platform itself and the entirety of the combat system is just, not built for effective teamwork.
Why would it work out here even if it's made a necessity?

Ghommie (approved):

PR #11361, opened 2020-03-06
Fixes bartender drink slide spill A

### LINKED ISSUE #11342: Drink flinging removing liquids
## Round ID:
21874

## Reproduction:
throw a drink with drink flinging enabled watch how it becomes empty even though it should of kept its liquid inside.

Ragolution:
Forces drink throw call to check for bartender throwing before spilling contents. Feels like a lazy fix, but it works. Refer to #11362 for different method.

Resolves #11342

### DIFF
--- a/code/modules/reagents/reagent_containers.dm
+++ b/code/modules/reagents/reagent_containers.dm
@@ -156,8 +156,6 @@
 		reagents.clear_reagents()
 		if(QDELETED(src))
 			return
-	if (!bartender_check(target)) // Needs to check for Bartender Slide before spilling.
-		reagents.clear_reagents() // Clear reagents of thrown container.
 
 //melts plastic beakers
 /obj/item/reagent_containers/microwave_act(obj/machinery/microwave/M)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/reagent_containers.dm
```diff
@@ -156,7 +156,8 @@
 		if(QDELETED(src))
 			return
```
Ghommie, line 159:
```suggestion
		reagents.clear_reagents()
		if(QDELETED(src))
			return
```

code/modules/reagents/reagent_containers.dm
```diff
@@ -156,7 +156,8 @@
...
 
-	reagents.clear_reagents()
+	if (!bartender_check(target)) // Needs to check for Bartender Slide before spilling.
+		reagents.clear_reagents() // Clear reagents of thrown container.
```
Ghommie:
```suggestion
```

Ghommie (changes requested):

code/modules/reagents/reagent_containers.dm
Ragolution: That seems like it'll break a bunch of shit. Are you sure? This instance of reagents.clear_reagents seems to be the main method of garbage collection for broken & thrown containers. At a glance, this probably would break beaker throw-splashing, too.

Ragolution:
Tested and works as intended, so far as I can tell.

Ghommie (approved):

PR #11362, opened 2020-03-06
Fixes bartender drink slide spill B

### LINKED ISSUE #11342: Drink flinging removing liquids

Ragolution:
Forces spill/splash proc out of loop before it can splash_reagents. Refer to #11361 for different method.

Resolves #11342


thestubborn:
give me a drink, bartender

Ghommie:
The other one is better.

PR #11364, opened 2020-03-06
Fixes more xenobio GC memes

silicons:
mob.LAssailant is now a weak reference to prevent hard references from preventing GC.


code/modules/mob/living/living.dm
```diff
@@ -295,7 +295,7 @@
 		if(!iscarbon(src))
 			M.LAssailant = null
 		else
-			M.LAssailant = usr
+			M.LAssailant = WEAKREF(src)
```
Ghommie:
```suggestion
			M.LAssailant = WEAKREF(usr)
```

code/modules/mob/living/simple_animal/slime/life.dm
```diff
@@ -186,12 +186,13 @@
 	if(M.stat == DEAD) // our victim died
 		if(!client)
 			if(!rabid && !attacked)
-				if(M.LAssailant && M.LAssailant != M)
+				var/mob/living/carbon/their_attacker = M.getLAssailant()
```
Ghommie, line 189: Any problem if we call the weakref `Resolve` here without the need of this proc?

Ghommie (changes requested):

code/modules/mob/living/simple_animal/slime/life.dm
silicons: not really it's more convenience than need and a built in null check in that proc

Ghommie (approved):
Merge conflicts.

PR #11375, opened 2020-03-07
Kills screen-wide EMPs

Auris456852:
## About The Pull Request

This makes a small line change to make the EMP reaction's size be determined logarithmically, rather than linearly. That meaning, there's diminishing returns on EMPs. As a consequence, there's not TOO much reason to try to make a very big EMP, like with maxcaps before toxins became research... Though I'm sure some of you, who like to get the largest numbers on things possible, can still reach some ungodly numbers with payloads.

## Why It's Good For The Game

Because you won't be able to EMP the whole station with one grenade anymore. You should, of course, be welcome to reach for new heights, but you shouldn't be able to lolEMP the entire station. Even if it costs a fair deal of uranium, a good miner can get 15 sheets in no time, assuming the ORM is upgraded in a timely manner.

## Changelog
:cl:
balance: Made EMP reaction growth logarithmic.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
+++ b/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
@@ -114,9 +114,12 @@
 
 /datum/chemical_reaction/emp_pulse/on_reaction(datum/reagents/holder, multiplier)
 	var/location = get_turf(holder.my_atom)
-	// 100 multiplier = 4 heavy range & 7 light range. A few tiles smaller than traitor EMP grandes.
-	// 200 multiplier = used to be 8 heavy range & 14 light range. 4 tiles larger than traitor EMP grenades. Shouldn't be all that much higher now.
-	empulse(location, round(log(2.25, multiplier)), round(log(1.75, multiplier)), 1)
+	// 50 multiplier = 4 heavy range & 7 light range. A few tiles smaller than traitor EMP grandes.
+	// 100 multiplier = 5 heavy range & 10 light range.
+	// 200 multiplier = 7 heavy range & 14 light range. 4 tiles larger than traitor EMP grenades.
+	// 300 multiplier = 8 heavy range & 17 light range. Still rather significant, considering that you can get dozens of bluespace beakers 30 minutes in with a competent crew.
+	// 900 multiplier = 12 heavy range & 30 light range. Still less than 300 before this commit.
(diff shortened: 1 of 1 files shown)


monster860 (commented):
for 200 multiplier its 6 heavy and 9 light.

image

shellspeed1:
It might be worth tweaking emps so that after a certain point they get much stronger and do more damage to exos and borgs. As it is, emps which are supposed to be the counter to exosuits are well. junk at best especially with upgraded cells and parts.

silicons:
hot take

I actually like large emps 
Why not make a system for more granular EMP power rather than a hard log cap?

Auris456852:
Theoretically a sound plan, but we'd likely need to change the empulse function in its entirety for that... Not that I looked into its inner workings. It's likely possible to scale EMP damage with the size of it, but unless we gave a buff to all EMP-based weapons in general, which is a delicate balance I would rather be careful with, it would only make chem-made EMPs notably better against silicons, as the others all call the empulse function straight up, with no changes in size. This only changes the calculation on the chem reaction.
If I had to make a suggestion on that front, which I may get on in the near or far future, just make the heavy part drain a percentage of their cell's max power to deal with the borgs and exos? Constant + percentage if you want it to counter slime cell ones better, too, which have little max power, but regenerate... But that'd be a bit too much methinks. @shellspeed1 

~~i either don't have enough math skill for that or enough knowledge in english to know exactly what you mean by granular @kevinz000~~

Ghommie:
Logarithms alone make the growth curve a bit too drastic here, the gain of a 480 units reaction here is not even the double of the 48 units one. Feel free to use different functions, but personally I'd say a rational exponent is enough.

Ghommie (changes requested):
The logarithms alone make the growth curves a bit too dramatic here, the gain of a 480 units reaction here is not even the double of the 48 units one. Feel free to solve the issue with different functions of your choice.

Auris456852:
Going over this, it appears I may have miscalculated things previously as well...
And the comment had lied to me too, though I should've been able to tell that one sooner - through some debugging (making it use multiplier alone for the EMPulse), I found that a 50/50 iron/uranium grenade has the multiplier of 50. However, that very same grenade previously had the range of the 100 multiplier listed in the comment. As one would expect when trying to use exponents or logarithms, this changes things a bit. I changed it to work progressively worse past the 50 multiplier point instead.

Ghommie (approved):

PR #11409, opened 2020-03-10
Particle Defender no longer just the HoS taser but a shotgun/Stimulants protect against the tased effect.

necromanceranne:
## About The Pull Request

1. People apparently didn't notice that the particle defender is a child of the main electrode. So it's been, per pellet, stronger than every other taser and doing 40 damage per pellet + the strong taser effect. It also had a fucking 7 tile range so it could be used to tase an entire screen of people. It's been like this for a good long while. **_What the ever loving fuck._**

2. Adrenals were not combating the tased effect so even IF you did get tased, you still had the slowdown as well as the damage over time. Now stimulants grant you protection from that effect. It's almost like spending 8 TC on an anti-stun should do something.

This doesn't help you against knockdowns or stun batons.

## Why It's Good For The Game

I fucking hate this gun. But even in my spite filled state I cannot bring myself to remove it because I like the absurdity of a pump action energy gun. Also it makes cool noises like BWOOSH every time I shoot it and it activates my catbrain neurons.

I also want to spend 8 tc on adrenals and know they work. That's a nice thing I think.

## Changelog
:cl:
balance: Particle Defender is now a much more sane weapon.
balance: Tasers are no longer ignoring stimulants.
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -128,6 +128,7 @@
 
 /datum/status_effect/electrode
 	id = "tased"
+	alert_type = null
 	var/slowdown = 1.5
 	var/slowdown_priority = 50		//to make sure the stronger effect overrides
 	var/affect_crawl = FALSE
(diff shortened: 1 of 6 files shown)


code/datums/status_effects/debuffs.dm
```diff
@@ -155,7 +156,11 @@
 /datum/status_effect/electrode/tick()
 	var/diff = world.time - last_tick
 	if(owner)
-		owner.adjustStaminaLoss(max(0, stamdmg_per_ds * diff)) //if you really want to try to stamcrit someone with a taser alone, you can, but it'll take time and good timing.
+		var/mob/living/carbon/C = owner
+		if(HAS_TRAIT(C, TRAIT_TASED_RESISTANCE))
```
silicons, line 160: not the best way to do this but since it ticks every second it isn't that horrible i guess..
actually wait yeah it slows down, it's much better to just check it on apply and have the trait add wipe any tased effect when it's added.

necromanceranne: Alright cool beans.

silicons:
"But even in my spite filled state I cannot bring myself to remove it because I like the absurdity of a pump action energy gun."

i'm going to be honest that sums up my own thoughts about the literal oldtaser shotgun very, very well

Ghommie (approved):

PR #11430, opened 2020-03-11
Tweaks for the transfer vote

Putnam3145:
## About The Pull Request

1. Makes the autotransfer vote have a proper user ("server") like the rest of the server.
2. Doesn't make the autotransfer vote make no-votes count as not just votes to transfer, but *more votes than actually voting to transfer*.

## Why It's Good For The Game

Consistency and that second decision just seems super questionable to me.

## Changelog
:cl:
tweak: Autotransfer vote now requires actual transfer votes to transfer.
/:cl:

### DIFF
--- a/code/controllers/subsystem/autotransfer.dm
+++ b/code/controllers/subsystem/autotransfer.dm
@@ -20,7 +20,7 @@ SUBSYSTEM_DEF(autotransfer)
 /datum/controller/subsystem/autotransfer/fire()
 	if(maxvotes > curvotes)
 		if(world.time > targettime)
-			SSvote.initiate_vote("transfer",null) //TODO figure out how to not use null as the user
+			SSvote.initiate_vote("transfer","server")
 			targettime = targettime + voteinterval
 			curvotes += 1
 	else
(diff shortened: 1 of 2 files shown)


Ghommie (approved):

SynnGraffkin:
@Putnam3145 You can probably do what I did (with the help of others) on Hyper when I ported it from Austation and just made non-votes = 1 vote transfer. I had to do it that way because the whole factor thing was funky in general and works fine on TG code but didn't work fine on Hyper which is just an unofficial Cit 2 server. Or you can just change the factor numbers in general so they cap out at 1. I believe the original reasoning behind it eventually going over 1 was to have it where rounds eventually have to end, but that is just speculation. 

image

Putnam3145:
I know the main objection I put forth there was thinking that no-votes count more than *actual* votes toward one of the options is questionable, but another reason is that I just sorta think that no-votes counting *at all* is questionable. A no-vote is a "i don't feel strongly one way or the other" thing, but it silently makes no-votes actually votes for leaving. I object to that.

Ghommie (approved):

PR #11438, opened 2020-03-12
nuclear fixes the blindness issue by making eye_blind() max() rather than add

silicons:
## About The Pull Request

similar to how status effects work, there's no reason this proc should be stacking infinitely anyways, i wish all of this were status effects tbh

## Why It's Good For The Game

blindness is bad and this'll only keep happening, yeah i know maybe update stat shouldn't be called multiple times a tick but until someone goes and fixes the entire codebase to not require so many calls it's safer than sorry.

## Changelog
:cl:
experimental: eye_blind now is like Stun() and similar basic status effect procs, and only max()es the blindness duration to the amount rather than adding
/:cl:

### DIFF
--- a/code/modules/mob/status_procs.dm
+++ b/code/modules/mob/status_procs.dm
@@ -19,14 +19,10 @@
 /mob/proc/set_dizziness(amount)
 	dizziness = max(amount, 0)
 
-///Blind a mobs eyes by amount
-/mob/proc/blind_eyes(amount)
-	blindness(amount)
-
 /**
   * Sets a mob's blindness to an amount if it was not above it already, similar to how status effects work 
   */
-/mob/proc/blindness(amount)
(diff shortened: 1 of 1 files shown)


code/modules/mob/status_procs.dm
```diff
@@ -21,7 +21,17 @@
 
 ///Blind a mobs eyes by amount
 /mob/proc/blind_eyes(amount)
-	adjust_blindness(amount)
+	blindness(amount)
+
+/**
+  * Sets a mob's blindness to an amount if it was not above it already, similar to how status effects work 
+  */
+/mob/proc/blindness(amount)
```
Ghommie: If it is only used in one place, five lines above, why should it have its own proc?

Ghommie (approved):

PR #11439, opened 2020-03-12
[WIP] Arcade Component.

Ghommie:
## About The Pull Request 
Title, so far only the battle arcade has been done. Hopefully it won't be a Four Dimensional Frankenstein like the virtual reality component was, but I suspect they are both equally wacky.

## Why It's Good For The Game
I have plans to implement those arcade games somewhere else as part of an overhaul.

## Changelog
None.


Ghommie:
Busy doing other things atm, will be back at this later.

PR #11444, opened 2020-03-12
Rebalances the PA Rifle

keronshb:
## About The Pull Request

This shouldn't be the always go to for blobs

## Why It's Good For The Game

Blobs have a lot to worry about, but this takes the cake on being the easiest thing to get to fight the blob.  It requires no effort to get and the wielder(s) are in a safe distance away from any nauts or the blob itself.  It removes too many tiles off of the blob at once and it has absolutely zero chance of fighting back.

## Changelog
:cl:
balance: Rebalances the PA Rifle
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/misc/beam_rifle.dm
+++ b/code/modules/projectiles/guns/misc/beam_rifle.dm
@@ -58,7 +58,7 @@
 	var/aoe_mob_range = 1
 	var/aoe_mob_damage = 20
 	var/impact_structure_damage = 20
-	var/projectile_damage = 20
+	var/projectile_damage = 40
 	var/projectile_stun = 0
 	var/projectile_setting_pierce = TRUE
 	var/delay = 30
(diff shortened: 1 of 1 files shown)


ghost:
for blob this only needs nerfs to structural damage not to mob/direct damage and the aoe fire (which doesn't hurt bloobs)

actually hitting a person with pa is a legendary feat

keronshb:
> 
> 
> for blob this only needs nerfs to structural damage not to mob/direct damage and the aoe fire (which doesn't hurt bloobs)
> 
> actually hitting a person with pa is a legendary feat

Good thing to keep in mind.  I updated the commit, does that look better?

silicons:
No because this nerfs it all around 
If this is meant to target blobs there's already a proc that's meant to reduce damage against certain structures
Also go fix the reflections if you can

That said feel free to knock the shot count down to 5 instead of 10.

keronshb:
Alright @kevinz000 does this look better?

I'm having a hard time finding that proc you mentioned.  This way it keeps the damage VS mobs, but it loses the amazing structural damage VS Blobs and has 5 uses instead of 10.

silicons:
@keronshb imo all that needs to happen is:

1. reflectable gets set to yes 
2. 5 shots instead of 10
3. remove piercing because i've realized this is kind of a meme
4. /obj/item/projectile/beam/beam_rifle/proc/get_damage_coeff(atom/target) look at the blob part change 0.65 to 0.4

there's no need for nerf to anything else, they're already a niche weapon barely worth using except in specific cheese cases and removing the piercing solves that, set all the other values back.

keronshb:
> 
> 
> @keronshb imo all that needs to happen is:
> 
>     1. reflectable gets set to yes
> 
>     2. 5 shots instead of 10
> 
>     3. remove piercing because i've realized this is kind of a meme
> 
>     4. /obj/item/projectile/beam/beam_rifle/proc/get_damage_coeff(atom/target) look at the blob part change 0.65 to 0.4
> 
> 
> there's no need for nerf to anything else, they're already a niche weapon barely worth using except in specific cheese cases and removing the piercing solves that, set all the other values back.

All done @kevinz000 BUT there's an issue I tagged you w/ on discord.  I'll repeat it here:

1) When I aim at the blob it reflects /the aiming beam/ now and I didn't even put the reflect under the aiming beam.  It spams the chat with "The aiming beam has been reflected off of the blob"

2) Every like... 5 seconds there's a 1/2 second where it WONT reflect the beam.

PR #11476, opened 2020-03-15
Completely overhauls how blocking, shields, reflecting, and anything in between is handled. Has some small gameplay effects.

silicons:
## About The Pull Request
attack_type is now a bitflag for more granular control.
*Charger stands now have a leap attack flag rather than thrown object, much like xenos.*

check_shields() and check_reflect() and the matching hit_reaction() and IsReflect() on mobs and objects are dying for the new check_block() and run_block() procs on mobs. Those are wrappers for do_run_block(), which call a component signal and call run_block() on items. All return values are |'d OR'd with bit operators, unless something interrupts the chain (like if RTA fully teleports you).

check_block() just virtual-checks if and how someone would blocks something without effects. run_block() actually does the full "sequence". This is implemented by do_run_block() having its first argument be FALSE if it's only a check and TRUE if it's a real attack.

Returning a nonzero value no longer automatically blocks damage. The procs now have to return BLOCK_SUCCESS to fully block damage/etc. Might need a percentage based support system later, but for now, this will work.

The resulting return is now a bitfield, supporting things like SHOULD_REFLECT, SHOULD_PASSTHROUGH and some technical ones for code handling.

IsReflect() and stuff is being moved into the run block, things will typecheck for reflectable energy projectiles. While more "annoying" for the coder, I believe this is going to be better in the long run for code modularity/how easy it is to implement complex blocking interactions/effects (which I will need later).

- [x] WIP - Modify projectiles to automatically handle certain flags instead of things like sleeping carp having to manually handle it.

- [x] WIP - Implement a priority system instead of static order for block checks on items/comsigs. For example, we might want shields to trigger before reactive armor.

- [x] WIP - Proof of concept changes to reactive armor as a demonstration of block flags. Might need to make this in another PR depending on how unbalanced it turns out to be :^)

## Why It's Good For The Game

More modular blocking system for more complicated effects later on, getting rid of 4 procs and making it into effectively 2 (as 2 are wrappers), and adds support for things like turrets/simplemobs (just as an example) checking if someone's shielded. 

Probably using this for the full on overhauling coming later after this is done.

## Changelog
:cl:
experimental: Completely reworked how shields, blocking, and similar are handled in code.
add: reactive teleport armor doesn't magically delete projectiles, they keep flying after blocked since you dodged it
add: anything that said "user deflects the attack" in general now actually deflects bullets instead of deleting them.
/:cl:


code/modules/mob/living/living_defense.dm
```diff
@@ -309,14 +294,14 @@
 			M.Feedstop()
 		return // can't attack while eating!
 
-	if(HAS_TRAIT(src, TRAIT_PACIFISM))
+	if(HAS_TRAIT(M, TRAIT_PACIFISM))
```
silicons, line 309: fixes a bug

code/game/objects/items/twohanded.dm
```diff
@@ -1180,7 +1182,7 @@
 	if(iscyborg(target))
 		..()
 		return
-	if(target.check_shields(src, 0, "[user]'s [name]", MELEE_ATTACK)) //No message; check_shields() handles that
+	if(target.run_block(src, 0, "[user]'s [name]", ATTACK_TYPE_MELEE, 0, user) & BLOCK_SUCCESS) //No message; check_shields() handles that
```
Linzolle: comment should technically be changed to run_block but thats super minor

Linzolle (approved):

PR #11478, opened 2020-03-15
[TESTMERGE ONLY] comments out wrongdirmovedelay

silicons:
Makes space/melee combat in general a PITA and nearly impossible if you're the one chasing, for testmerge something proper is done about this.


silicons:
@deathride58 please fix this this is seriously fucked

PR #11483, opened 2020-03-15
mining borg nerfs - "they should be outclassed lategame"

Hatterhat:
## About The Pull Request
- borg point card gone
- borg premium KA gone
## Why It's Good For The Game
widely agreed borg philosophy: "they should be good earlygame but outclassed lategame"
@BlackMajor: "what if I just didn't listen to that?"
## Changelog
:cl:
del: Mining cyborgs are no longer physically capable of claiming points nor wielding a premium accelerator.
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -173,6 +173,34 @@
 		R.module.basic_modules += S
 		R.module.add_module(S, FALSE, TRUE)
 
+
+/obj/item/borg/upgrade/advcutter
+	name = "mining cyborg advanced plasma cutter"
+	desc = "An upgrade for the mining cyborgs plasma cutter, bringing it to advanced operation."
+	icon_state = "cyborg_upgrade3"
+	require_module = 1
+	module_type = list(/obj/item/robot_module/miner)
+
+/obj/item/borg/upgrade/advcutter/action(mob/living/silicon/robot/R, user = usr)
+	. = ..()
+	if(.)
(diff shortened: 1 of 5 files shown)


DeltaFire15:
I don't believe that is the borg design philosophy. Its more of "Darn good at what they do but inflexible due to lack of hands and need for organics for most upgrades". Look at for example medborgs. They're good at what they do, and can be quite a good replacement for medstaff aswell when upgraded, but they get problems when encountering anything not prepared / equipped for, 
 as opposed to carbons who can adjust themselves to the situation. Therefore I don't really agree with the removal of the plasmacutter and PKA upgrades, however the crusher could probably be made an upgrade instead of a roundstart available tool due to it being quite strong.

silicons (changes requested):
let them keep the crusher, APC i'm pretty ambivalent on but premium accelerator needs to go.

necromanceranne:
Borgs are meant to be assets to the crew and specialized in their jobs, with equipment to help overcome their shortcomings due to a lack of modular gear without roboticist help.

I can understand removing the cutter and the premium KA, the former being basically a brute laser gun right now and the latter being powercreep bullshit. But the crusher is not actually that bad at all. It's about as useful as the standard PKA.

BlackMajor (approved):
Acceptable balancing.
Adding the Premium KA in the first place was overdoing it.
Point card existed with the intent on making a point bank later for it (which I never did. I'll add it back when I do)

silicons (approved):

PR #11492, opened 2020-03-16
Makes nymphomania no longer cause inviolable arousal

Putnam3145:
## About The Pull Request

See title. Nymphomania does nothing mechanically now.

## Why It's Good For The Game

The old thing was waaaay worse than just having it do nothing.

Why not remove it? People use it as a kind of weird "likes ERP" indicator. I think it's probably better to keep it but have it do nothing, at least so that people don't get mad.

## Changelog
:cl:
del: Nymphomania no longer has an effect.
/:cl:

### DIFF
--- a/code/datums/traits/neutral.dm
+++ b/code/datums/traits/neutral.dm
@@ -92,7 +92,6 @@
 	name = "Nymphomania"
 	desc = "You're always feeling a bit in heat. Also, you get aroused faster than usual."
 	value = 0
-	mob_trait = TRAIT_PERMABONER
 	gain_text = "<span class='notice'>You are feeling extra wild.</span>"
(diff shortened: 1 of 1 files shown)


PersianXerxes:
Perhaps the description should be changed, if you want it to be more of an "ERP flag"? Especially since the current description is based on its old mechanics, hah.

Ghommie:
Just remove it, the preference can already be stated in the flavor text and there is no special interaction for having this quirk at this point.

Ghommie (changes requested):
Just remove it (and make sure to do a savefile conversion or we'll get "undefined quirk" stack trace warnings for months to come), the preference can already be stated in the flavor text and there is no special interaction for having this quirk at this point.

PR #11507, opened 2020-03-16
Plastitanium Rapier: Sleepy Edition

necromanceranne:
## About The Pull Request

I felt like the rapier was struggling to find a good niche from just being a full armor piercing weapon. See, the issue with the rapier is that even if it was to go into a fight with it's main intended opponent, someone wearing full riot gear, it usually performed only a smidge above what and esword would in the same situation and then struggle against any other opponent. The only time it would be far more useful is if you faced against an opponent wearing deathsquad-level armor, in which case, you are likely in a far worse situation than whether or not you can puncture your opponents defenses.

This change hopefully gives it a far more utilitarian use compared to the esword. It effectively is a disabling sword. It can definitely kill the shit out of people and you absolutely can and should use it like this. But you might only want to subdue an opponent, in which case, this is a very useful tool as well.

It now boasts the ability to put you to sleep. With 30 stamina damage on hit a probability chance based on stamina loss to knock you clean out. It also completely bypasses block entirely since apparently giving it 200 armor pen just does that. Don't ask me, this was per kevin's suggestion and I thought it was neat.

## Why It's Good For The Game

It wasn't cutting it as an esword equivalent. Now it's a far more unique weapon. At least in my opinion.

Also, new sprites.

Screenshot 2020-03-17 06 32 16

Screenshot 2020-03-17 07 07 03

## Changelog
:cl:
balance: Altered the plastitanium rapier from a perfect AP but weaker esword equivalent to a more unique sleep inducing melee weapon.
add: Touched up all the relevant sprites.
/:cl:


Linzolle:
all around good changes imo, glad you added the sheating sound effect haha.

code/game/objects/items/melee/misc.dm
```diff
@@ -143,33 +143,64 @@
...
 
 /obj/item/melee/rapier/get_worn_belt_overlay(icon_file)
 	return mutable_appearance(icon_file, "-rapier")
 
+/obj/item/melee/rapier/attack(mob/living/target, mob/living/user)
+	. = ..()
+	if(iscarbon(target))
+		var/mob/living/carbon/H = target
+		var/loss = H.getStaminaLoss()
+		to_chat(H, "<span class='warning'>You feel extremely tired...</span>")
```
Ghommie: This is going to cause chat spam with every attack, are you sure about it?

code/game/objects/items/melee/misc.dm
```diff
@@ -143,33 +143,64 @@
...
+		H.drowsyness += 60
```
Ghommie: That's a lot of drowsyness, maybe you should put a cap on the amount.

code/game/objects/items/melee/misc.dm
```diff
@@ -143,33 +143,64 @@
...
+		H.Dizzy(10)
+		H.adjustStaminaLoss(30)
+		if((loss > 40) && prob(loss)) // if above 40, roll for sleep using 1% every 1 stamina damage
```
Ghommie, line 198: Am I wrong or is this likely to knock the daylights out of a carbon target only after 3 or 4 hits?

code/game/objects/items/melee/misc.dm
```diff
@@ -143,33 +143,64 @@
...
+			to_chat(H, "<span class='warning'>Your legs give out and you buckle over, unable to remain awake...</span>")
```
Ghommie: Is it intended for already sleeping targets to receive this message when repeatedly struck by the rapier?

Ludox235:
Can't wait to abuse the shit out of this as contractor

code/game/objects/items/melee/misc.dm
necromanceranne: I'll cut down on the chat messages as a whole but drowsiness doesn't really have very good message feedback at all. I found. (and no it isn't by the by)

code/game/objects/items/melee/misc.dm
necromanceranne: You're not wrong. An esword can also put you into actual crit in the same amount of hits and the only way to stop it is if you're wearing absurdly high armor. These are priced identically. Additionally, stun batons are 25 stamina loss. Both batons and this sword bypass armor. It's if anything on par with equivalent weaponry but obviously has an edge yet is very easily recognizable if you have it. It's huge, you have to wear it in a distinct purple sheath, you need to get into melee, and it cannot block bullets unlike the esword.

necromanceranne: As a note, an easy way to nerf it if it proves too much is to make the threshold to sleep 70, so it requires three hits at the least to sleep, but the change is meant to make you far stronger in a dual than your opponent.

necromanceranne:
Made some changes based on Ghom's concerns.

Ghommie (approved):

BobJoga:
IMHO i don't see how this is good. Why would a sword... A sharp thing, cause you to get knocked unconscious? I don't think this fits the item well at all.

PR #11508, opened 2020-03-17
micro optimization for atmos add_to_active/react_queue

silicons:
## About The Pull Request

|= into [] = 

## Why It's Good For The Game

slightly faster, more memory usage but who the fuck cares it won't be that bad
this is potentially called millions of times so eh, every bit counts right?

## Changelog
go away

### DIFF
--- a/code/controllers/subsystem/air.dm
+++ b/code/controllers/subsystem/air.dm
@@ -275,7 +275,7 @@ SUBSYSTEM_DEF(air)
 	if(istype(T) && T.air)
 		turf_react_queue[T] = TRUE
 		if(currentpart == SSAIR_REACTQUEUE)
-			currentrun[T] = T
+			currentrun[T] = TRUE
 
 /datum/controller/subsystem/air/proc/remove_from_react_queue(turf/open/T)
 	turf_react_queue -= T
--- a/code/controllers/subsystem/air.dm
+++ b/code/controllers/subsystem/air.dm
@@ -273,7 +273,7 @@ SUBSYSTEM_DEF(air)
 
 /datum/controller/subsystem/air/proc/add_to_react_queue(turf/open/T)
 	if(istype(T) && T.air)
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/air.dm
```diff
@@ -274,10 +273,9 @@ SUBSYSTEM_DEF(air)
 
 /datum/controller/subsystem/air/proc/add_to_react_queue(turf/open/T)
 	if(istype(T) && T.air)
-		turf_react_queue |= T
+		turf_react_queue[T] = T
```
Ghommie:
```suggestion
		turf_react_queue[T] = TRUE
```

code/controllers/subsystem/air.dm
```diff
@@ -274,10 +273,9 @@ SUBSYSTEM_DEF(air)
...
 		if(currentpart == SSAIR_REACTQUEUE)
-			currentrun |= T
-	return
+			currentrun[T] = T
```
Ghommie:
```suggestion
			currentrun[T] = TRUE
```

code/controllers/subsystem/air.dm
```diff
@@ -257,8 +257,7 @@ SUBSYSTEM_DEF(air)
 		T.add_atom_colour("#00ff00", TEMPORARY_COLOUR_PRIORITY)
 		#endif
 		T.excited = TRUE
-		active_turfs |= T
-		SSair_turfs.currentrun |= T
+		active_turfs[T] = SSair_turfs.currentrun[T] = T
```
Ghommie:
```suggestion
		active_turfs[T] = SSair_turfs.currentrun[T] = TRUE
```

Ghommie (changes requested):
These imply associative checks aren't done.

Ghommie (approved):

PR #11515, opened 2020-03-18
A dynamic rework: target threat levels

Putnam3145:
## About The Pull Request

This PR adjusts dynamic to more accommodate storytellers and indefinitely-long rounds by changing what threat and threat level mean. Instead of being "points left to spend for antags" and "points at the start of the round", threat level means "threat that the mode tries to stay at" and threat means "amount of threat posed by enemies to the station".

## Why It's Good For The Game

TG-dynamic isn't really designed for rounds as long as ours, and storytellers are too samey. This solves both problems, or at least is the framework to do such.

I think dynamic as it is right now is fine, actually; but this might be a better solution, and the code now exists, which is the real important thing. Even if this is rejected (and it should be, if the terrible error it consistently has persists), then the work I've done is still around, so that's all good.

## Changelog
:cl:
tweak: Dynamic reworked to be more "storyteller-like": threat is now how threatened the station is right now and threat level is how much the mode wants the station to be threatened.
tweak: Random storyteller is now truly random.
tweak: Story storyteller now gives a nice rising-action-climax-falling action curve for threat level.
tweak: "Chaotic" storyteller now simply ramps up threat level as round goes on.
add: "Classic" storyteller, basically doing what "random" did before.
/:cl:


code/game/gamemodes/dynamic/dynamic_rulesets_roundstart.dm
```diff
@@ -802,7 +802,6 @@
 /datum/dynamic_ruleset/roundstart/bloodsucker
 	name = "Bloodsuckers"
 	config_tag = "bloodsucker"
-	persistent = TRUE
```
Putnam3145, line 805: Ah, I didn't comment on the commit message here. This straight up does nothing. but make it (uselessly!) process with the default processing.

Arturlang: Good to know, gamemodes have an awfully lot of ways that would seem to cause them to end on the antag death, so i wasnt really sure which one it was.

code/game/gamemodes/dynamic/dynamic_storytellers.dm
```diff
@@ -39,6 +39,23 @@ Property weights are:
 	var/event_injection_cooldown_middle = 0.5*(GLOB.dynamic_event_delay_max + GLOB.dynamic_event_delay_min)
 	mode.event_injection_cooldown = (round(CLAMP(EXP_DISTRIBUTION(event_injection_cooldown_middle), GLOB.dynamic_event_delay_min, GLOB.dynamic_event_delay_max)) + world.time)
 
+/datum/dynamic_storyteller/proc/calculate_threat()
+	var/threat = 0
+	for(var/datum/antagonist/A in GLOB.antagonists)
```
Putnam3145, line 45: (it's like this everywhere else in the code and i have no idea if it's absolutely necessary but didn't want to risk it--this vs var/a in GLOB.antagonists and then typecasting the next line)

code/game/gamemodes/dynamic/dynamic_storytellers.dm
```diff
@@ -39,6 +40,24 @@ Property weights are:
...
+		if(A && A.owner && A.owner.current && A.owner.current.stat != DEAD)
```
Ghommie:
```suggestion
		if(A?.owner?.owner.current && A.owner.current.stat != DEAD)
```

code/modules/antagonists/devil/devil.dm
```diff
@@ -91,6 +91,7 @@ GLOBAL_LIST_INIT(devil_suffix, list(" the Red", " the Soulless", " the Master",
 	job_rank = ROLE_DEVIL
 	//Don't delete upon mind destruction, otherwise soul re-selling will break.
 	delete_on_mind_deletion = FALSE
+	threat = 30
```
Ghommie: Devils are only that threatening when ascended.

code/modules/events/meteor_wave.dm
```diff
@@ -16,6 +16,7 @@
 	startWhen		= 6
 	endWhen			= 66
 	announceWhen	= 1
+	threat = 15
```
Ghommie, line 19: Considering meteor shields, unidirectional waves and on-station asteroids ruins, the threat should be slightly lower.

code/modules/mob/living/simple_animal/hostile/mushroom.dm
```diff
@@ -12,6 +12,7 @@
 	response_help  = "pets"
 	response_disarm = "gently pushes aside"
 	response_harm   = "whacks"
+	threat = 0.3
```
Ghommie: All these critters do is attack each other in a highlander battle royale, and considering dozens of dozens of these shrooms can be spawned with hydroponics, this is a bit exploitable.

Ghommie (changes requested):
Guess it should be testmerged first.

code/modules/events/meteor_wave.dm
Putnam3145: This is only calculated during the meteor storm itself, mind. Basically the purpose of this threat increase is to allow the crew to focus on meteors for a bit, it doesn't represent the damage to be done as much as how much the crew needs to focus on it during.

code/modules/antagonists/devil/devil.dm
Putnam3145: No excuse for not including scaling since I did for vampires, yeah.

Ghommie (approved):

PR #11520, opened 2020-03-18
stops validhunting geneticists from knowing exactly who is an is a vampire (and allows vampires genes)

Seris02:
## About The Pull Request

can't have one without the other really

## Why It's Good For The Game

because kevinz told me to, he should know the reason why

## Changelog
:cl:
balance: bloodsuckers being unable to accept genes
/:cl:


Arturlang:
Earlier, bloodsuckers had the no DNA trait, so i had removed that, thinking it would fix it, thanks for getting to this before i did.

Ghommie (approved):

PR #11539, opened 2020-03-20
Rewrote waffle traitor flavor text

Putnam3145:
## About The Pull Request

Makes it more clear that the common freeform objective is not a free murderbone objective.

## Why It's Good For The Game

People were using it as an excuse to murderbone when, really, only tiger co is an excuse to murderbone. Murderbone's not against our rules, but we do explicitly say it's bad sportsmanship, so replying to "you can do anything you want" with "then i'll murder everybody" should not be encouraged.

## Changelog
:cl:
tweak: Waffle co objective rewritten to make more clear it's not a murderbone objective
/:cl:


Ghommie (approved):
Ehhh, fine.

PR #11551, opened 2020-03-20
Precision Weapons: 360 Noscope Edition

necromanceranne:
## About The Pull Request

Some weapons now have perfect accuracy even fired from the hip and turning in combat mode. Why? Because most of them are supposed to be used running away, in a chase, or at extreme ranges, or drawn and fired from the hip very quickly. While this definitely might not be the gritty, intense realism of an imprecise shooter, most of these weapons were practically nonfunctional unless fired point blank.

Except the mag weapons. Those are precision weapons BECAUSE THEY'RE A RAIL GUN. Also now have something differentiating them from the WT550 I guess. Hipfiring your mag pistol like the show pony you are.

Before anyone panics, this does not affect dual-wielding. Just turn inaccuracy and being out of combat mode.

## Why It's Good For The Game

180 bullet deviation on weapons with one shot that are meant to be used while chasing people is not good and most of these guns were victims of a system they were never designed to fit into. 

The sniper rifle in particular had some weird issues with inaccuracy and using it's scope functionality due to huge range the mouse would be moving across, so sniper rifles were mostly being ignored for the anti-tank pistol or used point blank.

## Changelog
:cl:
Balance: Some firearms are now far better when fired from the hip or while on the move in combat mode.
/:cl:


Ghommie (changes requested):
There is no need to add a new flag, I already added a `inaccuracy_modifier` variable you can set to 0 to achieve the same results a long time ago.

necromanceranne:
Fair enough, I'll get to fixing that then.

necromanceranne:
I'm gonna close this and reopen it with the value changes instead of re-editing everything.

PR #11554, opened 2020-03-20
absolutely ensures that screenshake can never, ever cause sleep() style latency in attacks

silicons:
not sure if animate sleep()s but i've been noticing that melee/ranged attacks sometime take their dear time to process, let's uh, make sure that won't happen.


Ghommie (approved):
¯\\\_(ツ)_/¯

PR #11555, opened 2020-03-20
Nerfs piggyback and fireman carrying

silicons:
## About The Pull Request

You can no longer use items while fireman carrying.
You are now slowed by 1 while piggyback carrying.
In piggybacking, the rider is dazed for 1 second after unbuckling (no item usage or pickup).
Being forcefully knocked off now dazes for 4 seconds in addition to the knockdown in both cases.

## Why It's Good For The Game

It's pretty powerful enough to have a meatshield already, having them also able to melee is terrible, and piggybacking is already better than fireman for meatshielding because it doesn't take a hand.
This has been a long time coming.

## Changelog
:cl:
balance: Fireman carrying now makes the person being carried unable to use items. Piggybacking now slows down the person being ridden. In both cases, the person riding will be dazed when forcefully dismounted, and dazed for a second if dismounting from piggybacking
/:cl:

### WARNING THIS IS SHITCODE AND RIDING DATUMS ARE NOT GOOD EXAMPLES OF HOW TO CODE AND THIS IS MY FAULT THIS SHOULD PROABLBY BE IMRPOVED IN THE FUTURE


Linzolle:
seems reasonable to me

code/datums/components/riding.dm
```diff
@@ -27,6 +27,15 @@
 	RegisterSignal(parent, COMSIG_MOVABLE_UNBUCKLE, .proc/vehicle_mob_unbuckle)
 	RegisterSignal(parent, COMSIG_MOVABLE_MOVED, .proc/vehicle_moved)
 
+/datum/component/riding/Destroy()
+	var/mob/M = parent
+	if(!istype(M))		//we absolutely can not runtime.
+		stack_trace("Warning: parent was not a human on [type]/Destroy().")
```
Ghommie: When the parent is deleted, its components are deleted too. This will end up in useless stack traces every time a vehicle is deleted.

Ghommie (changes requested):

Ghommie (approved):

PR #11572, opened 2020-03-22
Clockie Vanguard Spell - rebalance for stamcombat

DeltaFire15:
## About The Pull Request

This modifies the vanguard spell for clock cultists. Before, it only sucked up stuns which we have nearly completely removed and set stamloss to 0 once at the very beginning of it. Now, it doesn't set stamloss to 0 immediately, but instead regenerates 6(!, inform me if you think this is too much but its supposed to make stamina / stun weaponry pretty much ineffective for its duration and just fuck them over when it ends) stamina per tick (30 per second), but causes 50% of staminaloss it healed to hit the affected person when it runs out. Fun fact, the description of vanguard told people it sped up stamina regen, yet it only did the aforementioned 'set to 0' at the beginning. This should make it a viable tool in our stamcombat system.

## Why It's Good For The Game
Currently, vanguard, a spell supposedly quite important for combat, is very ineffective due to it only setting stamloss to 0 when its casted, and the stunsucc effect being quite useless with our combat system. Therefore, this aims to expand vanguards 'fight now, ignore stuns for now, backfire later if you did get hit and still fight at that time' from just stuns to stamcombat aswell, with its constant stamregen making it a effective tool in a clock cultists arsenal, but having the same disadvantage as for stuns (partial backfire on spellend).

## Changelog
:cl:
balance: clockie vanguard now quickly regenerates stam while active (as its description always told you,)
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -82,6 +82,7 @@
 
 /datum/status_effect/vanguard_shield/tick()
 	progbar.update(duration - world.time)
+	owner.adjustStaminaLoss(-2)
 
 /datum/status_effect/vanguard_shield/on_remove()
 	var/vanguard = owner.stun_absorption["vanguard"]
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -48,20 +48,24 @@
 	status_type = STATUS_EFFECT_REPLACE
 	alert_type = /obj/screen/alert/status_effect/vanguard
 	var/datum/progressbar/progbar
+	var/stamhealed = 0 //How much stamina did we regenerate?
 
 /obj/screen/alert/status_effect/vanguard
 	name = "Vanguard"
-	desc = "You're absorbing stuns! Your stamina is greatly increased, but not infinite. 25% of all stuns taken will affect you after this effect ends."
+	desc = "You're absorbing stuns aswell as quickly regenerating stamina, but be careful: 50% of stamina restored and 25% of stuns absorbed will affect you after this effect ends."
 	icon_state = "vanguard"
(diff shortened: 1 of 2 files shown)


code/datums/status_effects/buffs.dm
```diff
@@ -72,16 +76,18 @@
...
-	owner.setStaminaLoss(0, FALSE)
 	progbar = new(owner, duration, owner)
 	progbar.bar.color = list("#FAE48C", "#FAE48C", "#FAE48C", rgb(0,0,0))
 	progbar.update(duration - world.time)
 	return ..()
 
 /datum/status_effect/vanguard_shield/tick()
 	progbar.update(duration - world.time)
+	var/oldstamloss = owner.getStaminaLoss()
+	owner.adjustStaminaLoss(-6) //up to 30 stam / second for now, lets see...
```
ghost, line 89: this is 3/second

ghost:
life ticks are once every 2 seconds
better idea may be to give them physiology stamina immunity
https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/code/modules/mob/living/carbon/human/physiology.dm
this only works on human subtypes but only human subtypes have stamina so

DeltaFire15:
Localtested it repeatedly, it heals pretty exactly 30 stam / second.

code/datums/status_effects/buffs.dm
DeltaFire15: Runs on status effect ticks (0.2 seconds I believe), not life ticks, it just calls a value adjustment proc.

silicons (approved):
life is every second now because 2 seconds was terrible and someone finally got smart and replaced it some time ago
status effect ticks every second by default

PR #11583, opened 2020-03-22
Tiger Carp migrations (New Random Event and Mob)

Ignari-contrib:
Space is filled with various little monsters out to get you, well whats one more? Introducing the new Tiger Carps! With a new color scheme these little terrors can appear during their own carp migration event. They are more aggressive than the plain space carps, yet not as tough as the megacarps. Adds a new colorful entity to the infinite black outside the station.

Adds: New Tiger Carp Mob and Carp Migration event
Tiger Carps: Set to 50 Health, deal 17 damage on bite, has a new color scheme and does 60 Object damage per hit.


Putnam3145 (approved):

Ghommie (changes requested):
With a few adjustments, these event datums should be made into a subtype of `round_event_control/carp_migration` and `round_event/carp_migration`, because this is bad copypsta.

PR #11585, opened 2020-03-23
does the quirky thing where a beam with origin/target in a container doesn't cut the map in half because byond is quirky and things inside things are x/y/z=0

silicons:
## Why It's Good For The Game

this is why you have beams going across the map
that's Bad(tm)

## Changelog
:cl:
fix: beams should no longer go across the map and mess everything up if their source or target isn't on a turf.
/:cl:

### DIFF
--- a/code/datums/beam.dm
+++ b/code/datums/beam.dm
@@ -92,7 +92,7 @@
 	return ..()
 
 /datum/beam/proc/Draw()
-	if(!get_turf && (!isturf(target.loc) || !isturf(source.loc)))		//byond is quirky! if we aren't getting turf tell it to just Not(tm)
+	if(!get_turf && (!isturf(src.target.loc) || !isturf(src.source.loc)))		//byond is quirky! if we aren't getting turf tell it to just Not(tm)
 		return
 	var/atom/origin = get_turf? get_turf(src.origin) : src.origin
 	var/atom/target = get_turf? get_turf(src.target) : src.target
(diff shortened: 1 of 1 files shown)


code/datums/beam.dm
```diff
@@ -90,6 +92,10 @@
 	return ..()
 
 /datum/beam/proc/Draw()
+	if(!get_turf && (!isturf(src.target.loc) || !isturf(src.origin.loc)))		//byond is quirky! if we aren't getting turf tell it to just Not(tm)
+		return
+	var/atom/origin = get_turf? get_turf(src.origin) : src.origin
+	var/atom/target = get_turf? get_turf(src.target) : src.target
```
Ghommie: This is probably detrimental and confusing, name these two variables in a different, not omonimous way and early return if one or another are null (which shouldn't be).

code/datums/beam.dm
```diff
@@ -90,6 +92,10 @@
 	return ..()
 
 /datum/beam/proc/Draw()
+	if(!get_turf && (!isturf(src.target.loc) || !isturf(src.origin.loc)))		//byond is quirky! if we aren't getting turf tell it to just Not(tm)
+		return
```
Ghommie: Remove these.

code/datums/beam.dm
```diff
@@ -15,6 +15,8 @@
 	var/beam_type = /obj/effect/ebeam //must be subtype
 	var/timing_id = null
 	var/recalculating = FALSE
+	/// If either origin or target is in a container, do get_turf instead of just not beaming.
+	var/get_turf = FALSE
```
Ghommie: This is a pain in the buns and might break a few beams.

Ghommie (changes requested):
TL;DR: just using the turf coordinates for drawing the beam.

Ghommie (approved):

PR #11596, opened 2020-03-24
[Balance Pass] - Stunbaton (Because no one wants to do a single line …

Feasel:
…change to make this not shit)

I'm putting it at 35 as Kevinz suggest, rather it be 49 but w/e.

The "Intuitive" system will be staying, but now it won't suck ass to actually be used as a stamina draining weapon.

### DIFF
--- a/code/game/objects/items/stunbaton.dm
+++ b/code/game/objects/items/stunbaton.dm
@@ -15,7 +15,7 @@
 	attack_verb = list("beaten")
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 50, "bio" = 0, "rad" = 0, "fire" = 80, "acid" = 80)
 
-	var/stamforce = 25
+	var/stamforce = 35
 	var/status = FALSE
 	var/knockdown = TRUE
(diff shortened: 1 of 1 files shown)


necromanceranne:
A weapon that can knockdown and slow (knockdown is a slow, force shoveup definitely a slow) and do as much damsge as a desword really doesn't feel like the best rebalance despite carbons having natural stamina regen. Mostly because in combat mode the victim of this weapon is likely not going to be regenning.

Damage doesn't really excuse stunbatons being such an all-rounder weapon, and frankly I don't find them shit at all. They are very strong actually so long as you use them correctly and remember that they are complimentary to your other security gear. Like tasers are to the disabler function.

silicons (commented):
i specifically did not want to touch this @Ghommie 
that said if i did this i'd set it to 30 instead of 35
yeah batons are in a .. weird spot, but i have nowhere to go here because it really seems like it's either too strong or two weak and 25 was on the weak side. 30 would make them 4 hits instead of 5, 35 would make them 3-4 hits. 
shrug.

silicons (approved):

PR #11611, opened 2020-03-25
Makes so turning in place updates riders when riding someone's back

monster860:
this PR is completely untested

## About The Pull Request

Makes so turning in place updates riders when riding someone's back

## Why It's Good For The Game

because you turn and the person riding you doesn't turn and it looks broken

### DIFF
--- a/code/datums/components/riding.dm
+++ b/code/datums/components/riding.dm
@@ -48,7 +48,7 @@
 /datum/component/riding/proc/set_vehicle_dir_layer(dir, layer)
 	directional_vehicle_layers["[dir]"] = layer
 
-/datum/component/riding/proc/vehicle_dir_changed(from, to)
+/datum/component/riding/proc/vehicle_dir_changed(from_dir, to_dir)
 	spawn
 		handle_vehicle_offsets()
 		handle_vehicle_layer()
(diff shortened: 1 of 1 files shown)


Arturlang:
Please... test it

monster860:
Testing is for the weak

Ghommie (approved):

IronEleven:
Not a fan, I and others use it to represent sitting on laps.

code/datums/components/riding.dm
```diff
@@ -47,6 +48,11 @@
 /datum/component/riding/proc/set_vehicle_dir_layer(dir, layer)
 	directional_vehicle_layers["[dir]"] = layer
 
+/datum/component/riding/proc/vehicle_dir_changed(from_dir, to_dir)
+	spawn
```
Ghommie, line 52: Wait, what the hell?

monster860: basically the signal fires *before* the dir is actually changed so uh this is necessary

Ghommie: Have you considered adding a "dir override" argument to both procs below and see if that's doable at least?

silicons: addtimer's better tbh  if you gotta use this

Linzolle:
I've gotta agree with IronEleven, it can be used for RP tbh.

monster860:
if you want to do lap-sitting you can use pixel-shifting and shizz. The default behavior for this kind of thing should be the one that *doesn't* look broken.

PR #11627, opened 2020-03-26
"Unofficial" Combat Rework Part 3: Blocking/Parrying 

silicons:
PR open for discussion and initial review, most of the variables are there but exact values aren't set. 
Includes/Requires #11476 

## About The Pull Request

Adds new a new directional blocking and parrying system bound to hotkeys G/F (yes, G block F parry) respectively.

- [ ] Combat mode mouse spinning needs to be slowed down a bit, not sure what to set the delay as yet.

- [x] Block/Parry sound effects and visual effects

**I'm not going to say specific numbers for now but they're all easily editable.**

Directional block:
Items with this function can be used to directionally block.
This usually incurs the following penalties:
The inability to attack during this
The inability to attack for about 0.4 seconds after this 
You being slowed down during this 
Constant, non-buffered (might make it buffered) small stamina drain from this.
Blocking has three damage "sections" in total.
Each item capable of this has "absorption" amount, a ratio for damage that exceeds that, and a final damage threshold in which any damage exceeding that is not impacted by blocking at all. 
A block is considered fully successful if all damage is blocked (so damage is equal to or lower than absorption).
All damage blocked (not incoming, but successfully blocked) incurs stamina damage on the user, split between the arm holding the blocking item and their chest at a certain ratio. This will happen even if you get a full block.

- [ ] Implement block on riot shields as a proof of concept.

Parrying: 
Items with this function can be used to parry. This is sort of like dark souls/whatever parrying but **nothing about this is hashed out yet, unlike block, because I still can't say for sure how well this'll end with the amount of lag people suffer here.**

Parrying gets a startup and winddown time which doesn't allow for the parry to trigger at all. During the active time a certain frame/second is considered the "perfect" parry time, and there's also a leeway on either side of it that makes it still considered perfect. Anything outside of perfect suffers from efficiency falloffs based on how much you missed it. Startup/winddown will not trigger this at all.

This isn't hashed out as of yet but this will proabbly operate off efficiency to reduce incoming damage, perfect parries would fully block the attack while less-than-perfect will reduce damage.

I am likely adding a hefty clickdelay post-parry to discourage spamming it every time you are able to. It'll also likely have a movespeed delay disadvantage and stamina cost but I will probably need to discuss this in great detail with @BlackMajor 

- [x] Successful parries will allow the user to reflex-counter the attack, depending on how it's set for the item, ranging from just blocking the attack (or partially blocking), to disarming the attacker, knocking down the attacker, or an automatic counterattack. This will be all customizable on the item.

- [ ] Implement parries on energy swords as a proof of concept

## Why It's Good For The Game

After a long time of people complaining about RNG blocking and several people including I believe @deathride58 and @BlackMajor speaking about a block/parry system I decided to make one.

RNG's bad, I'm hoping the two combined will make combat more.. varied, but that's yet to be determined.

I'm hoping this can be testmerged in around 1-2 weeks, if things go well. Currently working on the backend of the whole thing, probably doing an actual testable implementation afterwards.

## Changelog
:cl:
add: Directional blocking has been added, keybound to G. This will reduce a portion of incoming damage if done with eligible items at the cost of stamina damage incurred to the user based on damage blocked as well as while active, as well as in most cases preventing the user from doing any attacks while this is active.
add: Parrying has been added, keybound to F. Timing-based counterattacks, effect heavily dependent on the item, WIP.
/:cl:


code/modules/mob/living/living_blocking_parrying.dm
```diff
@@ -0,0 +1,139 @@
...
+			. = GLOB.block_parry_data["[type_or_id]"] = new type_or_id
+	else		//text id
+		return GLOB.block_parry_data["[type_or_id]"]
+
+/proc/set_block_parry_data(id, datum/block_parry_data/data)
+	if(ispath(id))
+		CRASH("Path-fetching of block parry data is only to grab static data, do not attempt to modify global caches of paths. Use string IDs.")
+	GLOB.block_parry_data["[id]"] = data
+
+/// Carries data like list data that would be a waste of memory if we initialized the list on every /item as we can cache datums easier.
```
Putnam3145, line 20: isn't this what elements are for,

silicons: can you imagine flooding SSdcs elements with something like armor datums 

'cause I can't =|

that said a lot of other globals probably belong in subsystems lmao

silicons: also it wouldn't work well

when I use ID systems it's with the intent of being able to **easily** runtime modify it. 
Yeah you CAN do that with elements but it's like.. why?

necromanceranne:
source (1)

necromanceranne:
Also kev I don't know if you noticed but I made limbs take 75% stamina damage instead of 50% and they have only 50 health. If a shield absorbs more than 70 damage in one hit it could straight up cripple the limb and basically not doing anything at all.

I'd highly recommend the stamina damage not be scaled but flat based on the attack weapons weight. Lighter weapons being easily blocked while heavier weapons being much harder, which makes weapons such as knives be slightly less useful for getting past an opponents blocks but things like hypereutactic blades and toolboxes (which are functionally makeshift maces or flails) basically batter people even though shieds.

This is to more easily replicate momentum and kinetic force behind weapons, give value even to weapons with heavy stamina costs for swinging them.

lolman360:
nerf to high ping players

silicons:
hey guess what we're done :joy:

code/game/objects/items/shields.dm
```diff
@@ -172,20 +196,7 @@
...
-		if(CHECK_BITFIELD(shield_flags, SHIELD_TRANSPARENT) && (AM.pass_flags & PASSGLASS))
-			return BLOCK_NONE
-	if(attack_type & ATTACK_TYPE_THROWN)
-		final_block_chance += 30
-	if(attack_type & ATTACK_TYPE_TACKLE)
-		final_block_chance = 100
-	. = ..()
-	if(. & BLOCK_SUCCESS)
-		on_shield_block(owner, object, damage, attack_text, attack_type, armour_penetration, attacker, def_zone, final_block_chance, block_return)
+	max_integrity = 450
```
silicons, line 204: i'm not sure i should be buffing it that much but the reasoning behind buffing integrity is: 

shields currently deduct the real damage of something, not blocked or post-blocked damage.
so say you block 200 lasers
logically you take 10 0damage and the shield takes 100
right now no, the shield takes 200 damage.

code/modules/mob/living/silicon/silicon_defense.dm
```diff
@@ -120,6 +120,7 @@
 	flash_act(affect_silicon = 1)
 
 /mob/living/silicon/bullet_act(obj/item/projectile/P, def_zone)
```
silicons, line 122: redid how stuff works here because the rng stuff was frankly Fucking Terrible so pointing this out before someone screeches balance change

silicons: notably: stuff is now knocked down from cyborgs consistently if a projectile has either 10 dmage on it or 20 stamina damage on it

code/modules/mob/mob.dm
```diff
@@ -1086,3 +1086,10 @@ GLOBAL_VAR_INIT(exploit_warn_spam_prevention, 0)
 	for(var/obj/item/I in held_items)
 		if(I.item_flags & SLOWS_WHILE_IN_HAND)
 			. += I.slowdown
+
+/**
+  * Mostly called by doUnEquip()
+  * Like item dropped() on mob side.
+  */
+/mob/proc/on_item_dropped(obj/item/I)
```
silicons, line 1094: is this a good idea
who knows

silicons: is it a good idea to just turn this into a component signal?

silicons:
OH, LIMITATION:
When adding block/parry, the coder has to manually add ITEM_CAN_BLOCK or ITEM_CAN_PARRY to item_flags on things!

I have considered making it automatic by having the flags on blcok parry datum and having init assign the flags if it's there buuut I don't particularly want to for overhead purposes so keep this in mind!

code/game/objects/items/twohanded.dm
```diff
@@ -298,6 +300,45 @@
...
+	parry_failed_clickcd_duration = CLICK_CD_MELEE
+	parry_cooldown = 3 SECONDS
+
+// KEV, WHY AREN'T YOU JUST PUTTING IT IN THE DEFINITION?!?!?!?!
+// I'LL TELL YOU WHY, BECAUSE BYOND DOESN'T CONSIDER "[(1<<0)]" A "CONSTANT EXPRESSION"
+// WHAT EVEN IS MORE CONSTANT THAN 1 BITSHIFTED TO THE LEFT BY 0 PLACES AS A STRING?
+/datum/block_parry_data/dual_esword/New()
+	// more efficient vs projectiles
+	block_stamina_efficiency_override = list(
+		"[ATTACK_TYPE_PROJECTILE]" = 4
```
silicons: someone tell me if i should just bite the bullet and add defines for text versions of attack types so we don't have to start the practice of list defining in New()
because both seem horrible but that just might be less horrible lol

code/game/objects/items/melee/misc.dm
```diff
@@ -191,10 +233,9 @@
 	. = ..()
 	if(iscarbon(target))
 		var/mob/living/carbon/H = target
-		var/loss = H.getStaminaLoss()
 		H.Dizzy(10)
 		H.adjustStaminaLoss(30)
-		if((loss > 40) && prob(loss)) // if above 40, roll for sleep using 1% every 1 stamina damage
+		if(CHECK_STAMCRIT(H) != NOT_STAMCRIT)
```
necromanceranne, line 238: Just gonna leave a little note here to say I asked kev to change this to what it is now, okies thankies -w-

Linzolle (approved):

PR #11648, opened 2020-03-27
one line fix to a box mapping issue

Ghommie:
## About The Pull Request
Evidence Storage airlock access, so detectives can access it.

## Why It's Good For The Game
This will close #11644.

## Changelog
None.


zeroisthebiggay:
doesnt this make it so only the detective can access it unless secoffs get skeleton access since four is detoffice

Putnam3145 (approved):

PR #11664, opened 2020-03-28
Stuff about power

Putnam3145:
## About The Pull Request

Overhauls how APCs work: they now drain directly from the power; if there's left over, they fill their cells; if the power they drain isn't sufficient, drains from the cell instead. Also changed priority: 1. environment 2. equipment 3. lighting 4. cell; before it was 0. cell (all power came from the cell) 1. environment 2. lighting 3. equipment.

## Why It's Good For The Game

It's a bit of a preview for stuff I want to do anyway with power.

## Changelog
:cl:
tweak: APCs now drain power directly before filling their cell, instead of doing this strange thing where it drains its cell then fills it back up every tick
balance: APCs turn off lights before equipment, now
/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -84,7 +84,7 @@
 	var/lastused_equip = 0
 	var/lastused_environ = 0
 	var/lastused_total = 0
-	var/main_status = 0
+	var/main_status = 0 // Whether or not there's external power. 0 is "none", 1 is "insufficient", 2 is "charging".
 	powernet = 0		// set so that APCs aren't found as powernet nodes //Hackish, Horrible, was like this before I changed it :(
 	var/malfhack = 0 //New var for my changes to AI malf. --NeoFite
 	var/mob/living/silicon/ai/malfai = null //See above --NeoFite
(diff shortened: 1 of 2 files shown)


code/modules/power/apc.dm
```diff
@@ -1271,114 +1271,111 @@
...
+	if(cell && !shorted)
+		if(cur_excess > 0)
+			var/charging_cell = min(cur_excess, cell.maxcharge * GLOB.CHARGELEVEL)
+			cell.give(charging_cell)
+			add_load(charging_cell)
+			lastused_total += charging_cell
+			longtermpower = min(10,longtermpower + 1)
+			if(chargemode && !charging)
+				chargecount++
+				if(chargecount == 10)
```
Ghommie, line 1323:
```suggestion
				if(++chargecount == 10)
```

Ghommie (approved):

code/modules/power/apc.dm
Putnam3145: Okay for once I'm actually going to say "no, that's disgusting". There's a chance that this might actually make it run faster because byond is not in any way sane, but putting a comparison and a ++ in the same line is, to be frank, fucking disastrous for readability and I'd summarily reject it in any other language. Hell, I'd remove ++ from this language if I could, I don't know why I used it here except that people would complain if I didn't.

Ghommie: You are right. I didn't find anything worth to note on that quick review so I just dropped that stupid suggestion.

PR #11680, opened 2020-03-29
Ports a rewritten version of BAYSTRUMENTS

silicons:
## About The Pull Request

Everything under /instruments/synthesized_instruments is a direct copypaste from bay that I'm using right now so ignore it during review.
/instruments/musician.dm is also a copypaste I'm using right now while making this stuff

TL;DR
- I DO NOT KNOW IF THIS STILL WORKS ON 513. IF IT DOES NOT, WE ARE GOING TO HAVE A PROBLEM.
- Refactors sound code to a subsystem to be able to reserve sound channels using a stack, reserved by datums, etc etc
- Instruments completely reworked into new bay instruments. I will need help making a UI for it, because if all else fails I /guess/ I can use old UI but I do want tgui-nexi but I'm as of right now too new to javascript to do that so whatever

The things that you, the players will notice:
- Instruments will support MUCH faster songs, due to no sound channel being wasted during playback as any sound of the same "tone" will just use the same channel
- I can probably boost max song length a bit. Maybe. Probably. Maybe.
- If you try to have multiple people play instruments at once you might run into problems due to how the sound system works. I can either resolve this by making it so only a certain amount of instruments can play and too many players mean that they start cutting each other off, OR try to allocate in a way that maximizes the number of instruments that can play at the same time for bands and stuff (I say bands but with how byond sound works you'll be lucky if you even get two instruments playing at the same time and sounding even marginally good). There will be 128-256 at LEAST channels reserved for.. well, unreserved channel grabbing, so normal game sound effects can still play without instruments and "static" channels saturating them
- This WILL respect CHANNEL_HIGHEST_AVAILABLE. Yes, your admin midi spam and vore digestion noises won't be interrupted by this system :)))
- **MORE INSTRUMENTS WILL BE AVAILABLE. BIKE HORN INSTRUMENT IS BUFFED. Hell there's an instrument for alarm sounds or something, this works because of how this dynamically generates notes/loops rather than requiring one .ogg for each note**
- ALL LEGACY (so right now) INSTRUMENTS WILL BE FULLY (read: mostly) SUPPORTED IN THE NEW SYSTEM. It might sound ugly as sin, and suffer the usual bugs of "running out of sound channels", but hey, at least they exist.
- Instruments are now on a subsystem. Server lags? You suffer, not everyone else, like what would happen before.

## Why It's Good For The Game

Instruments are fun, and I want to play crappy converted undertale music while playing security.

## Changelog
:cl: 
add: Baystation instruments are here.
/:cl:

## This will also be ported to RP server on completion.


silicons:
"instruments on a subsystem you suffer not everyone else" is a lie just saying i just realized this 
i COUULD make it proper on a subsystem 
but the benefit isn't worth the trouble honestly 
i could also have just not subsystemized it at all instead of doing the weird hybrid subsystem processed decay while having a while loop process everyhting else 

@Putnam3145 @deathride58 @Ghommie give opinions please on if i should just bite the bullet and make it either subsystem or not subsystem instead of in-between

code/modules/instruments/instrument_data/_instrument_key.dm
```diff
@@ -0,0 +1,20 @@
...
+	src.key = key
+	src.deviation = deviation
+	src.frequency = frequency
+	if(!frequency && deviation)
+		calculate()
+
+/datum/instrument_key/proc/calculate()
+	if(!deviation)
+		CRASH("Invalid calculate call: No deviation or sample in instrument_key")
+	#define KEY_TWELTH 0.083
```
silicons: dont ask why this is named this
i put this in like 2 years ago as a replacement for baystation literally calling the define, and i quote, "Q", so yeah atleasti t's somewhat more descriptive :joy:

Putnam3145: why is it 0.083 instead of 1/12, which is kinda why it's called "KEY_TWELTH"

silicons: it's precalculated a 3 decimal place/digit accuracy because if i did (1/12) byond is going to compute 0.08333333333333333333333333333333333 every time something initialized

and while yes, that's uh, technically not too much of an issue it's like y
i'm raising it to a bit higher accuracy tho.

silicons: fuck it changed it it's only on init

code/controllers/subsystem/sounds.dm
```diff
@@ -0,0 +1,87 @@
...
+	/// List of all available channels with associations set to TRUE for fast lookups/allocation.
+	var/list/available_channels
+
+/datum/controller/subsystem/sounds/Initialize()
+	setup_available_channels()
+	return ..()
+
+/datum/controller/subsystem/sounds/proc/setup_available_channels()
+	available_channels = list()
+	for(var/i in 1 to using_channels_max)
```
Ghommie, line 23: Any reason not to use the `CHANNEL_HIGHEST_AVAILABLE` define instead of a static with the same value?

code/controllers/subsystem/sounds.dm
```diff
@@ -0,0 +1,87 @@
...
+
+/// Frees all datumless channels
+/datum/controller/subsystem/sounds/proc/free_datumless_channels()
+	free_datum_channels(DATUMLESS)
+
+/// NO AUTOMATIC CLEANUP - If you use this, you better manually free it later! Returns an integer for channel.
+/datum/controller/subsystem/sounds/proc/reserve_sound_channel_datumless()
+	var/channel = random_available_channel_text()
+	if(!channel)		//oh no..
+		return FALSE
```
Ghommie, line 55: I think you forgot to remove the selected channel from `available_channels` here, but I might be wrong.

code/controllers/subsystem/sounds.dm
```diff
@@ -0,0 +1,87 @@
...
+	using_channels[channel] = DATUMLESS
+	LAZYINITLIST(using_channels_by_datum[DATUMLESS])
+	using_channels_by_datum[DATUMLESS] += channel
+	return text2num(channel)
+
+/// Reserves a channel for a datum. Automatic cleanup only when the datum is deleted. Returns an integer for channel.
+/datum/controller/subsystem/sounds/proc/reserve_sound_channel(datum/D)
+	if(!D)		//i don't like typechecks but someone will fuck it up
+		CRASH("Attempted to reserve sound channel without datum using the managed proc.")
+	var/channel = random_available_channel_text()
```
Ghommie, line 66: Idem.

Ghommie (approved):
Seems good overall. Not an expert about musical scores though.

code/controllers/subsystem/sounds.dm
silicons: ehh mostly just for varediting if i or someone for whatever reason wants to privately reserve upper channels for their own ends mid-round

PR #11707, opened 2020-04-01
Punches land with the same consistency across races, punches cheaper, king hits are nastier

necromanceranne:
## About The Pull Request

~~This probably was because of combat mode code fucking this over. I have no idea why it was affecting xenos of all things because this was exclusively human species shit but I guess it's fine now.~~

This has gone from a fix to a rebalance because god it was really just punching being dumb on citadel due to stamina combat.

Now your miss chance is a calculation of 10+half your stamina and brute combined. You only roll for a miss if you roll your punchdamagelow. This is a really dumb way to get miss chance because it has to fucking roll twice basically but whatever.

The reason this is now consistent is because you literally have an obscenely large variance on punch values on most species. This should now be fixed by using 10 as the base, so now everyone is equal to a human on base punch chance.

In addition, punches cost 3.5 stamina per punch. This is the same value used for Gloves of the North Star's stamina heal. Which means they can punch infinitely. I don't think martial artists are using standard punch stamina costs anyway so this is almost certainly not going to affect glove scarpies, who may actually be right now capable of regenerating stamina through punching. Iunno. That's silly. 

But this really only affects everyone who isn't a martial artist basically.

In addition to all this, punching someone outside of combat mode is a 1.5x multiplier to your damage. Why? Because having a higher punch stun threshold is usually only something that humans have and they benefit from, but races with low stun thresholds will end up only getting marginal benefit from this. Now everyone benefits, but it ends up being a lot swingier on damage. 1.5 times 1 is still only 1.5. (Also meant you couldn't miss before)

**Edit:** And because of this modifier, a stun punch is now impossible before your target has already taken 50 stamina damage and is rolled based on a percentage of your stamina. Or else king hitting someone while they're out of combat mode is an almost certain knockdown and that's dumb as fuck. But this means you are more likely to be floored at higher stamina loss. Sprinting while in a boxing match is probably a bad idea.

## Why It's Good For The Game

~~I wanna punch shit.~~

I not only want to punch shit, I want to fucking redo all this fucking punching balance, starting with skooma and it's power to oneshot you because boy I was super fucking kind to it when I did the port and now I've gotta come back and end it once and for all.

Also why was a welder using less stamina than a punch but three times the damage and printed from an autolathe?

## Changelog
:cl:
balance: Punching should now be considerably more consistent across species.
balance: Punching is cheaper on stamina use.
balance: Punching someone out of combat mode now multiplies the damage by x1.5, just like hitting them with a blunt object!
balance: Stun punches are only possible if your target has suffered more than 50 stamina damage, with a higher probability at higher stamina damage.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1477,6 +1477,8 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
 		user.adjustStaminaLossBuffered(5) //CITADEL CHANGE - makes punching cause staminaloss
 
 		var/damage = rand(user.dna.species.punchdamagelow, user.dna.species.punchdamagehigh)
+		var/puncherstam = user.getStaminaLoss()
+		var/puncherbrute = user.getBruteLoss()
 
 		//CITADEL CHANGES - makes resting and disabled combat mode reduce punch damage, makes being out of combat mode result in you taking more damage
 		if(!(target.combat_flags & COMBAT_FLAG_COMBAT_ACTIVE) && damage < user.dna.species.punchstunthreshold)
@@ -1494,7 +1496,7 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
 			if(atk_verb == ATTACK_EFFECT_KICK) //kicks never miss (provided your species deals more than 0 damage)
 				miss_chance = 0
 			else
-				miss_chance = min((user.dna.species.punchdamagehigh/user.dna.species.punchdamagelow) + user.getStaminaLoss() + (user.getBruteLoss()*0.5), 100) //old base chance for a miss + various damage. capped at 100 to prevent weirdness in prob()
+				miss_chance = min(puncherstam + (puncherbrute*0.5), 100) //probability of miss based on stamina + half of brute. capped at 100 to prevent weirdness in prob()
 
 		if(!damage || !affecting || prob(miss_chance))//future-proofing for species that have 0 damage/weird cases where no zone is targeted
 			playsound(target.loc, user.dna.species.miss_sound, 25, TRUE, -1)
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1474,7 +1474,7 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
 			else
 				user.do_attack_animation(target, ATTACK_EFFECT_PUNCH)
 
-		user.adjustStaminaLossBuffered(5) //CITADEL CHANGE - makes punching cause staminaloss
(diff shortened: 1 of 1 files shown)


necromanceranne:
Actually this wasn't at all because of combat mode. It was because cit's punches drain the fucking SHIT out of your stamina. I'm gonna change the calculation on the miss chance a second time hold up cats.

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1494,7 +1498,7 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
 			if(atk_verb == ATTACK_EFFECT_KICK) //kicks never miss (provided your species deals more than 0 damage)
 				miss_chance = 0
 			else
-				miss_chance = min((user.dna.species.punchdamagehigh/user.dna.species.punchdamagelow) + user.getStaminaLoss() + (user.getBruteLoss()*0.5), 100) //old base chance for a miss + various damage. capped at 100 to prevent weirdness in prob()
+				miss_chance = min(clamp(((puncherstam + puncherbrute)*0.5) - ((punchedstam + punchedbrute)*0.5), 0, 100)) //probability of miss based on half your stamina and brute total against half their stamina and brute total. Capped at max 100 and min 0 to prevent weirdness in prob()
```
Ghommie: > ((puncherstam + puncherbrute)*0.5) - ((punchedstam + punchedbrute)*0.5)

By this new logic, punches shouldn't ever miss anymore.

necromanceranne: I might shrink the bonus down a little more just to avoid completely removing the miss chance. But the intent is that two people fighting each other should probably be able to punch one another unless they're stamdraining themselves sprinting, and if you're trying to punch someone who is freshfaced it'll start incurring miss chance if you are already stamina drained/injured.

necromanceranne: ~~Also because of basing miss chance on the division of punchhigh and low, when you have incredibly high variance on your punches, the numbers get...fucky. for instance, mushroom people miss more than humans because their punches had higher variance. As did military synths, corporate agents and some golems. This only got sillier when skooma was involved, with some of these races accuring weird rounding numbers or just flat out missing more. So removing the punchhigh/punchlow from the calc was more or less to keep it relatively sane.~~

Speaking of skooma had I not changed it originally it would have resulted in skooma users missing with near certainty. Thank fuck for that.

Sadly there is no consistency on punch stun thresholds so unfortunately skooma can also cause some races to stun instantly on punch. Also skooma in some races can oneshot people.

But who is going to feed golems skooma anyway? Especially plasteel golems.

necromanceranne: I MEANT HIT MORE NOT LESS AAAH IT'S LATE

Golems and fucking tiny miss chance but it still gets weird with skooma. Look my point is that the calculation dumb and so am I.

Ghommie (approved):

PR #11708, opened 2020-04-01
makes sergal markings better

Seris02:
## About The Pull Request

title

the chest is no longer one solid color on the sides

colors are made lighter so you can actually get the color you want instead of an always darker color

## Why It's Good For The Game

looks better

## Changelog
:cl:
tweak: the sergal markings
/:cl:


Ghommie (approved):
ok.

PR #11726, opened 2020-04-03
Dwarfy message spam QoL.

Ghommie:
## About The Pull Request
Title, this species is awful to play with the constant message spam, as somebody told me.
Also going a little over and tweaking some of those downer effects to be slightly less lethal, considering the reduced message warnings.

## Why It's Good For The Game
Chat messages spam is awful.

## Changelog
:cl:
tweak: Dwarves shouldn't have their chat constantly spammed by booze withdrawal and filth anymore now.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/dwarves.dm
+++ b/code/modules/mob/living/carbon/human/species_types/dwarves.dm
@@ -142,7 +142,7 @@ GLOBAL_LIST_INIT(dwarf_last, world.file2list("strings/names/dwarf_last.txt")) //
 				to_chat(owner, "<span class = 'warning'>Someone should really clean up in here!</span>")
 				last_filth_spam = world.time
 		if(26 to 50)
-			if(prob(7)) //And then the probability they vomit along with it.
+			if(prob(6)) //And then the probability they vomit along with it.
 				to_chat(owner, "<span class = 'danger'>The stench makes you queasy.</span>")
 				owner.vomit(10) //I think vomit should stay over a disgust adjustment.
 		if(51 to 75)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/species_types/dwarves.dm
```diff
@@ -136,57 +138,67 @@ GLOBAL_LIST_INIT(dwarf_last, world.file2list("strings/names/dwarf_last.txt")) //
...
 	switch(filth_counter)
 		if(11 to 25)
-			if(prob(25))
-				to_chat(owner, "<span class = 'danger'>Someone should really clean up in here!</span>")
+			if(last_filth_spam + 40 SECONDS < world.time)
+				to_chat(owner, "<span class = 'warning'>Someone should really clean up in here!</span>")
+				last_filth_spam = world.time
 		if(26 to 50)
-			if(prob(30)) //Probability the message appears
+			if(prob(7)) //And then the probability they vomit along with it.
```
Putnam3145: probably convey that this was originally a 0.2*0.3 chance (though that would be prob(6), hmm)

davidchan13:
This is good, but what Dwarves really need is a HUD icon to judge their sobriety similar to Food meter. Perhaps everyone could use some option HUD widget for that but for Dwarves its really needed since they can die of liver failure if sober too long.

If I were to pick what the HUD element was, make it a Beer mug:
If they are at the correct state of drunk, its full and frothy
As they start to sober the mug starts to empty and no more froth
When they are sober its completely empty
If they are dangerously sober it starts to crack and break
If they overly drunk the mug is overflowing or even tips over.

JTGSZ:
The easiest HUD to do would just be a transparent black square with maptext on it.

Ghommie:
Thanks for the ideas.

code/modules/mob/living/carbon/human/species_types/dwarves.dm
Ghommie: It's not such a big change after all, no?

silicons (approved):

PR #11731, opened 2020-04-03
quick bandaid patch for comms console topic()

silicons:


silicons:
this broke and im busy

PR #11768, opened 2020-04-06
uplink content updates with the current population

timothyteakettle:
## About The Pull Request
uplink content updates with population, so items that are locked by population 
this is hopefully good (however there is probably a better way to do this that I am not aware of?)

it effectively works by checking the current population and the population at when you last opened the uplink, and if they're different, getting the contents of the uplink again. also if you try buying something from an opened uplink but the current population has changed to not allow you to buy it, it'll stop you buying it.

## Why It's Good For The Game
imagine wanting your gamer population locked item but half the station latejoin after you get your uplink

also probably good if #11763 gets merged due to more things being population locked in the uplink and latejoiners shouldn't be stopping you from getting it?

## Changelog
:cl:
tweak: uplink content updates with current population
/:cl:

### DIFF
--- a/code/datums/components/uplink.dm
+++ b/code/datums/components/uplink.dm
@@ -119,13 +119,13 @@ GLOBAL_LIST_EMPTY(uplinks)
 		return
 	active = TRUE
 	if(user)
-		ui_interact(user)
 		//update the saved population
 		var/previous_player_population = saved_player_population
 		saved_player_population = GLOB.joined_player_list.len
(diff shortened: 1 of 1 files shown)


necromanceranne:
If a bunch of people cryo and the pop drops over time, this would also mean that antags not having purchased their items aren't sitting on pop inappropriate equipment. Works both ways, so that's pretty snazzy.

silicons (approved):

PR #11791, opened 2020-04-09
Makes spells activation only output one message instead of three.

Ghommie:
## About The Pull Request
I should refactor the action code a little so can_cast is not called three times because of the action button icon updates sometime later..

## Why It's Good For The Game
Fixing a whoopsie.

## Changelog
Nope.

### DIFF
--- a/code/modules/spells/spell.dm
+++ b/code/modules/spells/spell.dm
@@ -448,9 +448,8 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell)) //needed for th
 				to_chat(user, "<span class='notice'>You can't cast this spell here.</span>")
 			return FALSE
 
-	if(!skipcharge)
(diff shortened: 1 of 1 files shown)


code/modules/spells/spell.dm
```diff
@@ -449,7 +449,7 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell)) //needed for th
 			return FALSE
 
 	if(!skipcharge)
```
Putnam3145:
```suggestion
	if(!skipcharge && !charge_check(user, silent))
	    return FALSE
```

Putnam3145 (changes requested):
(obviously don't just commit this suggestion, you get what I mean)

Ghommie:
Closed in favor of the above PR.

PR #11793, opened 2020-04-09
Show LOOC toggle

Akrilla:
No message appeared for toggling your LOOC from the preferences tab.


Ghommie (approved):
Oh gosh the old code.

PR #11805, opened 2020-04-10
/carbon adjustStaminaLoss is now armor-piercing *as intended* before our modularized bandaid

silicons:
## About The Pull Request

title

## Why It's Good For The Game

adjustXLoss procs are supposed to be armor piercing (that's a lie: it's not just armor piercing per se, it just ignores the zone's usual armor checks, physiology and species checks. it still respects limb damage modifiers like on robotic limbs but i'm not doing anything about that because that would require redesigning the damage system of the game which is not really necessary/important right now for me to do), this one is no different. if you want stuff to respect armor use apply_damage instead (or we can make a new overall damage proc that does the same if necessary).

## Changelog
:cl:
experimental: adjustStaminaLoss on carbons is now a direct setter-ish proc instead of something that checks physiology/species mods, as per intent.
/:cl:

### THIS WILL BUFF STUNBATONS, AS THEY USE THIS.

### DIFF
--- a/modular_citadel/code/modules/mob/living/carbon/damage_procs.dm
+++ b/modular_citadel/code/modules/mob/living/carbon/damage_procs.dm
@@ -15,7 +15,7 @@
 	if(!forced && (status_flags & GODMODE))
 		return FALSE
 	var/obj/item/bodypart/BP = isbodypart(affected_zone)? affected_zone : (get_bodypart(check_zone(affected_zone)) || bodyparts[1])
-	if(amount > 0? BP.receive_damage(0, 0, amount) : BP.heal_damage(0, 0, abs(amount)))
+	if(amount > 0? BP.receive_damage(0, 0, amount * incomingstammult) : BP.heal_damage(0, 0, abs(amount)))
 		update_damage_overlays()
 	if(updating_health)
 		updatehealth()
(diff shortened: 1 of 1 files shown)


modular_citadel/code/modules/mob/living/carbon/damage_procs.dm
```diff
@@ -14,7 +14,12 @@
...
 		return FALSE
-	apply_damage(amount > 0 ? amount*incomingstammult : amount, STAMINA, affected_zone)
+	var/obj/item/bodypart/BP = isbodypart(affected_zone)? affected_zone : (get_bodypart(check_zone(affected_zone)) || bodyparts[1])
+	if(amount > 0? BP.receive_damage(0, 0, amount) : BP.heal_damage(0, 0, abs(amount)))
+		update_damage_overlays()
+	if(updating_health)
+		updatehealth()
+	update_stamina()
 	if((combat_flags & COMBAT_FLAG_HARD_STAMCRIT) && amount > 20)
 		incomingstammult = max(0.01, incomingstammult/(amount*0.05))
```
necromanceranne, line 24: What incomingstammult actually do anymore? Is this used anywhere else or was it just in the line you deleted? Because I'm fairly sure this was what was adjusting for stamina damage diminishing returns in adjust_damage once you hit hard stamina crit.

silicons: oh shit, woops

Ghommie (approved):

PR #11806, opened 2020-04-10
removes hotspot expose from skateboards

silicons:
## About The Pull Request

removes hotspot_expose(1000,1000)

because that took 1/2.5 = 40% of the air
heated it to 1000k
and mixed it back in

yeah that's why things were so hot.

## Why It's Good For The Game

why should i get to burn everyone in a hallway by going over a few tables.

## Changelog
:cl:
del: skateboards no longer superheat the air (as fast, they still do)
/:cl:


Arturlang:
Huh, how bad was it? I tested it a bunch and it wasnt that bad

Ghommie (approved):

PR #11820, opened 2020-04-11
Shield bashing

silicons:
## About The Pull Request

Adds shield bashing
Default behavior: 5 second cooldown, uses some attacker stamina, does stamina + brute to victims, knocks victims back 2 tiles in the direction of the user's lunge knocking them down if they hit anything solid.
Shield bashing is a 3 wide sweep-attack. It hits anyone within the line of 3 tiles infront of the user.
You can also directly right click someone on the floor to slam the shield into them, disarming them and doing the same amount of damage

Oh and if you don't do this on harmful intent you do a slightly weaker (no brute/stamina damage and pushes less, however still does other effects) attack for actual riot control.

## Why It's Good For The Game

Fun alt shield attack, why not? Can't be that OP. Except for the part where this is a sweep attack..
If it's too bad I can reduce knockback to 1.

## Changelog
:cl:
add: Shield bashing has been added
/:cl:


code/game/objects/items/shields.dm
```diff
@@ -3,11 +3,151 @@
...
+		return FALSE
+	if(world.time < last_shieldbash + shieldbash_cooldown)
+		to_chat(user, "<span class='warning'>You can't bash with [src] again so soon!</span>")
+		return FALSE
+	if(isliving(target))		//GROUND SLAAAM
+		if(!(shield_flags & SHIELD_BASH_GROUND_SLAM))
+			to_chat(user, "<span class='warning'>You can't ground slam with [src]!</span>")
+			return FALSE
+		bash_target(user, target, NONE, harmful)
+		user.do_attack_animation(target, used_item = src)
```
Linzolle, line 121: The linter doesn't seem to like this line for some reason

Linzolle: something to do with the "used_item = src"

silicons:
@Linzolle

Linzolle (approved):

PR #11865, opened 2020-04-15
Shuffles resist embed higher up on the resist chain

silicons:
## About The Pull Request

this also fixes it uh, nto working due to me breaking it lmao

## Why It's Good For The Game

Wish there was a better way to do this but in the majority of cases where you have a resistable item in you it's more important to get it out than to resist everything else

@deathride58 please think of a way to make this better because this can result in people critting themselves trying to resist resting status, but if I don't do this there's no other interface option for resisting embeds.

## Changelog
:cl:
tweak: Embed resist is now higher up on the resist attempt chain.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -645,11 +645,6 @@
 		attempt_resist_grab(FALSE)
 		// Return as we should only resist one thing at a time. Give clickdelay if the grab wasn't passive.
 		return old_gs? TRUE : FALSE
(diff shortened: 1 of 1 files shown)


Ghommie (commented):
This screws over people that are both on fire and with something embedded in them a little, it's not like you can stop drop and roll by clicking a href in the chat anyway. :laughing:

Ghommie (approved):

PR #11866, opened 2020-04-15
Shuffles around and buffs resisting embedded objects

silicons:
## About The Pull Request

Legcuffs no longer prevent this
Checks for MOBILITY_USE instead of MOBILLITY_MOVe
Does not require not being on fire
1x pain instead of 1.5x pain

## Why It's Good For The Game

Barely used feature, what's the point if tapping B after one single embed with a spear sends you to just above crit? I get you're supposed to remove it "carefully" but it's punishing enough without this.

## Changelog
:cl:
tweak: Resisting embedded objects is now less damaging/restrictive on its use.
/:cl:

### DIFF
--- a/modular_citadel/code/modules/mob/living/carbon/human/human.dm
+++ b/modular_citadel/code/modules/mob/living/carbon/human/human.dm
@@ -2,16 +2,15 @@
 	return
 
 /mob/living/carbon/human/resist_embedded()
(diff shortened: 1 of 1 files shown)


Ghommie (approved):

Ghommie:
Merged it just because of the legcuffed fix, honestly trying to yank a sharpnel out should be more punishing, but embedded objects already kill fast enough here.

PR #11871, opened 2020-04-15
minor nerf to aranesp's stamina regen

ghost:
## About The Pull Request
aranesp now heals 10 stamina damage per tick instead of 18

## Why It's Good For The Game
18 stamina damage healing per tick with no downsides if you take a few units of tricordrazine is kinda quirky. such a simple recipe shouldn't produce antag-level gear. 
aranesp is still very very good for healing stamina damage, just not at the same level as antag items. might give people a reason to use other minor stamina healing chems for maximum stamina healing

## Changelog
:cl:
balance: aranesp heals 10 instead of 18 stamina per tick
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drug_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/drug_reagents.dm
@@ -353,7 +353,7 @@
 
 /datum/reagent/drug/aranesp
 	name = "Aranesp"
-	description = "Amps you up and gets you going, fixes all stamina damage you might have but can cause toxin and oxygen damage."
+	description = "Amps you up and gets you going, fixes stamina damage you might have but can cause toxin and oxygen damage."
 	reagent_state = LIQUID
 	color = "#78FFF0"
 	pH = 9.2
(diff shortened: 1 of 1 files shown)


timothyteakettle:
still a meme because this is mostly powergay'd by slimepeople

ghost:
> still a meme because this is mostly powergay'd by slimepeople

no it's just powergayed

i'm too lazy to make aranesp something completely different but 10/tick, while still very good, is less good than literally stimulants without superspeed

necromanceranne:
There generally isn't anything wrong with a stamina heal but the issue is that non-aranesp 
drugs have depreciated since stamina combat. Aranesp had a niche to fill during stun combat era where it was uniquely only healing stamina damage. Now it has a much more significant role to fill and things like meth prove far weaker. This is kind of backwards given meth causing far nastier side effects than aranesp.

While it does need three chems that are somewhat annoying to make in their own right (epinephrine, atropine and morphine), two of them are available roundstart in abundance, atropine can be macro'd once you get tier 4 parts and morphine isn't that big a deal at all.

silicons (approved):

Ghommie (approved):

PR #11929, opened 2020-04-19
Shielded Component

Ghommie:
## About The Pull Request
Basically shielded hardsuit juice appliable to all clothing, items and living mobs too, in various forms.

## Why It's Good For The Game
May be of use for a future PR or two.

## Changelog
No, it's backend for now.


Linzolle (approved):

code/modules/clothing/suits/wiz_robe.dm
```diff
@@ -195,10 +195,9 @@
 	icon_state = "battlemage"
 	item_state = "battlemage"
 	recharge_rate = 0
+	max_charges = INFINITY
 	current_charges = 15
-	recharge_cooldown = INFINITY
```
Linzolle, line 199: Don't we not want the battlemage armour recharging over time?

Ghommie: Already handled by the zero `recharge_rate` 2-3 lines above.

PR #11946, opened 2020-04-20
Fix wormholes event flinging people into asteroids.

BlackMajor:
Webedit PR, I didn't test this but this oughta fix it. Probably :^)

Side effect of this fix is that wormholes probably won't spawn in solars now either. Good thing I guess?

### DIFF
--- a/code/modules/events/wormholes.dm
+++ b/code/modules/events/wormholes.dm
@@ -22,7 +22,8 @@
 
 /datum/round_event/wormholes/start()
 	for(var/turf/open/floor/T in world)
-		if(is_station_level(T.z))
+		var/area/A = get_area(T)
+		if(is_station_level(T.z) && !A.outdoors)
 			pick_turfs += T
 
 	for(var/i = 1, i <= number_of_wormholes, i++)
--- a/code/modules/events/wormholes.dm
+++ b/code/modules/events/wormholes.dm
@@ -24,7 +24,7 @@
 	for(var/turf/open/floor/T in world)
 		if(is_station_level(T.z))
 			var/area/A = get_area(T)
-			if(!A.outdoors)
+			if(A.outdoors)
 				continue
 			pick_turfs += T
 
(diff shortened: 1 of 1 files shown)


code/modules/events/wormholes.dm
```diff
@@ -22,7 +22,8 @@
 
 /datum/round_event/wormholes/start()
 	for(var/turf/open/floor/T in world)
-		if(is_station_level(T.z))
+		var/area/A = get_area(T)
+		if(is_station_level(T.z) && !A.outdoors)
```
Ghommie:
```suggestion
		if(is_station_level(T.z))
			var/area/A = get_area(T)
			if(!A.outdoors)
				continue
```
This way we only get the area after confirming the turf is on station level for performance, since this is a world loop after all.

Ghommie (changes requested):

code/modules/events/wormholes.dm
BlackMajor: Thanks fam

code/modules/events/wormholes.dm
```diff
@@ -23,6 +23,9 @@
 /datum/round_event/wormholes/start()
 	for(var/turf/open/floor/T in world)
 		if(is_station_level(T.z))
+			var/area/A = get_area(T)
+			if(!A.outdoors)
```
Ghommie:
```suggestion
			if(A.outdoors)
```
I did a typo.

Ghommie:
By the by, this sounds like a feature to me and not an issue, the risk/reward factor for wormholes jaunting is fairly thrilling already. If you don't want to risk being spaced, just stand still. But I'll see what other have to say about it.

BlackMajor:
> By the by, this sounds like a feature to me and not an issue, I find the risk/reward factor for wormholes jaunting thrilling. If you don't like them, just stand still.

The biggest gripe folk were having is that /mineral turfs, unlike regular wall turfs hide bodies underneath them. So if you get lodged in an asteroid and die, nobody can see you, unlike inside a wall.

silicons:
This PR is fine considering I don't think the asteroids should even be in there.

PR #11947, opened 2020-04-20
Ports /vg/ atmos map for boxstation

MrJWhit:
## About The Pull Request

Ports /vg/'s boxstation as reasonably as I could make it, see here: http://game.ss13.moe/minimaps/images/maps/Boxstation

Some notes to take into consideration:
-There's no extra glass/metal in atmos at roundstart with this
-There's three atmos suits
-I placed a single large scrubber in atmos as well

## Why It's Good For The Game

Gives a _bunch_ of space for atmos techs to play around in. Makes upgrading atmos something that can actually be done, and with tangible benefits to it.
I do feel ok with the things to take into consideration part. 
-The mats at roundstart is honestly too much and is rarely used, if ever. 
-The current builds for Lambda and Kilo, our newest maps, both include two atmos suits at roundstart. That, coupled with that this is straight up just what /vg/ has, means I feel comfortable putting that forward.
-The large scrubber was always something that people toyed with moving to atmos, and I honestly couldn't think of something else to place on the bottom for atmos tools

## Changelog
:cl:
balance: Ports /vg/ atmos into boxstation
/:cl:

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -3098,7 +3098,10 @@
 /turf/open/floor/plating,
 /area/maintenance/fore/secondary)
 "agd" = (
-/obj/machinery/atmospherics/pipe/manifold4w/general/visible,
+/obj/item/radio/intercom{
+	name = "Station Intercom (General)";
+	pixel_x = -30
+	},
 /turf/open/floor/plasteel,
 /area/engine/atmos)
 "age" = (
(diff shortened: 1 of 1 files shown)


DeltaFire15:
Three atmos hardsuits looks like waaay too much maybe reduce that to the default value, atmosians  have 'normal' advanced firesuits in their lockers aswell after all.
Also the air to waste pump starts as default on which just puts some more pressure on the waste loop, maybe have that start off.

MrJWhit:
All right, that *should* be better now. Reducing the amount of atmos suits also just makes the hallway look less god awful, I want to try and keep things consistent with /vg/ where reasonable.

PR #11958, opened 2020-04-20
Makes ling kill objectives kill-once

Putnam3145:
## About The Pull Request

See title.

## Why It's Good For The Game

Lings don't really have a compelling reason to *require* permakilling. It's convenient and such, but shouldn't necessarily be required.

## Changelog
:cl:
tweak: Lings kill-once instead of permakill.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -430,7 +430,7 @@
 		objectives += destroy_objective
 	else
 		if(prob(70))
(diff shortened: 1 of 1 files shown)


LetterJay:
Only issue this can bring up is meta grudging

Putnam3145:
We don't have CMD so it wouldn't even be metagrudging. Should be noted that changelings, uniquely, have the best tools to prevent being caught out despite the lack of CMD

Ghommie (approved):

PR #11988, opened 2020-04-23
Fixes tackle gloves not removing their components

silicons:
## About The Pull Request

So here's the thing: I don't know how it works on /tg/ I haven't checked inventory code there in a while but
Here 
dropped() is called AFTER something is already removed from the slot.

## Changelog
:cl:
fix: Tackle gloves now properly remove their components on drop.
/:cl:


Arturlang:
Maybe make it check the gloves are actually in the glove slot too?

Ghommie (approved):

PR #11989, opened 2020-04-23
Nerfs confusion

silicons:
## About The Pull Request
Random direction --> maximum 50 chance for random direction when over 40
Compounded 1.5 * confusion chance for turning 90 degrees to just confusion
Compounded 3 * confusion chance for turning 45 degrees to just confusion * 2

## Why It's Good For The Game

Confusion's ridiculously strong. It's an interesting effect but it shouldn't nullify movement.

This should probably be made into some sort of status effect in the future but for the moment I think this is a fair change that would allow more usage of it without making it completely debilitating.

Probably needs a testmerge.

## Changelog
:cl:
balance: Confusion effects have been significantly nerfed.
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -81,7 +81,7 @@
 
 	if(L.confused)
 		var/newdir = NONE
-		if((L.confused > 50) && prob(min(L.confused * 0.5), 50))
+		if((L.confused > 50) && prob(min(L.confused * 0.5, 50)))
 			newdir = pick(GLOB.alldirs)
 		else if(prob(L.confused))
 			newdir = angle2dir(dir2angle(direction) + pick(90, -90))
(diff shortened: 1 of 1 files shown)


Ludox235:
Nobody asked for this :/

Ghommie:
Confusion being ridicolously strong is actually an interesting factor, it doesn't nullify movement, but still severely restricts people from gracefully waltzing around halls, though it seems to last quite a bit too much for things like flashes or reagents, they are not designed that well.

TL;DR: Sumotori Dreams was a funny game.

TooFewSecrets:
Specifically, Changeling REEEEEEing totally shutting movement down in an AOE for a full minute through bowman headsets is stupid. Also applies to most other sources; confusion is a very drive-by status effect for how much it cripples you, and not exactly hard to get access to.

Ghommie (approved):

PR #12034, opened 2020-04-26
Maybe fixes vehicles blocking projectiles

silicons:
## About The Pull Request

With how buckled mob movement handling works, we cannot guarantee that every buckled mob is on the same tile by the time Crossed() is called (Moved() itself is called after movement, but crossing/etc are called on the initial atom's actual Move() step, while buckled mob handling comes after. The original atom Move()s, then Move()s any buckled mobs, then Moved() is called on buckled mobs one by one on a reverse order to the original. Look at atoms/atom_movement.dm for more information.)
Therefore the best way to make projectiles work that isn't going to take a fuckhuge refactor that I don't even know how to do (because even if we override behavior for Move() it's what it is for a reason, to imitate byond movement, if we just change it we're introducing issues that need to be resolved again in the future, and since we can't just change how byond pixel movement works they're not going to be resolvable) is to just have them check for buckled mobs when they hit an obj.

## Why It's Good For The Game

Secways/atvs/vehicles in general really should not be blocking shots.

## Changelog
:cl:
fix: Vehicles should no longer block projectiles. Should.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -315,6 +315,8 @@
 		objs += O
 	var/obj/O = safepick(objs)
 	if(O)
+		if(length(O.buckled_mobs))
+			return pick(O.buckled_mobs)
 		return O
 	//Nothing else is here that we can hit, hit the turf if we haven't.
 	if(!(T in permutated) && can_hit_target(T, permutated, T == original, TRUE))


Linzolle (approved):

Linzolle:
I don't have any better ideas

PR #12045, opened 2020-04-27
Identification components + chameleon clothing from uplink is now obfuscated

silicons:
## About The Pull Request

Adds identification components to set up systems where functionality of items or features are obfuscated unless the user knows how to use it.

Chameleon clothing from uplink is now obfuscated to non-traitors unless analyzed with a deconstructive analyzer

Chameleon gloves from uplink are now insulated.
Chameleon ID obfuscated too. **Those can not be analyzed with deconstructive analyzers.** This is to better match their old functions, aka not being usable/revealing at all unless you're a traitor. The chameleon action buttons were the change that revealed them, now it's no more.

deconstructive analyzers now have a deep scan mode.

## Why It's Good For The Game

Identification components might be useful in the future not just for obfuscating stuff like antag items.

Chameleon clothing: While I usually hate mechanically obfuscating antag gear and stuff I think it's reasonable to do it to this given them being useless as hell half the time. They're too weak and it's a bit hard to do a gimmick/have any stealth if any random joe who takes your gloves off instantly knows you're antag man.
Chameleon gloves from uplink are made insulated to make the kit worth a bit more.

## Changelog
:cl:
add: Identification components added.
balance: Chameleon kits from uplinks now include insulated gloves. Included items are not usable by non-traitors until ran through a deconstructive analyzer on deep scan mode.
/:cl:

### DIFF
--- a/code/modules/research/destructive_analyzer.dm
+++ b/code/modules/research/destructive_analyzer.dm
@@ -101,7 +101,7 @@ Note: Must be placed within 3 tiles of the R&D Console
 	if(!istype(loaded_item) || !istype(linked_console))
 		return FALSE
 
-	if (id && id != RESEARCH_MATERIAL_RECLAMATION_ID)
+	if (id && id != RESEARCH_MATERIAL_RECLAMATION_ID && id != RESEARCH_DEEP_SCAN_ID)
 		var/datum/techweb_node/TN = SSresearch.techweb_node_by_id(id)
 		if(!istype(TN))
 			return FALSE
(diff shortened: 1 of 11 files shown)


CameronWoof:
I'd like to put in a request out ahead of this that staff comes up with a ruling on when it is reasonable to use this deep scan mode, as my prescient future sight already sees disagreements about whether there's any way to use it without getting bwoinked for metagaming.

CameronWoof:
But this is a very good idea. This can be used to solve a lot of the "having to pretend not to know" issues that come up with traitor items.

code/modules/research/destructive_analyzer.dm
```diff
@@ -101,7 +101,7 @@ Note: Must be placed within 3 tiles of the R&D Console
 	if(!istype(loaded_item) || !istype(linked_console))
 		return FALSE
 
-	if (id && id != RESEARCH_MATERIAL_RECLAMATION_ID)
+	if (id && id != RESEARCH_MATERIAL_RECLAMATION_ID && id != RESEARCH_DEEP_SCAN_ID)
```
Ghommie, line 104: This could be a switch() instead of an else if cascade. But whatever.

Ghommie (approved):

PR #12050, opened 2020-04-29
movespeed is now no longer an integer 

silicons:
## About The Pull Request

AAAAAA
movespeed isn't integer y'all

## Changelog
:cl:
fix: multiplicative movespeed config is no longer integer only!
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -214,6 +214,7 @@
 /datum/config_entry/keyed_list/multiplicative_movespeed
 	key_mode = KEY_MODE_TYPE
 	value_mode = VALUE_MODE_NUM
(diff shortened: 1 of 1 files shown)


code/controllers/configuration/entries/game_options.dm
```diff
@@ -214,6 +214,7 @@
 /datum/config_entry/keyed_list/multiplicative_movespeed
 	key_mode = KEY_MODE_TYPE
 	value_mode = VALUE_MODE_NUM
+	integer = TRUE
```
Putnam3145: are you sure setting integer to TRUE does not in fact make it an integer

silicons: WAIT NO FUCK

PR #12061, opened 2020-04-30
a few mostly antag-related nerfs

kiwedespars:
## About The Pull Request

just a bunch of number tweaks on what i believe to be especially strong right now.

## Why It's Good For The Game

there's a lot of dumb stuff that needs to get changed up.

- fleshmend is one of the two strongest things about lings, and i feel like you shouldn't be undetectable if you take the two best abilities lings have to offer.
- combat defibs are undisarmable, disarming stunbatons that do 50 stamloss while also bypassing block chance, sort of crazy.
- ebow dealt 60 stam damage on a really low recharge. i get it that it's 12 tc, but for it to disarm, knockdown, and make it impossible to get up from without soft stamcritting yourself, that's sort of much.
- emagged hyposprays became better sleepy pens.
- syndie gamerkit becomes less of a "it's cheap and it's really really really useful, so let's get it" and more of a choice. 100u of 3-per-tick healing in an injector is pretty strong, especially when it heals stamloss faster than stimulants do. syndicate mediborgs exist for a reason, and they're like 50 tc to give /slightly/ better nanites, compared to the original 4 tc.

## Changelog
:cl:
- lings get loudness attached to fleshmend, so having fleshmend + adrenals outs you as a ling
- combat defibs do less stamdamage. 50 -> 20
- stimpacks inject 25 at a time instead of 10.
- ebow stam damage nerfed. 60 -> 40
- emagged hyposprays are no longer instant inject. 0.5 for self-inject, 1.5 seconds for injecting others.
- syndie medkit price nerfed.  4 tc -> 15 tc
/:cl:

### DIFF
--- a/code/game/objects/items/defib.dm
+++ b/code/game/objects/items/defib.dm
@@ -448,7 +448,7 @@
 	if(do_after(user, isnull(defib?.disarm_shock_time)? disarm_shock_time : defib.disarm_shock_time, target = M))
 		M.visible_message("<span class='danger'>[user] zaps [M] with [src]!</span>", \
 				"<span class='userdanger'>[user] zaps [M] with [src]!</span>")
-		M.adjustStaminaLoss(50)
+		M.adjustStaminaLoss(20)
 		M.DefaultCombatKnockdown(100)
 		M.updatehealth() //forces health update before next life tick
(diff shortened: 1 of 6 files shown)


deathride58 (changes requested):
Atomize your changes. These are all mostly unrelated balance changes that should be contained in separate PRs

PR #12085, opened 2020-05-01
makes the uranium mat two times less radioactive

Ghommie:
## About The Pull Request
Title. Web edit.

## Why It's Good For The Game
I must admit it's unnaturally strong and without decay. Also putnam's tweaks made radioactivity not inefficiently suck hay bales as much as it did before.

## Changelog
:cl:
balance: nerfed the radioactivity of custom uranium made items by two times.
/:cl:

### DIFF
--- a/code/datums/materials/basemats.dm
+++ b/code/datums/materials/basemats.dm
@@ -72,7 +72,7 @@ Unless you know what you're doing, only use the first three numbers. They're in 
 
 /datum/material/uranium/on_applied(atom/source, amount, material_flags)
 	. = ..()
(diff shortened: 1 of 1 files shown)


Trilbyspaceclone (changes requested):
Its better to just remove the radiation form uranium as a whole for a load of reasons.
Its just not good for the game at all

lolman360:
honestly making our current uranium "depleted uranium" and minimally/not radioactive would be the better option and then maybe adding a radioactive form obtainable through other sources (centrifuge machine or something idk) but that's more work than just making it less rads

PR #12122, opened 2020-05-05
gives mining winter coats similar storage to other mining suits

ghost:
## About The Pull Request
pka, large tanks, resonator, and both mining scanners

## Why It's Good For The Game
they're actual mining gear now

## Changelog
:cl:
tweak: made the mining winter coat's storage list more inline with other mining suits
/:cl:


Ghommie (approved):
.

PR #12123, opened 2020-05-05
The great material rebalance

silicons:
## About The Pull Request

As per WIP R&D/lathing/station production draft: https://hackmd.io/@CitadelStation13/rJ9_8GHKL

RND efficiency now caps out at 60% cost at tier 4 instead of 12.5%
Ore efficiency now caps out at 150% at tier 4 instead of 200%
Print times and busy times made consistent for R&D machinery, and reduced a bit
Autolathe print time reduced drastically

## Why It's Good For The Game
The old values were absolutely insane and made material balance just about nonexistent. While a total amplification of around 160%*150% = 240% is strong, it's nowhere as bad as the old 1600%.
Autolathe print time reduced drastically (I'll be adding print queues to autolathes and R&D machinery and porting them to tgui-next in a later PR by the by) to make them more relevant versus protolathe printing.

I'll probably be doing a mini pass of individual materials costs to standardize some things later down the line.

## Changelog
:cl:
balance: Material amplification by stock parts upgrading has been drastically reduced.
balance: Autolathes and protolathes print faster.
fix: Protolathe print time is now the same as their busy-state print-lockout time.
/:cl:


code/game/machinery/autolathe.dm
```diff
@@ -206,7 +206,7 @@
 				busy = TRUE
 				use_power(power)
 				icon_state = "autolathe_n"
-				var/time = is_stack ? 32 : 32*coeff*multiplier
+				var/time = is_stack ? 10 : 10 * coeff * multiplier
```
Ghommie: Hey, care to make a variable for this and set it to 20 for the public lathe, so I won't have to?
I don't believe simple freeloaders should benefit from reduced printing times this much.

Ghommie (approved):
Almost all fine.

code/game/machinery/autolathe.dm
silicons: sure

PR #12136, opened 2020-05-07
Fixes prosthetic causing bloodsuckers to bug out

Arturlang:
Makes you not be stuck in torpor due to having prosthetics, and loose blood due to regeneration accounting for it.
## Changelog
:cl:
fix: Fixes being stuck in torpor as a bloodsucker due to having a damaged prosthetic, and loosing blood due to it.
/:cl:


Ghommie (changes requested):
I suggest you to move these helper procs to` living_defense.dm` and `carbon_defense.dm` as well as an argument used to check wheter robotic or organic limbs should be counted.
Otherwise all fine.

Ghommie (approved):
Eh, whatever fine.

PR #12161, opened 2020-05-08
Fixes the maploader

silicons:
so it turns out when I did rotation maploading it caused some issues that didn't come to light until I tried to use the admin reload map thing.
right.

:cl:
fix: The maploader can once again crop maps, although the functionality may be ... different.
/:cl:

### DIFF
--- a/code/modules/mapping/reader.dm
+++ b/code/modules/mapping/reader.dm
@@ -1,6 +1,12 @@
 ///////////////////////////////////////////////////////////////
 //SS13 Optimized Map loader
 //////////////////////////////////////////////////////////////
+/*
+ * Notes:
+ * This does NOT support map files where more than one Z is in each grid set, as defined in the file.
+ * MultiZ map files ARE supported (however very much discouraged), as long as each gridset only contains one z.
+ * This assumes that for the most part, map files are properly formed, either DMM or TGM standard formats. If you feed it bad data, expect to crash.
+ */
 #define SPACE_KEY "space"
 
 /datum/grid_set
@@ -8,6 +14,12 @@
 	var/ycrd
 	var/zcrd
 	var/gridLines
+
+/datum/grid_set/proc/height(key_len)
+	return length(gridLines)
+
+/datum/grid_set/proc/width(key_len)
+	return gridLines[1] / key_len
 
 /datum/parsed_map
 	var/original_path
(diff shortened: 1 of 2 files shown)


code/modules/mapping/reader.dm
```diff
@@ -60,25 +72,37 @@
 	placeOnTop = FALSE as num,
 	orientation = SOUTH as num,
 	annihilate_tiles = FALSE,
-	z_lower = -INFINITY as num,
-	z_upper = INFINITY as num
+	crop_relative_to_game_world = TRUE
 	)
-	var/datum/parsed_map/parsed = new(dmm_file, x_lower, x_upper, y_lower, y_upper, z_lower, z_upper, measureOnly)
```
silicons, line 66: with orientation there's no way to do this without breaking, if people want to crop maps and math it out manually they can imeplement it custom instead of use load_map().

code/modules/mapping/reader.dm
```diff
@@ -532,6 +561,7 @@
 	// fallback: string
 	return text
 
-/datum/parsed_map/Destroy()
-	..()
-	return QDEL_HINT_HARDDEL_NOW
+/datum/parsed_map/vv_edit_var(var_name, var_value)
+	if(var_name == NAMEOF(src, dmmRegex) || var_name == NAMEOF(src, trimQuotesRegex) || var_name == NAMEOF(src, trimRegex))
```
silicons, line 565: static vars, no one should be editing these

code/modules/mapping/reader.dm
```diff
@@ -60,25 +72,37 @@
...
+/**
+  * Parse a map, possibly cropping it.
+  * Do not use the crop function unless strictly necessary.
+  * WARNING: Crop function crops based on the tiles you'd see in the map editor. If you're planning to load it in in a different orientation later, you better have done the math.
+  * It's recommended that you do not crop using this at all.
+  */
 /datum/parsed_map/New(tfile, x_lower = -INFINITY, x_upper = INFINITY, y_lower = -INFINITY, y_upper = INFINITY, z_lower = -INFINITY, z_upper = INFINITY, measureOnly = FALSE)
+	_parse(tfile, x_lower, x_upper, y_lower, y_upper, z_lower, z_upper, measureOnly)
+
+/datum/parsed_map/proc/_parse(tfile, x_lower, x_upper, y_lower, y_upper, z_lower, z_upper, measureOnly)
```
silicons, line 91: dmmregex and other things are statics, i'm pretty sure they can't collide but for the sake of performance no one should be trying to load/parse two different maps at the same time anyways.

Ghommie (approved):
Big yikes. Has this been tested, kevinz?

silicons:
@Ghommie yes

PR #12201, opened 2020-05-10
Purging unused dogborg items.

Ghommie:
## About The Pull Request
How much I longed for the day I would remove a good chunk of dogborg items. A bit let down there is still a good chunk of dogborg sleeper rubbish lingering around but that'll be dealt with later hopefully.

## Why It's Good For The Game
Removing unused features that are probably not going to make a comeback, not on my agenda at least. Also this will close #12198.

## Changelog
:cl:
del: Removed unused dogborg items.
/:cl:


Vizoplez:
Nevermind, editing because I actually went and read the code.

PR #12214, opened 2020-05-11
emp protection field.

Ghommie:
## About The Pull Request
WIP. May add more fields later, slowly.

## Why It's Good For The Game
I have a few ideas that kinda rely on these fields.

## Changelog
None yet.


shellspeed1:
Why would we use a field to protect against EMPs? I was under the impression faraday cages were how we currenlty do it so, a force field doing it doesn't really make sense.

Ghommie:
I was lazy.

PR #12235, opened 2020-05-12
Adds a choice to whether or not you're setting someone's bank account job to a job when assigning jobs from hop console.

silicons:
## Why It's Good For The Game

Makes custom job titling a pain without, as you have to instead manually assign accesses.
Vendoreconomy is a mistake.

## Changelog
:cl:
add: HoP consoles now ask if you want to set someone's bank account assigned job to their new rank.
/:cl:

### DIFF
--- a/code/game/machinery/computer/card.dm
+++ b/code/game/machinery/computer/card.dm
@@ -467,7 +467,7 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 						to_chat(usr, "<span class='alert'>No log exists for this job.</span>")
 						updateUsrDialog()
 						return
-					if(inserted_modify_id.registered_account && alert("Would you like to set their bank account's assigned job to this?", "Bank Account Job", "Yes", "Yes") == "Yes")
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/card.dm
```diff
@@ -467,7 +467,7 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 						to_chat(usr, "<span class='alert'>No log exists for this job.</span>")
 						updateUsrDialog()
 						return
-					if(inserted_modify_id.registered_account)
+					if(inserted_modify_id.registered_account && (alert(usr, "Would you like to set their bank account's assigned job to this?", "Bank Account Job", "Yes", "No") == "Yes"))
 						inserted_modify_id.registered_account.account_job = jobdatum // this is a terrible idea and people will grief but sure whatever
```
Ghommie, line 471: Should be moved to a new proc with waitfor set to FALSE, otherwise that alert() will halt the current function, possibly causing some runtimes or issues if the ID were removed or changed in the meanwhile

EDIT: it should check if the inserted modify ID is still the same after the input was given, as well as a canUseTopic call.

Ghommie (changes requested):

PR #12238, opened 2020-05-13
Some quick economy fixes and improvements.

Ghommie:
## About The Pull Request
Allowing players to withdraw the entirety of an account's savings by inputting any amount higher than what they have.
Fixing some jank with the vendor UI that made offstation (read, free) vendors unusable by those without an account, as well as premium items appearing as discounted when they actually aren't.
Fixing `custom_premium_price` not being set. Adding some parenthesis to discount suffixes.

## Why It's Good For The Game
Fixing some issues with a recently merged feature.

## Changelog
:cl:
fix: Fixed a few things with economy.
/:cl:


silicons (changes requested):
let's put crabphone in another pr considering it's a 2 line change and the potential of what we're allowing by enabling an "everyone lynch me" item that fucks everyone in the span of just a couple of minutes when we're already having issues with economy

PR #12240, opened 2020-05-13
makes the improvised shotgun require 2-hands to fire instead of 1

ghost:
## About The Pull Request
title

## Why It's Good For The Game
basically the double barrel is supposed to be unique in that it can be used 1-handed unlike other shotguns, but the improvised is a subtype of the double barrel so it also got the 1-handed change when db was made 1-handed again. this puts the shotgun and rifle on a more equal standing - clunkier gun that uses one hand or strong shooty gun that uses two? - and makes the bartender's shotgun more unique

## Changelog
:cl:
balance: improvised shotguns require both hands to fire instead of just one
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/revolver.dm
+++ b/code/modules/projectiles/guns/ballistic/revolver.dm
@@ -321,6 +321,7 @@
 	icon_state = "ishotgun"
 	item_state = "shotgun"
 	w_class = WEIGHT_CLASS_BULKY
+	weapon_weight = WEAPON_HEAVY
 	force = 10
 	slot_flags = null
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/improvised


necromanceranne:
This makes the sawn-off heavy as well. I don't think it would be entirely sane to make sawn-offs require two hands.

More importantly I don't actually understand why a single shot tube gun is in any way an overpowered weapon, short of it using shotgun ammo, that it needs the combat shotgun treatment. There already exists a pr up that makes improvised shotguns use a weaker alternative ammo, which seems more reasonable and does actually differentiate the improvised weapons from the double-barreled.

The only actual overpowered aspect of it, that it combined too strongly with rubbershot, I nerfed back when I made stamina and lethal separate for crit.

ghost:
> This makes the sawn-off heavy as well. I don't think it would be entirely sane to make sawn-offs require two hands.
> 
> More importantly I don't actually understand why a single shot tube gun is in any way an overpowered weapon, short of it using shotgun ammo, that it needs the combat shotgun treatment. There already exists a pr up that makes improvised shotguns use a weaker alternative ammo, which seems more reasonable and does actually differentiate the improvised weapons from the double-barreled.
> 
> The only actual overpowered aspect of it, that it combined too strongly with rubbershot, I nerfed back when I made stamina and lethal separate for crit.

idk i just like the mosin better flavor wise it's pretty pog

silicons:
@kappa-sama Is the issue alpha striking, or being able to use this with a shield or another weapon?

ghost:
alpha striking for the oneshot  but i see you've got it handled for now

PR #12252, opened 2020-05-14
Clockies now can summon shields to bash nonbelievers into orbit.

DeltaFire15:
## About The Pull Request
example

Adds ratvarian shields into the tier two scripture tier.
Gameplaywise, these are simillar to riot shields (except not transparent), though they are repaired with brass instead of titanium, aswell as having high bullet and melee armor in exchange for negative laser and energy resist (as most clockwork gear does),
Though, they do have a more or less unique mechanic, which is that they absorb the damage from blocked attacks (they still damage the shield, so be careful), and allow the user to perform empowered bashes using that energy. Current value for this is a maximum bash multiplier of _x4_ , using _90_ absorbed damage, aka _30_ damage absorbed per multiplier 'tier', leading to about _20_ brute damage, around _60_ stam and _8_ knockback if at the maximum threshhold. 
Bashes will only make use of stored energy if in harm intent, and only actually use it up if at least one poor sod gets hit with the bash. If there is more power than even a bash with highest multiplier would use, it will only use the amount needed to get that multiplier. Having the shield break loses all stored energy.
And no, clockies can't just beat eachother up to generate infinite charge.

Either way, if a nonclockie somehow got their hands on a shield, they can use it to normally bash (charge will never be used though) and block... though there may be a nasty surprise waiting for them if they actually do manage to block an attack.

Shield sprites shamelessly scavenged from clockwork marauders, I ain't the best at spriting so feel free to tell me if you have a better sprite, though this one should suffice.

On another note, this PR also fixes a small thing in the shieldcode itself which only specifically allowed titanium for repairs instead of taking the repair_material var into account.

## Why It's Good For The Game
Clockies as of now are rather lacking of actual combat equipment that is not just utility things (just the spear), leading to them requring to always steal weapons from somewhere, and get absolutely steamrolled if not the case (unless the whole station got converted before the siege phase).
This PR aims at giving clockies a somewhat increased combat capability without needing to ''acquire'' equipment from elsewhere, while introducing a gimmick to the shield that I think would be fun to have (and thematic to the clock cult)

## Changelog
:cl:
add: Clockies can create shields with empowerable bashes.
fix: Shield repairing now works properly for materials other than titanium.
/:cl:


ghost:
could you add it to summon equipment instead of having its own spell for consistency and simplicity of the spell list? very cool shield though kinda poggy

DeltaFire15:
> can you just add it to summon equipment instead of having its own spell? very cool shield though kinda poggy

I did think about just adding it to summon equipment but decided against it because of how the summon spell is coded ( aswell as call_weapon deleting the item if you dismiss it and just creating a completely new one, which is fine for the spear but problematic if the values actually matter.)
Alternatively, if it was added to it like the armor is (one set of armor every x seconds, you'd have the problem you could generate infinite shields with a quite low poweruse.
These would be the only two ways to effectively implement it into the summon equipment spell, as you can only cast it exactly once, which is why I decided on making it its own seperate spell (unless I wanted to code another call_weapon spell that stored it somewhere if dismissed and only deleted it if destroyed, but that'd be more work than just using a seperate scripture.)

ghost:
ok was just curious

silicons:
DEUS VULT

code/modules/antagonists/clockcult/clock_items/clock_weapons/ratvarian_shield.dm
```diff
@@ -0,0 +1,84 @@
...
+	var/dam_absorbed = 0
+	var/bash_mult_steps = 30
+	var/max_bash_mult = 4
+
+/obj/item/shield/riot/ratvarian/examine(mob/user)
+	if((is_servant_of_ratvar(user) || isobserver(user)))
+		desc = clockwork_desc
+		desc +="\n <span class='inathneq_small'>The shield has absorbed [dam_absorbed] damage, multiplying the effectiveness of its bashes by [calc_bash_mult()]</span>"
+	. = ..()
+	desc = initial(desc)
```
Linzolle, line 23:
```suggestion
	. = ..()
	if((is_servant_of_ratvar(user) || isobserver(user)))
		. += clockwork_desc
		. +="\n <span class='inathneq_small'>The shield has absorbed [dam_absorbed] damage, multiplying the effectiveness of its bashes by [calc_bash_mult()]</span>"

```

DeltaFire15: Not feasible as it is supposed to only display clockwork desc if examined by a observer / clockie, not normal desc + clockwork desc.
Clockie / normal desc is also supposed to be above the other special stuff for examine (e.g. materials, if its damaged, etc) ..() does, with only the multiplier acceptable to be below. This means desc needs to be edited, as it is used in ..()

The descriptions in comparison: 
With the change:



Without the change:



I can partially comply with this and attach the part about the multiplier via . +=, but desc itself requires the previous implementation. (And the partial change would be quite a bit more ugly as it'd require a check to change desc, then the . = ..() outside the if, and then another if with the same check since it has the same conditions for being displayed)

code/modules/antagonists/clockcult/clock_items/clock_weapons/ratvarian_shield.dm
```diff
@@ -0,0 +1,84 @@
...
+	playsound(owner, 'sound/magic/clockwork/anima_fragment_death.ogg', 50)
+	new /obj/item/clockwork/alloy_shards/large(get_turf(src))
+
+/obj/item/shield/riot/ratvarian/user_shieldbash(mob/living/user, atom/target, harmful)
+	if(harmful && is_servant_of_ratvar(user)) // No fun for non-clockies, but you can keep the normal bashes. Until you try to block with it.
+		var/actual_bash_mult = calc_bash_mult()
+		shieldbash_knockback = round(initial(shieldbash_knockback) * actual_bash_mult, 1)   //Modifying the strength of the bash, done with initial() to prevent magic-number issues if the original shieldbash values are changed
+		shieldbash_brutedamage = round(initial(shieldbash_brutedamage) * actual_bash_mult, 1) //Where I think of it, better round this stuff because we don't need even more things that deal like 3.25 damage
+		shieldbash_stamdmg = round(initial(shieldbash_stamdmg) * actual_bash_mult, 1) //Like 20 brute and 60 stam + a fuckton of knockback at the moment (at maximum charge), seems mostly fine? I think?
+		if(..()) //If this bash actually hit someone
```
Ghommie:
```suggestion
		. = ..() //If this bash actually hit someone
		if(.)
```

code/modules/antagonists/clockcult/clock_items/clock_weapons/ratvarian_shield.dm
```diff
@@ -0,0 +1,84 @@
...
+			if(actual_bash_mult > 1)
+				playsound(user, 'sound/magic/fireball.ogg', 50, TRUE, frequency = 1.25)
+				dam_absorbed -= calc_bash_absorb_use()
+			return TRUE
+		else
+			return FALSE
```
Ghommie:
```suggestion
		return
```

code/modules/antagonists/clockcult/clock_items/clock_weapons/ratvarian_shield.dm
```diff
@@ -0,0 +1,84 @@
...
+	else
+		damage -= 5
+	return ..()
+
+/obj/item/shield/riot/ratvarian/shatter(mob/living/carbon/human/owner)
+	playsound(owner, 'sound/magic/clockwork/anima_fragment_death.ogg', 50)
+	new /obj/item/clockwork/alloy_shards/large(get_turf(src))
+
+/obj/item/shield/riot/ratvarian/user_shieldbash(mob/living/user, atom/target, harmful)
+	if(harmful && is_servant_of_ratvar(user)) // No fun for non-clockies, but you can keep the normal bashes. Until you try to block with it.
```
Ghommie: I suggest an early return if it is either non-harmful or the user isn't a clockie instead.

Ghommie (changes requested):

code/modules/antagonists/clockcult/clock_items/clock_weapons/ratvarian_shield.dm
DeltaFire15: Did exactly that.

code/modules/antagonists/clockcult/clock_items/clock_weapons/ratvarian_shield.dm
DeltaFire15: Did this

code/modules/antagonists/clockcult/clock_items/clock_weapons/ratvarian_shield.dm
DeltaFire15: Aaand, logically did that aswell.

Ghommie (approved):

PR #12267, opened 2020-05-15
Reworks cryo knockout times

silicons:
## About The Pull Request

Cryo cells knocked out at a factor of 0.00125 * 2000 * bodytemperature deciseconds.
This means the colder you were, the less affected you were by the knockout.
This means that upgrade parts, which multiplied the 0.00125 part of the calculation, actually made it knock out for longer.

The equation has been adjusted to 30 seconds base, with upgrades from tier 1 to 4 decreasing it to up to 10 seconds base, which is then multiplied by (4ln(bodytemperature)-20) for temperature factor of 1.2 at standard to around 2-2.5 at 0K
This means at tier 1, if you're using a standard 73K cell , you're going to be knocked out for around 35 seconds. Keep going towards 2.7K and you'll start to suffer 60-75 seconds. Incentive to not have temperatures so cold without upgrades.

## Why It's Good For The Game

Doesn't make sense for upgrades to make knockout time worse, and doesn't make sense for it to be worse the higher temperature it's at.
Brings some new disadvantages to setting it to a super low temperature too.
Feel free to comment if you think these values are too low/high.

## Changelog
:cl:
balance: Cryo knockout time has been reworked.
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -188,7 +188,7 @@
 	if(air1.gases.len)
 		if(mob_occupant.bodytemperature < T0C) // Sleepytime. Why? More cryo magic.
 			// temperature factor goes from 1 to about 2.5
-			var/amount = max(1, log(T0C - mob_occupant.bodytemperature) - 3) * knockout_factor * base_knockout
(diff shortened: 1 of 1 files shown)


Ghommie (approved):

Ghommie:
If people end up perma-sleeped I'm blaming you. :smirk:

PR #12274, opened 2020-05-16
adds a snowflakey-ish(??? maybe???) way to modify gun damage so we don't need more shot types for improvised shotguns 

silicons:
## About The Pull Request

title, adds a var to gun

## Why It's Good For The Game

is it really i can't tell if this is shitcode someone school me but yeah this is for https://github.com/Citadel-Station-13/Citadel-Station-13/pull/12226

## Changelog
it's a code update for one var that almost no one cares unless they're already reading this in the code file why would i make a changelog

### DIFF
--- a/code/modules/projectiles/ammunition/_firing.dm
+++ b/code/modules/projectiles/ammunition/_firing.dm
@@ -32,7 +32,7 @@
 		BB.def_zone = user.zone_selected
 	BB.suppressed = quiet
 	
-	if(istype(fired_from, /obj/item/gun))
+	if(isgun(fired_from))
 		var/obj/item/gun/G = fired_from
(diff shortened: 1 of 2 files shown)


Linzolle:
seems... like a reasonable way to me?

code/modules/projectiles/ammunition/_firing.dm
```diff
@@ -31,6 +31,10 @@
 	else
 		BB.def_zone = user.zone_selected
 	BB.suppressed = quiet
+	
+	if(istype(fired_from, /obj/item/gun))
```
Ghommie:
```suggestion
	if(isgun(fired_from))
```

Ghommie (approved):

PR #12293, opened 2020-05-18
New gun flag for dual wielding exclusion

silicons:
## About The Pull Request

Title
Prevents a gun from being used as part of dual wielding (the loop skips over it without any effect) or being dual wielded (will not initiate the loop)

## Why It's Good For The Game

Better way to do this if you do not want to use WEAPON_HEAVY (which REQUIRES two free hands to use a gun.)

## Changelog
:cl:
code: Gun flags have been added with a dual wielding exclusion flag.
/:cl:


Ghommie (changes requested):
I'm pretty sure the `WEAPON_MEDIUM` already does that, except it doesn't work the other way around and only stops the gun from being fired offhand. This seems pretty unnecessary.

PR #12297, opened 2020-05-18
Makes gun weight akimbo restrictions work the other way around.

Ghommie:
## About The Pull Request
Yea, This means you can't barrage your three pistols while plinking mobs with a mag rifle should you have two extra arms somewhat... But this is probably for the best until I code in inaccuracity if firing bigger guns not wielded with both hands (two handed component), but that'll also mean moving other functions away from attack_self.

## Why It's Good For The Game
Simplier alternative to #12293.

## Changelog
:cl:
tweak: gun weight checks work the other way around when akimbo-ing.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -215,7 +215,7 @@
 	var/loop_counter = 0
 
 	bonus_spread += getinaccuracy(user) //CIT CHANGE - adds bonus spread while not aiming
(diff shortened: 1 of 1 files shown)


silicons:
add some documentation to weapon weights, please, if you want it to be used more.
personally i'd rather everything be flagged but I suppose a whole flag variable is not necessary.. yet.

PR #12302, opened 2020-05-19
Skill calculation fixes, tweaks, antag and brain damage modifiers.

Ghommie:
## About The Pull Request
Reworked the skill action modifier formulas to use a base multiplier value summed by another multiplier value determined by a 0-1 score based on current value/level, maximum, threshold and competency multiplier, see files. This fixes surgery skill bonus going up to 3.5 instead of 2.5, and also makes negative multipliers more unlikely.

refactored skill trait flags into a list of text entries so I won't have to do it later once we bump the 32-bit limit.

Antags can now have skill modifiers to help them in their tasks.
Brain damage now has two negative skill modifier (they do stack) depending on its severity. Might change it in the future probably.

tweaked how surgery and tool skills to award experience based time spent performing the action rather than number of times that action was performed. Repeatable surgery steps award less xp.

## Why It's Good For The Game
Finalizing our own skills framework. Next steps should be reworking the popup into a tgui-next menu to help players navigate through the mess that's skill modifers and then flesh it out with more skills

## Changelog
:cl:
fix: fixed surgery skill providing up to 3.5x speed boost instead of 2.5x
add: Antags can now have skill modifiers.
/:cl:


silicons (approved):
looks good

PR #12315, opened 2020-05-20
Fixes Styptic Powder & SilverSulf OD effects

DeltaFire15:
## About The Pull Request
An overdose on styptic / silversulf was supposed to a) null out their healing, b) deal light liver damage and c) make the reagent metabolise out really fast.
While c) worked as intended, a) and b) were ONLY being called on the first tick (overdose_start), allowing an ongoing OD to heal as previously, which is... not the intended behaviour. This PR fixes all of that. (In addition, silversulf was dealing brute, not burn, as its OD effect, which I assume was a mistake so I fixed that aswell, as it wouldn't null its healing otherwise)

## Why It's Good For The Game
Fixes good.
Proofreading and testing is too.

## Changelog
:cl:
fix: fixed silversulf & styptic OD effects
/:cl:


DeltaFire15:
Aw fuck travis machine broke

Linzolle (approved):

PR #12337, opened 2020-05-22
Adds the "Space Diner" space ruin.

EmeraldSundisk:
## About The Pull Request

Puts #12224 to use. For those who don't like surprises (or need to see it for development reasons), an image of the ruin can be found **HERE**:
https://cdn.discordapp.com/attachments/336748540690825216/713496556468436992/SPOILER_spessdiner.PNG

If it sounds (or looks) familiar, it should - the "lore" is that it was accidentally translocated along with CogStation. It too has been updated to account for TGcode functionality, although not to the same degree as your standard station. Even so, it features a basic air system, culinary utilities, and most notably a teleporter.

(Fixed things, good to go now! Or...should be at least.)

## Why It's Good For The Game

The Space Diner functions sort of like a smaller Space Hotel - a relatively safe area of space the crew can easily visit if granted access to the teleporter room. However, the lack of ghost roles means latejoins won't automatically know it exists (and by that extension won't know there's another free teleporter out in space). Also it's a space ruin. Can't have enough of those.

## Changelog
:cl:
add: Adds the "Space Diner" space ruin.
/:cl:


silicons:
actually you can have enough of those

silicons:
but sure

PR #12355, opened 2020-05-24
Adds two swords that can be crafted.

ghost:
## About The Pull Request
Title. One is the wooden sword, and the other is its upgraded version, the steel sword.

## Why It's Good For The Game
More items for people to hit each other with.

## Changelog
🆑
add: The wooden sword and steel sword have been added as weapons that can be crafted
/🆑


MrJWhit:
A 16 force, sharp weapon that requires some wood and plasteel is _way_ too powerful.

ghost:
> A 16 force, sharp weapon that requires some wood and plasteel is _way_ too powerful.

I'll nerf it in that case.

MrJWhit:
:+1: looks good to me then

necromanceranne:
Why not make this use material datum and printable from the autolathe.

KathrinBailey:
You could make the wooden one deal stamina damage, it'd be like Mount & Blade training swords then.
Not saying you have to but heck, it'd be pretty neat. Or deal stamina and a small amount of brute. Not much stamina, just a little.
Either way cool PR +1 I wanna make craftable maces now.

YakumoChen:
I think we have craftable maces already

Trilbyspaceclone:
Craftable maces are a space ruin disk.
As for this pr 1+ I really like the idea. Seems neet

Detective-Google:
> A 16 force, sharp weapon that requires some wood and plasteel is _way_ too powerful.

16 force is only one more than a welder which you can print in droves. a 12 damage weapon has no reason to be crafted.

MrJWhit:
We've also been having issues previously with powercreep for craftable/improv weapons, a toolbox only deals 12 damage.

silicons:
maybe i'm anti fun but i'm failing to see the point of two specific recipes
do we have support for material datum based recipes yet, if so why aren't we using that, and what's the point of specifically defining a wooden sword with 6 force?

ghost:
I think I'm going to do something else with the sword concept actually, improve upon it. Opinions are too varied at the moment to think this would be good for the game.

PR #12376, opened 2020-05-27
Skewers no longer cause deadchat spam

DeltaFire15:
## About The Pull Request
The forced speech from the skewer triggered even if the skewee was dead, causing immense amounts of deadchat spam if someone repeatedly failed lifting someone else of a skewer. This fixes that issue.

## Why It's Good For The Game
.. do I really need to elaborate?
Fix man good.

## Changelog
:cl:
fix: Skewer-freeing-fail no longer forces speech if the skewee is dead.
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_structures/traps/brass_skewer.dm
+++ b/code/modules/antagonists/clockcult/clock_structures/traps/brass_skewer.dm
@@ -113,7 +113,7 @@
 		"<span class='danger'>You start tenderly lifting [skewee] off of [src]...</span>")
 		if(!do_after(user, 60, target = skewee))
 			skewee.visible_message("<span class='warning'>[skewee] painfully slides back down [src].</span>")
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/clockcult/clock_structures/traps/brass_skewer.dm
```diff
@@ -113,6 +113,8 @@
 		"<span class='danger'>You start tenderly lifting [skewee] off of [src]...</span>")
 		if(!do_after(user, 60, target = skewee))
 			skewee.visible_message("<span class='warning'>[skewee] painfully slides back down [src].</span>")
+			if(skewee.stat == DEAD)
```
Ghommie:
```suggestion
			if(skewee.stat >= UNCONSCIOUS)
```

Ghommie (approved):

PR #12392, opened 2020-05-29
chemistry/bio bags nerf and little pill volumetric tweak.

Ghommie:
## About The Pull Request
Title. chemistry and bio bags are now normal sized (can still be worn on the pocket slot).
The chemistry bag now uses the volumetric storage system (won't affect their UI since they have `display_numerical_stacking` TRUE). raised the ceiling for pills volume by one unit since I heard 25u pills are considered small and not tiny volume-wise.

## Why It's Good For The Game
First of all, chemistry and bio bags being tiny items is pretty much wrong for containers of that capacity (at least until a foldable component is made or something), moreso since we already gone a good length to enforce a stricter storage management design.
Second, I forgot about chemistry bags when I tweaked pill bottles.

## Changelog
:cl:
balance: chemistry/bio bags are now normal sized (can still be worn on the pocket slot though)
balance: chemistry bags now use the volumetric storage system. (doesn't affect their UI)
/:cl:


code/game/objects/items/storage/bags.dm
```diff
@@ -382,14 +382,14 @@
...
+	slot_flags = ITEM_SLOT_BELT|ITEM_SLOT_POCKET
 	resistance_flags = FLAMMABLE
 
 /obj/item/storage/bag/chemistry/ComponentInitialize()
 	. = ..()
 	var/datum/component/storage/STR = GetComponent(/datum/component/storage)
-	STR.max_combined_w_class = 200
-	STR.max_items = 50
+	STR.storage_flags = STORAGE_FLAGS_VOLUME_DEFAULT
+	STR.max_volume = 100
```
silicons: please put volumetric storage stuff in the defines file for sizes, with the old w_class system you're keeping track of 5 things
with volumetric storage you're keeping track of a big range of integers. i do not think there's anything to be lost by having it in one file for easy tweaks

code/modules/reagents/reagent_containers/pill.dm
```diff
@@ -27,9 +27,9 @@
 
 /obj/item/reagent_containers/pill/get_w_volume()
```
silicons: so why aren't we just using a linear scaling from a minimum of 1 to a maximum of [x] at [y] volume? the volumetric storage system fully supports decimals, you know.

Ghommie: well, I actually didn't know, so thanks for telling. I'll make it scale linearly for sure later.

code/game/objects/items/storage/bags.dm
silicons: also how many units is this supposed to be?
i'd rather not have chemistry be too nerfed in terms of reagent capacity considering pill bottles are borderline meme right now with how weird it works due to the thresholding.

PR #12394, opened 2020-05-29
Photosynthesis element.

Ghommie:
## About The Pull Request
Basically wood golems, podpeople and shadowpeople blended into an element.

## Why It's Good For The Game
I was bored. For eventual badminning and future trickeries.

## Changelog
Should have no impact in game (except shadowpeople being able to regen inside closets now which is fine imo, and maybe not taking post mortem damage from light sources, whatever)

### DIFF
--- a/code/controllers/subsystem/dcs.dm
+++ b/code/controllers/subsystem/dcs.dm
@@ -1,7 +1,6 @@
 PROCESSING_SUBSYSTEM_DEF(dcs)
 	name = "Datum Component System"
 	flags = SS_NO_INIT
-	wait = 20
 
 	var/list/elements_by_type = list()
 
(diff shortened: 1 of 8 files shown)


necromanceranne:
images (13)

code/controllers/subsystem/dcs.dm
```diff
@@ -1,6 +1,7 @@
 PROCESSING_SUBSYSTEM_DEF(dcs)
 	name = "Datum Component System"
 	flags = SS_NO_INIT
+	wait = 20
```
silicons: is locking dcs processing into 2 second ticking a good idea?
i think it might be better to atleast pass something like time between fires/time after last fire and just have stuff scale to world.time instead of relying on this, like how i made projectiles. 
or actually, no, not world.time, but scaling the stuff in the actual element to the wait variable of the subsystem
my concern isn't like, keeping timing, but rather what changes in the future will do if we need this to tick at 1 second at some point.

Ghommie: Yea, I'll just make the components/elements process with SSobj instead, since they are all supposed to process at 2 seconds intervals right now, not 1, my bad. :stuck_out_tongue:

PR #12397, opened 2020-05-30
Ports TGMC timer cooldowns

Rohesie:
Porting it to /tg/ as well.
For the lazy, a quick, cheap and dirty way of adding cooldowns.
Turned a random cooldown timer I found into this system to provide an example of use-case.

I'd wait for it to be merged on /tg/ first, after the reviews.


Rohesie:
This is ready and merged on tg

PR #12433, opened 2020-06-03
Powers Reimagined : Void Volt

DeltaFire15:
## About The Pull Request
Adds the Void Volt scripture to the scripture tier (tier two) of clockwork cultists.
It is a chanted scripture, which means it can be channeled for quite a bit of time for increased effects, or well, in this case, more 'pulses' of effects, to a maximum of _20_ per cast, with _1_ second of time between individual pulses after _5_ seconds of initial invocation time.
Every pulse, the chant drains power from any powered object within _8_ tiles of the invoker, draining _100KJ_ from guns such as cells and APCs, aswell as _1MJ_ from any other powered objects within range such as cells, mechs, borgs, etc., which is handled as _100KJ_ for power-feedback purposes.
After performing the drain pulse, the invoker suffers _0.005_ times the drained power amount as burn damage, leading to _0.5_ damage per drained object per pulse (in general). 15 times the suffered damage is then added to the clockwork cult's global power pool. Also, a maximum damage of 20 can be suffered by a singular pulse, to not have the chant be immediate suicide in powerrich areas. It takes about _10_ pulses to drain a gun or normal cell, and about _20_ - _40_ to drain a normal APC (depending on how well the powernet is working). APCs are shorted out of they end up empty after one of these pulses, though any other equipment suffers no ill effects aside from this.
This chant also allows for multiple (adjacent) cultists to chant together, giving the chant _2_ tiles of range increase per additional cultist involved, to a maximum of _2_ times the starting range, so _16_
When multiple cultists are chanting this chant together, the damage dealt is split evenly between all of them.

This PR also tweaks around the clockcult power-drain helper (which is used for nothing but void volt as of now), aswell as reordering scriptures of the scripture tier to be a bit more organised (with no gameplay effect), aswell as making additional cultists actually chant for multi-invoker chants.

## Why It's Good For The Game
Void Volt in the old design document looked like an interesting utility scripture to have in a cultists arsenal, and with this slightly modified implementation of it, it shoud provide an useful tool to cause localised power outages stationside, combat hostile mechs, or somewhat mitigate the effect of energy-weaponry being used in high amounts (though of course those can just be recharged).
All in all, an interesting, though somewhat situational, chant for clock cultists and ratvarian chaplains.

## Changelog
:cl:
add: Adds Void Volt to the tier two scriptures of clockies
tweak: Changed around the sorting of some clockwork spells (No actual gameplay impact)
tweak: Modifies the clockie power drain helper a bit, it was not being used prior to void volt anyway
tweak: Adds an interaction for chants being used with multiple casters
/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_helpers/clock_powerdrain.dm
+++ b/code/modules/antagonists/clockcult/clock_helpers/clock_powerdrain.dm
@@ -1,13 +1,14 @@
 //horrifying power drain proc made for clockcult's power drain in lieu of six istypes or six for(x in view) loops
-/atom/movable/proc/power_drain(clockcult_user, drain_weapons = FALSE) //This proc as of now is only in use for void volt.
+/atom/movable/proc/power_drain(clockcult_user, drain_weapons = FALSE) //This proc as of now is only in use for void volt
 	var/obj/item/stock_parts/cell/cell = get_cell()
 	if(cell)
 		return cell.power_drain(clockcult_user)
-	return 0 //Returns 0 instead of false to symbolise it returning the power amount in other cases, not TRUE aka 1
+	return 0 //Returns 0 instead of FALSE to symbolise it returning the power amount in other cases, not TRUE aka 1
 
 /obj/item/melee/baton/power_drain(clockcult_user, drain_weapons = FALSE)	//balance memes
 	if(!drain_weapons)
 		return 0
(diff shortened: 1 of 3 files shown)


lolman360:
100kw or 100kj?

DeltaFire15:
Right, powercells & stuff use joules, so yeah it's KJ / MJ for the drain.

PR #12441, opened 2020-06-04
Fixes addictions.

DeltaFire15:
## About The Pull Request
Sooooo, turns out ifs nested in switch statements ONLY support CONSTANT numbers, NOT variables. Therefore, I moved the switch statement back to using if-elses and, hooray, it works.

## Why It's Good For The Game
Fix man good maybe? Especially since addictions will never exit stage one nor be removed otherwise.

## Changelog
:cl:
fix: Fixes addictions
/:cl:


Ghommie (approved):
I thought it did support variables, how weird...

PR #12504, opened 2020-06-09
[WIP]Bob's Terrible Underwear Rework

BobJoga:
## About The Pull Request

**TLDR:** This PR replaces the current non-sensical way that underwear is handled with inventory based underwear, adding a new inventory slot for underwear and replacing the old underwear sprite accessories with actual items.

**More detailed explanation:**
image
*(ignore the weird shifted underwear slot icon, i've fixed it)*

As you can see, i messed around with the HUD until it had a layout i'm satisfied with. The current uniform slot has been replaced with the underwear slot, the uniform slot being moved to above that slot, and the neck slot has been shifted to the top right. The way underwear works now is pretty much like this: All underwear kinda works like accessories to each other, you can attach your bra to your panties to wear them in the same slot. The reason i did it like this is because adding three separate slots for each type of underwear seemed not only limiting, but also counter-intuitive since it would clutter the UI a lot and significantly complexify the code. I think i could make an expand UI button or toggle slot button if i went with such an approach, but i decided against it.

image
image

I'd say the PR is almost done, but there are a few issues which i'll list here that worry me which i'd say i'll need help with:
1. **Underwear colors don't display proper.**
Using the color var is out of the question due to the appending system i've decided on. I had to make a very asinine way of displaying clothing - essentially making the clothes not use their own icons - instead, i used an overlay, which can be colored without affecting the whole sprite of all appended parts + the "master" underwear which has them appended. This is obviously a terrible, asinine solution, which might weigh down on the server a lot and is generally inefficient to work with.

**My solution:**
What i'm probably going to do is implement some sort of variation of the polychromic element on the underwear. This isn't completely ideal, since it means i'll have to sift through every colorable underwear and get the proper HTML values, but it is what it is.

2. **Appended underwear doesn't calculate which mob overlay to use proper.**
While the "master" underwear will handle digitigrade sprites fine, the appended underwear might have problems. I've made some dumb, crackhead code to "fix" it, but i'm quite sure it won't handle every problem and will be quite buggy.

**My solution:**
I'll try and see if i can modify the human icon code so it can call the worn_overlays proc proper on each piece of appended underwear.

3. **The inventory slot doesn't update when first joining the round.**
While the slot works proper and displays everything well on the round and stuff, when first joining a game it isn't updated correctly, and you see it as being empty when it really isn't, until you fix that by unequipping and equipping the underwear again.

**My solution:**
This one truly stumps me. I have no idea how i could fix this one - it's quite minor but still annoying enough for me to really want a fix.

4. **A lot of the code is wacky and nonsensical.**
If you look at the code, you'll come to the conclusion that i am a buffoon, which - yes it's correct.
The way the global underwear list is handled is quite dumb, and the blacklisted underwear doesn't use a config but is set on the initialize underwear proc. This might lead to many oddities in the character setup and just general stupidity when handling the code. While i think i could get away with making underwear based on the loadout, it would be unfamiliar for players, and create general confusion, on top of making it a bit hard for spawned human mobs to get random underwear.

**My solution:**
If it's deemed appropriate by other coders and the community, i might go with the loadout solution and just make spawned human mobs not have underwear by default. This would make the code significantly less complex and easier to handle in general, and make it more like the rest of clothing, but i've detailed the sacrifices here.

5. **Genitals don't update properly.**
While taking off underwear will technically work eventually to display your fun parts, it seems to be generally wonky in updating and might result in oddities like your weiner not showing up after taking your boxers off.

**My solution:**
Currently not really any idea - i'll try and fuck around with underwear code to see if i can find a way to fix it.

## Why It's Good For The Game

Underwear code is a crime against humanity, and severely limits creativity. With proper underwear, it's easier to handle genital code, and it also means silly stuff like tactical panties in the uplink and etc. While it might stress the server a tiny bit more, i feel that the gain from it is significantly higher than the disadvantages.

## Changelog
:cl: Bob Joga
refactor: Underwear code has been entirely overhauled. The inventory now has a dedicated underwear slot and underwear is actually treated as clothing items. For more info, see PR #12504 on the github.
/:cl:


Investigator77:
Now add a second ear slot)))

PR #12509, opened 2020-06-11
slimes can change their body color using alter form

timothyteakettle:
## About The Pull Request
alter form has a new option which allows slimes to pick a body color

## Why It's Good For The Game
makes alter form nicer because you can change so many things but not your colour, which is quite weird.

## Changelog
:cl:
add: slimes can now change their color using alter form
/:cl:


timothyteakettle:
so uh, in response to comments about how it affects their stealth, they cannot change their name, and you can change literally everything else about them, then throw on some clothes, this shouldn't be an issue.

.. you can literally add/remove tauric bodies from slimes.

PR #12513, opened 2020-06-12
[s] Fixes a major HTML injection vulnerability

deathride58:
Title. Does as it says on the tin. This PR makes it so that what you type into the flavortext box is *exactly* what you get, instead of how it currently works where the flavortext box just lets you write raw html because it html_decode()s and just blindly saves and displays that. The only place flavortext gets html_decode()'d with this PR is when you try to modify it, where it's used to make sure the flavortext appears exactly as you wrote it from top to bottom.

## Changelog
:cl: Bhijn
fix: raw HTML can no longer be used in flavortext
/:cl:


Linzolle (approved):
oh lord

silicons:
QUIRKY

PR #12514, opened 2020-06-12
the abstract idea of shadows can no longer be radioactive

silicons:
## About The Pull Request

everyone has something called an emissive blocker, added to their visual contents.
the better fix is to move it out of contents but someone else can do that. this fix works because there's no POSSIBLE reason you would want it to be radioactive in the first place.

## Why It's Good For The Game

invisible radiation is bad.

## Changelog
:cl: kevinz000
fix: emissive blockers can no longer be radioactive.
/:cl:


Ghommie (approved):
I guess I'll have to do the same for the fov holders.

PR #12520, opened 2020-06-13
Lobotomy can no longer give magic resilience traumas 

silicons:
## About The Pull Request

It does 50 brain damage though, up to 100.

## Why It's Good For The Game

We're trying to move away from cloning as a revival method.
Having the only way to fix lobotomy-class traumas have a 50% chance of making someone's round potentially an unlivable hell forever is not a great thing.

## Changelog
:cl: kevinz000
balance: Lobotomy no longer has a 50% chance of giving you a nigh-unremovable trauma, but does 50 brain damage even on success. On failure, it will give you a lobotomy-class trauma.
/:cl:


Linzolle (changes requested):
update the description for lobotomy to take the changes into account

Linzolle (approved):

PR #12524, opened 2020-06-13
Update dependencies.sh

hugedogfart:
you stink youre gay


kappa-sama:
this is why open source is a good idea

PR #12548, opened 2020-06-16
ports several tg events including fugitives

timothyteakettle:
## About The Pull Request
this ports the following tg round events:

brain trauma: someone gets a random brain trauma on the station, values were changed so no magic resilience on the trauma is possible

fake virus: people get fake virus symptoms but no actual virus

wisdom cow: a cow that speaks wisdom appears, due to lack of tg skills, instead of awarding skill xp it is milkable for liquid wisdom, which is basically mannitol but slightly better (also this means you can vv cows/goats now to set what they produce honk)

stray cargo pod: random cargo supply pod falls to the station with its location announced, rare variant drops a syndicate supply pod instead containing uplink stuff

fugitives: several fugitives appear on station, after ten minutes a hunter ship appears, with ghost roles able to be taken to become a hunter and catch the fugitives!

## Why It's Good For The Game
more roundevents is good

## Changelog
:cl:
add: Adds the brain trauma event, where one player gets a random brain trauma!
add: Adds the wisdom cow event, where the wisdom cow appears on the station!
add: Adds the fake virus event, where people get fake virus symptoms.
add: Adds the stray cargo pod event, where a cargo pod crashes into the station.
add: Adds the fugitives event, where fugitives are loose on the station, and it's the hunters jobs to capture them.
/:cl:


timothyteakettle:
is it worth separating this into two PRs, one with fugitives, and one with all the others, just because fugitives is very large compared to the other events which are small

PR #12552, opened 2020-06-17
improves replica pods and botany fertilizer

timothyteakettle:
## About The Pull Request
chem dumping is gone:
saltpetre and diethlyamine now work like robust harvest, they stop random mutations, except they add to potency/production and yield respectively, except they work as compound modifiers, meaning they will actually improve the plant's stats
unstable mutagen is slightly less likely to give a positive stat upgrade when mutating a plant, to avoid it being used to cheese plant stats to maximum

replica pods don't suck anymore:
you no longer get turned into a podperson by replica pods unless the replica pod seed is at potency 90 or higher
maturation speed is now 20 from 10, meaning it takes even longer for replica pods to grow
all blood types work on replica pods now, meaning slimes can be replica podded
pods take clone damage depending on their potency, from 25 to 75 depending on how low potency is

## Why It's Good For The Game
killing chem dumping means you can't get every plant to high stats quickly without first getting high tier parts
this basically kills botany speedrunning while still allowing people to do what they'd normally do, just over a longer timeframe

making replica pods not turn you into a pod means pod cloning is now a perfect substitute for cloning

## Changelog
:cl:
tweak: diethylamine/saltpetre now apply modifiers to plants each time they're ready to be harvested instead of when they're applied
tweak: replica pods only make you a podperson when the seed is 90+ potency now
/:cl:


DeltaFire15:
Ehhh podcloning should have some tradeoff at least because with this it is better than cloning in all regards (time, no clonedam / mutations, ease of setup, harder to prevent as you only need a blood sample and good luck hunting down every last syringe of that, etc). The only (arguably) negative it had before was turning you into a podperson, but with this gone too it just seems a biit too stronk.

timothyteakettle:
this pr does not add any bonuses to replica pods, it simply assures that people come out as their own race, and makes them even slower to grow, becoming a podperson mechanically is an upside because it's outright stronger than every roundstart race, this does not buff pods in anyway whatsoever (even though I got told podpeople should be buffed)

edit: ill probably give them clone damage + make pods even slower, remove roundstart replica pod seeds from vendors so they have to be mutated

timothyteakettle:
removed replica pods from vendor, increased maturation to 20, added clone damage, made mutagen on average deal negative upgrades over positive ones (by dividing the positive value by 1.25 for all stat mutations)

davidchan13:
Are you trying to make botany as bad as it is on tg

timothyteakettle:
no but I want to stop chem dumping being a thing and to allow replica pods to not suck

the only good change they did was killing chem dumping so people can't speedrun botany at stupid speeds

you are still given epic freedom with the dna manipulator unlike with tg where it's extremely tedious

want high plant stats on every seed? wait for fucking parts instead of being able to cheese botany with a single chem dispenser

silicons (dismissed):

silicons:
good change, instant chemdumps are bad.
i would be interested in something that applies tray chemicals over time (which would be a flat buff from this pr), but the instant stuff should go.

timothyteakettle:
ill look at applying chems over time in a second pr with a bigger scope over botany and all the seeds that aren't replica pods

TheObserver-sys:
Like I said several times in the discord, what you are looking for, to prevent instant chem dumping, is an adoption of TG's fertilizer system, which I've also said, barring mutagen, is the only good goddamn thing about NuBotany that wasn't something I helped come up with.

TripleShades:
The first part I like, but the pod person changes I feel are a bit much, as nobody bumrushes pod person cloning in the first place? Keep it as it was except for the slime compatibility and clone damage, in my opinion.

timothyteakettle:
the point is to make podcloning similar to actual cloning in terms of downsides, except it works for snowflakey races now and stops people going 'but me dont wanna be pod1111'

TheObserver-sys:
So, having tested it, here's what I can say I like, and what I definitely dislike about it.

_The Good Stuff:_
1. Max Stats will essentially just take 10 doses of 5u saltpetre and dieth apiece, making it a little more friendly towards the chem dispenser/biogenerator. This is good, because doing starts off the biogen alone is agony currently. We should consider adding dieth to the biogen, and removing ammonia, so the biogen is capable of helping run botany fully.

2. Being able to restore normal people from pods is generally a nice thing, especially if we make it take a small bit of effort to grab one. Having the threshold for plant people to be so high is also not a terrible idea, if only because from a purely statistical standpoint, they are basically human+

_The Bad Stuff:_
The growth of stats are exceptionally slow, and really just make assembling a truly *strong* plant even more of a slog than it already was. Even with using base stat discs, it still takes a very long time to produce a good quality plant. This is definitely going to scale really poorly too, running head first into the problem of intense micro management until science gives you the means to make good discs, and Science tends to forget service about half of the time, unlike other departments.

Krysonism:
Only snowflakes and /tg/ validhunters do not want to be an awesome cabbage man.

timothyteakettle:
.. we're a snowflake server, there's hardly anyone who wishes to have their race forcibly changed into a plant, and you can still become an epic cabbage man if you so desire.

silicons:
stale, open a new pr if you want with things addressed
i'd recommend working with @TheObserver-sys on this, they had good ideas.

PR #12558, opened 2020-06-18
Drastically buffs punch hit chance

silicons:
## About The Pull Request

Also, fixes pugilist not actually overriding the chances.

From (stamina + brute damage) * 0.5 (in a typical punch fight this means 60% of hits miss with 50 stam and 50 brute, that's uh, kinda high! to max(stamina * 0.5, brute * 0.5) which in the same situation would be 35% chance to miss.

## Why It's Good For The Game

Yeah I get that punches are supposed to be a last resort thing and have some special handling but as is, it's a bit weak unless your target is already on the ground. While it's realistic for people to falter after ~5 punches traded on each side, it's a bit unfun to have basically all your hits be nullified within a few seconds.

@necromanceranne I would like your opinion on this :heart:

## Changelog
:cl: kevinz000
balance: calculations for punch hit chance has been drastically buffed in favor of the attacker.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1492,10 +1492,10 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
 
 		var/miss_chance = 100//calculate the odds that a punch misses entirely. considers stamina and brute damage of the puncher. punches miss by default to prevent weird cases
 		if(user.dna.species.punchdamagelow)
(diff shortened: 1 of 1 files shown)


necromanceranne:
Wouldn't this mean a pugilist kicking still has only a 10% chance? Pugilist is meant to override the baseline punch and kicks override everything else. Wouldn't it be more sensible to put kicking higher up on the checks and make the pugilist check the else if?

Also yeah sure I don't really mind either way.

silicons:
@necromanceranne :+1:

Linzolle (approved):

PR #12562, opened 2020-06-18
Lots of ports

SmArtKar:
## About The Pull Request

Here I am porting lots of things from bee and skyrat

## Why It's Good For The Game

New great stuff is always good!

## Changelog
:cl:
add: Added "Problem Computers" where you need to solve math tasks for cargo/RnD points
add: Added Autodoc, a machine which will automaticaly insert implants into you
add: Added Hardsuit Armblade, an illegal attachment for security
add: Added Hydraulic Blade, which is an armblade without ability to pry open things. Needs illegal tech too!
add: Added BS miners, machines able to passively mine some resources.
/:cl:


silicons:
Sorry, but we're probably not going to be accepting the majority of these:
Autodoc - Completely negates the requirement of having someone else insert said implants/organs/whatnot for you. Surgery is balanced on at the very least having a buddy system. We specifically removed self surgery a few years back and there's no reason to readd it now.
Bluespace miner - ... no. No. Would completely negate material balance and is unnecessary at this time.
Problem computers - Cookie clickers go explicitly against what we are trying to design for science and cargo, emphasis on science. While the current state of techwebs is pretty bad, adding something with the same complexity and "skill" requirements of Telescience is not going to happen.

Please keep in mind that both of those servers have a very different design and balance philosophy than us, and the codebase *we* branched off of, /tg/station, and I hate to say this but just because something is fun doesn't mean it deserves a place. While progress is somewhat slow, we'd rather play the slow game than throw balance out the window for some short-lived fun that will powercreep the everloving crap out of departments and the station.

MrJWhit:
I fully agree with what kevinz said, and as a side note, I don't understand what the armblade stuff is good for.

ghost:
yeah we already have fake armblades the funny tech hydroblade serves no purpose besides bloat.
and security doesn't really need hardsuit knives for anything and would probably either never get used or end up not being used when someone gets banned for stabbing people with it on a green...

ShadeAware:
Bring the Autodoc to Skyrat 👀

deathride58 (changes requested):
Mass porting like this isn't acceptable without damn good reason. Features are not a good reason to mass port features in a single PR. Atomize your changes.

PR #12572, opened 2020-06-21
Disables the stray syndicate crate event

DeltaFire15:
## About The Pull Request
This disables the subtype of the stray crate event that drops at least 30tc of syndicate gear on the station in a random location by setting its max occurances and weight to 0.
Left in the code for if it's needed by any admins.

## Why It's Good For The Game
Pretty sure that nonantags getting nullcrates was one of the reasons they were removed. Having them now spawn as a random event is.. not the best way to go about it, especially since 30tc of tot gear can cause a lot of annoyances.

## Changelog
:cl:
del: Syndicate crate event cannot fire as a random event anymore.
/:cl:

### DIFF
--- a/code/modules/events/stray_cargo.dm
+++ b/code/modules/events/stray_cargo.dm
@@ -85,8 +85,8 @@
 /datum/round_event_control/stray_cargo/syndicate
 	name = "Stray Syndicate Cargo Pod"
 	typepath = /datum/round_event/stray_cargo/syndicate
-	weight = 6
-	max_occurrences = 1
+	weight = 0
+	max_occurrences = 0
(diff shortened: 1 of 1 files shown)


Detective-Google:
good pr

timothyteakettle:
i want to make round events a config at some point anyway because it seems better that way but anyway:

if it's that much of an issue then i can understand it's disabling, but really it could just have its weight set to 1 (same rarity as slaughter demon), its max occurences is already 1, while i'm against having null crates as a thing, having an extremely rare round event spice up a round seems fine to me

Ghommie (approved):

PR #12574, opened 2020-06-21
Buying a 'random uplink item' can no longer force objective / class changes

DeltaFire15:
## About The Pull Request
Due to objective classes and the objective reroll being added, random item purchases could choose the 'reroll class' item, which is.. slightly unfortunate to whoever has it happen to them, as random item is often chosen for 'I still have tc, gimme some shit', not 'oh no I had 2 tc and wanted an item, but now I got weird objs instead'
This blacklists the 0tc contract reroll item from being able to be chosen by the random item 'bundle'

## Why It's Good For The Game
Having people get forced into another class when they just wanted a random item (or did the big mistake of hitting the button with 0 or just few tc, as the contract reroll is the only item with that cost besides the random item button)

## Changelog
:cl:
balance: Choosing a random item in your uplink will no longer sometimes reroll your contract.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items/uplink_bundles.dm
+++ b/code/modules/uplink/uplink_items/uplink_bundles.dm
@@ -193,6 +193,8 @@
 			var/datum/uplink_item/I = uplink_items[category][item]
 			if(src == I || !I.item)
 				continue
+			if(istype(I, /datum/uplink_item/bundles_TC/reroll)) //oops!
+				continue
 			if(U.telecrystals < I.cost)
 				continue
 			if(I.limited_stock == 0)


code/modules/uplink/uplink_items/uplink_bundles.dm
```diff
@@ -193,6 +193,8 @@
 			var/datum/uplink_item/I = uplink_items[category][item]
 			if(src == I || !I.item)
 				continue
+			if(istype(I, /datum/uplink_item/bundles_TC/reroll)) //oops!
```
Ghommie, line 196:
```suggestion
			if(I.type == /datum/uplink_item/bundles_TC/reroll) //oops!
```
I doubt we are making any subtype out of it.

code/modules/uplink/uplink_items/uplink_bundles.dm
```diff
@@ -193,6 +193,8 @@
...
+				continue
```
Ghommie, line 197: Actually, just remove this entry from the possible_items list afterward instead of typechecking for it every loop, more efficient. At least until a new variable to blacklist items from the pick random choice is made.

DeltaFire15: Hm, unsure on if that'd actually be more efficient than just moving this check a few lines down right before the adding it to the list, as checking every item that's going to be added to the list would have a worst-case and average case of n, whereas having to go through the entire list afterwards (because it's saved as items, not typepaths, so can't just use the -= operator easily), then removing the specific item that matches the type would have a average case of  something between n and 2 * n, worstcase being 2 * n, n being the amount of suitable items.

Ghommie: It seems I didn't make myself understood. What I meant to suggest was actually removing the reroll item after the whole `for` loop is done, BUT apparently I missed that `possible_items` keys are instances and not typepaths or text strings.
So TL;DR this is fine as is, I was distracted.

Ghommie (approved):

PR #12585, opened 2020-06-23
Two updates to Snow Snaxi with one balance change

MrJWhit:
## About The Pull Request

Three things being changed:
-Removing the air alarm in tcomms.
-Removing a trash bin in genetics and mining.
-Makes another cargolathe in the cargo area.

## Why It's Good For The Game

Air alarm removal- It goes off roundstart because it's too cold, there's nothing that can be done about it, and Boxstation and Deltastation (the only two maps I did a look in) don't have air alarms in the Tcom Stat area, so it feels fine.
Trash bin- Those were temporarily placed there while disposals was being introduced. These got skipped over, no biggie, but just popping them.
Cargolathe- There isn't a way for cargo techs to actually obtain mats without breaking into mining.

## Changelog
:cl:
del: Removed air alarm in Snow Snaxi in Tcomms Sat
del: Removed trash bins in genetics and mining
tweak: Gives cargo techs a cargolathe
/:cl:


MrJWhit:
My everything broke, I'm just going to delete and remake.

PR #12592, opened 2020-06-24
felinids nya when tabled

timothyteakettle:
## About The Pull Request
felinids nya when tabled

## Why It's Good For The Game
so we can laugh at ~~kevin~~ felinids

## Changelog
:cl:
add: felinids now nya when tabled
/:cl:

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -134,6 +134,8 @@
 	if(!ishuman(pushed_mob))
 		return
 	var/mob/living/carbon/human/H = pushed_mob
+	if(iscatperson(H))
+		H.emote("nya")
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/tables_racks.dm
```diff
@@ -134,6 +134,8 @@
 	if(!ishuman(pushed_mob))
 		return
 	var/mob/living/carbon/human/H = pushed_mob
+	if(iscatperson(H))
```
LetterN, line 137:
```suggestion
	if(iscatperson(H) && prob(50))
```

LetterN (changes requested):

code/game/objects/structures/tables_racks.dm
timothyteakettle: considered it but id rather they just always nya when tabled because it's funny and not gameplay impacting

nik707 (approved):

Issue #12602, opened 2020-06-25
ChemMaster pill type button,and vending mechine icon are missing.


Sonic121x:
## Reproduction:
1:Chem master.
![Chem](https://user-images.githubusercontent.com/64410541/85647531-3f9e9900-b6d1-11ea-8a13-030ef657d66b.png)
2:Vending mechine.
![Vend](https://user-images.githubusercontent.com/64410541/85647532-40cfc600-b6d1-11ea-813e-168f16a138d5.png)
other icon are fine.
![Tech](https://user-images.githubusercontent.com/64410541/85647535-41685c80-b6d1-11ea-915c-cbfbdb2b2137.png)

Ghommie:
I'm pretty sure something gone wrong with the new asset cache PR.

LetterN:
fuck, fixing time then!

PR #12609, opened 2020-06-25
(testmerge for now) Revert "asset cache + better minimaps"

Ghommie:
Apparently the asset cache PR broke more than just a few things (see #12602). We should go for a fix soonish.


Arturlang:
TGUI 3.0 PR has a working asset cache port

PR #12610, opened 2020-06-26
Lying down now respects the direction you're facing

Detective-Google:
## About The Pull Request
if you're facing right your character will fall to the right
if you're facing left your character will fall to the left
if you're facing north or south you will have the old behavior of flopping randomly between right and left.

## Why It's Good For The Game

it can be a little mood killing to flop about randomly 6 times for the 50% chance to land on the right direction for your definitely not erp(tm)

## Changelog
:cl:
tweak: Lying down is better 
/:cl:

### DIFF
--- a/code/modules/mob/living/living_mobility.dm
+++ b/code/modules/mob/living/living_mobility.dm
@@ -96,14 +96,13 @@
 		mobility_flags &= ~MOBILITY_STAND
 		setMovetype(movement_type | CRAWLING)
 		if(!lying) //force them on the ground
-			if(src.dir == 1)
-				lying = pick(90, 270)
-			if(src.dir == 2)
-				lying = pick(90, 270)
-			if(src.dir == 4)
-				lying = 90
-			if(src.dir == 8)
-				lying = 270
+			switch(dir)
+				if(NORTH, SOUTH)
+					lying = pick(90, 270)
+				if(EAST)
+					lying = 90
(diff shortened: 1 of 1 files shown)


Detective-Google:
I tested this with different types of mobs and things and none of them seemed to fuck everything up and cause runtimes

code/modules/mob/living/living_mobility.dm
```diff
@@ -96,7 +96,14 @@
...
 		if(!lying) //force them on the ground
-			lying = pick(90, 270)
+			if(src.dir == 1)
+				lying = pick(90, 270)
+			if(src.dir == 2)
+				lying = pick(90, 270)
+			if(src.dir == 4)
+				lying = 90
+			if(src.dir == 8)
+				lying = 270
```
Ghommie:
```suggestion
			switch(dir)
				if(NORTH, SOUTH)
					lying = pick(90, 270)
				if(EAST)
					lying = 90
				else //West
					lying = 270
```

Ghommie (changes requested):

code/modules/mob/living/living_mobility.dm
Detective-Google: @Ghommie wouldn't it make more sense for North/South to be the else condition?

Ghommie (approved):

PR #12619, opened 2020-06-26
slimes blood colour is their body colour

timothyteakettle:
## About The Pull Request
this adds in support for custom blood colours by having your blood colour be defined by a var in your species datum
it also makes it so when updating a character's body, if the character is a slime, their blood colour is updated to their body colour

## Why It's Good For The Game
better character customization is good

## Changelog
:cl:
add: support for custom blood colours implemented, slimes blood colour now equivalent to their body colour
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/update_icons.dm
+++ b/code/modules/mob/living/carbon/update_icons.dm
@@ -193,9 +193,6 @@
 
 /mob/living/carbon/update_body()
 	update_body_parts()
-	// snowflakey thing: update blood colour if a slimeperson
-	if(isjellyperson(src))
-		dna.species.exotic_blood_color = "#" + dna.features["mcolor"]
 
 /mob/living/carbon/proc/update_body_parts()
 	//CHECK FOR UPDATE
--- a/code/modules/mob/living/carbon/update_icons.dm
+++ b/code/modules/mob/living/carbon/update_icons.dm
@@ -68,7 +68,7 @@
 	var/dam_colors = "#E62525"
 	if(ishuman(src))
 		var/mob/living/carbon/human/H = src
-		dam_colors = bloodtype_to_color(H.dna.blood_type)
+		dam_colors = H.dna.species.exotic_blood_color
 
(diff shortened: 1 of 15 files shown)


Ragolution:
Does this effect the coloration of blood bags at all?

code/modules/mob/living/carbon/update_icons.dm
```diff
@@ -193,6 +193,9 @@
 
 /mob/living/carbon/update_body()
 	update_body_parts()
+	// snowflakey thing: update blood colour if a slimeperson
+	if(isjellyperson(src))
+		dna.species.exotic_blood_color = "#" + dna.features["mcolor"]
```
Ghommie: Don't you have anywhere else (like species/proc/handle_mutant_bodyparts) to move this string of code to?

Ghommie (dismissed):

code/modules/mob/living/carbon/update_icons.dm
timothyteakettle: ill try moving it into there, didn't think about that

timothyteakettle:
also before this is merged I want to attempt to fix a blood-related issue with clothing as it probably fits within the scope of all the other blood-related changes i'm making here

timothyteakettle:
also in reply about blood bags, no, this does not affect them currently, they'll fill up using the default blood colours
I could probably make it so it updates the colours to whatever the most recent blood added is though, or attempt to make it blend the existing colours

edit: I stand corrected it does change the colour of blood bags!

code/modules/mob/living/carbon/update_icons.dm
timothyteakettle: placed it in the handle_mutant_bodyparts method for jellypeople

timothyteakettle:
it now applies the correct colour when splashing blood from beakers onto turfs, and splashing blood onto clothing

Ghommie (approved):

PR #12625, opened 2020-06-27
maybe fixes edge cases where half the server has their client input system broken

silicons:
this is tested

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -268,9 +268,7 @@ GLOBAL_LIST_EMPTY(external_rsc_urls)
 	else
 		prefs = new /datum/preferences(src)
 		GLOB.preferences_datums[ckey] = prefs
-	if(SSinput.initialized)
-		set_macros()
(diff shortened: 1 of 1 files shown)


code/modules/client/client_procs.dm
```diff
@@ -268,9 +268,7 @@ GLOBAL_LIST_EMPTY(external_rsc_urls)
 	else
 		prefs = new /datum/preferences(src)
 		GLOB.preferences_datums[ckey] = prefs
-	if(SSinput.initialized)
-		set_macros()
-	update_movement_keys(prefs)
+	addtimer(CALLBACK(src, .proc/ensure_keys_set), 0)	//prevents possible race conditions
```
Ghommie, line 271:
```suggestion
	INVOKE_ASYNC(src, .proc/ensure_keys_set)	//prevents possible race conditions
```

Ghommie (changes requested):

Ghommie (approved):
Was told the callback is run on the successive tick, not on the same one. So all fine.

PR #12632, opened 2020-06-28
Ups melee stam cost multiplier from 0.8 back to 1, made stam cost for throwing mobs scale with mob_size

Ghommie:
## About The Pull Request
Now that I loosened up the tediousness of the combat system a while ago, allowing people to keep fighting at low stamina albeit longer delays and lower damage, I think it's fine to undo the melee stamina cost reduction from #9532

## Why It's Good For The Game
See above, "soft stam crit" is not an equivalent of being a sitting duck anymore..

## Changelog
:cl:
balance: The stamina cost multiplier for swinging melee weapons against mobs has been brought back to 1 from 0.8
balance: The stamina cost for throwing mobs now scales with their mob size variable.
/:cl:

### DIFF
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -269,12 +269,12 @@ GLOBAL_LIST_INIT(shove_disarming_types, typecacheof(list(
 
 //stamina cost defines.
 #define STAM_COST_ATTACK_OBJ_MULT	1.2
-#define STAM_COST_ATTACK_MOB_MULT	0.8
+#define STAM_COST_ATTACK_MOB_MULT	1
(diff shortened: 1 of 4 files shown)


ShadeAware:
You made the combat system less tedious... only to turn right back around and make it even more tedious in just a different way? I never understood your thought process behind these "Fuck your fun" style PRs.

Ghommie:
On second thought, reducing the stamina recovery as a whole can lead to more likely overextertions from many aspects of the game, while I was mainly focused on melee.
Updated the PR accordingly.

ShadeAware:
This is **far** better than it was. 👍

PR #12636, opened 2020-06-28
[testmerge only] slime players are mad at me

silicons:
not even sure if this fixes it but this was the last thing that touched colors so it can't hurt


Ghommie:
I'm already fixing the jelly people issue in #12633, no need for a temp revert.

timothyteakettle:
also #12637 fixes all the blood related issues

PR #12641, opened 2020-06-29
machine construction no longer accesses merge_type after stack deletion

silicons:
may or may not fix a bug, probably a good thing anyways.


Ghommie:
Issue properly fixed in #12645.

PR #12648, opened 2020-06-29
bluespace tray + bluespace jar

timothyteakettle:
## About The Pull Request
first change: adds bluespace trays, able to hold 30 items instead of 7, printable from the service lathe, researchable through science

second change: adds bluespace jars, a reskin of the pet carrier which allows human sizes inside, smashes when thrown releasing whats inside, taking longer to load than a regular pet carrier, and only holding one mob maximum
(also reworked pet carriers to by default check move resist and hostility of the mob)

image
image

## Why It's Good For The Game
small qol for cooks and the jar is just generally memey

## Changelog
:cl:
add: bluespace tray added, allowing twice as many items as the regular tray, printable at the service lathe, researched through science
add: bluespace jar added, a kind of pet carrier that allows human sized mobs inside, and smashes when thrown, researched and printed through science
/:cl:


Ragolution:
I can understand the tray but that jar seems OP as hell and similar to the old pokeballs that were removed; As for naming conventions, why not a Metamaterial Tray? Bluespace is pretty down there for a +7 item carry cap.

timothyteakettle:
the jar is essentially a weaker version of a bsbb, can only hold one thing, takes far longer to load things into, takes less time to break out of, the only difference is it smashes when thrown and won't suffocate people inside it (also anything that can't go in a bsbb won't go in the jar)

the bluespace tray doubles the capacity, it seems like a small difference just because the regular trays capacity is tiny

davidchan13:
How bout some bluespace buckets for botany?

Ragolution:
should add a resist verb for players to get out, otherwise it's clowncar

timothyteakettle:
you are able to break out just like how you can break out of a bsbb except in less time, if you would look at the code (or my prior comment which mentioned it)

i considered buckets but they have no special behaviour so just ask for bluespace beakers or have them added to the service lathe

Ragolution:
Reading is lame and reading code is lamer. Going by your 'resists' thing I assumed it was a check for simplemobs, but if you have one that's a good. Elsewise, I didn't read the reply to the other person. B)

~~Alternatively: Make an admin one that's clowncar.~~

I still have concerns on whether or not this is good by balance standards, but that's up to Memetainers. My opinion still stands on giving these as Metamaterial trays since they're already limited to 14 items and see the most use in the kitchen.

Owai-Seek:
14 seems pretty low for bluespace tech considering plant bags hold, what, 60 plants? Honestly I'd take any content for service anymore. Bless.

timothyteakettle:
plant bags are a good point, guess ill make them fit more items because there's really no harm in letting them hold a tonne of foods, especially as tray damage scaling doesn't seem to be a thing anymore

timothyteakettle:
since the testmerge i got some feedback and made some changes: i've halved the resist time so it's now ten seconds to escape, compared to 20 of a bluespace bodybag/regular pet carrier, and the air inside should now be returning enough oxygen to not choke

Ragolution:
GIVE ME AN ADMIN CLOWNCAR ONE

timothyteakettle:
you can just vv ones escape time to be an impossibly large number and load time to be 0

timothyteakettle:
weird edgecase where exiting could get you stuck inside another object, remnant bug from pet carrier code which the bluespace jar uses, i changed it so exiting a pet carrier (and thus jar) works the same as a bsbb by getting the turf of the jar which should stop the issue

code/game/objects/items/pet_carrier.dm
```diff
@@ -192,9 +206,51 @@
...
+/obj/item/pet_carrier/bluespace/update_icon_state()
+	if(open)
+		icon_state = "bluespace_jar_open"
+	else
+		icon_state = "bluespace_jar"
+
+/obj/item/pet_carrier/bluespace/throw_impact()
+	//delete the item upon impact, releasing the creature inside (this is handled by its deletion)
+	if(occupants.len)
+		src.loc.visible_message("<span class='warning'>The bluespace jar smashes, releasing [occupants[1]]!</span>")
```
silicons: no src.

code/game/objects/items/pet_carrier.dm
```diff
@@ -192,9 +206,51 @@
...
+	qdel(src)
+	playsound(src, "shatter", 70, 1)
+	..()
```
silicons: . = ..() and qdel below tbh

code/game/objects/items/pet_carrier.dm
```diff
@@ -192,9 +206,51 @@
...
+		icon_state = "bluespace_jar_open"
+	else
+		icon_state = "bluespace_jar"
+
+/obj/item/pet_carrier/bluespace/throw_impact()
+	//delete the item upon impact, releasing the creature inside (this is handled by its deletion)
+	if(occupants.len)
+		src.loc.visible_message("<span class='warning'>The bluespace jar smashes, releasing [occupants[1]]!</span>")
+	qdel(src)
+	playsound(src, "shatter", 70, 1)
```
silicons, line 247: playsound after qdel is bad

code/game/objects/items/pet_carrier.dm
```diff
@@ -192,9 +206,51 @@
...
+	..()
+
+/obj/item/pet_carrier/bluespace/handle_internal_lifeform()
```
silicons: do what cloning did and have a static gasmix instead of this
also consider supporting non humans/stuff that breathe other gas but idc

timothyteakettle: instead decided to go and use traits so they're resistant while in the jar, because it's a far more efficient method for the pretty-much-same-effect

timothyteakettle:
also as an extra note I made it small now because it's literally weaker than bsbb in load time (4 seconds vs instant), escape time (10 seconds vs 20), and capacity (1 creature vs god-knows-how-many)

PR #12662, opened 2020-07-01
makes bows and arrows more feasible improv weapons

ghost:
## About The Pull Request
1) pipe bow crafting time lowered from minimum of **65** seconds spent in crafting menu to **20**
2) ashwalker bow crafting time lowered to 12 seconds from 20, because the quiver takes 8 seconds to make.
3) all arrow crafting times lowered from **3** seconds to half a second, excluding ashwalker arrows which require **1.5** seconds
4) bowstring crafting time lowered from **20** seconds to **5** seconds
5) station bow/arrow do not require **durathread** to craft; as they are improv weapons, they instead use cloth, a material that can be scavenged
6) ashwalker sandstone slab requires **20** seconds to craft instead of _**60**_. really? 60 seconds? they spawn with at least 3 of those off of memory and they aren't consumable...

## Why It's Good For The Game
maybe they'll get used if it's feasible to make them in the first place. i don't want to talk about the fact the pipe bow does >>>>>>>>>15<<<<<<<<<< damage but it's fine since it's a non-gun ranged improvised weapon

## Changelog
:cl:
add: cloth string to replace durathread string
del: durathread string
balance: All bows and arrows have had crafting times significantly reduced, coming out at up to 6 times faster crafting speeds. Improvised bows no longer require durathread; instead, they use cloth materials.
/:cl:

### DIFF
--- a/code/datums/components/crafting/guncrafting.dm
+++ b/code/datums/components/crafting/guncrafting.dm
@@ -9,7 +9,7 @@ k// PARTS //
 	icon_state = "riflestock"
 
 /obj/item/weaponcrafting/string
-	name = "bowstring"
+	name = "wound thread"
 	desc = "A long piece of thread with some resemblance to cable coil."
 	icon_state = "durastring"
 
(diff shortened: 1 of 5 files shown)


Trilbyspaceclone:
They are not meant to be feasible, when I made them they were balanced around being trash weapons.

ghost:
> They are not meant to be feasible, when I made them they were balanced around being trash weapons.

why should i care

silicons:
@Trilbyspaceclone not being feasible weapons is fine
taking 60+ second sto craft makes it a bad meme

PR #12679, opened 2020-07-03
Ok ok hear me out: Implements unarmed parry.

silicons:
## About The Pull Request

**Only works on unarmed attacks.**

Allows humans to parry punches, counter-punching with 100% hit chance and staggering the attacker if it's perfect.
4 stamina per use, no downsides other than a 3 second cooldown for doing this and not allowing punching for a full second after you initiate it.

known bugs: doesn't work for boxing gloves because it overrides the usual unarmed attack to bypass some things like clickdelays.

## Why It's Good For The Game

Uhhhhh.

## Changelog
:cl:
add: Unarmed parry is now a thing.
/:cl:


NopemanMcHalt:
~~Ability to two-hand every single item when?~~
~~Ability to two-hand bare arms when?~~
~~Citadel fully becoming Dark Souls when?~~

ghost:
could also attach unarmed parrying to pugilism and make pugilism slightly more common; if i'm remembering correctly boxing should give pugilism since it's a martial art, so that would let you parry with boxing gloves. this could also allow you to change cqc's RNG block/counter in throw mode to a stronger parry.

if unarmed parry was attached to pugilism, then it should probably be slightly buffed (since pugilism is relatively uncommon currently) and could be used as a large gap in power if you have enough skill. 

perhaps later on security officers could be spawned with pugilism, but i don't really know how to feel about sec being uniquely superior to other jobs in ways that can't be directly replicated

Linzolle (changes requested):
There's a weird interaction where if you try to parry while holding an item that cannot normally parry you instead perform the unarmed parry, with the item in your hands. I don't think it actually uses the item's force to damage with the parry but its still a very odd interaction and would prefer if it didn't work like that, unless that's intended.

necromanceranne:
Parry with foot.

Linzolle (approved):

PR #12692, opened 2020-07-04
Sheep + Loom Ports

Owai-Seek:
## About The Pull Request

Ports Loom Code from Yogstation
Ports Sheep from Yogstation
https://github.com/yogstation13/Yogstation/pull/6972

Adds a variable for sheets being usable with the loom, instead of said items having to be a subtype of cotton.
Sheep can be ordered from cargo.
Sheep can be sheered every ~3 minutes for wool.
4 Wool turns into a roll of cloth. (Could be changed later.)
Sheep can be milked.

## Why It's Good For The Game

Honestly, I just like having more animals. I think the loom stuff makes a lot more sense, as it doesn't rely on something being a cloth subtype. It also paves the way for possible wool products to be made through cargo, or maybe even the Cheese system Yogs has.

## Changelog
:cl:
add: Sheep
tweak: Replaced TG loom code with Yogstation Loom code.
imageadd: Sheep and wool stack icons.
/:cl:

dreamseeker_Ojeh6gOhQN


code/modules/mob/living/simple_animal/friendly/farm_animals.dm
```diff
@@ -354,6 +354,89 @@
...
+			icon_dead = "sheep_sheared_dead"
+			new /obj/item/stack/sheet/wool(get_turf(src), 16)
+			if(stat == CONSCIOUS)
+				icon_state = icon_living
+			else
+				icon_state = icon_dead
+			addtimer(CALLBACK(src, .proc/generateWool), 3 MINUTES)
+
+		return
+	..()
```
silicons:
```suggestion
	return ..()
```

code/modules/mob/living/simple_animal/friendly/farm_animals.dm
```diff
@@ -354,6 +354,89 @@
...
+			return
+		user.visible_message("[user] starts to shave [src] using \the [O].", "<span class='notice'>You start to shave [src] using \the [O]...</span>")
+		if(do_after(user, 50, target = src))
+			if(shaved)
+				user.visible_message("[src] has already been shaved!")
+				return
+			user.visible_message("[user] shaves some wool off [src] using \the [O].", "<span class='notice'>You shave some wool off [src] using \the [O].</span>")
+			playsound(loc, 'sound/items/welder2.ogg', 20, 1)
+			shaved = TRUE
+			icon_living = "sheep_sheared"
```
silicons, line 422: this and the icon stuff below it should probably be inserted into update_icon_state() or whatever it's called now logic @Owai-Seek 
so that proc would update the icon based on if it's sheared or not
and in the shearing proc (here) you just call update_icon() which will call both update_overlays() which is irrelevant to you/for this and also update_icon_state() which will update the icon state

code/modules/mob/living/simple_animal/friendly/farm_animals.dm
```diff
@@ -354,6 +354,89 @@
...
+		if(shaved)
+			to_chat(user, "<span class='warning'>The sheep doesn't have enough wool, try again later...</span>")
+			return
+		user.visible_message("[user] starts to shave [src] using \the [O].", "<span class='notice'>You start to shave [src] using \the [O]...</span>")
+		if(do_after(user, 50, target = src))
+			if(shaved)
+				user.visible_message("[src] has already been shaved!")
+				return
+			user.visible_message("[user] shaves some wool off [src] using \the [O].", "<span class='notice'>You shave some wool off [src] using \the [O].</span>")
+			playsound(loc, 'sound/items/welder2.ogg', 20, 1)
```
silicons:
```suggestion
			playsound(src, 'sound/items/welder2.ogg', 20, 1)
```

code/modules/mob/living/simple_animal/friendly/farm_animals.dm
```diff
@@ -354,6 +354,89 @@
...
+			shaved = TRUE
+			icon_living = "sheep_sheared"
+			icon_dead = "sheep_sheared_dead"
+			new /obj/item/stack/sheet/wool(get_turf(src), 16)
+			if(stat == CONSCIOUS)
+				icon_state = icon_living
+			else
+				icon_state = icon_dead
+			addtimer(CALLBACK(src, .proc/generateWool), 3 MINUTES)
+
```
silicons, line 430: get rid of the extra line because i'm nitpicky and like code being pretty

code/modules/mob/living/simple_animal/friendly/farm_animals.dm
```diff
@@ -354,6 +354,89 @@
...
+/mob/living/simple_animal/sheep/attackby(obj/item/O, mob/user, params)
+	if(stat == CONSCIOUS && istype(O, /obj/item/reagent_containers/glass))
+		udder.milkAnimal(O, user)
+		return TRUE
+	else
+		return ..()
+
+/mob/living/simple_animal/sheep/Life()
+	. = ..()
```
silicons:
```suggestion
/mob/living/simple_animal/sheep/BiologicalLife()
	if(!(. = ..()))
		return
```

silicons: ^ the above will also make it so you do'nt need the stat == CONSCIOUS check, as biologicallife returns false if the thing is dead or deleted

code/modules/mob/living/simple_animal/friendly/farm_animals.dm
```diff
@@ -354,6 +354,89 @@
...
+				icon_state = icon_living
+			else
+				icon_state = icon_dead
+			addtimer(CALLBACK(src, .proc/generateWool), 3 MINUTES)
+
+		return
+	..()
+
+/mob/living/simple_animal/sheep/proc/generateWool()
+	if(stat == CONSCIOUS)
```
silicons, line 435: early return (this mean if stat is not conscious return immediately instead of putting the code below into the block with the stat check)

code/game/objects/structures/loom.dm
```diff
@@ -10,32 +8,25 @@
...
-			user.show_message("<span class='notice'>You weave \the [W.name] into a workable fabric.</span>", MSG_VISUAL)
-	return TRUE
+/obj/structure/loom/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/stack/sheet))
+		var/obj/item/stack/sheet/W = I
+		if(W.is_fabric && W.amount > 3)
+			user.show_message("<span class='notice'>You start weaving the [W.name] through the loom..</span>", 1)
+			if(W.use_tool(src, user, W.pull_effort))
+				new W.loom_result(drop_location())
+				user.show_message("<span class='notice'>You weave the [W.name] into a workable fabric.</span>", 1)
```
silicons, line 23: both this and above, use \the intead of the as \the makes byond automatically do it if needed (pretty sure you don't need it at all but to be safe \the works)
also for show_message you might want to do TRUE instead of 1 if it's a boolean for the second argument

code/game/objects/structures/loom.dm
```diff
@@ -10,32 +8,25 @@
...
-	if(W.use_tool(src, user, W.pull_effort))
-		if(W.amount >= FABRIC_PER_SHEET)
-			new W.loom_result(drop_location())
-			W.use(FABRIC_PER_SHEET)
-			user.show_message("<span class='notice'>You weave \the [W.name] into a workable fabric.</span>", MSG_VISUAL)
-	return TRUE
+/obj/structure/loom/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/stack/sheet))
+		var/obj/item/stack/sheet/W = I
+		if(W.is_fabric && W.amount > 3)
```
silicons, line 19: early return, instead of putting the loom weaving logic below the block that checks check that if it isn't fabric or its amount isn't 3 tell the user that their thing isn't valid and return immediately

silicons: also while i'm here, the 3 show messages would likely be better as to_chat, unless you really want it so blind people can't use it but i don't care either way, either way works.

code/game/objects/structures/loom.dm
```diff
@@ -10,32 +8,25 @@
...
+			user.show_message("<span class='notice'>You start weaving the [W.name] through the loom..</span>", 1)
+			if(W.use_tool(src, user, W.pull_effort))
+				new W.loom_result(drop_location())
+				user.show_message("<span class='notice'>You weave the [W.name] into a workable fabric.</span>", 1)
+				W.amount = (W.amount - FABRIC_PER_SHEET)
+				if(W.amount < 1)
+					qdel(W)
```
silicons:
```suggestion
				W.use(FABRIC_PER_SHEET)
```
also why does this use FABRIC_PER_SHEET instead of 3 when the code above checks for amount over 3?

silicons (changes requested):
the above, if you need help ping me on discord you probably know who i am

Owai-Seek:
I basically just copied it over and tested everything to make sure if it worked/didn't bug out. It'll take me a bit to read through the requested changes and figure out how to do them since I can only -kind of- understand code.

code/modules/mob/living/simple_animal/friendly/farm_animals.dm
Ghommie: mobs still use their old update icons procs iirc.

code/modules/mob/living/simple_animal/friendly/farm_animals.dm
Ghommie: `if(!(. = ..()))` looks like something out of a loosely esoteric programming language though.

Ghommie:
Solve conflicts.

silicons:
stale

silicons:
@Owai-Seek

PR #12722, opened 2020-07-09
lowers wisdom cow's spawnrate to standard and makes it 66% less spammy

ghost:
## About The Pull Request
title

## Why It's Good For The Game
circuit spambots weren't enough
also liquid wisdom is very poggy

## Changelog
:cl:
tweak: wisdom cow is half as common and is wise enough to lag the server 66% less
/:cl:


timothyteakettle:
speak chance is a mistake from while I was porting it (I stuck it up way higher than it was to test it), weight is same as tg although because they have way more round events than us, it probably means it doesn't spawn as much

PR #12736, opened 2020-07-11
hey let's not have 25% micro borgs going around killing people

silicons:
should fix the expand module exploit


Ludox235:
speedmerge

Linzolle (approved):
finally

PR #12745, opened 2020-07-12
somewhat codifies some repository procedures and maintainer guidelines that haven't been written down anywhere

silicons:
alternatively we throw these on the wiki/rules page but i figure we might as well put it in github because that's where we work and it doesn't make sense to put maintainer stuff on a place coders aren't going to look at half the time (considering a lot of coders don't necessarily play here actively)


Linzolle:
looks good to me.

PR #12750, opened 2020-07-12
replaces plushies in the loadout menu with a choice box

timothyteakettle:
## About The Pull Request
removes non-chaplain plushies in the loadout menu and places a choice box that lists out every plushie including snowflakes
awakened plushie and the non-hugbox version of the narsie plushie are not selectable

choice box is a choice beacon but without the pod falling, it just gives you the thing in your hands

also choice beacons got slightly optimised so they save the list of contents instead of recalculating it every time they're used

## Why It's Good For The Game
slight optimisation, removes loadout bloat, this is a small qol thing

## Changelog
:cl:
add: plushies in the loadout have been replaced with a box that lets you choose one instead
/:cl:

### DIFF
--- a/code/game/objects/items/miscellaneous.dm
+++ b/code/game/objects/items/miscellaneous.dm
@@ -184,6 +184,9 @@
 		//snowflake plush
 		var/obj/item/toy/plush/snowflake_plushie = new(get_turf(M))
 		snowflake_plushie.set_snowflake_from_config(choice)
+		M.temporarilyRemoveItemFromInventory(src, TRUE)
+		M.put_in_hands(new choice)
+		qdel(src)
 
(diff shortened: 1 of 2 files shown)


timothyteakettle:
if there's any donator-only plushies or stuff like that, tell me and ill make a proper blacklist that it can't choose from (also if picking snowflake plushies is an issue because of donator loadouts having some specific custom plushies in it, i could just make it regular ones only shrug)

Strazyplus:
**Very nice**, only con is that now you only get one plushie and can't breed them (Without having a second is what I meant)

The ez solution is to make the plushie box purchasable from the vending machine, if you really really need two. or just order from cargo like you are meant to.

PR #12751, opened 2020-07-12
there's only so many times i can pass by a section of terrible code before i can't take it anymore so here is your vore rewrite

silicons:
rewrites vore into components and signals because i'm so, so sick and tired of this

cleans it up too because what the fuck


silicons:
oh does this mean i finally reached rock bottom as per the slippery slope image from 2016

> vore pr

code/modules/vore/vore_digest_act.dm
```diff
@@ -50,7 +48,7 @@
 /obj/item/clothing/accessory/medal/gold/captain/digest_act(...)
 	return FALSE
 /obj/item/clothing/suit/armor/digest_act(...)
-	return FALSE
+	return FALS
```
Linzolle, line 51: ???

nik707: gamer

LetterN:
## it's happening
[]

Echopsychotic:
is this PR still being worked on?

PR #12771, opened 2020-07-14
[TESTMERGE ONLY] bandaid fix the stunbaton issue, take fucking three 

silicons:
I don't like this. This doesn't fix whatever the root cause is, only decreases timing resolution so it's nearly impossible to trigger


silicons:
doesnt work

PR #12780, opened 2020-07-15
Even more engiborg power creep

BlackMajor:
Except not really.

This just changes the engiborg gripper to let it do more, y'know, engineering related things.
Rather unimpactful to rounds.
Thought it was a bit dumb my man engiborg's better at robusting than they are at actual meaningful construction.

-Tank dispenser given an attack_robot, so borgs can actually take tanks out of tank dispensers
-Engiborg holofan changed back to combifan (like it was to begin with before the holofans split)
-Engiborg gripper has an standard use now interacts with held item, dropping items has been moved to AltClick. This was necessary to get held item UI to work.
-Engiborg gripper grab list updated with the following:
All wall frames (newscaster, cameras, apcs, air alarms etc. Yes I'm aware engiborgs could already wall mount these with a screwdriver, but a gripper is so much more intuitive.)
Lights bulbs.
Vending machine restock units.
All /electronics subtypes (air alarm electronics, firelock electronics, etc. Again, yes I'm aware the engiborg has a pseudo circuit, but that doesn't cover all bases, and sometimes you just want to use a circuit that's on the ground, like the airlock electronics that fell out of an airlock that's already programmed)
Conveyer belts and conveyer switches. I'm a mad man I know.

And lastly, the following additions to the gripper are ones I'd like to be argued, since they're the most impactful.
Gas tanks. On the plus side, engiborg can actually put plasma tanks into radiation collectors, downside is bomb assembly (but they can't detonate them).
Material sheets and tiles, as well as use them in-hand. This means they can build walls and floors out of materials, as well as assemble associated material recipes.
Plasma and Plasteel are blacklisted however, unless emagged.
Stock parts. They can already pick them up with their RPED, but the fact they can pick up cells now means they can stick them in chargers first before making chemistry cry about the upgraded cells being empty. No, this does not mean you can take a cell out of your fellow borg.... you can stick one in though.

Let me know if there's any other engineering related items borgos should be allowed to pick up.

### DIFF
--- a/code/game/machinery/cell_charger.dm
+++ b/code/game/machinery/cell_charger.dm
@@ -103,7 +103,17 @@
 	removecell()
 
 /obj/machinery/cell_charger/attack_ai(mob/user)
+	if(!charging)
+		return
+
+	charging.forceMove(loc)
+	to_chat(user, "<span class='notice'>You remotely disconnect the battery port and eject [charging] from [src].</span>")
+
+	removecell()
 	return
+
+/obj/machinery/cell_charger/attack_robot(mob/user)
+	attack_ai(user)
 
 /obj/machinery/cell_charger/emp_act(severity)
 	. = ..()
--- a/code/game/objects/items/robot/robot_items.dm
+++ b/code/game/objects/items/robot/robot_items.dm
@@ -765,7 +765,8 @@
 		/obj/item/vending_refill,
 		/obj/item/stack/sheet,
 		/obj/item/stack/tile,
-		/obj/item/stack/rods
+		/obj/item/stack/rods,
+		/obj/item/stock_parts
 		)
 	//Basically a blacklist for any subtypes above we dont want
 	var/list/cannot_hold = list(
--- a/code/game/objects/items/robot/robot_items.dm
+++ b/code/game/objects/items/robot/robot_items.dm
@@ -758,7 +758,7 @@
 		/obj/item/circuitboard,
 		/obj/item/light,
 		/obj/item/electronics,
-		/obj/item/tank/internals,
+		/obj/item/tank,
 		/obj/item/conveyor_switch_construct,
 		/obj/item/stack/conveyor,
 		/obj/item/wallframe,
--- a/code/game/objects/items/robot/robot_items.dm
+++ b/code/game/objects/items/robot/robot_items.dm
@@ -746,8 +746,8 @@
 ***********************************************************************/
 
 /obj/item/weapon/gripper
-	name = "circuit gripper"
-	desc = "A simple grasping tool for inserting circuitboards into machinary."
+	name = "engineering gripper"
+	desc = "A simple grasping tool for interacting with various engineering related items, such as circuits, gas tanks and conveyer belts. Alt Click to open held item UI"
 	icon = 'icons/obj/device.dmi'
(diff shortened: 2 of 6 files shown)


Ludox235:
Speed merge. Now.

yourdoom9898:
In the worst case scenario with the borgs being able to interact with toxins, you'd have a borg release a spicy burn mix can, but that's griefing and a large violation of asimov laws, so is it really a issue?

BlackMajor:
> In the worst case scenario with the borgs being able to interact with toxins, you'd have a borg release a spicy burn mix can, but that's griefing and a large violation of asimov laws, so is it really a issue?

Engiborgs can already do this. They can already do every part of toxins except for putting gas in the handheld tanks.

Tupinambis:
I'm alright with this, a borg should be able to setup the engine when no engineers are available. More toys is good for making a bland round fun.

One thing I'd also like to see is replacing the current engiborg tools with the combined equivalent. (Borg Hand Drill, Borg Jaws of Life.) The list of modules on engiborgs takes up half the damn screen so anything to ease the real estate would be nice.

lolman360:
honestly a single tool module that you could just use ""inhand"" to switch tools (a la goon borgs) would probably be a lot more convinient because you could basically always keep it in borg hotbar with only sacrificing 1 slot instead of the 6 we have rn (screwdriver crowbar wirecutter wrench welder multitool)

BlackMajor:
> honestly a single tool module that you could just use ""inhand"" to switch tools (a la goon borgs) would probably be a lot more convinient because you could basically always keep it in borg hotbar with only sacrificing 1 slot instead of the 6 we have rn (screwdriver crowbar wirecutter wrench welder multitool)

Yeah but hotswapping between hotbar tools ala number keys 1,2,3 is faster than pressing Z then mousing, selecting or cycling through tools to get the one you want.

Example. Screwdriver crowbaring a bunch of windows at once.

Probably not a bad idea to just make them 2 in one tools kinda like jaws of life and hand drills.

ghost:
you already know how i feel about buffing a role so overpowered it has to have specific snowflake hardcounters to make them killable in the first place

BlackMajor:
> you already know how i feel about buffing a role so overpowered it has to have specific snowflake hardcounters to make them killable in the first place

Yeah but none of these changes are combat relevant.

Ludox235:
Blackmajor, you could maybe make them able to take batteries in their gripper to put them in a battery charger. Though you'll have to mess with it so engiborgs can't remove cells from other borgs and basically be better roboticist

BlackMajor:
> Blackmajor, you could maybe make them able to take batteries in their gripper to put them in a battery charger. Though you'll have to mess with it so engiborgs can't remove cells from other borgs and basically be better roboticist

I've wanted to do something like this for a while so they can stick them in APC's/chargers but ah.
Danger zone y'know.
Might do it anyway if I can figure out a reasonable way to override not interacting with borgs.

BlackMajor:
> Blackmajor, you could maybe make them able to take batteries in their gripper to put them in a battery charger. Though you'll have to mess with it so engiborgs can't remove cells from other borgs and basically be better roboticist

I did it.

silicons:
BASED??? BASED????

PR #12799, opened 2020-07-16
fixes a tiny pickle edge case

timothyteakettle:
## About The Pull Request
fixes a tiny edge case where if you were turned into a pickle by badmin means and then transformed into something else through the use of something like a wabbajack statue, you would drop your corpse on the floor

essentially this just rewrites it to not rely on the mob being destroyed, and just splits the two transformation cases into being handled by death, and the pickle transformation proc, stopping any issues with other transformations while keeping the smite safe

## Why It's Good For The Game
fixes a small bug

## Changelog
:cl:
fix: fixes a small pickle related issue
/:cl:


timothyteakettle:
oh right i was also told to update the icon

Ludox235:
Screenshot_1113

I will never forget you for making this mistake

code/modules/mob/living/simple_animal/pickle.dm
```diff
@@ -19,11 +19,10 @@
 		var/mob/laugher = A
 		laugher.emote("laugh")
 
-/mob/living/simple_animal/pickle/Destroy(force, kill_body = TRUE)
+/mob/living/simple_animal/pickle/death()
+	..()
 	if(original_body)
-		if(kill_body)
-			original_body.adjustOrganLoss(ORGAN_SLOT_BRAIN, 200) //to be fair, you have to have a very high iq to understand-
+		original_body.adjustOrganLoss(ORGAN_SLOT_BRAIN, 200) //to be fair, you have to have a very high iq to understand-
```
Ragolution, line 25: 200 brain damage is lethal, so they'll always spawn as a corpse.

```suggestion
		original_body.adjustOrganLoss(ORGAN_SLOT_BRAIN, 100) //to be fair, you have to have a very high iq to understand-
```

timothyteakettle: the point is that if the pickle dies, the person inside dies

the reason this is done because all simple animals when they die.. .. make a simple animal corpse, however making them retrievable as their self is a nice thing to do if they've been hit by badmemery

(if you get wabbajacked you don't have an original body and so you just die like normal afraidly)

if you smite them a second time or shoot them a second time to undo the pickle transformation, they will not spawn a corpse, it'll just be normal them

nik707 (approved):

PR #12800, opened 2020-07-16
Fixes swarmers not being able to eat any items at all

DeltaFire15:
## About The Pull Request
As of now, a check ran for swarmers attempting to eat an item would check if the swarmer contained a mob/living, as opposed to the target, which obviously always returned true because the swarmer is of that type, leading to them being unable to eat any items.
This PR fixes this.
Locally tested.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Swarmers can once again eat items as long as there's nothing living in them.
/:cl:


Ludox235:
Speedmerge

PR #12807, opened 2020-07-17
lowers stray cargo pod weight from 2x normal to 1/2 normal

ghost:
## About The Pull Request
i don't really like this event but i won't remove it out of respect

## Why It's Good For The Game
event that's just "someone on station gets free antag/sec gamer gear"? cringe!

## Changelog
:cl:
tweak: quartered the weight of the stray cargo pod event from 2x normal to 1/2 normal
/:cl:

### DIFF
--- a/code/modules/events/stray_cargo.dm
+++ b/code/modules/events/stray_cargo.dm
@@ -2,7 +2,7 @@
 /datum/round_event_control/stray_cargo
 	name = "Stray Cargo Pod"
 	typepath = /datum/round_event/stray_cargo
-	weight = 20
+	weight = 5
 	max_occurrences = 4
 	earliest_start = 10 MINUTES
 


timothyteakettle:
i'm fine with its weight being reduced however it doesn't give antag loot (access restrictions still apply on whats dropped too if it somehow does give sec stuff)

the one that /does/ give antag loot got disabled

ghost:
> i'm fine with its weight being reduced however it doesn't give antag loot (access restrictions still apply on whats dropped too if it somehow does give sec stuff)
> 
> the one that /does/ give antag loot got disabled

ok i didn't know if because it was a subclass it would get picked anyways but i just dislike events that don't directly create conflict and the supply pack one is "damn bro here's a free cargo crate i guess"

it's neat but shouldn't be a round staple

PR #12816, opened 2020-07-18
ports part 1 of tg wounds

timothyteakettle:
## About The Pull Request
this ports tg wounds (main pr is: tgstation/tgstation#50558)
please god only testmerge this it's not humanely possible for me to have fixed every bug in this extremely large port

if your bones randomly explode for no reason or wounds are untreatable please yell at me I have tested this non-extensively in a local to make sure nothing horrendously breaks

## Why It's Good For The Game
bones go brrrr

## Changelog
:cl:
add: due to recent innovative research in the medical field, you now have bones
/:cl:


code/game/objects/items/stacks/medical.dm
```diff
@@ -223,10 +220,40 @@
 			to_chat(user, "<span class='notice'>[M] is at full health.</span>")
 			return FALSE
 		user.visible_message("<span class='green'>[user] applies \the [src] on [M].</span>", "<span class='green'>You apply \the [src] on [M].</span>")
-		M.heal_bodypart_damage(heal_brute)
+		return heal_carbon(M, user, heal_brute, heal_burn)
 		return TRUE
```
Linzolle: remove this return lol

timothyteakettle: woops

code/datums/status_effects/wound_effects.dm
```diff
@@ -0,0 +1,186 @@
...
+	/// The right leg of the limping person
+	var/obj/item/bodypart/r_leg/right
+	/// Which leg we're limping with next
+	var/obj/item/bodypart/next_leg
+	/// How many deciseconds we limp for on the left leg
+	var/slowdown_left = 0
+	/// How many deciseconds we limp for on the right leg
+	var/slowdown_right = 0
+
+/datum/status_effect/limp/on_apply()
```
Linzolle, line 37: needs parent call

code/datums/status_effects/wound_effects.dm
```diff
@@ -0,0 +1,186 @@
...
+		return FALSE
+	var/mob/living/carbon/C = owner
+	left = C.get_bodypart(BODY_ZONE_L_LEG)
+	right = C.get_bodypart(BODY_ZONE_R_LEG)
+	update_limp()
+	RegisterSignal(C, COMSIG_MOVABLE_MOVED, .proc/check_step)
+	RegisterSignal(C, list(COMSIG_CARBON_GAIN_WOUND, COMSIG_CARBON_LOSE_WOUND, COMSIG_CARBON_ATTACH_LIMB, COMSIG_CARBON_REMOVE_LIMB), .proc/update_limp)
+	return TRUE
+
+/datum/status_effect/limp/on_remove()
```
Linzolle, line 48: ditto

code/datums/status_effects/wound_effects.dm
```diff
@@ -0,0 +1,186 @@
...
+	var/datum/wound/linked_wound
+	alert_type = NONE
+
+/datum/status_effect/wound/on_creation(mob/living/new_owner, incoming_wound)
+	. = ..()
+	var/datum/wound/W = incoming_wound
+	linked_wound = W
+	linked_limb = linked_wound.limb
+
+/datum/status_effect/wound/on_remove()
```
Linzolle, line 124: again

code/datums/status_effects/wound_effects.dm
```diff
@@ -0,0 +1,186 @@
...
+	linked_wound = null
+	linked_limb = null
+	UnregisterSignal(owner, COMSIG_CARBON_LOSE_WOUND)
+
+/datum/status_effect/wound/on_apply()
```
Linzolle, line 130: again

Ludox235:
BUCKLE UP BUCKAROOS

Tupinambis:
At a conceptual level, I actually like how wounds *seem* to work from what i've read, though I don't like RNG stuff.

A key problem I see however, is how wounds end up playing out when solo roles and antags get into the mix.

Ashwalkers, Golems, Vets, Space hobos, and the like are especially vulnerable as they have no feasible way to fix their wounds and alone and away from the station.

Traitor is basically fucked if they get outed and injured as medical tends to not be very inclined to heal a traitor and not broadcast their location. A TC item isn't really a good solution here.

Nukeops and Wizard have similar issues, though they don't even get the luxury of being a crewmember to begin with.

---

TG code has had a consistent medical meta for an incredibly long time, with your flat health increases medkits and your healing chems that basically regen health overtime. It was a system that allowed any player to heal themselves entirely on their own and surgery was rarely ever necessary, This PR totally changes that dynamic.

On that note, this change should not be taken lightly at all, and should probably not be merged if at all until solutions to the solo/antag problem are available. (and bugs are fixed of course.) I've seen these problems in action on many bayesque downstreams and they often tend to have hacky solutions or else you just die fuck you, which isn't fun. **Especially when we don't have enabled respawns (that I know of)** One example would be halostation, where the work around to the sheer amount of organ damage constantly received is an injection of magical heal juice and if you don't have said juice you will just die because surgery is inaccessible.

In summary, the current medical system is the way it is for a reason, so it's important to keep that reason in mind when wanting to change it.

PR #12821, opened 2020-07-19
Increases projectile speed across the board by around 40-45%

silicons:
## Why It's Good For The Game

Dodging projectiles is a joke and after some thunderdome testing consisting of dodging turrets just by walking around I kinda realized that projectiles are a little too easy to dodge right now, especially with Citadel's sprint mechanics.

## Changelog
:cl: silicons
balance: projectiles are by default 17.5 tiles per second instead of 12.5
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -49,7 +49,7 @@
 	var/pixel_move_interrupted = FALSE
 
 	/// Pixels moved per second.
-	var/pixels_per_second = TILES_TO_PIXELS(12.5)
+	var/pixels_per_second = TILES_TO_PIXELS(17.5)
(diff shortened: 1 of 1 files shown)


MrJWhit:
/// The number of pixels we increment by. THIS IS NOT SPEED, DO NOT TOUCH THIS UNLESS YOU KNOW WHAT YOU ARE DOING. In general, lower values means more linetrace accuracy up to a point at cost of performance.

lolman360:
> 
> 
> ```
> /// The number of pixels we increment by. THIS IS NOT SPEED, DO NOT TOUCH THIS UNLESS YOU KNOW WHAT YOU ARE DOING. In general, lower values means more linetrace accuracy up to a point at cost of performance.
> ```

yes that's the comment for `var/pixel_increment_amount` on line 54

Linzolle (approved):

PR #12830, opened 2020-07-19
Fixes genewebs being impossibly impassable

Funce:
## About The Pull Request

This fixes the Genetic Spider Webs to use the 50% chance for mobs, rather than the current 100% no-pass rate. I can adjust this number if required.

## Why It's Good For The Game

Bug Fix

## Changelog
:cl: Funce
fix: Genetics spiderwebs are no longer impassable walls
/:cl:

### DIFF
--- a/code/game/objects/effects/spiders.dm
+++ b/code/game/objects/effects/spiders.dm
@@ -38,7 +38,7 @@
 
 /obj/structure/spider/stickyweb/CanPass(atom/movable/mover, turf/target)
 	if (genetic)
-		return
+		return TRUE
 	if(istype(mover, /mob/living/simple_animal/hostile/poison/giant_spider))
 		return TRUE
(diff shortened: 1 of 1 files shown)


lolman360:
this makes them always return true on `canpass` making them 0% chance to get stuck

Funce:
It was returning NULL before hand. So it was like 50% FALSE, otherwise NULL. So 100% FALSE.
This changes that so the base-line is TRUE

lolman360:
yes the baseline shouldn't be true that's not how canpass works if you want genetic spiderwebs to be not-snowflaked remove the `if(genetic)` check entirely

knowing that that check was placed there intentionally this is more a balance pr than a fix

Funce:
Huh, okay. Let me check that out.

------


Oh heck you're right let me change this to the original way I was going to do this.

PR #12832, opened 2020-07-20
do we want (probably) free memory savings trick question yes we do

silicons:
clears remove/add overlays list after compiling instead of leaving it around and nulls priority overlays if it's empty
no reason to leave the list around in a lot of cases especially if something is only done as an one-off overlay change during init.


lolman360:
`CitadelStationBot removed the Test Merge Candidate label 19 minutes ago`

Linzolle (approved):

PR #12834, opened 2020-07-20
Experimental hybrid clickdelay system (No seriously I'm trying my best to not call it Minecraft attack delays)

silicons:
## About The Pull Request

- [x] code in recharge bars for click and resist delay

Resist delays overhauled and standardized, now resisting does not care about clickdelay but still applies it

Clickdelays have been changed from a single proc and variable controlling everything to a two-stage last_action next_action system. next_action is basically old clickdelay, forcing the user to not attack until that time has passed. last_action however only tracks time of attack which means we can now have items that require the user to not have attacked within a period of time rather than only supporting items putting the user on cooldown for a period of time post-attack.

This'll probably introduce a lot of bugs so uh, yeah, bugfix galore? On the bright side, this is directly going to be useful for me to do those silly rhythm components and the next phase of combat changes, so whatever, worth it to me.

The code now needs to manually flush or discard last_action_immediate and next_action_immediate. This is why it's called two stage, and this is mostly to prevent race conditions from people putting sleep()s or things that sleep() in their item attack code. `last_action_immediate` and `next_action_immediate` are only checked at the very beginning of attack proc chains and the normal `last_action` and `next_action` variables are never set except at the very end of attack proc chains. This ensures that if you have an attack doing a custom delay it won't choke anything else in the proc chain that will check for delay, as anything else will check the non-immediate variables, which are only written to at the very end of the attack proc chain (the start/end of the proc chain are currently in mob/CommonClickOn). Click delay still needs to be set normally/last action still needs to be set at the end of procs, and while this is usually automatically handled in the base of object/mob attack procs in items if you make a custom item you will need to do your own handling!

There is a safety net that ensures last_action is set for MELEE mob attacks unless specifically overridden, though.

## Why It's Good For The Game

Better clickdelays, more support for complex stuff down the line, rather than only a tail-end-attack system.

## Changelog
:cl: silicons
refactor: clickdelay has been refactored into an experimental hybrid system. Check code/modules/mob/clickdelay.dm for more information.
balance: Resisting no longer checks clickdelay, but is standardized to a 2 second per resist system for most forms of resisting. It still sets clickdelay, though.
add: Meters have been added for estimating time until next attack/resist. Won't be that useful due to our clickdelay currently being very short, though. They're visible from your hand and resist HUD elements.
experimental: Most attacks and forms of attacking (minus unarmed because it's too much of a pain to refactor how hugs/gloves of the north star works) now check for time-since-last-attack rather than making it so you can't attack for said time. This means you can very quickly switch to a gun from a melee weapon, whereas in the old system a melee weapon would put you on lockout for 0.8 seconds, in the new system all the gun cares about is that you did not attack for at least 0.4 seconds.
code: All clickdelay setting/reading are now procs, so it should be trivial to implement another system where drawing/switching to a weapon requires you to have it out for x seconds before using it. I am not personally doing it at this point in time though because it will likely just annoy everyone with no real gain unless we do something like putting a 0.8 second switch-to cooldown for guns (which I did not, yet)
refactor: attack_hand has been refactored to on_attack_hand
remove: sexchems no longer impact click delay
/:cl:


silicons:
removed sexchem click delay impacts because i'm not going to bother with that and it's not even adding anything anyways

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -41,10 +41,7 @@
 	if(manifest)
 		. += "manifest"
 
-/obj/structure/closet/crate/attack_hand(mob/user, act_intent = user.a_intent, unarmed_attack_flags)
-	. = ..()
-	if(.)
-		return
+/obj/structure/closet/crate/on_attack_hand(mob/user, act_intent = user.a_intent, unarmed_attack_flags)
```
Linzolle, line 44: needs to call parent lol! crates cant be opened!

Linzolle: wait hol up let me look closer

Linzolle: yes

Linzolle (dismissed):
seems to be in a stable state now

PR #12836, opened 2020-07-20
haha ok the server is burning down time for some live debugging (maybe)

silicons:
adds reference tracking

don't merge this until i disable the #define as it has overhead and should only be enabled via testmerge when we need it


silicons:
extools fucky wucky, so.

PR #12847, opened 2020-07-21
Chat queue optimization

Putnam3145:
## About The Pull Request

#12838 has an obvious, easy, fast optimization to apply, and that is this: make it only do the slow bit when a client is around to hear it. Attempts at DLLizing it have led to sheer horror and nightmares beyond my ken.

## Why It's Good For The Game

Lag... bad?

## Changelog
:cl:
tweak: Slight optimization in chat code.
/:cl:


LetterN:
will you also do it on the chat datum itself? or is it fine as it is (it only gets called when the ss isn't running)

PR #12849, opened 2020-07-21
raises max keybind command length to 32 so controller users don't get kicked

silicons:


Linzolle (approved):

Linzolle:
imagine using a controller

PR #12863, opened 2020-07-22
clown mobss are now allied to each other and clownops are now allied to clown mobs

silicons:
## Why It's Good For The Game

clownops dying to their own kind is hilariously bad

## Changelog
:cl: silicons
add: clownops and clown mobs now share the same faction. HONK!
/:cl:

### DIFF
--- a/code/game/gamemodes/clown_ops/clown_ops.dm
+++ b/code/game/gamemodes/clown_ops/clown_ops.dm
@@ -58,6 +58,7 @@
 	if(visualsOnly)
 		return
 	H.dna.add_mutation(SMILE)
+	H.faction |= "clown"
 
 /datum/outfit/syndicate/clownop/leader
 	name = "Clown Operative Leader - Basic"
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/clown_ops/clown_ops.dm
```diff
@@ -58,6 +58,7 @@
 	if(visualsOnly)
 		return
 	H.dna.add_mutation(SMILE)
+	H.faction |= "clown"
```
Linzolle, line 61: shouldn't this be LAZYADD and not a bitwise operation since its adding a string to a list or does this actually work?

code/modules/mob/living/simple_animal/hostile/retaliate/clown.dm
```diff
@@ -37,6 +37,10 @@
 	var/banana_type = /obj/item/grown/bananapeel
 	var/attack_reagent
 
+/mob/living/simple_animal/hostile/retaliate/clown/Initialize(mapload)
+	. = ..()
+	faction |= "clown"
```
Linzolle, line 42: and im pretty sure here you can just do `faction = list("clown") ` in the type definition

Linzolle (approved):

PR #12876, opened 2020-07-23
tidies up species files

timothyteakettle:
## About The Pull Request
removes roundstart eligibility defines from all species subtypes because they all return TRUE when the base proc also returns TRUE meaning it's literally useless

moves xenos from furrypeople.dm into their own species file, xenos.dm (also moved a dummy definition into the dummy folder where it belongs, also from furrypeople.dm)

removes every single god damn definition of tail wagging and moves it into one set of procs covered by two new variables defined in the species datum and set in each subspecies when used

adds a species category variable to each species and moves race drinks into using a single proc instead of one for every single drink, and they now just check the species category stored in your species to be more efficient

removes isavian, isaquatic, ismoth, because they're all species that don't exist now, ismoth just linked to checking the insect species, which is done by isinsect, so all uses of ismoth have been replaced with isinsect

## Why It's Good For The Game
messy code is bad
non-needed procs repeated throughout species, is bad (like how we defined four procs for tail wagging in like 8 species files when you can do it all once in the base species file)
we shouldn't have is helpers for species that don't exist

## Changelog
:cl:
tweak: species code is now slightly less messy
/:cl:


code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
```diff
@@ -2400,31 +2373,21 @@ All effects don't start immediately, but rather get worse over time; the rate is
...
-	else
-		M.adjust_disgust(25)
-	return ..()
+	species_required = "shadow"
 
-/datum/reagent/consumable/ethanol/hollow_bone
+/datum/reagent/consumable/ethanol/species_drink/hollow_bone
 	name = "Hollow Bone"
 	color = "#FCF7D4" //(252, 247, 212)
 	description = "Shockingly none-harmful mix of toxins and milk."
```
ghost: non

i'm a useful member of the community

timothyteakettle: thank

PR #12881, opened 2020-07-24
since it's a little hard to fix the actual issue let's just make the projectile safety check that's causing stuff to freeze when something goes wrong more lenient so things don't freeze and people don't think anything is wrong

silicons:
to be honest it isn't THAT necessary

maybe i'll just revert to pixel projectiles v7 instead of this at some point

it'll be fine.


lolman360:
"hey mekhi? why are these disabler beams stuck in midair?"

Linzolle (approved):
nice

PR #12883, opened 2020-07-24
beam rifles are now practical in emitters

silicons:
## About The Pull Request

and no, they won't work in circuits and turrets for very obvious reasons

## Why It's Good For The Game

they don't do as much damage as a normal emitter shot but they have AoE and fire.
can't be that bad.
interesting things are interesting.

## Changelog
:cl: silicons
add: beam rifles now go into emitters properly
/:cl:


lolman360:
>no beam rifle turrets

boo

necromanceranne:
I think this causes some performance issues actually, Kev. I remember lexia doing this once and causing massive time dilation. Not sure what was up or if that was ever fixed but you may wanna be aware of that.

silicons:
@necromanceranne nah i fixed it in newest projectiles
...


_i think_

PR #12884, opened 2020-07-24
fixes a dumb bug to do with material crafting

timothyteakettle:
## About The Pull Request
the place you put down structures crafted with stacks like windows, is now your turf, instead of your drop location, because sometimes your drop location isn't your turf, and this causes you to be able to continually craft items onto the same spot which is bad when you have 50 plasma statues on one turf and set one on fire

## Why It's Good For The Game
i too like not having an entire hallway turn to dust because someone used a bug to build 50 plasma statues on one spot and shoot them

## Changelog
:cl:
fix: slight tweak to how material crafting works
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -233,7 +233,7 @@
 				return
 			T.PlaceOnTop(R.result_type, flags = CHANGETURF_INHERIT_AIR)
 		else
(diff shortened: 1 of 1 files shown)


timothyteakettle:
really anything that /checks/ your location should be using get_turf not drop_location because otherwise you end up with weird scenarios like how pet carriers used to release people inside the person holding the pet carrier if the thing inside resisted out because it placed them inside drop location lmao

PR #12887, opened 2020-07-24
Nerfs Changeling Adrenaline

CameronWoof:
## About The Pull Request

- Adrenaline Glands no longer produce changeling meth.
- Changeling Adrenaline is now ~25% less effective.
- It is now physically possible to reach the OD threshold of Changeling Adrenaline.
- Changeling Adrenaline no longer makes the user immune to the tased status effect.

Where to start? Changeling was disabled due to the problematic way it interacted with stamina combat (namely, its various abilities allowed it to effectively ignore stamina combat altogether), which made it an undisputed combat powerhouse. Changeling has a long list of unique powers and you wouldn't know it, because for the longest time, the only combination anyone used was adrenals + strained muscles.

Now that Changeling has returned after its long hiatus... that's still the case. Adrenals is such a safe and powerful pick that it utterly dominates the gamemode. Changeling balance is synonymous with adrenals balance. This PR aims to rectify this by heavily reducing the power of adrenals so that we might see a little more variety out of this particular antagonist, rather than the metagolem of adrenals and EMP round after round after round after...

## Why It's Good For The Game

... round after round after round after round after...

## Changelog
:cl:
balance: Changeling adrenaline has been nerfed. Try some of the many other abilities in the Genetic Emporium!
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/adrenaline.dm
+++ b/code/modules/antagonists/changeling/powers/adrenaline.dm
@@ -13,5 +13,5 @@
 
 //Recover from stuns.
 /obj/effect/proc_holder/changeling/adrenaline/sting_action(mob/living/user)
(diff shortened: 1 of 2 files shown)


Tupinambis:
no thanks.

silicons (changes requested):
remove taser resistance, fix the actual noslow bug with meth or make the meth only decrease some slowdown instead of immuning

this is final

CameronWoof:
I can't do what kevinz wants.

PR #12889, opened 2020-07-24
minor quirk changes

timothyteakettle:
## About The Pull Request
security officers including the warden & HoS can no use the following quirks:
blindness
monophobia (it can cause heart attacks)
poor aim (its almost like its their job to shoot people)
brain tumour
blood deficiency

blood deficiency now lowers blood by 0.2 per cycle instead of 0.275, this means it now is effectively inverse blood regen (natural regen is 0.1, and so with the quirk it becomes -0.1) this should stop it being as lethal

family heirloom no longer gives doctors an advanced analyser, it now picks between a regular analyser and a scanner plushie

## Why It's Good For The Game
why could doctors get an advanced analyser as their family item its a _negative_ quirk
why were sec allowed to choose quirks that could kill them without visiting medbay often, or stop them from doing their job

## Changelog
:cl:
tweak: nanotrasen has decided to fire all disabled members of the security division and confiscate certain sentimental items from doctors
/:cl:

### DIFF
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -14,7 +14,7 @@
 	if(NOBLOOD in H.dna.species.species_traits) //can't lose blood if your species doesn't have any
 		return
 	else
-		quirk_holder.blood_volume -= 0.275
+		quirk_holder.blood_volume -= 0.2
 
 /datum/quirk/depression
 	name = "Depression"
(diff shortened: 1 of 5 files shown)


Ludox235:
Nooo let me get my powergame advanced health analyzer without going CMO :((((

PR #12914, opened 2020-07-25
forensic scanners no longer pick up color in blood dna

silicons:
fix


timothyteakettle:
i was expecting forensic scanners to just read the last value that was dna in the blood not go through every value, is it possible for there to be multiple types of dna in a blood pool or something

Putnam3145:
holy shit who puts color in DNA (haha i did in 2017)

Linzolle (approved):

PR #12919, opened 2020-07-26
Dura belt tweaks

SiliconMain:
Now protects contents from rads like normal belts
Can hold big extinguishers because it can hold big tools
Description changed because I didnt like the old one
Nader belt unatomic but syndie webbing gets rad flags so I figured it should too
:cl:
tweak: Durathread belts now protect their contents from radiation, and can hold full sized extinguishers
/:cl:


Ludox235:
Maybe it's me being dumb dumb but shouldn't it be A RCD and not AN RCD?

Putnam3145 (approved):

PR #12930, opened 2020-07-26
so you think you can parry turrets

silicons:
## About The Pull Request

turrets now stagger their shots by a maximum of 2 seconds, randomized between 0 seconds and shot_delay/3 seconds (so by default, 0.5 seconds)

## Why It's Good For The Game

they're too predictable let's change that.

## Changelog
:cl: silicons
add: turrets now automatically stagger their shots. Happy parrying/blocking.
fix: turrets now speed_process, they were shooting slower than they should be
/:cl:


Tupinambis:
Turrets were a bit easy to cheese if you knew what you were doing even before parry. Good change.

Linzolle (approved):

PR #12942, opened 2020-07-27
Fixes the 'Naked' outfit

DeltaFire15:
## About The Pull Request
In the travelling-trader PR, the outfits added didn't actually have names, so they defaulted to "Naked" as name, which lead to one always ending up with the otherworldly surgeon outfit if one selected 'Naked'. This fixes that and gives the two added outfits actual names.

## Why It's Good For The Game
Fix man good

## Changelog
:cl:
fix: The 'Naked' outfit is no longer broken.
/:cl:


timothyteakettle:
fuck I didn't know they needed names

PR #12944, opened 2020-07-27
fixes some crafted crates containing items

### LINKED ISSUE #12815: Crate crafting bug
Issue reported from  Round ID: 23686 (Citadel Station)

Reporting client version: 513

if you make a blood/RCD/prosthetic crate with plasteel you get them full of items that usually are found in those crates.
Screenshot_1

timothyteakettle:
## About The Pull Request
adds a new var should_populate_contents to closets which if FALSE, means it wont call populate contents
makes it so the 'fake' versions of the rcd, prosthetic and blood crates don't call populate contents
makes the crafted versions of these crates the 'fake' type

closes #12815

## Why It's Good For The Game
fixes a bug, adds a thing to closets that was surprisingly not there

## Changelog
:cl:
fix: some crafted crates won't contain items now, and thus have stopped breaking the laws of physics
/:cl:


Linzolle (approved):
pain

PR #12956, opened 2020-07-28
Makes bloodsucker death not spam now actually working

Arturlang:
## Changelog
:cl:
fix:Makes bloodsucker death not as spammy. 
/:cl:


Arturlang:
Theres an unrelated issue with bloodsucker death, gib() that destroys the brain causes the victims mind to loose its loc and be forced to the arrivals shuttle, methinks

PR #12960, opened 2020-07-29
loadout changes part 1: adds subcategories to the loadout

timothyteakettle:
## About The Pull Request
part 1 of two planned things I want to do to the loadout
this adds subcategories
image
the order of the categories/subcategories is the order in which they're set out in a global list

also ports a few clothing sprites from variousc codebases and adds them to the loadout:
pink tutu
black dress
several kimonos
samurai outfit and helmet (they don't have armour, they're a costume)
a bath robe

it also moves all donator items into one section, they're still only visible and selectable if you're whitelisted to use them, otherwise if you have none the entire tab will just be empty, again makes finding your stuff easier

part 2 will be just adding more stuff to some of the categories that are kinda empty

**PLEASE NOTE: this changes all the loadout paths and so people will have to pick their loadout items again**

## Why It's Good For The Game
makes it easier to have more loadout items without it being an incredibly long list
also it's easier to find specific things in the loadout now

## Changelog
:cl:
add: more items in the loadout and loadout has subcategories now for easier searching
/:cl:


timothyteakettle:
forgot to remove some stuff I used for debugging, gonna remove it now

edit: also just tested donator items by vv-ing myself onto the bike horn one while in a local and it showed up and I could spawn with it

code/_globalvars/lists/loadout_categories.dm
```diff
@@ -0,0 +1,13 @@
+GLOBAL_LIST_INIT(loadout_categories, list(
+	CATEGORY_BACKPACK = list(SUBCATEGORY_BACKPACK_GENERAL, SUBCATEGORY_BACKPACK_TOYS),
+	CATEGORY_NECK = list(SUBCATEGORY_NECK_GENERAL, SUBCATEGORY_NECK_TIE, SUBCATEGORY_NECK_SCARVES),
+	CATEGORY_MASK = list("NOSUBCATEGORY"),
```
silicons: make this a define (NOSUBCATEGORY) please thanks

silicons: or just an empty list

timothyteakettle: ill define NOSUBCATEGORY because the way I have it working is having NOSUBCATEGORY be a single subcategory for the purpose of for loops, except it is a specific edge-case where the subcategory won't be shown and by default everything is in it for that main category

PR #12962, opened 2020-07-29
Another Pass Over Kilo

dapnee:
## About The Pull Request

Added an air alarm to science near the bepis, fixed the active tiles around the arrivals shuttle, and fixed the commissary APC being only renamed and not area stringed correctly

## Why It's Good For The Game

Fixing time spent on loading the map is always good.

## Changelog
:cl:
fix: changes commissary APC so it actually powers the room, adds a missing AIR alarm, arrivals no longer has active atmos tiles.
/:cl:


timothyteakettle:
suggestion: just delete kilo entirely

PR #12964, opened 2020-07-30
monkeys stop bleeding nonstop

timothyteakettle:
## About The Pull Request
okay so tg uses the probability of bleeding as the squareroot of the amount, we don't, so when the monkey bleed proc is called with 0 amount to bleed, it makes an image regardless as our bleed proc is different

it now does the same as humans and checks the bleed amount isnt 0 before calling bleed

## Why It's Good For The Game
fixes a bug

## Changelog
:cl:
fix: monkeys no longer continuously bleed everywhere
/:cl:


Linzolle (dismissed):
nice

timothyteakettle:
woops i pushed to the wrong branch that was for the 'end our suffering' pr where i brutally kill the thing known as combat mode

PR #12976, opened 2020-07-30
Durathread Reinforcement Kits

Hatterhat:
## About The Pull Request
image
craftable durathread reinforcement kits, for making your jumpsuit better at not getting you killed
(sets jumpsuit armor to durathread levels)
sprites by @Michiyamenotehifunana god bless
## Why It's Good For The Game
for years, one has had to sacrifice fashion for protection, or vice versa. no longer.
## Changelog
:cl:
add: Durathread reinforcement kits! Sprites by Toriate, sets jumpsuit armor to durathread levels, craft in the crafting menu.
/:cl:


Hatterhat:
i don't know what that changelog compile commit is doing in my perfectly good branch but i'm gonna fight it

code/game/objects/items/armor_kits.dm
```diff
@@ -0,0 +1,60 @@
...
+		if(firemax != TRUE)
+			C.armor.fire = 40
+			used = TRUE
+		if(acidmax != TRUE)
+			C.armor.acid = 10
+			used = TRUE
+		if(bombmax != TRUE)
+			C.armor.bomb = 5
+			used = TRUE
+
```
timothyteakettle, line 33: if you're only going to use meleemax/lasermax, etc once, it'd be better to just replace where the variables are used with the condition used to set it

code/game/objects/items/armor_kits.dm
```diff
@@ -0,0 +1,60 @@
...
+		if(used == TRUE)
```
timothyteakettle: no need for == TRUE you can just have if(used) because it's implicitly checking its true, you can also have !used instead of used != TRUE or used == FALSE for other places you've used it

code/game/objects/items/armor_kits.dm
timothyteakettle: honestly you could just use like C.armor.acid = max(C.armor.acid,10) and then compare the old and the new armour values to know if it was used or not in one statement edit: or even use a solution with for loop because its a list to reduce the number of lines

Hatterhat:
i'm going to fight git in its entirety

TheObserver-sys:
This? I like this. A question though: Does it work on the chameleon jumpsuit?

Hatterhat:
yeah probably

it'll have the equal/best armors from both chameleon jumpsuit and durathread jumpsuit, if thats relevant, dw about it

PR #12987, opened 2020-07-31
fixes sergal ears being backwards

silicons:
unless this is how it's supposed to be?

check icondiffbot please.


Linzolle:
i'm pretty sure that the opening on sergal ears is supposed to be on the back side
quote from the [wiki page](http://www.vilous.net/wiki/Sergals#Head_and_face) about them

> Ear holes cannot normally be seen from the side, because their ear holes always turn to the rear and are completely covered by rear pointing fur. In other words, their ears are NOT like the ears of Earth's creatures such as cats or foxes at all.

Poojawa:
Sergal ears are correctly orientated. This literally puts them on backwards

PR #12988, opened 2020-07-31
so it turns out mob burning checked for no oxygen instead of yes oxygen due to some other pr so you can only die of being on fire if you're in space in a hardsuit

silicons:
## Why It's Good For The Game

fix

## Changelog
:cl: silicons
fix: being on fire works again.
/:cl:


ghost:
based

Linzolle (approved):

PR #13013, opened 2020-08-03
N tvsg sebz Nezbere (or: a bunch of clockcult related stuff)

DeltaFire15:
## About The Pull Request
Aight, first and foremost. This is a lot of stuff, which mostly is just related by that it all has to do with clockies. If any things are so questionable any maintainer wants me to atomise them into their own PR, just tell me, and I'll do it. Other than that, feel free to give me feedback on any of these or suggest changes if you think any of this is not well balanced. Anyways, with that out of the way, this PR does quite a bunch of things:

1. Cultists now start with a replica fabricator in their backpack.. no brass though, due to power-cheese.

2. Kindle now is back to _15_ ds cast time, instead of _10_ ds. But also, the mute (when the stun ends) has been increased to _5_ seconds (instead of _2),_ and the stuttering has been increased to last _5_ seconds after the mute ends, instead of _3_

3. The ratvarian spear has been slightly modified to not be able to remove instead of add vitality under very specific circumstances.

4. The ratvarian spear can parry now. Short cooldown, very short parry that relies on you timing it right, but relatively low penalties. Low failure tolerance. Not going to list all the vars here, check the file for info.

5. A new implant has been added for clock cultists (and people who emag it): The brass claw implant is able to extend a very obvious claw on command, which gains combo continuously attacking one target, leading to _2_ bonus damage per hit after the first. Base damage is _15_, _5_ lower than the spear's damage. Or well, _25_ lower if used against silicons. Also cannot be thrown. Combo gain tripled when attacking hostile cultists. The combo effects only apply on mobs, not objects. Attacking objects in return also does not reset the combo counter. Combo as of now is capped at _18_ bonus damage after _10_ consecutive hits against the same target, causing _33_ total damage if the target somehow is still standing after ten consecutive hits. Can be summoned via the new sigil of rites, maximum of _4_ summons per shift. More on rites in a moment.

6. Clockwork Rites have been added. They require a Sigil of Rites to be summoned first (costs _1400_ power, aswell as requiring _2_ invokers). This sigil then can be used to perform rites on it. Generally, these require clockwork power, cells, materials and sometimes a target to perform various effects. Info on these can be gotten via interacting with a sigil, selecting a rite, then choosing 'show info' instead of 'cast'

Current rites are: 

- Rite of Advancement: Needs some power, a target (can be invoker), a power cell and a proximity sensor, aswell as any organ or cybernetic implant (save brains) to be placed on it. It will consume the materials to implant the organ or implant into the target's body, without any surgery needed. Only works on clock cultists.

- Rite of Woundmending: Needs some power, a target (cannot be invoker, but can be noncultists), a power cell, a health analyzer, and a flask of zelus oil (can be empty). This will, when completed, heal all wounds on the target, though it will cause _10_ toxins damage per wound healed.

- Rite of the Claw: Needs some power, a power cell, any cybernetic implant and a flash to summon a brass claw implant on the sigil. This can only be used _4_ times per round, so coordinate on who should acquire these.

Extensively tested on local, though I alone probably didn't find every single bug I might have accidentally created, so I'd recommend testmerging this first if it gets accepted until a few clock cult or clock eldritch affairs rounds happen.

## Why It's Good For The Game
1-3 are relatively small changes, which should in general be useful:

On 1: A clock fabricator is needed by cultists to make brass, their main building resource. Locking it to tier 2, even for starter cultists, is kinda meh. Converted cultists will have to wait till tier 2 anyways.

On 2: Short cast time to counteract the short mute is kinda ehhhh balancewise. Casttime increase together with a few seconds of extra mute / stutter should help the mode in general, without making 1-sec cast time kindling cultists able to easily kindle people once they have a tiny bit of distance between them and chasers.

On 3: Fix man good. Negative vitality gain aka vitality losses is meh for a weapon that should generate it.

Hoo boy, now onto the big parts:

On 4: The Ratvarian Spear has always been called somewhat weak in direct combat, save for the throw, which breaks the spear, leading to cultist often preferring noncult weapons as opposed to the supposed 'main' cult weapon. This should alleviate this issue by allowing skillful players to parry using the spear, alleviating some of the power issues. Parry counterattacks are even harder to pull off, having a 2ds timespan.

On 5: This poses an alternative weapon for cultists to use or combine with the spear, which trades damage and the ability to throw it causing a knockdown / high damage for the ability to gain combo, making it as effective as the spear after four successive hits on the same target. It also has a slight advantage at wounding enemies, at least after the first combo stack, but less precise sharpness. Should be fun to experiment with, without harming balance. If it does, the values can always be tweaked to fit in better.

On 6: Yeeah this is a very big one. Clock rites were something I wanted to do since quite a while, and they should allow for quite some interesting effects, whilst being a tiny bit more complex than 'use slab, make thing'. This also allows for a woundheal spell to not be as easily acquirable as sentinels compromise, as that would be a very strong too, though cultists do need something to handle their wounds as they will usually have issues actually getting medical things going on when the wounds start rolling in.
Anyways, rites in general also allow for more stuff to add without bloating the slab too much, and while having some more specific or fluff requirements than just 'have this amount of power'
For now, these three rites should show a few options for the effects one could make rites create, though there could be quite alot more, and it should be relatively easy to implement.

## Changelog
:cl:
balance: Roundstart cultists now start with a replica fabricator - no brass though, make your own.
balance: Kindle cast time: 10 > 15, mute after stun end: 2 > 5, slur after mute end: 3 > 5
fix: The ratvarian spear no longer adds negative vitality under very specific circumstances.
balance: The Ratvarian Spear can parry now! Short parries with low leeway, but low cooldown.
add: The brass claw, a implant-based weapon which gains combo on consecutive hits against the same target.
add: The sigil of rites, a sigil used to perform various rites with a cost of power and materials
add: The Rite of Advancement: Used to add a organ or cyberimplant to a clockie without need for surgery.
add: The Rite of Woundmending: Used to heal all wounds on another cultist, causing toxins damage in return.
add: The Rite of the Claw: Used to summon a brass claw implant. Maximum of 4 uses per round.

/:cl:


timothyteakettle:
reading through the list of changes these seem like cool changes, I want to check out the parry time on the spear to make sure it's not busted, but cool.

ghost:
please cap the claw damage at something like 2x or 3x spear damage

Ludox235:
Next add eminence abilities that aren't just overheating walls and placing funny markers :)

lolman360:
tbh eminence should be a person with aghost instead of just an ai so
1. there's counterplay, you can kill their super vulnerable body now (reebe gone).
2. now that there's counterplay you can justify giving them more power
3. it's 100% of the time better to accept a ghost eminence because ascending to eminence removes a set of boots on the ground, this absolves that issue

KeRSedChaplain:
to people yelling about eminence improvements, those are coming eventually, I've just been insanely lazy and haven't gotten around to it

PR #13015, opened 2020-08-03
Combat v4: Redesigning Stamina, Sprint, Combat mode

silicons:
## About The Pull Request

Stamina Combat:

- [x] Stamina buffer decoupled from main stamina health

- [x] tickrate stamina buffer HUD updates rather than every second

- [x] Stamina buffer recharge semi-linked to main stamina health - at just above stam-health stamcrit, expect 50% buffer capacity and 33% recharge, numbers subject to tweaking.

- [x] All attacks will follow a "use buffer or fail" system. **Attacking will not draw from main stamina health.**

- [x] **Soft Stamcrit removed** - Replaced with buffer maximum scaling

- [x] Stamina buffer given rapid recharge - with average attack cost being ~4.6 stamina per second, it will recharge at 1.5/second without combat mode and 3 with. It will be updated on the HUD elements every tick rather than every second for a smoother experience. The stamina buffer is still 30 points in size.

- [x] Shield blocking at the moment will directly go to stamina health.

- [x] All active actions that use stamina buffer will use the buffer

- [x] Gun stamina usage dropped to semi-negligible levels.

- [x] **All attack/defense modifiers (damage multipliers) removed from combat mode**, replaced with buffer recharge boost. The requirement for combat mode to do certain actions remains for now.

- [x] **All current randomized gun inaccuracies are removed** with replacement to be determined

- [x] Sprint changed to cost more with higher slowdowns

- [x] Sprint changed to use tiles/second speedup config instead of straight up multiplicative slowdown, which has proven time and time again to be too unnuanced for these purposes

- [x] movespeed modification tiles/second calculation system

## Why It's Good For The Game

https://hackmd.io/@CitadelStation13/SyoneIyeP

edit 2: **THE PR IS ACCURATE IN CASES OF IT CONFLICTING WITH THE DESIGN DOC.**

## Changelog
:cl: 

wip

/:cl:

edit 1: "you can't attack without stamina buffer doesn't that mean you can't kill anyone" - **No.** Look at the values for the recharge and the values for swings. There's no way you can't kill someone in 20 seconds with an extinguisher and no way you  can't kill someone unless they're going full on nanite-elite hardsuit powergame with an axe in the 8-10 seconds it'll take for it to drain buffer. This is going to be, of course, worse for you, if you are near stamcrit, but we can adjust values as needed. The point is to not have stamina-damaging weapons entirely neuter your attack output (because on the old system the buffer is very slow recharging and it relied on you being able to dip into your stamina health to attack) 

**Note that the buffer has no delay between action for it to recharge at the moment.**


moltoretardo:
Randomized guns inaccuracies removed

Godlike

Hatterhat:
https://www.youtube.com/watch?v=-MaCJZIBKGs

Tupinambis:
>Stamina buffer decoupled from main stamina health

Good. maybe it should be renamed to "Action stamina" or something though.

>Stamina buffer recharge semi-linked to main stamina health - at just above stam-health stamcrit, expect 50% buffer capacity and 33% recharge, numbers subject to tweaking.

I feel like this may be confusing and still give a feeling of "helplessness" by being low on stamina which already existed before. Personally i'd rather only have EITHER buffer capacity reduction or recharge reduction, not both. Or even better, have it be binary in that action stamina/stamina buffer functions normally unless you are put in stamcrit. 

>All attacks will follow a "use buffer or fail" system. Attacking will not draw from main stamina health.

Good. 

>Soft Stamcrit removed - Replaced with buffer maximum scaling

Good god thank fuck.

>Stamina buffer given rapid recharge - with average attack cost being ~4.6 stamina per second, it will recharge at 1.5/second without combat mode and 3 with. It will be updated on the HUD elements every tick rather than every second for a smoother experience. The stamina buffer is still 30 points in size.

30 points is a weird number  but otherwise this sounds good. Stamina has always mattered too much long term and not enough short term so it's good to see that change.

>Shield blocking at the moment will directly go to stamina health.

Shields feel far too strong right now, and honestly being able to "break" a block and effectively "Stagger" by eliminating their stamina buffer seems like a better way to handle it.

Maybe have a brief stamina buffer crit where it takes a moment to regenerate where the target is vulnerable?

>Gun stamina usage dropped to semi-negligible levels.

Thank fuck.

>All attack/defense modifiers (damage multipliers) removed from combat mode, replaced with buffer recharge boost. The requirement for combat mode to do certain actions remains for now.

Also thank fuck.

>All current randomized gun inaccuracies are removed with replacement to be determined

ALSO thank fuck.

>Sprint changed to cost more with higher slowdowns

Thank fuck x4.

>Being in stamina hardcrit will hardstun until you are back at 50% stamina health. However, stamina regeneration is massively boosted in stamina hardcrit. (experimental)

This is quirky but I get why it's the way it is. This is very similar to a "stun timer and then you get stamina to act again" but it would be nicer to just have the timer before instantly receiving stamina so it's less confusing if you don't know how it works.


Overall all these changes are good imo.

CameronWoof:
Obviously I'm in favor of this.

deathride58:
This is very interesting overall, though some of your notes in the PR desc, in particular, "attack/defense modifiers removed from combat mode", are very misleading since they directly conflict with what's stated in the design doc.
The changes outlined in the design doc seem to be fairly good overall, though. I'm very interested to see how it plays out on live.

HeroWithYay:
This is nice.
I like this.

deathride58:
Spent some time fooling around with this, here's my own takes on it

- stam buffer recharge is slow. incredibly fucking slow, especially when combined with attacking being capped by your stam buffer rather than stamina itself. "stam buffer given rapid recharge" my ass. like holy shit, it's somehow clunkier than stamina was during the first testmerge, and yes i'm using combat mode. for reference, the intent behind stamina buffer is for it to recharge with a sort of curve; the less stamina buffer you have, the faster it recharges, and if it's empty, then a recharge tick will just grant you about half of your stamina buffer, so that all you have to do for proper stamina management in most cases is wait for a single recharge tick then continue attacking (which, well, can't be done here)
- restricting attacking to only be affected by your stamina buffer takes out all the skill involved with managing your stamina, and removes the risk vs reward when it comes to digging into your own stamina pool to continue attacking (because hey, can't use your stamina pool anymore). a much better way to discourage this without lowering the skill ceiling as much as this change has would be to instead make it so that attacks digging into your main stamina make an unpleasant hud sound and gives you a longer click delay than you otherwise would have. also, this just makes the "lmao why does my spaceman have asthma" complaints worse since exhaustion kicks in much sooner
- staminaloss affecting the stamina buffer so much makes stamina-based weapons a lot stronger than they currently are. with the old stamina buffer mechanics, being affected by staminaloss didn't really bar you too much from fighting, since the stam buffer regenned incredibly fast when empty. again, lowering the skill ceiling here
- on the positive side of things, i do enjoy how it ends up feeling a little bit more like arma 3's ace3 stamina simulation (albeit without the long-term effects), which was a source of inspiration for the stamina mechanics. specifically, i like the way that running out of short-term stamina has a much more immediately-noticeable impact, since your character's mechanically exhausted at that point. even though the way it currently accomplishes this is one that lowers the skill ceiling by quite a bit (and also goes directly against feedback regarding the wall of exhaustion being too easy to hit), it'd be fairly easy to rework it to feel a lot more natural
- sprint cost being made mostly consistent is a pretty good change overall, but now it runs into the problem of sprinting at lower move speeds being pretty useless since it costs so much for so little gain. maybe set sprint cost up to have a speed floor or something, where there's a minimum guaranteed speed for sprinting, unaffected by further move delay changes (maybe have the minimum speed for sprinting be standard walk speed?)

Overall? It feels like you're listening to feedback regarding the stamina mechanics.... and then proceeding to double down on the exact things that most people complain about, some of those things being things that I myself have gripes about

PR #13032, opened 2020-08-05
podpeople obesity fix (this time citbot properly will do the feature delta)

### LINKED ISSUE #13023: Anthro plants will sometimes spend the entire round stuck with a fat moodlet
## Round ID:
Requires multiple rounds to be observed

## Testmerges:
yes

## Reproduction:

Play as an anthro plant and notice that, on about half of all the rounds played, you will be stuck with a near-permanent negative moodlet from being fat that can only be temporarily dispelled by entering the darkness, which damages you, or consuming lipolicide. The moodlet will come back as soon as you either re-enter the light or run out of the chemical. Some rounds this won't occur and you'll instead just have a permanent stuffed moodlet. 

I think it might be for the best to make it so that anthro plants get slightly less nutrient value from light, requiring them to sometimes eat, or to look into if being in the light simply automatically sets their hunger to fat instead of slowly regenerating their nutrient, or to make it so that the light has a cap on how high your nutrition can be restored by it, requiring eating to get to higher fullness levels. Since they're supposed to be weaker than the regular version of plant people. Also someone should look into if the pod cloning version of the species has the same issue or if it's just the round start version. I can't myself.

lolman360:
## Changelog
:cl:
fix: podpeople no fat when sunbathing.
/:cl:

closes #13023

### DIFF
--- a/code/datums/elements/photosynthesis.dm
+++ b/code/datums/elements/photosynthesis.dm
@@ -60,7 +60,7 @@
 			if(L.stat == DEAD)
 				continue
 			if(light_nutrition_gain)
-				L.adjust_nutrition(light_amount * light_nutrition_gain * attached_atoms[AM], NUTRITION_LEVEL_FULL)
+				L.adjust_nutrition(light_amount * light_nutrition_gain * attached_atoms[AM], NUTRITION_LEVEL_WELL_FED)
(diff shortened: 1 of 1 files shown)


Ludox235:
#13023 Make so it closes this.

Ghommie (approved):

PR #13033, opened 2020-08-05
how about we not gib slimepeople for having too much blood that is a little unfun eh now they'll just take damage and bleed the excess

silicons:
## Why It's Good For The Game

why the hell would you just gib people for >200% blood with no warning or damage who coded this?

## Changelog
:cl: silicons
balance: lethal blood now causes damaging bleeding instead of outright gibbing
/:cl:


lolman360:
salt pr

TheObserver-sys:
There is a warning, but it's like, 10%? before death, it's kinda goofy as fuck.

Ludox235:
salt pr

TomGarou:
*full disclosure I was also caught completely by surprise by this mechanic since it's appearance was an unexpected addition to the citadel code that gibbed me for gaining blood as a slime person using their toxin consuming process that seemed unusually harsh considering the lack of gibbing elsewhere.* 

Personally I agree with a change in this mechanic, or a removal. From what I have seen there is a general trend that things are moving away form gibbing on citadel, (C4, Emagged Recycler ect....) besides having too much blood shouldn't destroy items. A blood explosion isn't exactly on the same threshold as having C4 placed on you, something violent enough to damage the actual station infrastructure while affecting the target. In our current iteration that doesn't even gib you anymore. 

...perhaps having their heart rupture/get destroyed would be a decent middle ground? (maybe even allowing the upgraded prosthetic heart to handle more max blood?)

Sishen1542:
@silicons ok so basically this was added because teak didnt fully review all the wounds changes (its a quirky epic add from the wounds port, https://github.com/Citadel-Station-13/Citadel-Station-13/commit/e8fade4f7fa8e9acaac1bcb2ea2ae53063aabadb#)

slimes are extremely fragile as is due to poojblood and wounds (which ill be fixing soon) so i wouldnt be against removing the gib entirely, could give them slowdown for blood percentages above a certain limit instead.

ghost:
**not** salt pr

PR #13034, opened 2020-08-05
turns off fix/feature delta bot

silicons:
this isn't a feature we need at the moment maybe if we had a full team of coders and 20 prs a day this would be useful to keep people fixing bugs but honestly this just annoys contributors with no gain if you want to close a pr for being bad just close it don't use this as an excuse too.


Ghommie (approved):
Good riddance.

PR #13040, opened 2020-08-05
funny chair loses block chance and gains active block/parry

Sishen1542:
## About The Pull Request
chairs lose block chance since we are swapping over to active blocks/parries for items, too much to paste in here so just check code for values

## Why It's Good For The Game
removing old systems

## Changelog
:cl:
balance: gave chairs active block/parry in exchange for removal of block_chance
/:cl:

### DIFF
--- a/code/game/objects/structures/beds_chairs/chair.dm
+++ b/code/game/objects/structures/beds_chairs/chair.dm
@@ -319,10 +319,38 @@
 	throwforce = 10
 	throw_range = 3
 	hitsound = 'sound/items/trayhit1.ogg'
-	block_chance = 50
 	custom_materials = list(/datum/material/iron = 2000)
 	var/break_chance = 5 //Likely hood of smashing the chair.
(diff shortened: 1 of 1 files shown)


DeltaFire15:
Why exactly is a random chair better at blocking than an actual weapon aka electrostaves?

Sishen1542:
Chairs had a 50% block chance previously.
Also they have a 5% chance to break on attack/block

Hatterhat:
buff it to 10 force because its inevitably going to break and because its a whole ass chair you are bashing someone to death with

Sishen1542:
you can attack while blocking with chair, 10 force would make ripostes 25 unless I brought the multiplier down a bit.

Sishen1542:
chairs require 2 hands to pick up now, gave a few value tweaks and introduced a damage cap (anything over 20 force will go through no problem, so riot shields are better at handling the big weapons)

PR #12974, opened 2020-07-30
removes brainwashing from uplinks

Sishen1542:
## About The Pull Request
comments out brainwashing disk from uplinks

## Why It's Good For The Game
quirky funny surgery that is too easily abused by just giving your metafriends the wacky antag status

PR #13044, opened 2020-08-06
increases the cost of the brainwash surgery, reduces its success chance

timothyteakettle:
## About The Pull Request
cost: 5 tc -> 10 tc
success rate: 85% with hemostat -> 65% (also reduced the ghetto methods by a fair amount too)
closes #12974

## Why It's Good For The Game
too cheap / easy to perform, alien tech requires it to be actually researched first whereas this doesn't and it literally turns traitor into a conversion mode in the right hands

provides an alternative to outright removing it by actually attempting to nerf it a bit because content removal is bad, especially on an underused item that is basically never an issue

## Changelog
:cl:
tweak: brainwashing disks now cost 10 tc in the uplink instead of 5, and have a lower success chance when performed
/:cl:

### DIFF
--- a/code/modules/surgery/advanced/brainwashing.dm
+++ b/code/modules/surgery/advanced/brainwashing.dm
@@ -25,7 +25,7 @@
 	return TRUE
 /datum/surgery_step/brainwash
 	name = "brainwash"
-	implements = list(TOOL_HEMOSTAT = 85, TOOL_WIRECUTTER = 50, /obj/item/stack/packageWrap = 35, /obj/item/stack/cable_coil = 15)
+	implements = list(TOOL_HEMOSTAT = 60, TOOL_WIRECUTTER = 30, /obj/item/stack/packageWrap = 15, /obj/item/stack/cable_coil = 5)
 	time = 200
 	var/objective
 /datum/surgery_step/brainwash/preop(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
--- a/code/modules/uplink/uplink_items/uplink_roles.dm
+++ b/code/modules/uplink/uplink_items/uplink_roles.dm
@@ -46,7 +46,7 @@
 	Insert into an Operating Console to enable the procedure."
 	item = /obj/item/disk/surgery/brainwashing
 	restricted_roles = list("Medical Doctor", "Roboticist")
-	cost = 5
+	cost = 8
 
 /datum/uplink_item/role_restricted/clown_bomb
 	name = "Clown Bomb"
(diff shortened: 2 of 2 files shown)


moltoretardo:
Fixes none of the problems mentionned in the stated pr. Even better, now the poor fellas that get caught by the funky medical officer will have better chances of getting his jam fucked up by random traumas. The same metabuddying problem is still fully present with the easy “hrrngh give me augments please” and just giving your buddy free antag status.

timothyteakettle:
ah yes the same metabuddying that's present in literally every conversion mode, especially bloodsucker, and is generally sorted by admins when possible, thanks for the feedback

Ludox235:
Waaah waaah m-muh metabuddying
Bro? Adminhelp??? Allo???? Stop trying to delete shit when admins are supposed to deal with that shit???

moltoretardo:
Okay buuut admins don't deal with it tho. It was very recurrent at one point and nothing was ever done because it's by the rule. It's RP fellas. And yeah, Bloodsuckers suck even more than brainwashing and if you want to remove it too go ahead. It's something that just cannot be sorted unless you use the good old removal of the said item. It adds absolutely nothing of value to a round. Saying that deleting things is bad is also one of the point of view I can't understand. Why did we remove x-ray from geneticist? Can we please re-add it but harder to get? Why is public mining not here anymore, I can't go out on lavaland and powergame before the miner, can we re-add it too? Removing stuff is sometime necessary. One of the other argument was "why are people complaining if it's never done!!!" Yet one rather spicy round involving funny disk was done quite recently which spiked some drama (I'm not naming anyone since throwing people under the bus could be mean)  and I have some funny memories about the funny zpowder and forced brainwash gimmick some of the special players had. Traitors have some rather good gears so that they can solo accomplish their objectives with ease. Raising an army of antag just by sitting next to a table sounds rather funky.

Waaah Waaaah m-muh easy powertripping traitor gadget
Bro? Find a better gimmick? Allo? Stop saying that admins give a shit about that kind of stuff when it's already obvious they don't.

ghost:
@TheSpaghetti can you try to port child sex pr from skyrat up to citadel so i can add another shitpost to the pile of "Linking people their bad pr's"

ghost:
> @TheSpaghetti can you try to port child sex pr from skyrat up to citadel so i can add another shitpost to the pile of "Linking people their bad pr's"

no lol

silicons:
the solution to brainwashing is making it more detectable on health analyzers per what i discussed a while ago. it's still on my todo list but i might as well close this. if too much time goes on without me or someone else fixing it we may go through with these nerfs, however.

PR #13045, opened 2020-08-06
ports bronze (fake clockwork) airlocks and windows

silicons:
## About The Pull Request

Ports from tg the ability to make bronze airlocks and windows

## Why It's Good For The Game

clockwork fashion is good

## Changelog
:cl: ported from tg
add: bronze airlocks and windows can now be built
balance: i also tweaked bronze flooring to be cheaper.
/:cl:


MrJWhit:
I don't think this is a good idea.
By allowing things to be made more willy nilly for random ERP dens, it makes everything more normal. I don't think large scale antag aesthetics should be considered normal. They're a relatively rare antag, and so the aesthetics of them should also be rare with it. Keep unique things unique.

I also think being able to mass produce brass floors is a bad idea for the same reason.

Ragolution:
hmm yeah

PR #13052, opened 2020-08-07
fixes laptops not being able to be picked up

silicons:


moltoretardo:
Good pr for nameko’s gaming pc inc.

PR #13071, opened 2020-08-07
Adds doors to the west of box medbay

ghost:
## About The Pull Request

dreamseeker_3FpXDmlHXH

puts some doors on the west side of box medbay, medicine closet is now outside the security checkpoint and the table has been moved south by the shower

## Why It's Good For The Game

reduces suicide rate of medbay players by at least 10%

## Changelog
:cl:
tweak: Doors added to the west side of box medbay to make things a bit more manageable.
/:cl:

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -6467,6 +6467,12 @@
 	},
 /turf/open/floor/plasteel/white,
 /area/science/xenobiology)
+"alX" = (
+/obj/structure/disposalpipe/segment{
+	dir = 6
+	},
+/turf/open/floor/plasteel/white,
+/area/medical/chemistry)
 "alY" = (
 /obj/structure/cable{
 	icon_state = "1-2"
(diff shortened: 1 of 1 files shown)


Tupinambis:
I don't like the idea that random people can just bypass the lobby, nor the location of those doors, nor the lack of a button.

I'd rather give doctors free access to the maintenance tunnel adjacent to medbay and surgery, that way if they are on that side of the station they can have an easy route into medical without the need to add entirely new doors.

ghost:
These airlocks still require access and they're replacing windows, so it's arguably MORE secure than before.

I main medbay on a server where this change was made years ago and losing it is suffering. Box medbay is in a bit of an awkward spot due to being on an internal corner, the alt entrance makes it so you don't have to go all the way around that corner and all the way down medbay if you're coming from the west. Medbay chokepoints also really suck to deal with when SHTF because you have to fight against bad design slowing down traffic on top of an increased workload. A small maint exit wouldn't solve these problems, you'd end up going the long way but in the opposite direction.

edit: I forgot but there's already plastic flaps on the west side you can crawl under so this is basically just making that entrance more intentional.

PR #13078, opened 2020-08-08
Made traitor reroll reroll to MI13 only.

Putnam3145:
## About The Pull Request

Rerolling traitor goals will now ONLY give you "proper" objectives.

## Why It's Good For The Game

Anyone who is rerolling is probably doing it cause they don't want freeform anyway.

## Changelog
:cl:
tweak: Rerolling your traitor goals will ONLY give you "proper" objectives.
/:cl:


Tupinambis:
Ehhh I disagree with this, because what if you get MI13 objectives and want to reroll at that point? Then you are kinda screwed.

PR #13104, opened 2020-08-10
TGUI statpanel

silicons:
yay more wip prs from me that i won't finish for weeks

About the pull request:

i'll be honest this is probably shitcode and stylemistake will probably make something better on tg pretty soon but whatever

Statpanel now has a TGUI panel
verbs stays on info because i'm not porting every verb
MC stays on info (so byond stat) because i don't want to have to have debug systems using tgui for reasons
turf/alt click stays on info for now, because there's no feasible way to port it. byond grid system is uh, promising i guess, but at the least i'll probably make alt clicking force your tab to "Stat/Verbs" so you don't have to manually switch after alt clicking

Why it's good for the game:

stat() is absolutely horrid both for the server and for players and lags everyone out

- [ ] make EmbeddedWindow layout without resizing, title bar + close button, and use it
- [ ] figure out how to tick the updates, i don't want to use tgui subsystem, maybe I'll make Stat() do it by detecting if the user has the tgui panel or the normal stat/verb panel 
- [x] port status panel to it
- [ ] port admin tickets to it
- [ ] put an admin tab with some basic stuff like world.time, coordinates, etc
- [ ] port statclick voting to it
- [ ] port changeling, spells, genetics statclicks to it (which involves deciding how to handle ui_act and statclicks since i don't want this to be too hard to use from the BYOND/DM side)
- [ ] maybe make the DM interface not cbt to use so people can make datum verbs or something that use this haha wouldn't that be epic. this would include something to allow for adding custom tabs
- [ ] alt click forces to verb/info tab since it can't be ported

image


silicons:
(better off doing this on tg or something)

deathride58:
go all the way or no balls

silicons:
@silicons port tg couls system

PR #13107, opened 2020-08-10
The warcrime gun belongs in the armory

DeltaFire15:
## About The Pull Request
Hellgun single-packs are now classified as a armory cargo pack instead of a goodie.

## Why It's Good For The Game
This is a weapon that is really good as wounding, around every second hit or so. Why this would be classed a 'goodie' as opposed to an armory-weapon is beyond me.

## Changelog
:cl:
balance: hellgun single-pack classification: goodies -> armory
/:cl:

### DIFF
--- a/code/modules/cargo/packs/armory.dm
+++ b/code/modules/cargo/packs/armory.dm
@@ -223,3 +223,10 @@
 					/obj/item/ammo_box/magazine/wt550m9/wtrubber,
 					/obj/item/ammo_box/magazine/wt550m9/wtrubber)
 	crate_name = "auto rifle ammo crate"
+
+/datum/supply_pack/security/armory/hell_single
+	name = "Hellgun Single-Pack"
+	crate_name = "hellgun crate"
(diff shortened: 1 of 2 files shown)


Ludox235:
Cringe pr. Delete!

lolman360:
while you're at it bump up the cost too

Hatterhat:
its a 7 shot laser gun that does slightly more damage does it really need to be locked up like this

Ludox235:
> 
> 
> its a 7 shot laser gun that does slightly more damage does it really need to be locked up like this

The wound chance is absolutely fucking bruh.
I've tried it out and like every single shot has done either second or third degree burns, the second shot usually turning it into a catastrophic burn.
Which increases damage taken to that limb by 10%/20%/30%

lolman360:
>its a 7 shot laser gun that does slightly more damage does it really need to be locked up like this

yes burn wounds are super brutal

PR #13109, opened 2020-08-10
adds more heatpipes to snaxi sm

Detective-Google:
it needs more

## Changelog
:cl:
fix:sm
/:cl:

### DIFF
--- a/_maps/map_files/Snaxi/Snaxi.dmm
+++ b/_maps/map_files/Snaxi/Snaxi.dmm
@@ -1640,8 +1640,8 @@
 "aiS" = (
 /obj/structure/closet/secure_closet/CMO,
 /obj/machinery/keycard_auth{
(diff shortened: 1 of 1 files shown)


MrJWhit:
The issue isn't piping, or lackthereof. It's the fact that planetmos will just decide it doesn't like anyone, and stop working. That's why SM work on snaxi is more dangerous than any other station.

Ludox235:
That's why snaxi is such a fucking dumb map. Especially when as a cyborg the fucking SM explodes and people blame you when it's just the map being a shit.

Detective-Google:
> The issue isn't piping, or lackthereof. It's the fact that planetmos will just decide it doesn't like anyone, and stop working. That's why SM work on snaxi is more dangerous than any other station.

Except we literally tested it and found that this was the issue.

PR #13114, opened 2020-08-11
"it'll be fine Sables, nothing bad will happen if we remove the /mob blacklist from radiation"

silicons:
so it turns out i'm forgetful


Ludox235:
Horny radiation depot moment

PR #13128, opened 2020-08-11
sleeper window resize, sleeper window resizable, sleepers no longer work from the inside

Hatterhat:
## About The Pull Request
sleeper windows are now 450x650 and resizable, also no longer work from the inside *without the appropriate var
## Why It's Good For The Game
i like being able to read the sleeper UI actually
## Changelog
:cl:
fix: Sleepers are no longer able to work from the inside.
tweak: Sleeper windows can now be resized. Also they're better to read.
/:cl:


Arturlang:
Why not fix the damage type colors while you are at it so they correspond to their damage types instead of being always red?

Hatterhat:
fuck

PR #13138, opened 2020-08-12
Local Idiot Nerfs Traitor Equivalent To Old Tasers 

KrabSpider:
## About The Pull Request

Lowers the stamina damage and knockdown of energy bows to 80 (originally 160; which is well above the threshold for disarming — either way,) and 40 (originally 60); energy bows of all types no longer disarm on hit.
**Price knocked down to 8 TC from 12; have fun with your more widely available e-combo meta!**

## Why It's Good For The Game

Being able to deal a massive amount of stamina damage while also disarming someone of any weapon they are holding before with a ranged 1.5 cooldown weapon is kind of (_for lack of better wording_) nutty? If e-bows had some kind of ammunition, I'd  _kind of_  of understand why they are in the state they're in — but they aren't. It's an infinite charge PKA subtype that stamcrits in 2-3 shots and makes actually dealing with anyone that uses it a massive pain because by the time you get the sense to pick the weapon you are using up, you either A) had it stolen by said shooter OR B) been hit with another bolt and are in stamina softcrit (or just full on stamcrit). 

## Changelog
:cl:
balance: Energy Bow. Knockdown brought down to 80 from 160. Stamina damage brought down to 40 from 60. No longer disarms. 
balance: Uplink bows are now 8 TC (from 12). .
/:cl:


timothyteakettle:
lower the tc cost of the miniature energy crossbow in the uplink from 12tc to something like 8 if it's going to be made weaker

Tupinambis:
I get the nerf, but I'm with timothy. It will now take 4 shots to stamcrit (or 3 to soft but stamina is quirky until kev fixes that) as opposed to 3/2. AND the disarm is removed. I get it being unfun to fight, but 12 TC is very expensive for something that is now just a better disabler with knockdown. If the goal with the ebow is nonlethal takedowns (Which considering the changes to objectives, it probably is) then it should be more accessible to begin with.

On a related note, we should do a similar type of nerf to revolvers with the same appropriate TC reduction. (With more/cheaper speedloaders in the box to compensate as well would be nice)

KrabSpider:
> lower the tc cost of the miniature energy crossbow in the uplink from 12tc to something like 8 if it's going to be made weaker

>I get the nerf, but I'm with timothy. It will now take 4 shots to stamcrit (or 3 to soft but stamina is quirky until kev fixes that) as opposed to 3/2. AND the disarm is removed. I get it being unfun to fight, but 12 TC is very expensive for something that is now just a better disabler with knockdown. If the goal with the ebow is nonlethal takedowns (Which considering the changes to objectives, it probably is) then it should be more accessible to begin with.
>
>On a related note, we should do a similar type of nerf to revolvers with the same appropriate TC reduction. (With more/cheaper speedloaders in the box to compensate as well would be nice)

Acknowledged.

KrabSpider:
Changes made.

Tumesuo:
no

timothyteakettle:
this is ok

necromanceranne:
As the person who ported it, I suspect it has to do with the greater power of knockdowns over damage. However, the ebow still massively struggles to perform against more than one person, and the 1.5 fire delay is not a small drawback, but what keeps it from being able to combat more than one person without a lot of patience and timing.

At 8 tc it might even be more ridiculous, even at lower damage and even losing disarming.

You can buy two now. And that's filthy powerful with regards to kiting people back or keeping someone disabled. I know this because I deliberately nerfed the fuck out of the large ebow foe explicitly this reason.

PR #13140, opened 2020-08-12
removes an item

Detective-Google:


moltoretardo:
Speedmerge

timothyteakettle:
counter argument:
unknown
edit: also someone could rename anything to 'dildo' and embed it in you using tape

while i'm not against this i'm also not for it because there's literally no reason to either remove it or keep it
edit2: just remembered we nuked the dildo circuit model so yeah nuke these too

ghost:
changed my mind this is neither based nor unbased

Tumesuo:
no

Putnam3145:
this i agree with though

silicons:
not harmful enough for full removal at the moment.

PR #13176, opened 2020-08-15
Increases Kindle cast time and moves a bit of code

DeltaFire15:
## About The Pull Request
This PR does exactly two things:
1. Kindle now has a cast time of 25ds instead of 15ds
and
2. The belligerent scripture is now where it should be in the code, plus the scripture priorities of the driver tier are now correct once again.
Thats pretty much all.

## Why It's Good For The Game
About 1: Kindle as of now seems a bit too effective in combat, increasing the cast time by one more second to 2.5 seconds total should make it a bit less useful in combat whilst keeping its general use.

About 2: This is pretty much just a minor tweak to get the scripture to where it should be, with absolutely zero gameplay impact.

## Changelog
:cl:
balance: kindle cast time: 15ds -> 25ds
tweak: Moved the Belligerent Scripture to where it should be in the code
/:cl:


Ludox235:
Kindle spam in combat is emergent gameplay don't merge :( 

Unironically though kindle spam was absolutely stupid and I'm glad it's being nerfed

PR #13177, opened 2020-08-16
[FUCK] Cogscarabs can no longer do very scary stuff

DeltaFire15:
## About The Pull Request
Scarabs had some stuff they really shouldn't have due to some checks not working as intended. This makes their infinite-resources a thing admins will have to varedit onto the shell, instead of being TRUE by default and being set to false by a check.

## Why It's Good For The Game
COGSCARABS SUMMONING RATVAR IS NOT GOOD

## Changelog
:cl:
fix: Cogscarabs are no longer always Pogscarabs
/:cl:


Ludox235:
emergent gameplay don't remove

PR #13192, opened 2020-08-17
Buffs/hugboxes slaughter demons

Putnam3145:
## About The Pull Request

Makes slaughter demons gain damage and time-spent-moving-fast-when-leaving-bloodcrawl as they eat more people; also made it so that slaughter demons release all their victims on death, albeit with no hearts. They only get the buff if they eat stuff with hearts.

Laughter demons always get the buff, but the buff only increases the damage 1/3 as much, and in fact the laughter demon occasionally does *less* damage (though will, on average, do more). Japes.

## Why It's Good For The Game

Slaughter demon is too weak but also, if it actually kills you, removes you from the round with no hope of ever returning to it. This hopefully fixes the weakness, and definitely fixes the latter; a good medbay will be able to quickly revive all the killed individuals, as long as they prepare enough hearts.

## Changelog
:cl:
balance: Buffed slaughter demon: gets stronger as it eats people
balance: Nerfed slaughter demon: no longer permanently round-removes all who are eaten by it, instead releasing their now-heartless bodies
/:cl:

### DIFF
--- a/code/modules/antagonists/slaughter/slaughter.dm
+++ b/code/modules/antagonists/slaughter/slaughter.dm
@@ -65,6 +65,10 @@
 	var/wound_bonus_per_hit = 5
 	// How much our wound_bonus hitstreak bonus caps at (peak demonry)
 	var/wound_bonus_hitstreak_max = 12
+	// Keep the people we eat
+	var/list/consumed_mobs = list()
+	//buffs only happen when hearts are eaten, so this needs to be kept track separately
+	var/consumed_buff = 0
 
 /mob/living/simple_animal/slaughter/Initialize()
 	..()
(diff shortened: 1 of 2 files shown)


Ludox235:
I mean... The whole point of taking slaughter demons instead of laughter demons was they took the crew away forever instead of releasing them all.

code/modules/antagonists/slaughter/slaughter.dm
```diff
@@ -111,8 +115,44 @@
...
+	for(var/mob/living/M in consumed_mobs)
+		if(!M)
+			continue
+		var/turf/T = find_safe_turf()
+		if(!T)
+			T = get_turf(src)
+		M.forceMove(T)
+
+/mob/living/simple_animal/slaughter/proc/refresh_consumed_buff()
+	melee_damage_lower = 22.5 + (0.5 * consumed_buff)
```
lolman360, line 138: `melee_damage_lower += (0.5 * consumed_buff)`
. with upper as well

Putnam3145: that would lead to a quadratic curve instead of a linear one, each heart increasing damage by (0.5*hearts) instead of just 0.5; I know having 22.5 as a magic number is ugly as sin, and I could probably just make it a define, but I wrote it this way mostly so I could make laughter demon buff work differently.

Putnam3145:
> I mean... The whole point of taking slaughter demons instead of laughter demons was they took the crew away forever instead of releasing them all upon death.

Gonna be frank here: this is irrelevant. It sidesteps whether that's a good idea for a random event in the first place, basically. For wizard rounds, fine, they're chaotic and short, but slaughter demons can just spawn, at random.

keronshb:
> 
> 
> > I mean... The whole point of taking slaughter demons instead of laughter demons was they took the crew away forever instead of releasing them all upon death.
> 
> Gonna be frank here: this is irrelevant. It sidesteps whether that's a good idea for a random event in the first place, basically. For wizard rounds, fine, they're chaotic and short, but slaughter demons can just spawn, at random.

I think they're ultra-rare for that reason no?  Like out of the year I've been here I've only seen a midround Slaughter Demon once, maybe twice maximum.  I think one of them was bussed too.  I've seen more Solo Ops, which are also hyper-rare, than I have midround Slaughters.

Not that this is a bad change, but I think their spawnrate should probably go up if this goes through.

code/modules/antagonists/slaughter/slaughter.dm
```diff
@@ -207,7 +240,22 @@
 		if(3)
 			adjustBruteLoss(30)
 
-/mob/living/simple_animal/slaughter/laughter/proc/release_friends()
+/mob/living/simple_animal/slaughter/laughter/refresh_consumed_buff()
+	melee_damage_lower -= 0.5 // JAPES
+	melee_damage_upper += 1
```
ghost, line 245: is there a reason the slaughter and laughter damage changes are done differently

necromanceranne:
Are slaughter demons spawning regularly now? That absolutely shouldn't be happening.

keronshb:
> 
> 
> Are slaughter demons spawning regularly now? That absolutely shouldn't be happening.

As you say this I saw my 3rd and 4th midrounds two days in a row, which is a complete surprise.

code/modules/antagonists/slaughter/slaughter.dm
silicons: @Putnam3145

Putnam3145: primarily because laughter demons always get the buff

Putnam3145:
there's been a midround slaughter demon event since I started playing the game, yeah

PR #13201, opened 2020-08-18
makes it so you can craft armwraps roundstart again.

kiwedespars:
## About The Pull Request

regular armwraps no longer require durathread.

## Why It's Good For The Game

why would you go through all that trouble for a trait that buffs punches slightly and also makes it so you can't use guns.

## Changelog
:cl:
tweak: removed durathread from armwraps recipe.
/:cl:


necromanceranne:
I was wondering what happened to the recipe. It seems it wasn't a deliberate nerf from ghommie just the next most reasonable stack to choose. My advice would be making it require sutures as a stand in for string. Or tape.

You guys port tape yet?

kiwedespars:
good idea, we do have tape.

PR #13214, opened 2020-08-19
[s] Min and max are two very distinct procs...

DeltaFire15:
## About The Pull Request
Fixes the modified damage formula from #12658, and I will personally table @silicons for merging it without looking at it first.
Using max instead of min does not cap it at 20, it set the minimum amount of damage **any** amount of oil splashing does to 20. 
Obivously, thats not fine. This PR fixes that.

## Why It's Good For The Game
Cooking Oil still instakilling people. aswell as now being capped at a **minimum** damage of 20 instead of a maximum of 20 obviously ain't good

## Changelog
:cl:
fix: The cooking oil damage formula is no longer scuffed.
/:cl:


timothyteakettle:
FUCK

PR #13220, opened 2020-08-20
Expanding airlocks for the majority of stations

MrJWhit:
## About The Pull Request

Increases the majority of cycling airlocks from 1 wide, to 2 wide.
This does not include:
-Lambda and Snaxi, because those are actively being worked on at the moment
-BoxStation Engines, because there's a PR for one of the engines, and I'd rather do it all at once, then forget.
-Arrivals, departures, and all associated shuttles. I could be persuaded to make these adjustments, but, expanding arrivals can actually cause a nonzero amount of issues because it moves things around pretty significantly, and departures tends to have the extra space. Shuttle cycling airlocks aren't used to enter and exit the station basically ever.

## Why It's Good For The Game

With fastmos, carrying something into the station with a 1 tile wide movement, is actually just awful. 
Open the first airlock. It gets pushed out of your hand. Go back to grab it. Open the second airlock, more air pushes it out of your hand. Go back, and depending on the amount of air on that part of the station, wait for that to depressurize to bring the object it. While waiting, the first airlock closes, because it's a cycling airlock. Open the first one, the second airlock closes. Move forward to open the second airlock, and hopefully by this point the air doesn't push the object out of your hand again. Then move on with your day.
This fixes that.

## Changelog
:cl:
tweak: Increases the majority of airlocks by 1 tile.
/:cl:

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -5033,7 +5033,7 @@
 	},
 /obj/effect/mapping_helpers/airlock/cyclelink_helper,
 /turf/open/floor/plating,
-/area/solar/port/fore)
+/area/maintenance/solars/port/fore)
 "ajr" = (
 /obj/structure/table/wood,
 /obj/machinery/atmospherics/pipe/simple/supply/hidden,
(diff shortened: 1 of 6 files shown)


EmeraldSundisk:
Can we get some visual examples?

MrJWhit:
Sure, boxstation thing:
image


The delta station I had to rework one part, and this is it:
image

One of the few OmegaStation adjustments:
image

Random kilostation adjustment, this station was easy all around:
image

EmeraldSundisk:
Ah, interesting. #13205 should handle CogStation, but I don't mind taking that on if it doesn't.

MrJWhit:
If you want to update it, feel free, if you don't feel like it, I can handle it no problem. There isn't a rush or anything.

PR #13225, opened 2020-08-21
Heavily nerfs the TEG.

jack-avery:
## About The Pull Request

Nerfs the TEG's power generation heavily.

## Why It's Good For The Game

Before, it was way to easy to generate multiple MW of power with the TEG, especially with cheap setups.
This tweaks the power generation to be still powerful, so as to not make it completely worthless. Could still go further.
This is just initial testing, I guess. It's roughly 1/9 as powerful as before.

EDIT:
This changes the way that the TEG's power generation is calculated to be LOGISTICAL rather than LINEAR. You can expect to create about 100kW with a 3x3 build now, but that power generation increases rapidly as you increase the heat past the heat difference that a heater/freezer setup can create.
This is a much better handling of the TEG's power generation, thank you for the suggestions whoever suggested things

https://cdn.discordapp.com/attachments/161574200581029888/689392827545026564/unknown.png

## Changelog
:cl:
tweak: Thermoelectric Generator power output
/:cl:

### DIFF
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -42,6 +42,7 @@
 
 
 #define GENRATE 800		// generator output coefficient from Q
+#define LOGISTIC_FUNCTION(L,k,x,x_0) (L/(1+(NUM_E**(-k*(x-x_0)))))
 
 /obj/machinery/power/generator/process_atmos()
 
@@ -66,7 +67,7 @@
 				var/energy_transfer = delta_temperature*hot_air_heat_capacity*cold_air_heat_capacity/(hot_air_heat_capacity+cold_air_heat_capacity)
 
 				var/heat = energy_transfer*(1-efficiency)
-				lastgen += energy_transfer*(efficiency-0.41)
+				lastgen += LOGISTIC_FUNCTION(1000000,0.0034,delta_temperature,2000)
 
 				hot_air.set_temperature(hot_air.return_temperature() - energy_transfer/hot_air_heat_capacity)
 				cold_air.set_temperature(cold_air.return_temperature() + heat/cold_air_heat_capacity)
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -66,7 +66,7 @@
 				var/energy_transfer = delta_temperature*hot_air_heat_capacity*cold_air_heat_capacity/(hot_air_heat_capacity+cold_air_heat_capacity)
 
 				var/heat = energy_transfer*(1-efficiency)
-				lastgen += energy_transfer*efficiency
+				lastgen += energy_transfer*(efficiency-0.41)
 
 				hot_air.set_temperature(hot_air.return_temperature() - energy_transfer/hot_air_heat_capacity)
 				cold_air.set_temperature(cold_air.return_temperature() + heat/cold_air_heat_capacity)
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -42,7 +42,6 @@
 
 
 #define GENRATE 800		// generator output coefficient from Q
-#define LOGISTIC_FUNCTION(L,k,x,x_0) (L/(1+(NUM_E**(-k*(x-x_0)))))
 
 /obj/machinery/power/generator/process_atmos()
 


Tupinambis:
Not a bad start, but the TEG needs to have the 3x3 funny setup removed entirely from being functional before the TEGs numbers are adjusted honestly. Power generation and consumption across the board need rebalancing at some point, but currently it doesn't matter so much as it matters how easily the TEG can be cheesed.

EmeraldSundisk:
I don't think the problem is so much with TEGs as a whole as it is the tiny meme setups. If you have a larger one, involved to the point where it's supposed to serve as the station's main engine, it should still be able to do that.

lolman360:
as long as heaters/freezers into TEG are not thermodyamically >100% efficient then the mini teg setups will remain viable

and doing so would require touching heaters/freezers or biasing tegs towards favoring higher heat than just a temperature differential because the diff you can generate between t2 heater/t2 freezer is pretty big (i think it's about 800k? off the top of my head t2 freezer goes to 43.15 and heater is around ~700.)

jack-avery:
It should be noted this is one of a few things that should be done to bring the TEG to an acceptable position.
**I understand that just directly nerfing the TEG's power output does not fix it's strongest inherent flaws.**
BUT... it's wrong to say that this does not fix one of it's problems.
As of now, the power generation in comparison between freezers and room temperature is already enough to power the station.
This PR brings the power generation of that down to (based on rough mental math) 1/3 of the average station load.
Obviously, heaters can be used to bump that up to multiple MW.

**The next step, in my opinion, is outright making the TEG not work with heaters. _At all._**

Why?
The difference in effort between using a freezer instead of a cooling loop and a heater instead of a burn chamber is IMMENSE.
The burn chamber requires a lot more piping and setup while a cooling loop is just HE pipes in space.
Plus, it's not a stretch for this, since the TEG already stalls at 9500kPa-- while we have canisters that can handle heats trillions of times hotter than the core of the sun. We're not exactly dealing with a realism simulator, here.

tl;dr:
This is one of a couple things that MUST be done to bring the TEG down to size.
The power generated from the heat difference between a freezer and room temperature is too large, and this makes it so that when eventually the meme funny 3x3 heater/freezer TEG is outright **REMOVED** like it should be, the only way to make the TEG generate a considerable amount of power is to make a burn chamber for the hot loop.

MrJWhit:
So, as it currently stands, extra power is effectively pointless. Power is currently a binary "is this on, or is this off?". Stations take up between 150kW and 300kW of power at full capacity, and any engine can supply that, even an upgraded turbine can give enough power for an entire station.

Extra power doesn't help, or harm, the station. It's just there to push the limits of what engineers can do for fun and education. 

I don't see why this should be there a nerf in the first place, if there was going to be a power refactoring, I can see a different nerf, but as-is I don't know why this should change.

Why does the extra power matter? The TEG is, weirdly, a difficult engine for newer players. Getting a big number at an output is deeply rewarding to figuring out all of the wierdness associated with it. Plus with a higher number means that the variance of power output is easier to read for the new player.

**This doesn't even stop the 3x3 TEG memes**, if you wanted to do that, require a hard tempature difference threshold to generate power. Heaters and Freezers can only max out at a difference of ~1500k, so make it need a bigger threshold to work.

Ludox235:
reminder the TEG can do 10mw of power while a CO2 SM does around ~2MW even when at around 5000 MEV/CM3 from a CO2 setup

ghost:
Like MrJWhit said, what is the point of nerfing power output when that extra output provides no benefit? It can't be imbalanced by nature because it's extremely easy to power the station with any engine.

Additionally, the whole notion of "balancing" power is shortsighted because it hinges on the idea that every round has a full staff of competent players to maintain the engine(s), which isn't the case. The more sane way to "balance" power would be to provide incentives to maximize power output while making sure minimal setups can still power the station, not just reducing the output of existing engines because "same numbers = good".

MrJWhit:
> reminder the TEG can do 10mw of power while a CO2 SM does around ~2MW even when at around 5000 MEV/CM3

You can cheese a lot more out of that by putting rad collectors in the SM chamber itself.

MrJWhit:
With the change from a straight downscale of power generation to a logistic function, I am perfectly happy with this PR. It stops 3x3 from generating too much power, and keeps the TEG being used properly.

shellspeed1:
Pretty solid pr. A single 3x3 teg working as backup power for a part of the station is fine and it more than encourages proper teg burn chamber builds. I would like to point out to people complaining that this is too much power or it doesn't make sense you can get better numbers out of a pacman which does use power. This pr at least in my opinion would deal with the issue of teg bullshit while still making them viable for autism projects.

PR #13249, opened 2020-08-23
removes firedoor autoopen on touch

silicons:
## Why It's Good For The Game

it was pretty nice when you could actually use firedoors to pose an obstacle rather than having them be worthless outside of an actual fire.

## Changelog
:cl:
balance: firedoors no longer automatically open on touch when there's no pressure differences.
/:cl:


Tupinambis:
It was also very annoying accidentally triggering them. Good change.

PR #13253, opened 2020-08-23
Certain Chemsplosions no longer trigger while inside a mob.

ghost:
## About The Pull Request

Most reagent explosions no longer react inside mobs. Potassium/Water is nerfed in explosion strength.

## Why It's Good For The Game

Honestly, getting exploded after getting shot with multiple syringes and going into crit/getting gibbed from the explosions is kinda wack. Potassium/Water still works, but it's weaker.

## Changelog
:cl:
del: Explosive chems no longer react inside people.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
+++ b/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
@@ -50,8 +50,7 @@
 	name = "Explosion"
 	id = "potassium_explosion"
 	required_reagents = list(/datum/reagent/water = 1, /datum/reagent/potassium = 1)
-	strengthdiv = 10
-	mob_react = FALSE
+	strengthdiv = 50
 
 /datum/chemical_reaction/reagent_explosion/potassium_explosion/holyboom
 	name = "Holy Explosion"
--- a/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
+++ b/code/modules/reagents/chemistry/recipes/pyrotechnics.dm
@@ -44,18 +44,20 @@
 	required_reagents = list(/datum/reagent/nitroglycerin = 1)
 	required_temp = 474
 	strengthdiv = 2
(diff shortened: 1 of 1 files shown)


Sishen1542:
keep water/pot reacting in mob tbh

ghost:
> keep water/pot reacting in mob tbh

I suppose I could, but I'll have to nerf the resulting explosion.

ghost:
Pot/water now can still react in mobs, it's just that even with the best syringe, bluespace, the explosion will be very weak, and won't even delimb. As a result of that, its explosions generally have less of an oomph, but are still able to destroy structures and knock people down.

shellspeed1:
I mean, I wanted to propose a different solution but people don't seem to like it. Might as well anyway. What if instead of people getting gibbed by bombs and as such round removed, we made it flings their organs and limbs everywhere so a bored doctor could take all the bits and pieces and rebuild the patient?

ghost:
> I mean, I wanted to propose a different solution but people don't seem to like it. Might as well anyway. What if instead of people getting gibbed by bombs and as such round removed, we made it flings their organs and limbs everywhere so a bored doctor could take all the bits and pieces and rebuild the patient?

Honestly that works by itself as a solution to many things. The main thing I wanted to get rid of with this pr is getting exploded via a syringe. There's not really a counter to that, plus its not really fun to get hit by, as an antag, or as a non-antag.

Kraseo:
cringe

silicons:
not a fan of how this is handled, especially the pota/water nerf.

if it comes down to it it wouldn't be impossible to make explosions have an ignore/dampen explosive damage object-cache of sorts to make the syringes unable to instagib, but as is i don't think this is necessary.

ghost:
I think I see a way to do this differently, with several ideas. I'm just gonna close this and make a new pr for it.

PR #13255, opened 2020-08-24
you can choose limbs to be prosthetic/amputated

timothyteakettle:
## About The Pull Request
**as with my other customization prs, please testmerge this first**

prosthetic trait gone
you can choose limbs to be amputated and prosthetic now
you can only have up to two prosthetics though
having at least one prosthetic gives you one quirk point (having more than this doesnt give more)

this is going to get conflicts with my other prs i think, so uh, if i don't realise just yell at me

## Why It's Good For The Game
customization

## Changelog
:cl:
add: due to changes in policy, and several lawsuits, Nanotrasen has been forced to allow disabled people to sign up
/:cl:


timothyteakettle:
i'm gonna make it so having the paraplegic trait blacklisted means you can't have amputated limbs as that job either

timothyteakettle:
okay if you have paraplegic blacklisted then it regenerates your limbs

moltoretardo:
Do you still get the bonus trait point if you get your limbs regenerated. Some people like me could powergame it hjonk

timothyteakettle:
you don't get points for amputated limbs only prosthetic

timothyteakettle:
travis pls

timothyteakettle:
hey please don't merge or testmerge this i need to fix the issue with json encode/decode so it only runs when there's data there, and also i might aswell port rp prosthetics while im at it because qol

also aside from this i'm going to make sure that cloning: doesnt give free amputated limbs / doesn't regenerate amputated limbs / oh also prosthetics should be useable in hand to select the type or something because that would be neat

code/modules/client/preferences_savefile.dm
```diff
@@ -874,6 +874,9 @@ SAVEFILE UPDATING/VERSIONING - 'Simplified', or rather, more coder-friendly ~Car
 	WRITE_FILE(S["scars3"]						, scars_list["3"])
 	WRITE_FILE(S["scars4"]						, scars_list["4"])
 	WRITE_FILE(S["scars5"]						, scars_list["5"])
+	if(length(modified_limbs))
+		WRITE_FILE(S["modified_limbs"]				, safe_json_encode(modified_limbs))
+	WRITE_FILE(S["chosen_limb_id"],   chosen_limb_id)
```
silicons:
```suggestion
	if(islist(modified_limbs))
		WRITE_FILE(S["modified_limbs"]				, safe_json_encode(modified_limbs))
	WRITE_FILE(S["chosen_limb_id"],   chosen_limb_id)
```

code/modules/client/preferences_savefile.dm
```diff
@@ -514,6 +514,8 @@ SAVEFILE UPDATING/VERSIONING - 'Simplified', or rather, more coder-friendly ~Car
 	S["scars3"]							>> scars_list["3"]
 	S["scars4"]							>> scars_list["4"]
 	S["scars5"]							>> scars_list["5"]
+	if(length(S["modified_limbs"]))
+		safe_json_decode(S["modified_limbs"])					>> modified_limbs
```
silicons:
```suggestion
	var/limbmodstr
	S["modified_limbs"]													>> limbmodstr
	if(length(limbmodstr))
		modified_limbs = safe_json_decode(limbmodstr)
	else
		modified_limbs = list()
```
also you might want to sanitize it in sanitization proc too, with an equally runtime-safe setup

code/modules/client/preferences_savefile.dm
silicons: this ensures overwrite on empty

code/modules/client/preferences_savefile.dm
silicons: (this ensures if the saved data is empty/null they get their limbs wiped instead of any potentially old data lingering)

PR #13277, opened 2020-08-27
Reworks Anatomic Panacea

ghost:
## About The Pull Request

Reworks anatomic panacea into a buff effect. Rather than it injecting a cocktail of antihol/mannitol/mutadone/pentetic jelly it aims to make the skill directly purge the target. The ability to purge zombie tumor and alien ambryo remains. It also ups the cost from 1 point to 2 due to the increased usefulness, though I'm unsure whether or not the numbers and/or length of the ability are good, low or too high

## Why It's Good For The Game

Hopefully makes anatomic panacea worth choosing as an ability due to the increased effects it provides, as well as no longer purging all chems due to injecting yourself with pentetic jelly

## Changelog
:cl:
tweak: Changed anatomic panacea into a direct buff, instead of a chem injection
balance: Changed the values of anatomic panacea
imageadd: Added a new icon for panacea's buff alert
/:cl:


Detective-Google:
if this is gonna be snowflaked out of chemicals it should have a counter similar to fleshmend i.e fire

Hatterhat:
let me keep my good mutations from the wishgranter god damn you

ghost:
> 
> 
> if this is gonna be snowflaked out of chemicals it should have a counter similar to fleshmend i.e fire

I could do that, though fire/burn is the most common way to counter a lot of antag abilities. So perhaps something different?

Also, thoughts on the numbers? I'm not sure what a good amount would be



> 
> 
> let me keep my good mutations from the wishgranter god damn you

Does normal mutadone clear wishgranter mutations?

Hatterhat:
yes probably

PR #13284, opened 2020-08-28
Experimental optimization for get_hearers_in_view.

Putnam3145:
## About The Pull Request

Makes get_hearers_in_view use dview instead of screwing around with tile luminosity, makes it iterate through the returned list only once instead of twice with filters, and replaces length() with .len, which doesn't check for null (it's a LOCAL PROC VARIABLE IT'S NOT GONNA DIE RANDOMLY) and thus is ~1.5 times (!) faster.

This overall results in an overall speedup in benchmarking (i.e. making every single mob say a phrase over the radio repeatedly) of approximately 2x.

## Why It's Good For The Game

get_hearers_in_view causes a weird amount of lag spikes, so it's best it isn't bad.

## Changelog
:cl:
refactor: get_hearers_in_view does fewer hackish things and runs a bit faster
/:cl:

### DIFF
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -254,16 +254,11 @@
 	if(R == 0)
 		processing += T.contents
 	else
-		var/lum = T.luminosity
-		T.luminosity = 6
-		var/list/cached_view = view(R, T)
-		for(var/mob/M in cached_view)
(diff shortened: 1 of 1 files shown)


Putnam3145:
doesn't work LOL

PR #13122, opened 2020-08-11
load_delay var for /obj/item/ammo_box, nerfs shotgun stripper clips (hatter why'd you pr them)

lolman360:
allows you to set a `load_delay` var for `/obj/item/ammo_box`. on hitting a magazine (or gun with internal magazine) with a `/obj/item/ammo_box` with `load_delay`, it'll run a do_after for `load_delay` seconds.

shotgun stripper clips now have a load_delay of 0.6 seconds

shotgun stripper clips now only load one shot at a time- as far as i'm aware we don't have shotguns with internal box magazines, so you're either
1. loading a doublebarrel with a 4-round stripper clip. 
2. loading a triplebarrel with a 4-round stripper clip 
3. loading a tube magazine with a stripper clip. use a speedtube idiot.

all of these are fucking stupid and you should be fumbling around with them.

## Changelog
:cl:
balance: shotgun stripper clip nerf. ammoboxes can now accept a load_delay that happens when they attack a magazine, internal or external.
/:cl:

PR #13287, opened 2020-08-28
Reverts load delays because it's still a problem and by god am I just done at this point

silicons:
Reverts Citadel-Station-13/Citadel-Station-13#13122

not as good as i thought it was in retrospect it's time to take this out back


lolman360:
you can just remove it from shotgun stripper clips they still have their uses

PR #13288, opened 2020-08-28
meta brig remap

Detective-Google:
## About The Pull Request
Meta brig is way different now.

## Why It's Good For The Game
most maps haven't been changed in a long ass time 

I was in discord bugging people for feedback and nobody had anything concrete to say; if people suddenly hate this after the pr I will implode

## Changelog
:cl:
tweak: meta brig is different now
/:cl:


Detective-Google:
metabrig

Tupinambis:
"Nobody had concrete to say" yeah sure okay I gave you paragraphs of feedback on the map but you ignored all of it because you decided to block me on discord for some reason. 🥴 So glad I put all that effort in to give you constructive feedback only for you to claim nobody had any.

Tupinambis:
Quoting from what I said on discord: I love pretty much all of the center of your security design, I think decorating it with vendors, chairs, and the like would add a lot. It reminds me of cogmaps science area, which would be a good place to pull from.

Besides that I think maybe having perma on the RIGHT SIDE and everything else on the left would look and flow better. It might give perma more room as well. 

The maint room on the northeast side should be connected to the rest of sec maint, rather than being seperated, otherwise that part of maint will feel to small and not realistic to hide in. 

That hallway between maint and dorms is a bit long and empty, it might be good to add like an observatory or something there, since meta has a general lack of "vibe spots"

Execution is probably too small

Not having that maintenance area above the vault causes a very similar issue with maint feeling too small and making it harder to hide around places.

Overall I think your sec design would just benefit from being bigger and more comprehensive, since it's very compact but doesn't leave as much room for perma and maintenance.

YakumoChen:
A lot better than the last brig design but still has some flaws I can think of. The biggest one is that there's no 'lookout' window for the space above the armoury, you need to rely on cameras to so much as look north of the brig now, so you won't really be able to deal with anyone invading. I think the brig medbay and transfer should switch places honestly, and that would facilitate said lookout.

I like the circles around the brig, having lots of pathways and routes is good design. 

Kinda agree with above, I don't really know where all the maint around brig went and it kinda needs maint in general, it actually feels a bit too obnoxiously well-defended because of that.

Detective-Google:
awesome the fucking pool pr bricked this one and it's not letting me fix conflicts :)

PR #13290, opened 2020-08-29
Nukies can't get wall-boots anymore

ForrestWick:
## About The Pull Request
Why was this even a thing. Removes wall walk boots from nukie uplinks

## Why It's Good For The Game
Spawning as a nukie and raiding CC for gamer gear is bad. ALSO nukies have more than enough ways to breach the station or get through obstacles, they don't need fucking /noclip on top of these options, much less for 6tc when war ops exists.

## Changelog
:cl:
balance: removes wall walking boots from nukie uplink
tweak: removes wall walking boots from nukie uplink
/:cl:


Detective-Google:
low charges with no way to recharge without either waiting a long time for the station to have bluespace or buying loads of launchpads and breaking them; there are other ways to get out of the nukie base and I think the ideal solution is to add a ZTRAIT_STATION check for the boots.

MrJWhit:
The wall-walking boots don't use bluespace crystals, you have to buy special ammo for it.

Speaking of, the recharge should probably be included as well, otherwise you can buy the ammo without the actual item itself.

PR #13296, opened 2020-08-29
chasms no longer drop anchored things unless it's a mob

silicons:
## Why It's Good For The Game

kinda silly you can't put pipes over chasms and stuff

## Changelog
:cl:
tweak: anchored objects (not mobs!) are now immune to chasms
/:cl:


DeltaFire15:
Issue with doing it like that is that one can for example place machinery or anchored objects on lavaland next to a tendril, then kill it, and have the objects float there.
This is especially _problematic_ with quantum pads, as this'd cause people who teleport there to immediately get chasmd. There's probably other problematic machinery and objects, but this one comes to mind as the very first.

PR #13301, opened 2020-08-30
on someone's request i am making meteors slice through station ruins like butter so they are no longer free meteor shields because i want to make meteor timer 10 minutes instead of 4

silicons:
opinions wanted

basically this means anything asteroid-y turf doesn't count as a hit


lolman360:
on whose request, i need to know so i can send death threats over my spessman

MrJWhit:
Sup.

lolman360:
good suggestion

CameronWoof:
A warning you can work with in exchange for needing to heed the warning. Could work well.

timothyteakettle:
seems good

PR #13304, opened 2020-08-30
Revert "Buffs defib users by making them stop pulling automatically"

GrayRachnid:
Reverts hugbox. There's an upgrade for this.

#8235


GrayRachnid:
image
no fucking clue how these runtimes have anything to do with this pr

Poojawa:
First, it ain't hugbox so much as QOL, only affects user who has their hands full of paddles. Makes sense they wouldn't be pulling.

Secondly, it also prevents racing off into the sunset to defib at the same time.

Thirdly, update your branch.

GrayRachnid:
> 
> 
> First, it ain't hugbox so much as QOL, only affects user who has their hands full of paddles. Makes sense they wouldn't be pulling.
> 
> Secondly, it also prevents racing off into the sunset to defib at the same time.
> 
> Thirdly, update your branch.

There has always been a defib upgrade specifically to stop people from being shocked.

Besides, only one person can be pulling on them at any point in time, that person is usually the person defibbing them.

"It makes sense" in the real world but this is a video game and some things need to be balanced around what happens more often(In this case, the person pulling being the one defibbing).

And thanks for the tip

Poojawa:
There wasn't always an upgrade disk, Trilby added them

punishing someone who has UI lag is not fun at all, it's a minor QoL tweak for heat-of-the-moment situations. Just nice to have, really.

The-Sun-In-Splendour:
First, based pr
second, merge this

PR #13309, opened 2020-08-30
[tm only]Revert "separates tongues speech handling into accent datums" - ipcs can't talk

silicons:
Reverts Citadel-Station-13/Citadel-Station-13#13268


timothyteakettle:
the issue is I forgot to make the robot accents handle_speech return the speech args causing speech to become null, should i just make a revert of this revert and then a commit that implements it, i went through and it's the only one that doesn't return

PR #13310, opened 2020-08-30
Makes fusion analyze not assume we're still on fusion v5

Putnam3145:
## About The Pull Request

There's an entire list attached to EVERY GAS MIXTURE used ONLY FOR FUSION and it DOESN'T MAKE SENSE. this is STILL TRUE ON TG AS WELL. This makes it actually work right.

## Why It's Good For The Game

if we're gonna bloat the game with 60000 lists better make it work

## Changelog
:cl:
fix: Fusion scan is now actually useful.
/:cl:


html/changelog.html
```diff
@@ -50,6 +50,7 @@
 		-->
 		<div class="commit sansserif">
 
+			<h2 class="date">22 July 2020</h2>
```
Putnam3145:
```suggestion
```

PR #13313, opened 2020-08-30
Makes plastitanium rapier not silently poison for pacifists.

Putnam3145:
## About The Pull Request

Instead, it makes it LOUDLY poison. He he he.

## Why It's Good For The Game

Turns out this thing's pacifist safe, but best it have logging and not be completely silent.

## Changelog
:cl:
balance: Plastitanium rapier no longer silently sleeps with no chance at counterplay when used by pacifists.
/:cl:

### DIFF
--- a/code/game/objects/items/melee/misc.dm
+++ b/code/game/objects/items/melee/misc.dm
@@ -236,7 +236,7 @@
 	if(iscarbon(target))
 		if(HAS_TRAIT(user, TRAIT_PACIFISM))
 			visible_message("<span class='warning'>[user] gently taps [target] with [src].</span>",null,null,COMBAT_MESSAGE_RANGE)
-		log_combat(user, src, "slept", I)
+		log_combat(user, target, "slept", src)
 		var/mob/living/carbon/H = target
 		H.Dizzy(10)
 		H.adjustStaminaLoss(30)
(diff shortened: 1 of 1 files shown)


necromanceranne:
Actually uh maybe it's best that it doesn't work for pacifists at all. I'll fix it later. My bad. Afterattacks are weird with pacifists and I always trip up on this.

Putnam3145:
It ain't an afterattack, it's just an attack, which is probably the problem. Anyway, actually, this squares with all the other stuff pacifists can do (tasers, disablers etc.).

necromanceranne:
Don't even know my own code, guh huh.

That might be true, but I pictured the weapon actually injuring people to inflict the poison. But this is fine too honestly.

code/game/objects/items/melee/misc.dm
```diff
@@ -234,6 +234,9 @@
 /obj/item/melee/rapier/attack(mob/living/target, mob/living/user)
 	. = ..()
 	if(iscarbon(target))
+		if(HAS_TRAIT(user, TRAIT_PACIFISM))
+			visible_message("<span class='warning'>[user] gently taps [target] with [src].</span>",null,null,COMBAT_MESSAGE_RANGE)
+		log_combat(user, src, "slept", I)
```
necromanceranne:
```suggestion
		log_combat(user, src, "slept", target)
```

I think this is the bit that's hecc'd.

The-Sun-In-Splendour:
while you're at it make slicing throats not a thing pacifists can do :')

PR #13314, opened 2020-08-30
No more chemspam: Ports portions of ArcaneMusic's botany rework from TG.

TheObserver-sys:
## About The Pull Request
This PR aims to change how plants handle nutrients and mutations. This directly ports the relevant portions of TGBotany, without bringing the other systems into play.

What's new?:
A new gene known as instability. This stat determines if the other stats will be in flux, as well as rolling for new species, or at even more extreme levels, new traits and new chemical contents. No more depending on just strange seeds to get something fun, at the cost of your time. Instability has five thresholds in this system.
At 20+ Instability: Plants will weak roll their stats every life cycle, this is akin to 1 unit application of mutagen from the old hydroponics system.
At 30+ Instability: Plants, on harvest, have a chance of producing mutated offspring. This persists through every instability level.
At 40+ Instability: Plants will strong roll their stats every life cycle, akin to 2u+ applications of mutagen.
At 60+ Instability: Plants will attempt to roll new species every life cycle. Akin to a 5+ application of mutagen.
At 80+ Instability: Plants will roll for new traits and new chemical genes, every cycle the likelihood increasing as the instability does.

Autogrow functionality: Control-Clicking a tray will activate a somewhat power intensive autogrow mode, which will kill weeds, keep the tray watered, and kill pests, while cutting the reagent drain in your tray in half. Good for when you need to step away, and you don't have earthsblood to stabilize the tray for a while.

A reason to upgrade the hydroponics trays: Autogrow is pretty decent on its own, but, upgrading the hydroponics trays quadruples the water containment, raises the reagent capacity to 50, and most importantly, cuts the reagent drain dramatically, meaning less time fiddling with plant chems.

Reworked and New Fertilizers: 
Old fertilizers have been reformulated to be useful, and new fertilizers have come into being to interact with stats that you'd normally not be able to interact with outside the gene editor.
L4Z: Now lightly heals the plant, and increases plant instability, making it viable for actual use in obtaining new plants.

EZNutriment: Lightly adjusts instability, increases potency passively. Trays start with 10 units of this. A better bog standard.

Robust Harvest: Boosts yield and potency, drops instability. No longer applies the 30% bonus yield effect. 

Liquid Earthquake: Buffs plant production speed, but also welcomes additional weed growth into the party. Use sparingly.

Endro Grow: Reduces the yield and potency of your plant, but makes it far more resilient, increasing the endurance stat directly. 

Earthsblood: While earthsblood cannot be used to force permanent self sustain anymore, it is still very well worth obtaining. When applied to a tray, the plant inside will become immaculately healthy, plentiful, and potent, at the cost of absolute stability, making it nearly impossible to mutate a plant while it's still in the reagent tank.

Buzz Fuzz: Lightly heals the plant, kills pests, and increases potency. I honestly wasn't aware this existed until I started working on this.

Unstable Mutagen: Greatly boosts instability, but the tray gets poisoned for it. It's best to not use straight unstable mutagen, and instead mix it with Left 4 Zed and Ash to keep your plant healthy.

A new mode for the plant analyzer: Activating the Plant Analyzer in hand now allows you to switch between plant stats mode, and plant chems mode, which is useful to learning what new chemicals your plant has generated after a period of high instability. This can also be used on seeds and plants on the ground.

What this PR does not bring:
Grafting system
Cross Pollination System
Destruction of the DNA Manipulator.

What will need work if this PR gets merged:
Circuitry pertaining to botany robots
Plumbing pertaining to hydroponics trays.

## Why It's Good For The Game
This has been a long time coming, as maintainers and contributors have been looking for a way to remove the heavy chem cost and heavy chem spam nature of botany. By adding a stat that handles mutations, and adjusting chems and fertilizers to work within that, we can greatly reduce the need for a chem dispenser, and therefore open up the job to new botanists, while allowing returning/experienced botanists to achieve faster and more interesting results. This also changes the beginning of botany, mostly through not forcing people to rush Gaia if they do not wish to.

## Changelog
:cl: ArcaneMusic, with minor tweaks by TheObserver-sys
add: Adds and modifies fertilizers, as well as a new stat, Instability.
del: Removes the rather disused functionality of irrigation hoses, temporarily disables circuit use on hydroponics trays.
tweak: Plant Analyzer have been upgraded. Using one in hand will now switch between a stat view of your plant, and a chemical view of your plant.
tweak:Trays now accept and store reagents, as well as coming with an autogrow mode. However, these upgrades came at the cost of old self sufficiency, meaning you must attend to your plants a bit more often. 
tweak: Earthsblood, while not being able to gild trays anymore, has been found to still be quite powerful as a fertilizer.
/:cl:


triplezeta:
good pr

TheObserver-sys:
For the record, I welcome any tweaks maintainers wish to bring to this before it gets undrafted/testmerged, so long as the mutations still work, and the early potential for mutated plants in a harvest still work.

code/modules/plumbing/plumbers/autohydro.dm
```diff
@@ -42,7 +41,7 @@
 
 /obj/machinery/hydroponics/constructable/automagic/process()
 	if(reagents)
-		applyChemicals(reagents)
+		reagents.add_reagent(reagents)
```
lolman360, line 43: does this do anything, even

lolman360: actually if you're going to add the plumbing functionality to base trays you may just want to remove most/all of the code here, put it in the basic hydrotray, and have a var or something control autoharvest (because that's the only thing these trays will have different, apart from being unable to add reagents manually)

lolman360:
also suggestion: hydroponic tray mode that freezes processing, stopping growth, reagent consumption and etc.  maybe on control+shiftclick or multitool attackby or something. alternative to autogrow that doesn't consume as much power but isn't nearly as automated

code/modules/plumbing/plumbers/autohydro.dm
TheObserver-sys: I think for now, I'll definitely remove the autohydro, and then look at how TG's got plumbing set up so I can tweak the tray. If I recall, I *think* TG has the hydrotrays now built to be automatically plumbing compatible as of recently, but I'm also hesitant to make a move on the whole plumbing system because on TG's side it's got a lot of kinks that need to be worked out.

TheObserver-sys: Maybe I can add an autoharvest function to the base tray too, though that will be dangerous for plants at high instability, because it's very well possible to have a plant roll two chems that react and then explode.

TheObserver-sys:
I'd rather not make a full stasis option, but I'm willing to draw back the power consumption more on autogrow. What would you say is an acceptable power target?

Issue #13299, opened 2020-08-30
Baby slimes starting off with low nutrition after an adult slime split


BlueWildrose:
Issue reported from  Round ID: 24249 (Citadel Station)
Reporting client version: 513.1530

Slimes start at about roughly 328/1200 nutrition, more or less when they split.

The reason why this is a problem is because freshly splitted slimes now require three monkies to grow into an adult and split instead of two now.

This not only makes xenobiology already more time consuming, but it means more monkey spam.

Edit:
I'm told this was an intentional choice. If so, why such a drastic change?

Edit 2: 
I'm also told it wasn't. Never mind.

BlueWildrose:
Gonna make an attempt at looking into this further and fixing this.

BlueWildrose:
Problem solved.

PR #13319, opened 2020-08-31
Fixed newly created slimes being hungry

BlueWildrose:
## About The Pull Request

There's some new code in /mob/Initialize() that randomizes the starting nutrition of a newly created mob between a minimum and maximum value of 250 and 400 respectively. 

This code also affects slime mobs, making them hungry by their standards when it comes to nutrition, and therefore, each newly created slime requires 2 monkies to turn into an adult slime, and an additional monkey to split, **resulting in 3 monkies required to make a slime reproduce, instead of the standard 2**.

This fix adds a singular line of code that gives slime their own starting nutrition in /mob/living/simple_animal/slime/Initialize. It'll be between 650 and 800. I believe slimes at some earlier point were meant to always start at exactly 700 nutrition upon spawning, either by splitting or by being spawned in by admins or something, but if the element of randomness is getting involved with hunger, might as well include them in slimes too.

## Why It's Good For The Game

A change happened as mentioned above that accidentally nerfed xenobiology DRAMATICALLY, this unnerfs xenobiology back to what was normally expected.
This will conclude issue #13299 which I posted a while ago.

## Changelog
:cl:
fix: fixed slimes starting off hungry
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/slime/slime.dm
+++ b/code/modules/mob/living/simple_animal/slime/slime.dm
@@ -106,6 +106,7 @@
 	set_colour(new_colour)
 	. = ..()
 	AddComponent(/datum/component/footstep, FOOTSTEP_MOB_SLIME, 7.5)
+	set_nutrition(rand(650, 800))
 
 /mob/living/simple_animal/slime/Destroy()
 	for (var/A in actions)


Ghommie (approved):
Ah... Thank you.

PR #13332, opened 2020-09-02
makes magweapons emp-immune

silicons:
## Why It's Good For The Game

magweapons were overly nerfed and they're now hilarious mockeries that no one uses 
time to make them have a semi niche purpose again even if they're still garbage in comparison to most other weapons.

## Changelog
:cl:
balance: magweapons/magrifles are no longer drained by EMPs.
/:cl:


KeRSedChaplain:
Can't we just have mag weapons be like, the designated AP weapon the station gets, making them EMP-proof doesn't really fix much

PR #13335, opened 2020-09-03
Adds shivering/sweating notifications (in devman terms: readds the old temperature notifications)

deathride58:
## About The Pull Request
This PR readds the old temperature notification alerts alongside the current ones. For reference, the old temperature notifications are based purely off of your character's body temperature, indicating that they're shivering or sweating, while the current ones are based off the room temperature relative to your character's body temperature, indicating how the environment feels for your character.
Sprites are very rudimentary, and are based off of emojis. If an actual artist would like to replace them, you are absolutely 100% free to do so.

## Why It's Good For The Game
While the current notifications on their own are really good for giving you instant feedback for what's going on in the environment, there's quite a few things that directly modify your character's body temperature, which leads to... quite a few awkward oversights, like getting cryostung but not getting the notification because you're wearing a hardsuit, or getting blasted by a watcher beam, resulting in the temperature notification saying you're hot, when in reality, you're cold.

## Changelog
:cl: Bhijn
add: Readded the old method of temperature notifications in the form of a new pair of shivering/sweating notifications.
/:cl:


CameronWoof:
I like the idea, but sweating being communicated with pure blue drops of water isn't especially clear, and having two different icons show up to explain temperature isn't, either.

I would very heavily suggest paring this down to one icon, even if it means giving potentially less information. All the player *really* needs to know is if their body temperature is lowered or elevated in comparison to their normal resting temperature. I don't think fairly unimportant information like what temperature the room is around you (whether or not that is affecting you negatively) is worth cluttering the UI with extraneous images.

lolman360:
^. The only case I could really think of where having separate icons would be useful is if you're in a hot environment while at a low bodytemp and vice versa. Since bodytemp changes with the local environment, this really only happens with bodytemp chems (leporazine, hell ramen, frost oil)

Putnam3145:
body temperature doesn't change instantly with the environment.

PR #13340, opened 2020-09-04
Disables the 'random brain trauma' event

DeltaFire15:
## About The Pull Request
Title. The event is disabled via setting its weight to zero.

## Why It's Good For The Game
Issue about this event is that it is basically just the random appendicitis event, but more annoying, taking away from rolls of more interesting events whilst increasing the odds of ones that are just a annoyance to 'yikes'.
For example, an antagonist being selected by it. This can roll lobotomy-only-cureable-full-body-paralysis, or the trauma that makes you sleep every now and then, and so on. Best case it is a minor annoyance that can be fixed in a second, worse case it fucks the round over. Even appendicitis can be mitigated via getting toxheal chems, whilst this usually needs either surgery, or even a lobotomy to deal with if it rolls one of the more concerning traumas. And good luck getting those fixed if you are an antag, or it rolls you in a situation where medical isn't as easily available.

## Changelog
:cl:
del: The 'random brain trauma' event has been disabled.
/:cl:

### DIFF
--- a/code/modules/events/brain_trauma.dm
+++ b/code/modules/events/brain_trauma.dm
@@ -1,7 +1,7 @@
 /datum/round_event_control/brain_trauma
 	name = "Spontaneous Brain Trauma"
 	typepath = /datum/round_event/brain_trauma
-	weight = 25
+	weight = 0
 
 /datum/round_event/brain_trauma
 	fakeable = FALSE


timothyteakettle:
when porting this from tg i literally hugboxed it to not have magic resistant traumas and to roll the worse resistances far less

also no this isn't an excuse for not nerfing it, i feel the real issue is it's happening too often because people nerfed weights of everything they didn't like

PR #13341, opened 2020-09-04
Nerfs random brain trauma a good deal.

Putnam3145:
…e's doctors.

## About The Pull Request

Changes the following for spontaneous brain trauma:
1. Will only roll if there is at least one MD, CMO or paramedic.
2. Will only roll if they're on the station Z-level.
3. Won't randomly cause lobotomy-tier traumas.
4. Significantly lower chance of severe traumas.

## Why It's Good For The Game

It really is an annoying event, but there's other ways to deal, and I'd rather the doctors have something to do.

## Changelog
:cl:
balance: Made spontaneous brain trauma a good deal less annoying.
/:cl:

### DIFF
--- a/code/modules/events/brain_trauma.dm
+++ b/code/modules/events/brain_trauma.dm
@@ -6,7 +6,7 @@
 /datum/round_event_control/brain_trauma/canSpawnEvent(var/players_amt, var/gamemode)
 	var/list/enemy_roles = list("Medical Doctor","Chief Medical Officer","Paramedic")
 	for (var/mob/M in GLOB.alive_mob_list)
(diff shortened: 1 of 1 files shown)


DeltaFire15:
There already is the 'spontanious appendicitis' and ' heart attack' events, which together have a weight of 40. This does also not handle the possibility of full-body paralysis or things of simillar severity.

Putnam3145:
Alright, reduced chance of severe traumas, too.

DeltaFire15:
I also suggest capping the max occurances as both appendicitis and heart attack have (2 for heart attack, 4 for appendicitis, probably 2-3 would be good here) to prevent its high weight from blocking more interesting events from rolling, as it's a incredibly common event.

PR #13343, opened 2020-09-04
Increases the idle animation startup time for drakeborgs from approx. 1 second to approx. 30 seconds

deathride58:
This PR does exactly as it says on the tin. The idle animation that drakeborgs have is commonly cited as appearing way too jarring, considering idle animations are very rare for SS13 mobs (and the few idle animations that do exist are fairly static or otherwise simplistic, like how slimes jiggle slightly to indicate that they're alive). This PR resolves that complaint by increasing the startup time for that idle animation, from it's current 1 second timer, to a much less frequent 30 seconds.

The walk animation is another commonly-cited reason for drakeborgs appearing too jarring against the rest of the game, but this is out of the scope for this PR. The reason why it's jarring is due to the walk animation being fairly slow when taking into consideration the existence of sprinting and vtec upgrades, which results in the drakeborg just kinda gliding across the floor, like it's got rollerskates in it's feet or something. The three ways to potentially resolve this, in order from best, to worst, are as follows:

- 1: Commission the artist for a gallop animation for drakeborgs, and replace the current walk animation with the freshly-commissioned gallop animation. It's unknown how much was paid for drakeborgs to begin with, so the predicted cost of this is unknown. Additionally, the artist's commissions are currently closed as of writing, so it'd likely take a good few months before we get the opportunity to do this. But this option would directly resolve the jarring appearance of the drakeborg walk animation by making their movement animation more or less match the animations of dogborgs, which are much better suited to the context of SS13 (or, well, BYOND in general, really)
-  1.A: Have someone else sprite a short gallop animation for drakeborgs. If anyone's willing to volunteer for this, you are absolutely free to do so. Personally, I'd prefer the route of commissioning mizartz directly, since they honestly did a really good job with the animations, even if the context of the current animations don't do 'em the proper justice.
- 2: Speed up the drakeborg walk animation by cutting the frame time in half. This would be very easy to do, but would still run into the issue of appearing too jarring, but this time in the context of slower movement, since then it'd look like the drakeborg's just kinda shuffling in place to scoot over a tile or two.
- 3: Remove the drakeborg walk animation. This is the absolute laziest method of the three currently available. This should ideally only be done as an absolute last resort.

## Changelog
:cl: Bhijn
tweak: The idle animation for drakeborgs now takes approximately 30 seconds to start up, as opposed to 1 second.
/:cl:


ghost:
what is it with bhijn and noticing an issue and then instead of fixing the issue just making the issue slightly less of an issue but still an issue

i don't understand the thought process since bhijn currently has the most power out of anyone on the server and, as such, doesn't have to lick feet to try and do the tiniest changes

The-Sun-In-Splendour:
Delete the idle animation

PR #13344, opened 2020-09-04
Makes heat conduction work for the first time ever

Putnam3145:
## About The Pull Request

Yeah, turns out it never worked. Stop telling me monstermos made it stop working, it doesn't even work on TG.

## Why It's Good For The Game

It might actually not be. The game's balance does seem to be built around this system simply not working at all; note that I had to reduce the floor tile thermal conductivity to 1/10 of what it was, otherwise plasma fires get fixed on their own in minutes. We could try it, though.

Requires a change to the extools dll cause monstermos doesn't have thermal conductivity, whoopsie doopsie. This was never noticed because superconductivity does not work.

## Changelog
:cl:
fix: Made superconductivity work for the first time literally ever.
/:cl:


Arturlang:
fuck yes, ive been missing this

Putnam3145:
not sure how, since it never worked

Arturlang:
....from other servers

NalyKalZul:
Does this also work for negative temps?
Because i can see a new way to remove plasma fires then, Il just rewrite the tiles on fire in question.
Though said solution migth be worse in other aspects, ill have some fun with it striking a balance.

Putnam3145:
@Arturlang i'm not sure it ever worked on other servers

@NalyKalZul no, it actually stops working at precisely 30 celsius

PR #12957, opened 2020-07-29
removes malf disk from tendril loot

ghost:
## About The Pull Request
 🤡 

## Why It's Good For The Game
malf ai and illegal tech are both items that are worthless on lavaland and overpowered on station. 
however malf ai is way more shitty overpowered literal antag ability gear and illegal tech is kinda aight i guess since snaxi has abductor tech

FUCK those types of items my dude. it's like how we have the anti-magic items that are worthless on lavaland but absolutely CUCK the shit out of cult and wizard rounds if they're unlucky enough that a miner gets the funny anti-magic.

## Changelog
:cl:
balance: miners can no longer acquire funny antag item
/:cl:

PR #13365, opened 2020-09-07
Hey wait isn't there numbers between 11 and 15

DeltaFire15:
## About The Pull Request
Fixes randomdrop necropolis chests sometimes being empty, a few options having no loc argument when being created new, one option not being possible to roll because of having two ifs in a switch with the same condition, aswell as some other minor stuff.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Fixed wacky necropolis loot chest behavior
/:cl:


Putnam3145:
i'm pretty sure this is actually intentional behavior, just coded funny, but then again they could just be confused about case fallthrough

moltoretardo:
#12957 
Malf ai disks were removed because stealth malf traitor ai is cancer

PR #13373, opened 2020-09-09
Fixed toggleable polychromic hoods/helmets not polychroming with their suit

BlueWildrose:
## About The Pull Request
**Observations when it comes to the polychromic code:**
- There was code that took toggleable hoods into account.
- However, the proc to register the helmet was never called when the polychromic element recognized that the item in question had a toggleable helmet, meaning that the hood on the polychromic winter coat was never registered to be part of the polychromables.

This is my "I'm new to BYONDcode" way of fixing this. It seems to work. If there's even better solutions to this problem, or people see any flaws with this code, people are welcome to provide input there.

I assume @Ghommie might want to provide input on this, since this is their code that I'm editing.

## Why It's Good For The Game
Fixes are good. A hoodie not polychroming with their bottom half looks ugly. This essentially fixes the polychromic winter coat.

## Changelog
:cl:
fix: The Polychromic winter coat's hoodie will now polychrome, alongside any other new polychromic items with toggleable headwear.
/:cl:

### DIFF
--- a/code/datums/elements/polychromic.dm
+++ b/code/datums/elements/polychromic.dm
@@ -62,6 +62,7 @@
 			RegisterSignal(A, COMSIG_ITEM_WORN_OVERLAYS, .proc/apply_worn_overlays)
 			if(suits_with_helmet_typecache[A.type])
 				RegisterSignal(A, COMSIG_SUIT_MADE_HELMET, .proc/register_helmet)
+				register_helmet(A) //just call it directly it only needs to be called once, right?
 	else if(_flags & POLYCHROMIC_ACTION && ismob(A)) //in the event mob update icon procs are ever standarized.
 		var/datum/action/polychromic/P = new(A)
 		RegisterSignal(P, COMSIG_ACTION_TRIGGER, .proc/activate_action)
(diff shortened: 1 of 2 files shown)


BlueWildrose:
I think I know of a better way to implement this code without it looking so band-aidy, I'll make another commit to this later today

BlueWildrose:
Special thanks to @LetterN for helping me through BYONDcode better. This actually makes the signal work, rather than just bypassing the signal. Definitely will make any new code that interacts with winter coats function better too. As well as hardsuits.

BlueWildrose:
Well then.

The polychromic hoodie does work externally, but there seems to be some conflicts happening inside the interface right now, where it seems to be rendering a second hoodie icon on the toggle button in the interface. There's also cases where you can see a phantom hoodie, until it's clicked.

Look at this webm for an example.

https://lambda.sx/WGaf.webm

I have no idea where to begin with this, but if this is okay, then go ahead and merge it. Otherwise, I'm going to see about figuring out where to begin with this whenever I find the time to do so. It's not anything gamebreaking, it's just glitchy looking.

PR #13375, opened 2020-09-10
the emp rework (emp severity reduces as you move from epicentre, severity is 1-100)

timothyteakettle:
## About The Pull Request
only testmerge this
it has been tested but something is bound to go wrong

please tell me if any of the formulas need adjusting
some of them are done such that 100 severity is equal to an old heavy emp
some of them are done such that 100 severity is a bit stronger than an old heavy emp
etc

the full list of changes to emp severity formulas is found here, **it's very long** https://pastebin.com/iw50uQnE

the maximum distance is calculated as (power/7)^0.7 this is subject to change obviously
right now this means a 300/300 uranium iron reaction would have a max distance of 23 tiles
a 50/50 reaction would have a max distance of 6 tiles
a 10/10 reaction would have a max distance of 2 tiles

the severity is calculated as (max_distance / (distance+1)^0.3) * (100/max_distance)

## Why It's Good For The Game
emps suck
this makes emps not suck

## Changelog
:cl:
refactor: emps now work from 1-100 severity instead of 1/2 and the severity reduces as you move from the epicentre
/:cl:


timothyteakettle:
updated severity formula so it curves way better now
might make the severity be linear when it's 5 or less tiles range

timothyteakettle:
pr is fine i'm just sick of having to do empty commits to make travis realise it

timothyteakettle:
still fine travis just segfaulted

timothyteakettle:
if this isnt merged by the time i finish my wound fixes i will probably work on wave based emps in this pr too because it seems fun

PR #13380, opened 2020-09-10
(TESTMERGE THIS FIRST WOW) Made monstermos faster and more convenient

Putnam3145:
## About The Pull Request

The code's a bit bare on the byond side, sorry. What I did was change Monstermos to use a big-ass vector instead of doing a `new` allocation on the heap every single time a new gas mixture pops up, which also keeps all the gas mixtures in one contiguous location in memory. I also updated gas mixtures to use a vector instead of an array (turns out the performance is the same AT WORST) and updated most of the functions to use the standard library instead of bespoke for loops using the define. Also, I compiled the DLL with some fancy compiler flags that allow for vectorization of loops, so merging and multiplication is faster. On top of that, I moved the active turf processing and equalization logic *entirely* to C++. I'm not actually sure of that made anything faster.

## Why It's Good For The Game

1. Atmos lag is solved. Yes, solved. Local tests show that performance barely gets worse with higher amounts of active turfs; in fact, by "barely worse", I mean that the byond proc call overhead is now the only performance problem in turf sharing. There is no significant performance difference between 5000 turfs and 50 turfs being operated on; the performance at 5000 turfs is ~100 turfs/ms, vs the previous ~10.

2. C++ no longer needs to be dealt with in order to add new gases. Monstermos now gets the information on how many gases there are at runtime, then initializes all the "moles" vectors with that size. As previously mentioned, this turns out to be just as fast as arrays at worst, and is much more convenient.

## Changelog
:cl:
refactor: Atmos is free.
/:cl:


lolman360:
epic. extools lighting when?

PR #13382, opened 2020-09-10
hulks no longer break their hand from punching a wall 

silicons:
## Why It's Good For The Game

hulks already get shafted by stamina combat, no need to double down and make it truly worthless

## Changelog
:cl:
balance: hulks can smash again (walls no longer break their hands)
/:cl:


necromanceranne:
This change got reverted on tg as well because it's dumb.

PR #13389, opened 2020-09-13
Donator Item - Halo

LetterJay:


modular_citadel/code/modules/client/loadout/__donator.dm
```diff
@@ -510,3 +510,9 @@
 	slot = SLOT_IN_BACKPACK
 	path = /obj/item/bedsheet/cosmos
 	ckeywhitelist = list("grunnyyy")
+
+/datum/gear/donator/hisakaki
+	name = "cosmic space bedsheet"
```
timothyteakettle: wrong name which is important due to how loadout gears are saved

SandPoot:
`modular_citadel/code/modules/custom_loadout/custom_items.dm:559:error: 'icons/obj/custom_w.dmi': cannot find file`

ArchieBeepBoop:
Yep. @LetterJay Please change that to 'icons/mob/clothing/custom_w.dmi' instead.

PR #13393, opened 2020-09-15
[Ready] Donator Item - Custom Skirt 

LetterJay:
Still pending byond name, so don't merge.


modular_citadel/code/modules/client/loadout/__donator.dm
```diff
@@ -510,3 +510,9 @@
 	slot = SLOT_IN_BACKPACK
 	path = /obj/item/bedsheet/cosmos
 	ckeywhitelist = list("grunnyyy")
+
+/datum/gear/donator/customskirt
+	name = "custom atmos skirt"
+	slot = SLOT_IN_BACKPACK
```
timothyteakettle, line 516: pretty sure this should be SLOT_W_UNIFORM because if it can't go on the person it automatically gets placed in the backpack slot

modular_citadel/code/modules/custom_loadout/custom_items.dm
```diff
@@ -550,3 +550,12 @@
 	desc = "A fancy cloak embroidered with polychromatic thread in a pattern that reminds one of the wielders of unlimited power."
 	icon_state = "polyce"
 	poly_colors = list("#808080", "#8CC6FF", "#FF3535")
+
+/obj/item/clothing/suit/customskirt
+	name = "unzipped atmos skirt"
+	desc = " atmos jumpsuit skirt with the top half zipped down and wrapped around the waist."
```
timothyteakettle: unnecessary space at the start of the description

modular_citadel/code/modules/client/loadout/__donator.dm
LetterJay: Keeping it in backpack.

thakyZ:
Is there any possibility that the skirt can be made to go into the uniform slot (`i_clothing`) and behave like a uniform would normally? Because the concept of this was I would wear it as somewhat normal attire rather than having it be placed in the `o_clothing` slot...

thakyZ:
So basically I would like it to be able to be placed in this slot:
image
From my testing it can only be placed in this slot:
image

LetterJay:
Changed to be under clothing

silicons:
@LetterJay doesn't compile apparently

thakyZ:
`__donator.dm` needs to have the path changed as well

thakyZ:
Damn Segmentation fault...

thakyZ:
The Travis CI build seemed to fail, but compiled successfully... It seems to be a segmentation fault that was the cause, after shutting down, so I think it needs to be run again...

Update:
Ah I see... now there are those to sprite atlases...

PR #13397, opened 2020-09-15
Micro Powered Fans!

ArchieBeepBoop:
## About The Pull Request

Adds the ability to craft a version of the tiny fans found in shuttles and other areas of the station. The code contains its own sprites and it's an addon, not a mod of existing code.

## Why It's Good For The Game

This is more of a balance thing than anything else. Holofans are instant but not readily available. Micro fans take some time to build, require 5 plasteel sheets to create an assembly for, must be welded then cabled to properly function. Plus, containment is broken without power.

Screenies:

image
image

**MMMM, toasty tritium fire.**

image

**Unpowered, fire breaks free and destroys them.**

image

image

## Changelog
:cl:
add: Craftable Micro Powered Fans
/:cl:


The-Sun-In-Splendour:
can't you just make the fans found in the shuttles craftable instead
or at least make them a child of the fans found in the shuttles....

ArchieBeepBoop:
> can't you just make the fans found in the shuttles craftable instead
> or at least make them a child of the fans found in the shuttles....

I have, however, doing that has been found to be quite... Unbalanced according to people. Atmos also did not like how it technically made the holofans almost useless. Which is why I decided to make this a version that is different from the structure code based minifans and balanced it by making it consume power and also fail during a power outage.

Arturlang:
What if they are ATMOS SHIELDS instead, so it makes even a tiny bit of sense, god please, and make them fairly expensive so theyre a more annoying alternative to the tempoary ones and walls and doors, but stronger than normal holofans

ArchieBeepBoop:
> What if they are ATMOS SHIELDS instead, so it makes even a tiny bit of sense, god please, and make them fairly expensive so theyre a more annoying alternative to the tempoary ones and walls and doors, but stronger than normal holofans

Making them more expensive is actually quite easy. Just gotta change the amount of metal sheets required to build them and/or the amount of power consumed.

The-Sun-In-Splendour:
> 
> 
> > can't you just make the fans found in the shuttles craftable instead
> > or at least make them a child of the fans found in the shuttles....
> 
> I have, however, doing that has been found to be quite... Unbalanced according to people. Atmos also did not like how it technically made the holofans almost useless. Which is why I decided to make this a version that is different from the structure code based minifans and balanced it by making it consume power and also fail during a power outage.

how is it unbalanced to just make them functionally similar to the shuttle atmos fans except they can be wrenched and use power.

ArchieBeepBoop:
Also, 

> > > can't you just make the fans found in the shuttles craftable instead
> > > or at least make them a child of the fans found in the shuttles....
> > 
> > 
> > I have, however, doing that has been found to be quite... Unbalanced according to people. Atmos also did not like how it technically made the holofans almost useless. Which is why I decided to make this a version that is different from the structure code based minifans and balanced it by making it consume power and also fail during a power outage.
> 
> how is it unbalanced to just make them functionally similar to the shuttle atmos fans except they can be wrenched and use power.

I thought you meant just making the regular tiny fans a craftable in the metal sheet menu. I was told that was unbalanced through the research I've done due to the fact that the regular shuttle tiny fans were basically better than atmos fans as they did not use power and proceeded to turn it from a structure to a type of machinery. Structures cannot draw power, therefore making it a child of the original fan was already off the table from the start.

Fun fact, the tiny fans are code derived from the lavalands survival pods and were apparently originally meant to only be used there. The code was never split from them and properly turned into its own file when mappers started adding them everywhere due to how convenient they were.

Look it up: https://github.com/Citadel-Station-13/Citadel-Station-13/blob/be20861294f132d522d8002c989494223e502e4d/code/modules/mining/equipment/survival_pod.dm

ghost:
i feel like being able to make fans is a lesser evil than being able to do 160 damage per click with a 5 tile range by breaking one window somewhere on station

edit: make it a machine with cheap stockpart reqs instead of an inhand craft and it's Probably Fine

MrJWhit:
I'm not a super big fan of the idea initially, I'd rather see how it's handled elsewhere first to be honest.

ArchieBeepBoop:
I'm more than happy to turn this into a draft or close it for the time being if the maintainers wish so, at least until I've acquired more feedback from the actual use of these and how they are being used. Maybe even a neat little form.

MrJWhit:
I'm going to be honest, tinyfans are a bit of a sore point for me, because it's used as extremely lazy map design. Holofans require some amount of effort to get because it takes some weird materials, there aren't a lot around, they're limited to their usage.

This is just takes 5 metal sheets (unless I'm reading the code wrong), and, honestly I see that as way too easy to get access to, there should be some higher level tech associated with a fan that stops a plasmafire.

lolman360:
yeah 5 metal sheets is not at all expensive especially with oredoubling and power isn't really much of a consideration considering basic sm+solars already covers station base load with plenty to spare

if they were powered by an internal cell and had to be stuck in a recharger maybe?

MrJWhit:
image

ArchieBeepBoop:
I'll change it to plasteel tomorrow then and increase power usage alongside crafting time. How about 5 or 10 plasteel sheets?

ghost:
i like the idea but i'm pretty sure the best idea is to just make it a very very cheap machine (think ai pads)
in-hands with super long crafting timers are boring and tedious
if the fans are machine-type then they can use cheap materials but still be hard to mass produce because of storage and the fact you need an autolathe
i'd also recommend either making the board an autolathe produce-able or in early engitech

ArchieBeepBoop:
I'm currently reworking them into being a proper assembly.

ArchieBeepBoop:
New commit adds the assembly, welded sprite, plasteel cost and all the tooltips to properly assemble a micro powered fan.

code/game/machinery/poweredfans/fan_assembly.dm
```diff
@@ -0,0 +1,111 @@
...
+			2 = Welded in place
+			3 = Wires attached to it, this makes it change to the full thing.
+	*/
+
+/obj/machinery/fan_assembly/Initialize(mapload, ndir, building)
+	. = ..()
+	if(building)
+		setDir(ndir)
+
+/obj/machinery/fan_assembly/Destroy()
```
silicons: please get rid of this override

code/game/machinery/poweredfans/fan_assembly.dm
```diff
@@ -0,0 +1,111 @@
...
+	use_power = NO_POWER_USE
+	power_channel = ENVIRON
+	idle_power_usage = 0
+	active_power_usage = 0
+	max_integrity = 150
+	layer = ABOVE_NORMAL_TURF_LAYER
+	anchored = FALSE
+	density = FALSE
+	CanAtmosPass = ATMOS_PASS_YES
+	var/state = 1
```
silicons: please use defines instead of magic variables

code/game/machinery/poweredfans/fan_assembly.dm
```diff
@@ -0,0 +1,111 @@
...
+		return TRUE
+	return FALSE
+
+/obj/machinery/fan_assembly/deconstruct(disassembled = TRUE)
+	if(!(flags_1 & NODECONSTRUCT_1))
+		new buildstacktype(loc,buildstackamount)
+	qdel(src)
+
+/obj/machinery/fan_assembly/examine(mob/user)
+	..()
```
silicons: . = ..()
also don't do deconstruction_hints just . += since examine is a list return now of text

code/game/machinery/poweredfans/fan_assembly.dm
```diff
@@ -0,0 +1,111 @@
...
+/obj/machinery/fan_assembly/proc/deconstruction_hints(mob/user)
+	switch(state)
+		if(1)
+			to_chat(user, "<span class='notice'>The fan assembly seems to be <b>unwelded</b> and loose.</span>")
+		if(2)
+			to_chat(user, "<span class='notice'>The fan assembly seems to be welded, but missing <b>wires</b>.</span>")
+		if(3)
+			to_chat(user, "<span class='notice'>The outer plating is <b>wired</b> firmly in place.</span>")
+
+/obj/machinery/fan_assembly/update_icon()
```
silicons: update_icon_state()

code/game/machinery/poweredfans/poweredfans.dm
```diff
@@ -0,0 +1,63 @@
...
+	use_power = ACTIVE_POWER_USE
+	power_channel = ENVIRON
+	idle_power_usage = 5
+	active_power_usage = 10
+	max_integrity = 150
+	layer = ABOVE_NORMAL_TURF_LAYER
+	anchored = TRUE
+	density = FALSE
+	CanAtmosPass = ATMOS_PASS_NO
+	var/obj/machinery/fan_assembly/assembly = null
```
silicons: no = null needed

code/game/machinery/poweredfans/poweredfans.dm
```diff
@@ -0,0 +1,63 @@
...
+			if(!assembly)
+				assembly = new()
+			assembly.forceMove(drop_location())
+			assembly.state = 2
+			assembly.setAnchored(TRUE)
+			assembly.setDir(dir)
+			assembly = null
+			new /obj/item/stack/cable_coil(loc, 2)
+		else
+			new /obj/item/stack/cable_coil(loc, 2)
```
silicons: move this up a level and get rid of the previous cable coil spawn + else

code/game/machinery/poweredfans/poweredfans.dm
```diff
@@ -0,0 +1,63 @@
...
+	qdel(src)
```
silicons, line 27: is there no ..() logic that's needed?

code/game/machinery/poweredfans/poweredfans.dm
```diff
@@ -0,0 +1,63 @@
...
+	else
+		assembly = new(src)
+		assembly.state = 3
+	air_update_turf(1)
+
+/obj/machinery/poweredfans/power_change()
+	..()
+	update_icon()
+
+/obj/machinery/poweredfans/update_icon()
```
silicons: first icon state should be update_icon_state() instead of update_icon which shouldn't be overridden
second please don't update air status/air update turf in update_icon or update_icon_state or update_overlays, they're purely for updating visuals, it's bad practice

silicons (dismissed):
i agree with this pr but please fix the things listed

ArchieBeepBoop:
> i agree with this pr but please fix the things listed

Working on it! Sorry for all the shitcode @silicons . I was using the cameras as a base code for this as it was my very first actual contraption and not just general code so I didn't know where/what to follow as the example of a good machine code.

PR #13398, opened 2020-09-15
wound config values, wound bonus for how damaged a limb is

timothyteakettle:
## About The Pull Request
extra bonus added to injury rolls depending on how damaged a limb is (this is one of the configs) by default it's 0.333 as in 0.333x the bodyparts current damage is added to the roll
wound exponent lowered from 1.225 to 1.2 (this is the other config)

if you plan on changing the configs read what they do first, they're commented and are pretty easy to read where they're used too

also yeah did you know how damaged a limb is has no bearing on wounds prior to this
woops

## Why It's Good For The Game
wounds are not being applied enough nor consistently enough
this makes them even more consistent, and also more likely!

## Changelog
:cl:
tweak: wound exponent lowered slightly from 1.225 to 1.2
config: wound exponent and limb damage multiplier are now config values
/:cl:


timothyteakettle:
need to make one final change when i am available so it's in a merge-able state

Ryll-Ryll:
Actually there is a bonus to wound rolls if a limb is at its max damage, shown here https://github.com/timothyteakettle/Citadel-Station-13/blob/3db878484d5a9bc4037773237c39bbda90bea3ee/code/modules/surgery/bodyparts/_bodyparts.dm#L448

This sounds like a much better system since I kinda lazy'd out on that originally, you just may want to remove disabled_wound_penalty

PR #13400, opened 2020-09-16
adds two procs to find exterior tiles for projecting hull shields around for later pr

silicons:
@lolman360 

expensive as hell but shields shouldn't be cycling that fast anyways


lolman360:
epic i can probably code shields basically from scratch now

Putnam3145:
if it's actually a problem i can move this (among other tile stuff, probs) to C++ but it probably won't be an actual problem.

lolman360:
@Putnam3145 i'm going to add a cd on shield restarts so at worst it's going to be called every 30sec or so

PR #13402, opened 2020-09-16
copying prefs always regenerates icons now (attempt to stop preview icons being outright wrong)

timothyteakettle:
## About The Pull Request
title
yes it's an expensive proc but it's necessary when you're making constant changes to a character
if this becomes an issue somehow i'll look into a more efficient way of handling character previews by making dummies into simplemobs that can still wear clothes

## Why It's Good For The Game
attempts to make preview icons slightly more consistent

## Changelog
:cl:
tweak: character previews should be more consistent now
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -2634,8 +2634,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 				new_limb.replace_limb(character)
 			qdel(old_part)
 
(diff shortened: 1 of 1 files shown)


Azarak:
Consider update_body_parts() instead

timothyteakettle:
regenerate icons will update this amongst other things, and without extensive changes to dummies this is the simplest approach to solve the awful issues that come with dummies being reused

PR #13413, opened 2020-09-20
Properly standardizes fuid production rate.

Putnam3145:
## About The Pull Request

Fuid (sic) generation is far too fast right now cause I did some funny math way wrong and misunderstood what the constants were supposed to be. Now it bases it on a standard "every 10 seconds".

I also reduced milk generation rates to match that of goats/cows; before, it was almost twice as much to milk yourself as to milk goats/cows, which is off.

## Why It's Good For The Game

I fixed spatters being made from too small amounts a while ago, but now the amounts are always large enough. Not good.

## Changelog
:cl:
balance: G fuid production is now much lower.
/:cl:

### DIFF
--- a/code/__DEFINES/citadel_defines.dm
+++ b/code/__DEFINES/citadel_defines.dm
@@ -47,7 +47,7 @@
 #define BALLS_SIZE_DEF		2
 #define BALLS_SIZE_MAX		3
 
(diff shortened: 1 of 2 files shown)


lolman360:
your balls finally feel full again

PR #13418, opened 2020-09-22
bumps typing indicator timeout because 5 minutes is too little to write emotes

silicons:
don't ask


Putnam3145:
next increase the emote limit beyond 2000 characters. you know you want to

PR #13420, opened 2020-09-22
Light replacer can print at Engineering

thakyZ:
## About The Pull Request

I made it so that the light replacer, once researched can be printed at the Engineering Department's Protolathe. Engineering shares the duty with service of replacing lights, so it seems fitting.

## Why It's Good For The Game

As I stated before Engineering shares the duty of replacing lights with service. I tend to replace lights often as engineering as no one else tends to do it when they see a broken light. Plus a lot of lights are behind engineering doors that service doesn't always have access too.

## Changelog
:cl:
add: Added the ability to print the Light Replacer at the Engineering Protolathe
/:cl:


timothyteakettle:
based

thakyZ:
What's with the segmentation faults

timothyteakettle:
they happen sometimes but it's nothing to do with this pr

PR #13422, opened 2020-09-22
Fixes angels not being fast while flying

silicons:


code/modules/mob/living/carbon/human/species_types/angel.dm
```diff
@@ -1,4 +1,4 @@
-/datum/species/angel
+t/datum/species/angel
```
SandPoot: t

silicons: thanks

PR #13424, opened 2020-09-23
Makes cryo pods no longer replace already-won objectives.

Putnam3145:
## About The Pull Request

As title. It's kinda bullshit that you can finish an objective then WHOOPS the target you've *already dealt with* cryos and now you've lost.

## Why It's Good For The Game

This is kinda a bug anyway.

## Changelog
:cl:
tweak: Cryoing no longer unwins already-won objectives.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -355,23 +355,23 @@
 		cryo_items -= I
 
 	//Update any existing objectives involving this mob.
-	for(var/datum/objective/O in GLOB.objectives)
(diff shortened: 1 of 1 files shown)


DeltaFire15:
Looking at how is_considered_alive works, and how people generally ghost / close the game after entering cryo, this might lead to not yet completed objectives also staying if their target cryos.

Putnam3145:
hmm, you're right, lemme see how I could do this without kludging otherwise

PR #13432, opened 2020-09-25
Tweaks a thing

SiliconMain:
Plastic was something thoughtlessly copied over from skyrat, no one ever puts it in silo unless theres chemists and anyone can just decon plastic flaps anyways, QoL change.

:cl:
tweak: Minor adjustment to material cost of long range atmos analyzer
/:cl:


MrJWhit:
This is a weird tweak. 
Plastic is both difficult to get to and not difficult at all?
Why diamond?
Why so little?
Does the diamond have anything to do with the scanner, or is it just some high level material?

PR #13434, opened 2020-09-25
Blacklists A LOT of stuff from being contaminated

SiliconMain:
Done in an effort to reduce overhead.
There are so many hardmapped bullshit machines that cant be deconstructed and rebuilt to remove the contamination, and there really isnt a good way to go about doing it, especially when some of them are full tile and cant be cleaned by zeolites. 
Please discuss a way to go about this, at a minimum I want to make everything hardmapped in contamination proof, but I really dont want to nitpick through the mess that is every machine to see which ones can and cant be built.
:cl:
tweak: Lots of things can no longer be contaminated by radiation
/:cl:

### DIFF
--- a/code/game/machinery/_machinery.dm
+++ b/code/game/machinery/_machinery.dm
@@ -95,6 +95,7 @@ Class Procs:
 	flags_1 = DEFAULT_RICOCHET_1
 	flags_ricochet = RICOCHET_HARD
 	ricochet_chance_mod = 0.3
+	rad_flags = RAD_NO_CONTAMINATE
 
 	anchored = TRUE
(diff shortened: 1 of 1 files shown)


MrJWhit:
This is a **really** shit solution and isn't /that/ bad of a source of lag. 

If you want something better, make it so spraying water from a fire extinguisher on a machine can get rid the contamination by washing it off.

EDIT: You could also just make zeolites sprayers more common if it's that bad of a problem, we already have a solution that exists, players just don't use it it seems like.

Putnam3145:
> isn't /that/ bad of a source of lag.

it is the slowest thing on the server whenever it happens? it isn't *laggy*, though, yeah

timothyteakettle:
hi im going to let liquids and sprays go over certain machines please send me names of machines that this should apply to thanks

CameronWoof:
all machines

PR #13443, opened 2020-09-26
Gabbys new wooden floor tile types

ItzLollipop:
## About The Pull Request

Adds three new types of wooden plank floor tiles made by me! This has kind of been delayed I know.

It's almost been a year.

**_:agony:_**

unknown

## Why It's Good For The Game

Adds more turfs to make the game more visually appealing for newer and better environments.

## Changelog

:cl:
add: Three new turf tiles.
imageadd: Turf icons with multiple damage icons, with in-hand icons for each tile.
spellcheck: Edited the name and description to wooden.
/:cl:


ItzLollipop:
Think I should mention this but there is one thing that should be brought to attention, since the three base of the first type of wooden tile, they can be merged into the original type. This problem also occurs with carpet too, and my solution would just to make the original wooden tile to-| /turf/open/floor/wood/wood_small | as a subtype of the main bit which defines all of it, to solve this.

Detective-Google:
image
closer preview than the icon bot for those interested.

Detective-Google:
image
my only real feedback is that the borders on the diagonal (and to a lesser extent the other designs) feels a bit light, on the right here is a slightly darker version that I think might look better for reference.

ItzLollipop:
> 
> 
> image
> my only real feedback is that the borders on the diagonal (and to a lesser extent the other designs) feels a bit light, on the right here is a slightly darker version that I think might look better for reference.

Yeah I totally understand that, I was kind of thinking of doing smoothing for the diagonal one to add a border to make it look a little more fluid.

Oh I did not realize that is dark, weird.

Detective-Google:
they're the same normally  I just edited one to make a point



> > image
> > my only real feedback is that the borders on the diagonal (and to a lesser extent the other designs) feels a bit light, on the right here is a slightly darker version that I think might look better for reference.
> 
> Yeah I totally understand that, I was kind of thinking of doing smoothing for the diagonal one to add a border to make it look a little more fluid.
> 
> Oh I did not realize that is dark, weird.

PR #13453, opened 2020-09-27
microoptimization to get rad contents

silicons:
jesus christ atmos is less laggy than radiation


Putnam3145:
hate to tell you this but this is actually slower 

EDIT: why did i elaborate in-game and not here. Use the "get last member then len--" pattern to optimize this, checking length() with every loop like this is gonna be slower than the previous thing anyway, even accounting for the funny -=, I think

PR #13456, opened 2020-09-27
Removes pyroclastic slimes, but makes the anomaly fire way bigger

Putnam3145:
## About The Pull Request

As title. Makes no slime spawn from the pyroclastic anomaly explosion, but makes its plasma fire FAR hotter, by multiplying the oxygen content by 1.5 and the plasma content by 3. This essentially makes it equivalent to a highly above-average plasmaflood in terms of effectiveness.

Mutually exclusive with #13455.

## Why It's Good For The Game

This is approximately the same danger level as a flux anomaly, I'd say--which is to say, a few minutes with the tools given to engineering, and not much more than that.

## Changelog
:cl:
del: Sentient pyroclastic anomaly slimes removed
balance: Pyroclastic anomaly explosion buffed massively
/:cl:

### DIFF
--- a/code/game/objects/effects/anomalies.dm
+++ b/code/game/objects/effects/anomalies.dm
@@ -285,13 +285,7 @@
 /obj/effect/anomaly/pyro/proc/makepyroslime()
 	var/turf/open/T = get_turf(src)
 	if(istype(T))
-		T.atmos_spawn_air("o2=500;plasma=500;TEMP=1000") //Make it hot and burny for the new slime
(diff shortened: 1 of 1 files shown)


ForrestWick:
I'm against removing ghost role stuff. There's better ways to handle this than punishing everyone over the actions of so far three people afaik

keronshb:
Making the fire hotter just makes it burn through the hull faster, putting it out faster no?

PR #13469, opened 2020-09-29
Revert "C++ innovations: adds an "unstuck" step to SSair, AKA brute forcing LINDA"

Zandario:
Reverts Citadel-Station-13/Citadel-Station-13#13436


Putnam3145:
You can just remove the byond bit--probably easier if you do, as long as it's not explicitly called it won't ever happen.

Also, uh, what problem is this causing? I have a better implementation for Yogs, I didn't expect this to get merged, so I don't mind.

PR #13478, opened 2020-10-01
Further monstermos backend changes aka "wow I can just brute force EVERYTHING"

Putnam3145:
## About The Pull Request

1. Brought back air_turfs, primarily because having all the C++ code in one subsystem and all the byond code in another will make things far smoother on average. The default wait on air_turfs is 2 instead of 5, so fastmos is back I guess.
2. Added turf rescanning: my last attempt at this was highly experimental, but merged anyway, uh oh. This one's more thoroughly tested and vetted. Essentially, this step checks every turf in the world and adds them to active if it was already in active OR has differences from its neighbors. This happens in ~4 ms on average on my system, but then, I have 8 cores and a newer i7 than the server, so the performance impact might be larger than I think. Thankfully, I also included an in-game toggle for this, in the form of a var in the air_turfs subsystem.
3. Did away completely with the concept of a "current run" for active turfs. Every active turf is run through every time the function is called. This has not led to a performance impact, from what I can tell.

## Why It's Good For The Game

Smooth atmos is nice. We can change some of these things later (disabling rescanning, setting air_turfs to 5 are the most likely changes).

## Changelog
:cl:
add: LINDA now regularly scans every turf in the world for turfs that should be added to active. Yeah, all of them.
tweak: SSair_turfs is back, giving SSair a bit more leeway to work with.
tweak: Active turf and equalization processing now process every active turf, every time. Turns out this is *reasonably* fast.
/:cl:


Putnam3145:
it's good enough.

PR #13490, opened 2020-10-01
kind of easy to tell if you have a fake virus if it's ticking at 20 times the speed of a normal virus right

silicons:


timothyteakettle:
woops

PR #13491, opened 2020-10-01
missing health slowdown now scales to maxhealth

silicons:
## Why It's Good For The Game

makes it so more maxhealth also means you're more resistant to health based slowdown rather than, say, 100 damage at 200 maxhealth being the equivalent of 100 damage at 100 health, it'd now be equivalent to 50 damage at 100 health

## Changelog
:cl:
experimental: damage slowdown is scaled to percentage-based rather than absolute numbers
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1046,7 +1046,8 @@
 		return
 	var/stambufferinfluence = (bufferedstam*(100/stambuffer))*0.2 //CIT CHANGE - makes stamina buffer influence movedelay
 	if(!HAS_TRAIT(src, TRAIT_IGNOREDAMAGESLOWDOWN))	//if we want to ignore slowdown from damage, but not from equipment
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1046,7 +1046,8 @@
 		return
 	var/stambufferinfluence = (bufferedstam*(100/stambuffer))*0.2 //CIT CHANGE - makes stamina buffer influence movedelay
 	if(!HAS_TRAIT(src, TRAIT_IGNOREDAMAGESLOWDOWN))	//if we want to ignore slowdown from damage, but not from equipment
-		var/health_deficiency = ((maxHealth + stambufferinfluence) - health + (getStaminaLoss()*0.75))//CIT CHANGE - reduces the impact of staminaloss and makes stamina buffer influence it
+		var/scaling = maxHealth / 100
```
Putnam3145, line 1049: make 100 as the default maxhealth that this is scaled to a define?

Ghommie (approved):

PR #13498, opened 2020-10-02
Lingblood rework (Loudness is now an average, rather than a sum)

deathride58:
Title. A full writeup of why I'm making this change, and what I predict will come out of it, can be found here: https://hackmd.io/@deathride58/HkTyFgr8v

~~(this is currently untested because byond is just completely refusing to compile on my own hardware for some reason, but it should in theory work. theoretically.)~~ (edit: i figured out how to break byond's compiler, disregard)

## Changelog
:cl: Bhijn
tweak: Changeling loudness is now determined as an average of all their abilities, rather than the sum
tweak: To compensate for this, blood tests now require a loudness value of 1 or higher to detect ling blood. Additionally, blood test explosions are now triggered only when the loudness value is higher than 2.
/:cl:

### DIFF
--- a/code/__DEFINES/antagonists.dm
+++ b/code/__DEFINES/antagonists.dm
@@ -63,6 +63,10 @@
 #define CONTRACT_UPLINK_PAGE_CONTRACTS "CONTRACTS"
 #define CONTRACT_UPLINK_PAGE_HUB "HUB"
 
+//Lingblood stuff
+#define LINGBLOOD_DETECTION_THRESHOLD 1
+#define LINGBLOOD_EXPLOSION_MULT 2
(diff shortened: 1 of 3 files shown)


Hatterhat:
imagine not going loud the moment you get antag

Detective-Google:
never once have I had heating a ling's blood actually do anything

deathride58:
@Detective-Google yeah, this PR should hopefully fix that, since it'll be make lings have to deliberately go out of their way to pick up stealthy abilities if they wanna avoid blood tests, rather than how it is on the current code where lings simply have to avoid taking more than 2 loud abilities at a time if they wanna avoid blood tests

PR #13524, opened 2020-10-06
cid randomizer check bypass via whitelist

silicons:
someone seems to be unable to get through the randomizer check.
this will allow me to override it for that specific person.

security wise, every CID change by people on the whitelist is still alerted and recorded, so it's easy to tell if it's legitimate foul play even with someone bypassing it.


deathride58 (approved):

deathride58:
Actually wait. Doesn't the whitelist also take priority over bans, stickybans, goonchat's ban evasion detection, etc?

edit: nvm ignore me im dum

PR #13528, opened 2020-10-06
Robotic limbs 2 : Threshhold Boogaloo

DeltaFire15:
## About The Pull Request
So. I looked over bodypart code and thought 'Hey, wait, why are we using BODYPART ORGANIC || rendered_like_organic everywhere, when we can just make a BODYPART_HYBRID define, and add appropriate helpers + replace the checks everywhere.'
Anyways, a few hours later, and thats done, aswell as a bit of other stuff:

Other than the backend things that'll probably not really be too visible ingame, the biggest change is, that instead of robotic / hybrid limbs working after the principle 'If you go above [x] damage of a type, you cannot go below [x] damage of that type without surgery', they instead work via 'If you go above [x] damage of a type, you cannot go below [y] damage of that type without surgery', which should make them slightly less punishing for their users, whilst still being debilitating, especially if present on multiple limbs.
For species which get robotic limbs specifically, the x and y are 25 and 15 respectively, though they are an easy to modify define, so, if that seems like bad values, do tell me.
For other robotic parts, it's generally:
Borg limbs: High threshhold, but also high mindamage, so don't go above the threshhold
Advanced Prosthetics: Relatively low threshhold, but also very low mindamage, making them not very disruptive even if they did go past the threshhold
Surplus Prosthetics: Terrible.
Hybrid limbs can now also be injected by hypos on inject mode, though spraying won't work.
Additionally, I fixed up some surgeries having rather nonfunctional requirements, e.g. the robotic limb fixing surgery.

As usual, tell me if anything seems off or imbalanced yadda yadda.

Attendum: Don't mind the five billion commits there supposedly is on this, I pressed a wrong button when resolving merge conflicts. It's all fixed now, though!

## Why It's Good For The Game
Refactoring how synth-species limbs work should be useful for having an easier time doing things that only work on specific combinations, without needing a chain of ||s.
Making the threshhold value itself and the minimum damage it locks the limb at seperate makes robotic limbs a tiny bit less directly punishing for crossing the threhhold by even one point, while still keeping it somewhat debilitating especially if on multiple limbs at once, and allowing for better value differenciation for different types of limbs.
Everything else is generally just minor changes that either have a minor impact on the game and / or fix stuff up.

## Changelog
:cl:
balance: Robotic Limbs now each have their own damage threshhold values
balance: Robotic Limb damage threshholds are now seperated into threshhold itself and mindamage when passed
balance; Hybrid limbs can now be injected with hypos, but not sprayed (Still not healed by chems)
tweak: Brain surgery has been tweaked back to allowing robotic limbs, blacklisting IPC brains instead.
tweak: Robot brain surgery can now be used on organic heads, if there is a IPC brain in them somehow.
tweak: The robot limb heal surgery can now be used even if the target's torso is not robotic, as long as they have robotic limbs
refactor: BODYPART_ROBOTIC / BODYPART_ORGANIC checks replaced with helper-procs whereever possible.
code: Added a BODYPART_HYBRID define for robotic bodyparts that behave organic in some regards.
/:cl:


shellspeed1:
Dumb question, could we use this to make hybrid limbs for augmentation?

DeltaFire15:
Currently, you'd need to create a spare synthetic and take their limbs since you can't just print those as of now, but yes you can use them for augmentation.

PR #13555, opened 2020-10-11
fixes movespeed configurations

silicons:


code/controllers/configuration/config_entry.dm
```diff
@@ -27,7 +27,10 @@
 /datum/config_entry/New()
 	if(type == abstract_type)
 		CRASH("Abstract config entry [type] instatiated!")
-	name = lowertext(type2top(type))
+	if(isnull(name))
```
Ghommie:
```suggestion
	if(!name)
```
name is a text variable, `isnull("")` returns 0 unlike `!""` iirc (in the rare case someone changes config_entry/var/name value from null to "").

Ghommie (dismissed):

PR #13560, opened 2020-10-12
Changes spawn rate for wittel geysers from 1 in 27 to 1 in 20

shellspeed1:
## About The Pull Request
This changes the spawn rate for wittel from 1 in 27 to 1 in 20 leaving it as still rare but means that the chemical is actually available. 

## Why It's Good For The Game
Makes it less frustrating to get wittel which is already a difficult chem to get and use as it requires a miner to find multiple geysers, use a reinforced plunger on them, and lastly collecting the chemicals and taking them to the station which can be done with a beaker or a large plumbing project.

## Changelog
:cl:
balance: The spawn rate of wittel has been upped to 1 in 20 up from 1 in 27.
/:cl:


shellspeed1:
Actually this is a bad idea.

PR #13561, opened 2020-10-13
[s] Objective fixes : Cryo edition

DeltaFire15:
## About The Pull Request
Hey remember #13424? Yes?
Guess what, the thing I warned about happened, and people are getting free greentexts on kill / kill once / maroon / etc objectives if their target cryos. The only one of those that does not need their target in a certain state at roundend specifically is kill once.
So, I reworked that check a bit. Instead of using the default check_completion() proc, I added a check_midround_completion() proc that gets called by the cryopod, and if not overriden, just calls the check_completion() proc instead. Spoiler, it IS overridden in alot of objectives with targets that might accidentally lead to auto-greentexts if allowed to run their roundend check with a cryoing target, generally just with FALSE, but with a check for the actual midround completion state (won) for kill once. Hooray!

## Why It's Good For The Game
Free greentext man bad, fix man good.

## Changelog
:cl:
fix: Your target cryoing will no longer give you a free greentext.
/:cl:


code/game/gamemodes/objective.dm
```diff
@@ -313,6 +332,9 @@ GLOBAL_LIST_EMPTY(objectives)
 /datum/objective/protect/check_completion()
 	return !target || considered_alive(target, enforce_human = human_check)
 
+/datum/objective/protect/check_completion()
```
Ghommie:
```suggestion
/datum/objective/protect/check_midround_completion()
```

Ghommie (changes requested):

Ghommie (approved):

PR #13562, opened 2020-10-13
Fixes vamp shapeshifting a bit

DeltaFire15:
## About The Pull Request
Vamp shapeshifting as of now has a bunch of issues, mostly related with it persisting after losing species via special ways (e.g. borging), and simillar things. This should prevent those issues, especially considering it's vampire season.

The only place it's used in is the vampire species after all, and this poses a failsafe against it not being removed for whatever reason (aswell as actually typecasting the caster to a carbon when checking blood volume.. and fixing a check that was the wrong way...)

## Why It's Good For The Game
Fix man good

## Changelog
:cl:
fix: Vampire shapeshifting should now behave as intended
/:cl:


Ghommie (dismissed):

code/modules/mob/living/carbon/human/species_types/vampire.dm
```diff
@@ -153,14 +153,19 @@
 		to_chat(caster, "<span class='warning'>You're already shapeshifted!</span>")
 		return
 
+	var/mob/living/carbon/human/human_caster = caster
+	if(!human_caster?.dna?.species?.id == SPECIES_VAMPIRE)
+		to_chat(caster, "<span class='warning'>You don't seem to be able to shapeshift..</span>")
+		return
```
Ghommie: Actually, I think this portion should be located in _`/obj/effect/proc_holder/spell/targeted/shapeshift/bat/cast(list/targets,mob/user = usr)`_ (while replacing _`caster`_ with _`user`_) indented under this check: 

_```if(!(locate(/obj/shapeshift_holder) in targets[1]))```_

to stop further unwarranted behaviours found in `shapeshift/cast()`.
Also add a typecheck at the start of the species id check since you mentioned non-human mobs.

P.S. Lines 151-154 are unreachable, _`/obj/effect/proc_holder/spell/targeted/shapeshift/Restore`_ is run instead in that case, and need be removed.

Ghommie (changes requested):

code/modules/mob/living/carbon/human/species_types/vampire.dm
DeltaFire15: Mh, I don't think I want to stop them from transforming back (if they somehow get changed while in the mob holder, for example), just initiating the transform itself which Shapeshift() does, so having it in this proc should be fine imo.

151-154, while rare, can be reached if Shapeshift() gets called manually / by different sources, so better to have a failsafe regardless.

if(!human_caster?.[etc]) should be a sufficient check for human_caster not existing, e.g. them being non-humanoid / the typecast failing, no? Unless I misunderstood how the ? operator works there.

Ghommie: > Mh, I don't think I want to stop them from transforming back (if they somehow get changed while in the mob holder, for example), just initiating the transform itself which Shapeshift() does, so having it in this proc should be fine imo.

the _`if(!(locate(/obj/shapeshift_holder) in targets[1]))`_ line should take care of it anyway. (I forgot to include the NOT operator before, my bad).
 
> if(!human_caster?.[etc]) should be a sufficient check for human_caster not existing, e.g. them being non-humanoid / the typecast failing, no? Unless I misunderstood how the ? operator works there.

Pretty much. The _`?.`_ operator returns 1 only if the value is null (maybe 0 and "" now too). So a typecheck is necessary to avoid errors.

DeltaFire15: updated it & moved the species check to cast(), that fine now?

PR #13565, opened 2020-10-13
adds empty() proc to pipelines that purges all connected airs to avoid sdql2 and proccall torture (SPT) whenever atmos breaks and the station is -#1.IND degrees hot

silicons:
also useful for grief fixing

### DIFF
--- a/code/modules/atmospherics/machinery/datum_pipeline.dm
+++ b/code/modules/atmospherics/machinery/datum_pipeline.dm
@@ -208,7 +208,11 @@
 		stack_trace("[src]([REF(src)]) has one or more null gas mixtures, which may cause bugs. Null mixtures will not be considered in reconcile_air().")
 		return listclearnulls(.)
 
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/machinery/datum_pipeline.dm
```diff
@@ -208,7 +208,11 @@
 		stack_trace("[src]([REF(src)]) has one or more null gas mixtures, which may cause bugs. Null mixtures will not be considered in reconcile_air().")
 		return listclearnulls(.)
 
-/datum/pipeline/proc/reconcile_air()
+/datum/pipeline/proc/empty()
+	for(var/datum/gas_mixture/GM in get_all_connected_airs())
+		GM.empty()
```
Putnam3145:
```suggestion
		GM.clear()
```

Putnam3145 (changes requested):

PR #13578, opened 2020-10-16
removes majority of taurs, replaces with virgo taurs and virgo colour system

timothyteakettle:
## About The Pull Request
**this needs the migration test & wearing clothes to be tested, please do not merge this yet and holy shit testmerge it before merging**

all old synth taur sprites have been kept, and are renamed to be (old) and still use mutant colours for coloring, meaning if you have them selected, NO CHANGES WILL BE MADE TO YOU
the above change also applies to anyone with the tentacle or naga sprites (to my knowledge all our active taur players are using tentacle/naga/synth sprites, and also malric using drake i think but they seem okay with the new sprites)

(also for anyone who hates the virgo naga sprite, delams converted our existing naga sprite to use the better colour system, so we're not actually using the virgo naga sprite)

this is the UI when selecting a new taur sprite that has a marking (2 parts that can be separately coloured)
image

here's a bad example of the separation of colours (see light/dark green, one part is markings, the other is the primary)
image

## Why It's Good For The Game
our taurs kinda suck
being able to clearly and separately colour two parts of taur sprites without having to select mutant colours which was an awful concept is good

## Changelog
:cl:
del: removed a lot of taur sprites
add: added a lot of new taur sprites
add: added new colour system for sprite accessories which relies on the independent colouring of two parts that make up one accessory
/:cl:


timothyteakettle:
i went the wrong way about this
removal is not the right idea when i can instead convert stuff to color matrixes and simply move what is being used as a color matrix

PR #13582, opened 2020-10-17
adds advanced colouring mode: allows you to individually colour mutant parts

timothyteakettle:
## About The Pull Request
title

this system is complicated (we have no spriting standard for matrixed parts, so i had to painstakingly list out what sections were matrixed for each matrixed sprite, i.e. if only red was used, if green was used, if red AND green was used, and so on for R/G/B)
and then after all of this it has to be swapped around behind the scenes when choosing stuff so that your primary feature is always stored as red, secondary as green, tertiary as blue, but what you see in the selection will likely not be this, because a sprite with only a green matrix section, will be chosen through the primary colour, because i swap the primary href around for the secondary, which in turn is what stores green matrixed section values

tldr; it's complicated and has hurt me in many ways

## Why It's Good For The Game
improves customization a ridiculous amount by giving true freedom over the colour of mutant parts

## Changelog
:cl:
fix: a tonne of fixes to colourisation of parts, too many to name, including some sprite fixes
add: things now have their own individual primary/(secondary)/(tertiary) colours as required, and these can be modified by you
/:cl:


BlueWildrose:
oh dear goodness this is going to be awkward for the slime shapeshifting feature


AND also for the slime splitting feature

AND also for the magic mirror feature if you're a wizard probably

oh yeah there's also the pride mirror.


at this rate maybe changing the slime color in-game should shift the hue of the slime person entirely instead of just changing one color

timothyteakettle:
mirrors it shouldn't really matter too much (often not used aside for changing species and massive shitcode that i'll refactor at some point to be based off customization code)
slime customization: yeah if you change your body colour in-game i could probably make it change the colour of everything else or have a second option for all-colour

timothyteakettle:
okay its a toggle now, the migration is tested and works, the toggle is off by default, the migration doesn't touch your body colours and because the toggle is off your mutant parts are ALSO not affected

PR #13584, opened 2020-10-17
adds EMOTE_BOTH and EMOTE_OMNI which is both audible/visible and ignores checks entirely. subtler/subtle is omni and regular emotes are both.

silicons:
## About The Pull Request

this also makes it so you can't choose visible or audible for subtler/etc because, well, there's no need anymore.

## Why It's Good For The Game

i've been told it's annoying to not be able to see emotes by blind characters
i don't think this'll be abused.

## Changelog
:cl:
experimental: blind people can now see custom emotes/me emotes
experimental: subtler now ignores blind/deafen checks.
/:cl:

### DIFF
--- a/code/__DEFINES/say.dm
+++ b/code/__DEFINES/say.dm
@@ -76,8 +76,14 @@
 #define LINGHIVE_LINK 3
 
 //whether the emote is visible or audible.
+// Requires sight
 #define EMOTE_VISIBLE 1
+// Requires hearing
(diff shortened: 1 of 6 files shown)


BlueWildrose:
did you do this via webedit

you absolute madman

Ghommie (approved):
I can't put a finger on it, but it feels like a blind people buff...

PR #13585, opened 2020-10-17
Fixed species-specific drinks not giving a mood boost if right species

BlueWildrose:
## About The Pull Request
There was a bug where drinking a species-special drink didn't give a mood boost like it's supposed to. I'm fixing that.

What caused this?
The reagent event used was "on_mob_life" rather than "reaction_mob". 
"on_mob_life" is triggered while the reagent is inside the character's body. "reaction_mob" is triggered upon initially taking the reagent into your body.

Initially, the drink does not count as a "DRINK_SPECIES" up until it detects an appropriate species on_mob_life.

Drink mood boosts are only applied when you initially take a sip, and only via ingestion. Dictated by "/datum/reagent/consumable/reaction_mob".

This means that after setting itself to "DRINK_SPECIES", it becomes pointless because it's already passed that point where it gives the mood boost.

This pull request will correct the code to be "reaction_mob" instead. This pull request also changes the disgust default to 26 from 25, to make the "barfy face" appear on your first sip, so you know you're drinking something bad for you immediately.

This change also means you thankfully can't weaponize species drinks (beyond the booze factor) to make people vomit horribly using a syringe gun. Because with the current code, it was making you a _horrible stunlocked vomiting mess with even just five of an incorrect species drink reagent_, **because it was applying massive chunks of disgust for each tick instead of upon one sip**.

## Why It's Good For The Game

I just wanna be a happy slimeperson that's drinking my Jell Wyrm.

also fixes are good.

## Changelog
:cl:
fix: Fixed species-specific drinks not giving a mood boost if you are that species.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
@@ -2284,7 +2284,7 @@ All effects don't start immediately, but rather get worse over time; the rate is
 
 /datum/reagent/consumable/ethanol/species_drink/reaction_mob(mob/living/carbon/C, method=TOUCH)
 	if(method == INGEST)
-		if(C.dna.species && C.dna.species.species_category == species_required) //species have a species_category variable that refers to one of the drinks
+		if(C?.dna?.species?.species_category == species_required) //species have a species_category variable that refers to one of the drinks
 			quality = RACE_DRINK
 		else
 			C.adjust_disgust(disgust)
--- a/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
@@ -2279,7 +2279,7 @@ All effects don't start immediately, but rather get worse over time; the rate is
 ////////////////////
 /datum/reagent/consumable/ethanol/species_drink
 	var/species_required
-	var/disgust = 30
+	var/disgust = 26
 	boozepwr = 50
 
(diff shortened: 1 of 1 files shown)


Ghommie (dismissed):

code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
```diff
@@ -2279,15 +2279,16 @@ All effects don't start immediately, but rather get worse over time; the rate is
...
 
-/datum/reagent/consumable/ethanol/species_drink/on_mob_life(mob/living/carbon/C)
-	if(C.dna.species && C.dna.species.species_category == species_required) //species have a species_category variable that refers to one of the drinks
-		quality = RACE_DRINK
-	else
-		C.adjust_disgust(disgust)
-	return ..()
+/datum/reagent/consumable/ethanol/species_drink/reaction_mob(mob/living/carbon/C, method=TOUCH)
+	if(method == INGEST)
+		if(C.dna.species && C.dna.species.species_category == species_required) //species have a species_category variable that refers to one of the drinks
```
Putnam3145: minor nitpick but if this isn't fixed here i'll do it myself later
```suggestion
		if(C?.dna?.species?.species_category == species_required) //species have a species_category variable that refers to one of the drinks
```

BlueWildrose: Not a bad idea, I'll toss in your suggestion.

PR #13596, opened 2020-10-19
Donator - Vest and Shirt

LetterJay:
Should be ready, pending a description and name change


modular_citadel/code/modules/client/loadout/__donator.dm
```diff
@@ -522,3 +522,10 @@
 	slot = SLOT_HEAD
 	path = 	/obj/item/clothing/head/halo
 	ckeywhitelist = list("hisakaki")
+
+/datum/gear/donator\vest
```
SandPoot: \

Putnam3145:
```suggestion
/datum/gear/donator/vest
```

PR #13601, opened 2020-10-20
antagonistic items will no longer fail to be placed on you

timothyteakettle:
## About The Pull Request
title

works through the following process for critical items:
try adding it to any of the given slots as normal, if it fails, continue
do they have a backpack? if so, place it in
otherwise, can we forcibly give them a default backpack? if so, equip the backpack, then place the item inside
otherwise, message admins with information on the incident (this should never happen, as forcibly equipping a backpack should never fail)

## Why It's Good For The Game
no more "i didnt get my [antag tool] when i spawned in as [some antag]"

## Changelog
:cl:
tweak: antag items are now of critical importance and wont fail to be placed on the character
/:cl:

### DIFF
--- a/code/game/gamemodes/gangs/gang.dm
+++ b/code/game/gamemodes/gangs/gang.dm
@@ -197,7 +197,7 @@
 
 	if(gangtool)//Here is where all of the text occurs when a gang boss first spawns in.
 		var/obj/item/device/gangtool/G = new()
-		var/where = H.equip_in_one_of_slots(G, slots)
+		var/where = H.equip_in_one_of_slots(G, slots, critical = TRUE)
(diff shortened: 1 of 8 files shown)


shellspeed1:
Better idea, what about making the mission critical items free in the uplink for the first purchase?

timothyteakettle:
that only works for syndicate items, but yeah that's a good idea

The-Sun-In-Splendour:
How about giving them a one use implant that drops pod their antag items on them. Big warning window that says HEY ITS VERY VISIBLE.

Additionally you could literally just make reskins of it where it teleports with blood magic or whatever for cultists and such?

PR #13602, opened 2020-10-20
Adds a new "minor antag" roundstart to dynamic

Putnam3145:
## About The Pull Request

This adds a new type of round that dynamic rounds can roll, where, instead of picking a roundstart type and scaling it up, it picks individual antagonists until it reaches the target threat level, so that it can be e.g. two traitors and a bloodsucker, one heretic and a traitor, a bloodsucker and a changeling etc.

This can happen randomly, depending on storyteller + threat level, or it can happen if the roundstart ruleset picking fails for whatever reason (e.g. nobody has their damn prefs on).

## Why It's Good For The Game

Moooostly I think that this can help with certain preference issues, e.g. if "not enough people have traitor on" for the game mode or whatever it'll just fall back to checking everyone and rolling "enough" people. The question remains whether the threat/threat level paradigm is a useful measure as opposed to mere antag counts, but hey.

## Changelog
:cl:
add: A fallback for dynamic antag rolling that allows for it to just try between traitor, blood brothers, heretics, changeling, bloodsucker and devil until there are enough roundstart antags. This can also happen randomly anyway. Blood brothers and devil are disabled for now, but the code is there to enable them.
add: A new storyteller, "Grab Bag", that forces the above round type.
/:cl:

### DIFF
--- a/code/__DEFINES/dynamic.dm
+++ b/code/__DEFINES/dynamic.dm
@@ -1,7 +1,7 @@
 #define CURRENT_LIVING_PLAYERS	1
 #define CURRENT_LIVING_ANTAGS	2
 #define CURRENT_DEAD_PLAYERS	3
-#define CURRENT_OBSERVERS	    4
+#define CURRENT_OBSERVERS		4
 
 #define NO_ASSASSIN				(1<<0)
 #define WAROPS_ALWAYS_ALLOWED	(1<<1)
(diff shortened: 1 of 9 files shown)


YakumoChen:
Does all this still roll before jobs?

Putnam3145:
@YakumoChen minor antags in particular roll after jobs.

YakumoChen:
Minor meaning all the ones in this mode? Or do you mean like ghost role antags? If everything rolls after jobs that just means that sec/head mainstay players will never get a chance to be antag in this mode.

Putnam3145:
This mode will *usually* only be seen when people vote calm/grab bag or if not enough players can be put together for a roundstart antag setup in the first place--if that happens, then the sec/head mains will have been put in jobs that allow them to be that antag anyway, because it:

1. Rolls a roundstart antag
2. Attempts to start the antag, which involves putting the people with the pref on in jobs that allow it, randomly
3. If it can't get enough people for it, starts the minor antag roll, 1-5 minutes later

PR #13610, opened 2020-10-22
Makes the atmos subsystem pause correctly

Putnam3145:
## About The Pull Request

if active turfs gets too high then EQ starts taking literally all time away from active turfs which keeps active turfs from becoming inactive which makes that last forever. this fixes that. i'm dumb.

## Why It's Good For The Game

atmos subsystem doesn't work without this.

## Changelog
:cl:
fix: atmos subsystem no longer dies if there's too many gases
/:cl:


DeltaFire15:
Please look into the reason travis is failing on this PR.

Linzolle:
Is there a reason this hasn't been merged yet? It's been test merged for so long and everything breaks without it, feels like it should just be merged at this point unless it's some kind of dumb bandaid fix.

DeltaFire15:
> Is there a reason this hasn't been merged yet? It's been test merged for so long and everything breaks without it, feels like it should just be merged at this point unless it's some kind of dumb bandaid fix.

It reliably fails travis, so if it was merged, any new PRs would always fail it, which is pretty annoying.

PR #13617, opened 2020-10-23
loadout 2: electric boogaloo

timothyteakettle:
## About The Pull Request
loadout saved/loaded as json
data structure supports loadout save slots which you can swap between but for now all data is allocated to slot 1
the ability to colour polychromic items from the loadout, and have it save

## Why It's Good For The Game
polychromic colour saving

future support for:
things like name/desc changing on items
colouring non polychromic items
save slots

## Changelog
:cl:
add: loadout now has save slot support and colour choosing/saving for polychromic items
/:cl:


timothyteakettle:
json loadout works now i just need to add the funky colour selection / slot selection

timothyteakettle:
color can now be edited, the final issue seems to be the migration and the application of the change in color to the element, also once the colours shot up to 17 from 3 but i could never replicate it, uh, woops

PR #12994, opened 2020-08-01
removes dildo

ghost:
## About The Pull Request
circuit assemblies

## Why It's Good For The Game
sex circuits are permabannable

## Changelog
:cl:
del: dildo circuit assemblies
/:cl:

PR #13625, opened 2020-10-25
Reverts this circuit frame removal 

LetterN:
Reverts Citadel-Station-13/Citadel-Station-13#12994

asked by @silicons


MrJWhit:
Why? Does anyone need dildo circuits?

MrJWhit:
Oh, when I asked silicons, the response was "i said if they wanted to revert it to revert it".
Which, why?

silicons:
"asked by" means i said people can revert it if they want, as it isn't particularly harmful to keep

i don't personally care about this either way.

PR #13630, opened 2020-10-26
character size scaling

timothyteakettle:
## About The Pull Request
before you downvote this and cry because of the title, please read the pr description, please.

the size limits are 90% to 125%
these do not stack with the dwarf species, i.e. they just outright will not apply

if you go under 100% size, REGARDLESS OF BY HOW MUCH, you will lose 10 max health, meaning you can't gamer the debuff by being 99% size or 91% or something dumb in that regard

if you go above 100% size, nothing happens, you're just bigger, congratulations on having a bigger hitbox
the reason it's capped at 125% is because this is the size of the gigantism trait and id rather it not be bigger than that because i dont want people filling my screen with their size on main

note: sizes can and will get distorted, however it's not too much of an issue because the minimum size i put in is literally 90%

## Why It's Good For The Game
size customization in a sane manner when done correctly can be good

## Changelog
:cl:
add: you can now customize your size from 90% to 130%, going below 100% makes you have 10 less max health
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -519,10 +519,10 @@
 	min_val = RESIZE_DEFAULT_SIZE
 	integer = FALSE
 
-//Pun-Pun movement slowdown given to characters with a body size smaller than this value,
+//Penalties given to characters with a body size smaller than this value,
 //to compensate for their smaller hitbox.
 //To disable, just make sure the value is lower than 'body_size_min'
-/datum/config_entry/number/threshold_body_size_slowdown
+/datum/config_entry/number/threshold_body_size_penalty
 	config_entry_value = RESIZE_DEFAULT_SIZE * 0.85
 	min_val = 0
 	max_val = RESIZE_DEFAULT_SIZE
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -531,8 +531,8 @@
 //multiplicative slowdown multiplier. See 'dna.update_body_size' for the operation.
 //doesn't apply to floating or crawling mobs
 /datum/config_entry/number/body_size_slowdown_multiplier
-	config_entry_value = 0.25
-	min_val = 0.1 //To encourage folks to disable the slowdown through the above config instead.
+	config_entry_value = 0
+	min_val = 0
 	integer = FALSE
 
 //Allows players to set a hexadecimal color of their choice as skin tone, on top of the standard ones.
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -509,13 +509,13 @@
 
 //Body size configs, the feature will be disabled if both min and max have the same value.
 /datum/config_entry/number/body_size_min
-	config_entry_value = RESIZE_DEFAULT_SIZE
+	config_entry_value = 0.9
 	min_val = 0.1 //to avoid issues with zeros and negative values.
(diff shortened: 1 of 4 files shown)


MrJWhit:
No.

timothyteakettle:
thanks for your detailed constructive feedback i will use this to make my prs better

MrJWhit:
It's not worth it to open this door, dwarves are kind of balanced because of their hilarious booze need and the fact that they can't speak for shit. Even though they're still busted because of the hitbox.
Reducing hp isn't that serious of a penalty when you are less likely to be hit, because you're physically smaller and it's harder to just click on them. Stack that with cit's movespeed, and anyone half decent at dodging, is going to just crush with it.

10 less hp does what, exactly? Put them into crit 1 click faster? That's /not/ enough of a deterrent.

Size mechanics are inhearently broken unless you rework how the sprites and hitboxes are hit, you are inherently better at combat because of this.

timothyteakettle:
Being 10% smaller is too small a change for it to be incredibly strong, the 10 less max health is literally enough for it to be mechanically not worth using, to be convinced that having 10 less health is worth being 10% smaller from a purely mechanical point of view is silly.

I can play around with the values, but having 10 less maximum health is a huge debuff, it's literally 10% of your life /gone/ permanently, the comparison being you're a tiny bit smaller and so people who would click the outer edges of your sprite might now miss, which is comparable to just doing stuff like going bald to reduce your hitbox

timothyteakettle:
Ill add some reference images later today for 90%, 100%, and 125% size ranges on the same character, next to eachother to show the actual hitbox differences

deathride58 (changes requested):
the config stuff for character size caps should be left in, same with the config penalties for sizes smaller than certain values

timothyteakettle:
re-added config values for minimum, maximum, and the penalty threshold

timothyteakettle:
re-added slow stride formula on-top of the 10 hp reduction, however i changed the values of the config to have the multiplier be 0 for now, so it can be enabled/tuned by whoever touches configs

i.e. the support for it is there and it is up to other people to enable it if they wish for it to be used

deathride58 (dismissed):

YakumoChen:
Is this 10 HP based on the human max of 200 or the crit threshold of 100

timothyteakettle:
based on the maximum such that you go into crit 10 health earlier

timothyteakettle:
had to do a fix because there was an old issue with the original size code

Tupinambis:
I'm against this for consistency, visual clarity, visual quality, and gameplay reasons. I get why customization and better *showing* the size of people sounds good in theory, but I feel like it will just make gameplay more inconsistent and make things look uglier. I've seen allowing size scaling in action and I'm really not in any way a fan. It doesn't feel like the benefits are worth the drawbacks.

YakumoChen:
Sprite distortion is incredibly ugly.

silicons:
*takes a deep breath*

timothyteakettle:
sprites can look distorted (it warns you when selecting your sprite size!), this is typically worse when downsizing the sprite compared to upsizing the sprite, but the problem is minimised due to the cap of 90%

i feel like having this testmerged and getting feedback once people have actually seen and tried it would be nice

Echopsychotic:
I think the sprites would look better at 150 personally, again it's only a hitbox taker for the person and not much of a bad change to increase the size the sprites can take.

YakumoChen:
is there a way to make the hitbox static but not affect the size?

moltoretardo:
Flashbacks from cit 6 years ago 😨

PR #13644, opened 2020-10-29
[READY] Hybrid bodyparts - wound surgery edition

DeltaFire15:
## About The Pull Request
Since synthetics as of now can get wounds, but the wound-fixing surgeries were locked to organics only, this gives them access to those surgeries, with some organic steps replaced with mechanic ones where applicable.
Also fixes a typo in the mechanic_open step, and clarifies the required tool for two mechanical surgery steps.

~~Pending local testing to doublecheck this working aswell as some other hybrid bodypart interactions, but should work.~~
Localtested it a bunch and it works fine, additionally also made wound surgeries not qdel() when the wound is fixed since it's kinda weird to see the close step, now being available, and the surgery then just being deleted instead. All wound-related surgeries have checks in place for their steps if there's no wound there anymore anyways.

Edit: This also fixes organs being fed to people after inserting them (was because success returned 0 instead of 1), and designates most tools (welder, wirecutter, wrench, screwdriver, multitool) as surgery tools due to the bigger amount of people requiring robotic surgeries now. (makes them not bonk people if on help intent and having a active surgery), plus it makes the initiate() proc of surgery always return true if the doafter fails as not doing so causes some annoyances (starting to forcefeed / slapping them with whatever you used for the surgery / etc.). The behaviour for axtually failing is still the same.

## Why It's Good For The Game
Synths can get wounds, they should also be able to heal them via surgery instead of only via the agonizingly slow nonsurgical means.
Also most tools are commonly needed for surgery now that we have more robotic synths, therefore they should have the safeguard normal tools have (so you don't accidentally stab your patient's eyes out)
Not having people get slapped with / forcefed whatever you were using for the surgery if you step away without finishing the step sounds like a good thing too.

## Changelog
:cl:
balance: Biomechanical (hybrid) bodyparts now have access to wound-fixing surgeries.
tweak: A wound being fixed no longer just qdel()s surgeries connected to it.
tweak: Some robotic surgery steps are now a bit more clear.
fix: Organs no longer get fed to people after successfully being inserted into them.
tweak: Not completing the do_after of a surgery no longer causes you to attack the target with whatever you were holding.
/:cl:

### DIFF
--- a/code/datums/wounds/_wounds.dm
+++ b/code/datums/wounds/_wounds.dm
@@ -91,8 +91,6 @@
 	var/wound_flags = (FLESH_WOUND | BONE_WOUND | ACCEPTS_GAUZE)
 
 /datum/wound/Destroy()
-	if(attached_surgery)
-		QDEL_NULL(attached_surgery)
 	if(limb?.wounds && (src in limb.wounds)) // destroy can call remove_wound() and remove_wound() calls qdel, so we check to make sure there's anything to remove first
 		remove_wound()
 	limb = null
--- a/code/game/objects/items/devices/multitool.dm
+++ b/code/game/objects/items/devices/multitool.dm
@@ -21,13 +21,13 @@
 	force = 5
 	w_class = WEIGHT_CLASS_SMALL
 	tool_behaviour = TOOL_MULTITOOL
+	item_flags = SURGICAL_TOOL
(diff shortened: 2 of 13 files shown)


DeltaFire15:
Can't wait for this to merge conflict after half a month of sitting around.

PR #13683, opened 2020-11-14
Attempts to turn gift static blacklist into a global list

SandPoot:
## About The Pull Request
Title.

## Why It's Good For The Game
Travis Angwy.

## Changelog
:cl:
code: Swaps the gift static blacklist with a global list one.
/:cl:

### DIFF
--- a/code/game/objects/items/gift.dm
+++ b/code/game/objects/items/gift.dm
@@ -8,14 +8,14 @@
  * Gifts
  */
 
-GLOBAL_LIST_INIT(gift_item_blacklist, list(
+GLOBAL_LIST_INIT(gift_item_blacklist, typecacheof(list(
 	/obj/effect,
 	/obj/belly,
 	/obj/mafia_game_board,
 	/obj/docking_port,
 	/obj/shapeshift_holder,
 	/obj/screen
-	))
+	)))
 
 GLOBAL_LIST_EMPTY(possible_gifts)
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/gift.dm
```diff
@@ -8,13 +8,14 @@
...
-/var/static/blacklisted_items = typecacheof(list(
+GLOBAL_LIST_INIT(blacklisted_items, list(
 	/obj/effect,
 	/obj/belly,
 	/obj/mafia_game_board,
 	/obj/docking_port,
 	/obj/shapeshift_holder,
-	/obj/screen))
+	/obj/screen,
+	))
```
DeltaFire15:
```suggestion
	/obj/screen
	)))
```

DeltaFire15: I also recommend renaming the list to something like gift_item_blacklist so it doesn't confuse people / immediately conflict with the next person to name their global list blacklisted_items.

SandPoot: Seems right, when i get to my computer I'll change it.

SandPoot: I just think that the third parenthesis is wrong.

DeltaFire15: Weeell yes, but not if you readd the typecacheof in front of the list, because otherwise it'd be able to spawn any subtype of the blacklisted items

SandPoot:
Also, it would be nice if this got merged alongside  #13677.

PR #13693, opened 2020-11-20
Lazy change to give lings more DNA slots

shellspeed1:
## About The Pull Request
Lets lings turn into more people to have with the stealth antag part. 

## Why It's Good For The Game
Makes lings more sneaky. Also enables memes with turning the entire crew into each other to cause chaos more effectively. 

## Changelog
:cl:
balance: Lings now have a few more DNA slots.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -18,7 +18,7 @@
 
 	var/list/stored_profiles = list() //list of datum/changelingprofile
 	var/datum/changelingprofile/first_prof = null
-	var/dna_max = 6 //How many extra DNA strands the changeling can store for transformation.
+	var/dna_max = 99 //How many extra DNA strands the changeling can store for transformation.
 	var/absorbedcount = 0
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/changeling/changeling.dm
```diff
@@ -18,7 +18,7 @@
 
 	var/list/stored_profiles = list() //list of datum/changelingprofile
 	var/datum/changelingprofile/first_prof = null
-	var/dna_max = 6 //How many extra DNA strands the changeling can store for transformation.
+	var/dna_max = 99 //How many extra DNA strands the changeling can store for transformation.
```
Arturlang, line 21: If you intend to essentially remove the dna strand cap, you should be more clear about it in your changelog, and It's bad practice to just leave in code that is not really neccesary, I suggest that you get rid of this var and the places it's used so we do not have more unused code floating around.
Not to mention that your changelog is suggesting you only added a couple more slots, when in reality you made it essentially unlimited.

PR #13699, opened 2020-11-22
nanite updates: permanent nanites, hostile lockdown, anti-virus, resistance tweaking

silicons:
## About The Pull Request

adds 2 new programs that drain 0.5/tick each, preventing viral replica (but NOT infective exomotion) from working, and the other one will prevent consoles from modifying the nanites as well as the host themselves (not implemented, intended for use later)
nanite resistances made into variables, tweaked for new EMP system, shock/minor shocks tweaked accordingly.
**emp and shock drain/desync chances nerfed.**
stealth nanites no longer immune you from viral replica (why does it do that for a 0.2 cost program)
adds support for permanent nanites and unremovable nanites

## Why It's Good For The Game

more program types more program dynamics
will also need these functions very, very soon.

## Changelog
:cl:
tweak: nanite resistances tweaked
add: new nanite programs added for locking the user out from being modified by consoles or antivirals.
/:cl:

### DIFF
--- a/code/datums/components/nanites.dm
+++ b/code/datums/components/nanites.dm
@@ -288,7 +288,7 @@
 	severity *= emp_severity_mod
 	var/loss = (severity / 100) * (rand(emp_percent_deletion_lower, emp_percent_deletion_upper) * nanite_volume) + rand(emp_flat_deletion_lower, emp_flat_deletion_upper)
 	adjust_nanites(null, -loss)
-	if(prob(severity / emp_desync_mod))
+	if(prob(severity * emp_desync_mod))
 		cloud_id = 0
 	for(var/X in programs)
 		var/datum/nanite_program/NP = X
--- a/code/datums/components/nanites.dm
+++ b/code/datums/components/nanites.dm
@@ -189,10 +189,10 @@
 		permanent_programs += P
 		if(immutable)
 			P.immutable = TRUE
-		for(var/i in programs)
-			var/datum/nanite_program/E = i
+		for(var/e in programs)
+			var/datum/nanite_program/E = e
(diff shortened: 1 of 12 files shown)


DeltaFire15:
Stealth making you immune to viral replica is so there is a reason to install stealth except if you want to do very very bad things to the ones you install it in. Just how the 'you can't analyze the nanites' gives you 0.1 nanite regen so there is justifications to use it, and not just if you want to do bad stuff.

silicons:
viral replica is rarely used, stealth getting a buff is good imo

also i renerfed nanites vs emp/shock because i uhhh realized that my values were WAY too low lmao

code/datums/components/nanites.dm
```diff
@@ -191,28 +285,28 @@
 	holder.icon_state = "nanites[nanite_percent]"
 
 /datum/component/nanites/proc/on_emp(datum/source, severity)
-	nanite_volume *= (rand(60, 90) * 0.01)		//Lose 10-40% of nanites
-	adjust_nanites(null, -(rand(5, 50)))		//Lose 5-50 flat nanite volume
-	if(prob(40/severity))
+	severity *= emp_severity_mod
+	var/loss = (severity / 100) * (rand(emp_percent_deletion_lower, emp_percent_deletion_upper) * nanite_volume) + rand(emp_flat_deletion_lower, emp_flat_deletion_upper)
+	adjust_nanites(null, -loss)
+	if(prob(severity / emp_desync_mod))
```
DeltaFire15: I recommend doing this, since emp_desync_mod is < 1, therefore dividing through it would increase the probability, not decrease it.
```suggestion
	if(prob(severity * emp_desync_mod))
```

PR #13705, opened 2020-11-24
Buffs rad collectors

Putnam3145:
## About The Pull Request

Increased the power-from-radiation by 1.25x and increased power output by 1.25x, leading to a total increase of 56.25% buff.

## Why It's Good For The Game

With the way power is now, you need to set up the supermatter AND solars to get even a chance of powering the whole station. The supermatter's power output has been severely lacking historically anyway, so here's a patch to that.

## Changelog
:cl:
balance: Rad collectors now get 1.25x as much energy from radiation
balance: Rad collectors now put out 1.25x as much stored energy per tick
balance: Above two rad collector changes give a total 56.25% power output increase
/:cl:

### DIFF
--- a/code/modules/power/singularity/collector.dm
+++ b/code/modules/power/singularity/collector.dm
@@ -1,7 +1,7 @@
 // stored_power += (pulse_strength-RAD_COLLECTOR_EFFICIENCY)*RAD_COLLECTOR_COEFFICIENT
 #define RAD_COLLECTOR_EFFICIENCY 80 	// radiation needs to be over this amount to get power
-#define RAD_COLLECTOR_COEFFICIENT 100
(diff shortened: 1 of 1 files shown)


DeltaFire15:
The statement about power use generally would not be correct. Setting up either all solars (at least ~250kw, usually more) or the supermatter alone (~300-600kw on basic setups) is enough to keep the entire station powered.
(For comparison, the power use of all APCs together amounts to ~150-200kw at most, usually less, though a bit more while charging of course)

Looked into it some more, the issue currently present came from APCs always using as much power as they could for recharging their cell instead of just as much as is needed for the cell to get back to full, leading to it draining alot more power with my PR fixing the amount of powernet power they use for that process. It'll be fixed with the next testmerge update.

PR #13710, opened 2020-11-25
makes ores not destroyed by mining explosives

silicons:
## Why It's Good For The Game

it frankly isn't but would make toxins more useful than just for nuking the station with

## Changelog
:cl:
balance: in a cowardly move i have made ores not destroyed by explosions.
/:cl:

### DIFF
--- a/code/modules/mining/ores_coins.dm
+++ b/code/modules/mining/ores_coins.dm
@@ -318,12 +318,6 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
 	pixel_x = rand(0,16)-8
 	pixel_y = rand(0,8)-8
 
-/obj/item/stack/ore/ex_act(severity, target)
-	if (!severity || severity >= 2)
-		return
-	qdel(src)
-
-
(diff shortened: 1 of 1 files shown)


MrJWhit:
Feels too powerful.

Just set off explosions on lavaland for fun like the rest of us.

Soranayo:
I'm against this for a few reasons, first one being if a scientist does make a bomb and destroys half of lavaland with one, it makes miners ultimately useless. Second of all, if they do feel unneeded (And if it's actually a valid or powerful strategy, it will), it'll push players even more to focus on megafauna than they already are. Overall i think it's a bad idea.

DeltaFire15:
I *think* ores already shouldn't get destroyed by explosions, since they should be in the rock initially, which means the explosion affecting the tile destroys the rock, pops out the ore, and *technically* shouldn't break the ore. Let me test it locally to find out..

Tested locally, yes, bombs do not cause material loss as of now, even if the target rock is within the deva radius of the bomb.

PR #13712, opened 2020-11-26
burn infection changes

timothyteakettle:
## About The Pull Request
limbs at total infection wont stop you from healing the infection now
if you have preservation chems or are cold (and not resistant to cold) your infections will also stop getting worse

## Why It's Good For The Game
less lethal, actually treatable, less reliance on cloning

## Changelog
:cl:
tweak: limbs at total infection wont stop you from healing the infection
tweak: if you have preservation chems or are cold (and not resistant to cold) your infections will also stop getting worse
/:cl:

### DIFF
--- a/code/datums/wounds/burns.dm
+++ b/code/datums/wounds/burns.dm
@@ -25,16 +25,16 @@
 	var/sanitization = 0
 
 	/// Once we reach infestation beyond WOUND_INFESTATION_SEPSIS, we get this many warnings before the limb is completely paralyzed (you'd have to ignore a really bad burn for a really long time for this to happen)
-	var/strikes_to_lose_limb = 3
+	var/limb_decay = 3
+	var/datum/brain_trauma/severe/paralysis/sepsis
 
 
 /datum/wound/burn/handle_process()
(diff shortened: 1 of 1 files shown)


timothyteakettle:
gonna test this further because there's a case where it might not do what i expect

DeltaFire15 (changes requested):
Seems mostly good, though I'd suggest a early return before the switch handling infestation (l. 77) if the limb already is decayed, since it already gets constant tox damage caused at the very start of the proc, and some of the messages make no sense after the limb already hit full-on sepsis.

shellspeed1:
I mean, when a limb is dead like that it's supposed to be replaced, either with a mechanical limb or a fleshy one. Might be worth adding a fleshy torso replacement instead.

DeltaFire15:
@timothyteakettle bumping.

DeltaFire15:
Marking this as stale due to requested changes having been present for a considerable period of time with no author response.

DeltaFire15:
Closing this PR due to staleness. Feel free to reopen this once work on it is resumed.

PR #13718, opened 2020-11-30
podpeople now have green blood

ghost:
## About The Pull Request

Podpeople now have green blood.

## Why It's Good For The Game

More distinction between races.

## Changelog
:cl:
add: plant people bleed sap
/:cl:

PR #13720, opened 2020-11-30
podpeople have green blood but done correctly

timothyteakettle:
## About The Pull Request
gives podpeople green blood, does not make a new subtype for the blood (blood subtypes are really bad please stop making them!)
closes #13718 (it doesn't work and additionally makes a new blood subtype which means you won't even be able to replica pod podpeople)

## Why It's Good For The Game
more species flavor is good right

## Changelog
:cl:
tweak: podpeople blood colour changed to green
/:cl:


DeltaFire15 (approved):
Looking good to me.

PR #13728, opened 2020-12-03
adds 8 save slots 

silicons:
someone asked me to


timothyteakettle:
make it a config, also once that's done ill look into removing clutter by not showing un-used character slots, but instead using a 'new character' button that works to the same effect as clicking on an unused slot

PR #13731, opened 2020-12-03
[Testmerge?]Re-add Minerborg's Mining Points ID

CloverIsLucky:
## About The Pull Request

Re-adds the Minerborg's mining points card.

## Why It's Good For The Game

Removing the mining points ID from the minerborg's moduleset seems to be a mistake. There is no excuse to keep it removed from their moduleset.

## Changelog
:cl:
add: Minerborg points ID is back!
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -921,6 +921,7 @@
 		/obj/item/gps/cyborg,
 		/obj/item/weapon/gripper/mining,
 		/obj/item/cyborg_clamp,
+		/obj/item/card/id/miningborg,
 		/obj/item/stack/marker_beacon,
 		/obj/item/destTagger,
 		/obj/item/stack/packageWrap)


CloverIsLucky:
So this is my first time creating a pull request. I hope I did it well. I wrote "testmerge" in the title for that. If everything is okay, I hope this becomes a full merge.

MrJWhit:
Could you be more specific on why they should be reintroduced?

CloverIsLucky:
> 
> 
> Could you be more specific on why they should be reintroduced?

When we hit the lowpop, I hardly see anyone capable of claiming the mining points when miners, cargo techs or Heads aren't signed up yet. And minerborgs could become very useful again if making them able to purchase anything that a crew needs or wants from the mining vendor, specially, more survival pens or help purchasing premium items for example. This is why I made this PR. This PR isn't requesting re-adding the Premium KA module back, just the mining points ID.

Hatterhat:
if you readd mining borg point cards, consider adding an item that allows the points to be siphoned from their card (or any other card while we're at it) and added to another card, in this case a proper crewmember's

CloverIsLucky:
> 
> 
> if you readd mining borg point cards, consider adding an item that allows the points to be siphoned from their card (or any other card while we're at it) and added to another card, in this case a proper crewmember's

I have to check at that money syphoning machine code and the ID's code so I can see how to modify the borg mining points card to be siphoned.

DeltaFire15:
@silicons Opinion on this?

silicons:
@DeltaFire15 i'm struggling to see why they need one imo
what are they going to spend it on

silicons:
on the other hand there's also not much reason they don't need one so /shrug

DeltaFire15:
Thought about this PR a bunch and, thing is, mining borgs don't exactly have much use for points, save for buying five billion medipens to replace medbay, which isn't exactly something that great (plus if they get a module reset the card just gets deleted)
Soo, gonna close this. Though don't let this discourage you from making other PRs!

PR #13751, opened 2020-12-11
standardizes text formatting to rp server's |one character| instead of ||two|| because why did i even do the latter

silicons:
## Why It's Good For The Game

the latter is horrible

## Changelog
:cl:
tweak: text formatting now uses one character instead of two around the text to emphasize. 
/:cl:


lolman360:
you did it because the regex was simpler

buffyuwu:
you are a blessing on this earth

DeltaFire15 (approved):
Yeah ok

PR #13761, opened 2020-12-12
getting something_horrible'd now drops all items BEFORE you horribly die and turn into a pile of bones.

silicons:
drop items because full item deletion using this is kind-of silly.

:cl:
balance: dna melt drops all items being destroying you
/:cl:


DeltaFire15:
Pretty sure the way it is right now (instability being substracted) is correct, considering it is the negative of the current stability.
So, at stability 100, getting the not-as-bad events would have prob( 70 - (-100)) which'd be prob(170), while at -20 stability it'd have a prob of ( 70 - 20) = prob(50), so, penalizing negative stability.

silicons:
@DeltaFire15 OH IM STUPID WOOPS

DeltaFire15 (approved):

PR #13770, opened 2020-12-13
Adds a spontaneous guardian spirit event.

Putnam3145:
## About The Pull Request

1. Adds a spontaneous guardian spirit event. It can attach to *any* mob with a client, including silicons or sentience potions or whatever. It's rather rare--tied for lowest weight in the game--but becomes much more likely when it's already fired.
2. Gave guardian spirits an antagonist, for round-end display fun + dynamic tracking.

## Why It's Good For The Game

Injecting just a bit more weird into rounds is good, I think.

## Changelog
:cl:
add: Spontaneous guardian spirit formation
/:cl:


DeltaFire15:
I dislike having this as a normal random event, as giving random people guardians doesn't seem like the greatest idea.

If it was a magic event though? Yeah sure.

Putnam3145:
It's a terrible idea, which is kinda why I want it. Other idea was to give it to two people but make one of them some manner of antag, which would require more dynamic integration.

tiramisuapimancer:
can you make it so that with advanced neutral network gaming the game sufficiently tracks having a solo minion Master trigger event

BlueWildrose:
Guardian spirits will result in whoever dying with it getting reduced to bones, which will certainly result in a rather interesting predicament for the target of the guardian spirit involved. Unless that changes, this definitely has a benefit and a consequence.

The benefits being you have a guardian spirit that may or may not be powerful. The consequences is that it'll end up interrupting someone's ERP maybe, and that it'll dust your body when you die.

Hope the person behind that spirit wises up and gets a prescan, as well as a blood extraction for perhaps podcloning due to that.

This'll certainly make things interesting no doubt.

code/modules/mob/living/simple_animal/guardian/guardian.dm
```diff
@@ -183,7 +186,8 @@ GLOBAL_LIST_EMPTY(parasites) //all currently existing/living guardians
 			for(var/obj/item/W in summoner)
 				if(!summoner.dropItemToGround(W))
 					qdel(W)
-			summoner.dust()
+			if(dusts_master)
+				summoner.dust()
```
DeltaFire15, line 190: Might also wanna put the message for the dusting aswell as the part dropping all their items behind the dusts_master check.

DeltaFire15 (changes requested):
No dextrous stands for this event please.

LordSaladin:
random jojo event

Arturlang:
We need more wierd and cool stuff like this in general

PR #13772, opened 2020-12-13
adds a respawn/return to lobby system, tweaks cryo to preserve everything instead of just ""important items""

silicons:
## About The Pull Request

defaults are the following
- 5 minute delay if cryo'd
- 15 minute delay from observe/death
- can only respawn as non command/sec. can disable anything but assistant in config
- cannot respawn on the same slot or a previous character

- respawn bans exist

cryo now preserves everything and puts it into a neat little box for retrieval

## Why It's Good For The Game

respawn system as prior discussed with penalties/restrictions.

allows players to reenter the round

cryo deleting items was stupid as hell anyways

## Changelog
:cl:
add: players can now respawn/return to lobby as a ghost after a 15 minute (default) delay and rejoin on another character with some/many restrictions
add: cryo now preserves everything
/:cl:


code/modules/mob/living/death.dm
```diff
@@ -66,6 +66,10 @@
 	GLOB.alive_mob_list -= src
 	if(!gibbed)
 		GLOB.dead_mob_list += src
+	if(mind.key)
```
DeltaFire15:
```suggestion
	if(mind?.key)
```

ghost:
Does this actually enable respawn?

silicons:
added lower delay for cryoing out

@TheSpaghetti yes???

ghost:
> @TheSpaghetti yes???

Please disable it by default, people who die have to wait until the next round (if they don't get revived) **for a reason.**

Putnam3145:
> Please disable it by default, people who die have to wait until the next round if they don't get revived for a reason.

Which is? Consider the why, rather than just saying there is one. I'm all for keeping things the same if there's no reason to fix them, but there *is* some reason to fix it here, and I'm not sure what the rationale for this is, beyond "death should be punishing", which I'm not 100% sure is a compelling reason.

ghost:
> > Please disable it by default, people who die have to wait until the next round if they don't get revived for a reason.
> 
> Which is? Consider the why, rather than just saying there is one. I'm all for keeping things the same if there's no reason to fix them, but there _is_ some reason to fix it here, and I'm not sure what the rationale for this is, beyond "death should be punishing", which I'm not 100% sure is a compelling reason.

Metagaming, mostly.

silicons:
@TheSpaghetti disabling it by default is a moot point when i'm the one who's setting the configs unless it's for some kind of 4d chess for downstreams to not have it

this is something we're doing to lessen the impact of being reduced to a ghost. metagaming can be handled by admins where spotted.

silicons:
if the next main head admin thinks this is awful they can disable it

DeltaFire15 (approved):

PR #13774, opened 2020-12-13
Clown Car The Honkening

keronshb:
## About The Pull Request

Re-adds the clown car.  Bonks upon hitting an external airlock so no easy space kills.  Also ups the TC from 15 to 16.

## Why It's Good For The Game

More variety is good, this should make it less murderbony.

Special thanks to Augments and Saliith for their help!

## Changelog
:cl:
add: Re-adds the Clown Car to the clown uplink
balance: 15 >16 TC cost
balance: bonks on external airlocks
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items/uplink_roles.dm
+++ b/code/modules/uplink/uplink_items/uplink_roles.dm
@@ -60,7 +60,6 @@
 	cost = 15
 	restricted_roles = list("Clown")
 
-/*
 /datum/uplink_item/role_restricted/clowncar
 	name = "Clown Car"
 	desc = "The Clown Car is the ultimate transportation method for any worthy clown! \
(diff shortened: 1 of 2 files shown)


MrJWhit:
Honestly just set it to 20TC.
Also, yo changelogs doesn't actually say you added the clown car back.

keronshb:
> 
> 
> Honestly just set it to 20TC.
> Also, yo changelogs doesn't actually say you added the clown car back.

Kevin said 16, so I'll have it at 16 for now unless people feel like it's needed and maybe a poplock too.

Thanks for catching that, added.

ForrestWick:
iirc there were more issues than just the lack of bonking on airlocks

After experiencing it on Skyrat like 3 times, I will say that the fact this thing is an instant GBJ is retarded. Something should be done about that, some way for people to actually escape besides having to spam B and hope the clown car stops for like 30+ seconds

keronshb:
Ok the latest commit will actually prevent it from moving onto a space tile.

PR #13779, opened 2020-12-14
(probably) fixes all emotes being italicized 

silicons:
tin


Putnam3145:
This probably also makes subtle/subtler not italic, maybe; test it?

PR #13785, opened 2020-12-15
Makes grab bag ONLY make station antags; triples alien dynamic threat.

Putnam3145:
## About The Pull Request

1. Grab bag actually works as advertised, i.e. adding only on-station antags. 
2. Aliens were worth barely anything, as of now. During a round where aliens were damn near taking over it still tried to spawn a nightmare. That's unacceptable. Now they shouldn't coexist much at all.

## Why It's Good For The Game

1. Grab bag should work as advertised.
2. Seriously, xenos are way more threatening than the mode gives them credit for. 

## Changelog
:cl:
tweak: Grab bag works as advertised.
balance: Xeno threat in dynamic tripled.
/:cl:


BlueWildrose:
Someone should make it so alien queens can't ride ATVs and other vehicles while we're at it.

_(as funny as that was)_

PR #13800, opened 2020-12-16
Removes the Malf AI Doomsday module

BlueWildrose:
## About The Pull Request

This does one thing - removes the traitor AI's ability to doomsday.

Malf AI isn't being used as a gamemode anymore, and if anyone tries doing a malf AI gamemode after this they're probably gonna run into horrible things as a result.

## Why It's Good For The Game
Alright, here we go, time for me to get into discussion about this.

Some of you may have heard about, or been involved in a round where a few players were collaborating with at least two traitors, one firmly confirmed (the AI) and one not as much, though it did make the situation all the more suspicious - in making an engineering department micronation gimmick. When a few had discovered that the AI was indeed, malf/a traitor and announced it, the round proceeded to grow into a very chaotic state.

Throughout that round there was one particularly loud person in that micronation **who wasn't an EoC** advocating that the AI was fine constantly, constantly, constantly. Stating things like the AI wasn't going to hurt anyone, the AI and their cyborgs is part of their nation, delivering threats to anyone attacking anyone part of the nation, so forth.

End result of that round was that people treated the whole situation as an emergency and there was a whole lot of walls torn down, holes blown all around either by the AI blowing or roguifying machines up or the crew trying to make a path to the AI or into engineering. The one non-AI EoC died for good, someone revived the AI after someone had killed it although they didn't complete their objectives, so forth.

Afterwards, there was a debate in #ss13-main for a while about how traitor AIs should be treated. There was two admins having stances that contradicted one another as well, if I remember correctly. Some say people should not react to the AI in the way that the crew reacted that round, others delivered the opposite.

Here's my personal take on it:
**Traitor/Malf AIs in their current state are as threatening as bloodsuckers or xenos if ignored, and need to be terminated upon discovery.**
Much like bloodsuckers, the longer they stay alive, the more power they can accumulate. APCs hacked means more things they can do, such as reactivating cameras, blowing up machines, and ultimately disabling the crew's capability to stop their cyborg army through easier means - a cyborg army which they more than likely have because [malf AIs naturally have this ability to just magnetize ghosts to positronic brains simply by a ghost reading the AI's laws upon examining them](https://media.discordapp.net/attachments/647696379908456448/669327544016961536/unknown.png).

Consciously allowing a traitor AI who you know is a traitor AI to be alive is an extremely risky endeavor, because you do not know what the AI's objectives are and it's almost guaranteed they have to kill someone, as they can't receive stealing objectives. Killing the AI is currently the only way to stop them _securely_. Anyone _trusting_ such an AI is a _massive idiot_, **especially** since Rule 4 states that antagonists should be trying to do their objectives.

Even without the malf AI's abilities to destroy any means of breaking into its core like blowing up chemistry, they have the ability to turn the entire station into a major hazard simply by flooding plasma and shocking doors everywhere.

Additionally, since the AI counts as a "traitor" unlike the malf AI gamemode, this suggests that the AI can collaborate with other traitors in completing their objectives, which can only reinforce their capabilities. There was one round where someone was teleporting around with bluespace tomatoes while they were carrying a carded malf AI, who was **doomsdaying**. This combination made it near impossible to stop the doomsdaying AI.

So this leads to my PR here. Since there are some who feel that a **round-ending antagonist on the same level as xenos** should be given a chance, my suggestion is that we should tone down the malf AI to start. Starting off by stripping it of the ability to use the Doomsday device, which will end rounds if the crew is not prepared to deal with it in a timely fashion. This isn't the absolute solution to this problem and debate, but it's a start. It doesn't stop the AI from just flooding everything with plasma, it doesn't stop them from sending their cyborg army to murder everything, it doesn't stop them with their all-mighty access to the station of doing terrible things. It just takes away something that was meant for the Malf AI gamemode, which just carried over to traitor AI, and is the first step to address this dispute that players appear to have.

Perhaps there could be means later on about "deconverting" the AI of its antagonist status (which I believe someone else had the idea of) so that there are alternate methods besides killing such a powerful antagonist, and to allow the AI alternatives of "bad ends" similarly to their carbon counterparts, who have punishments ranging from permabrig, to pacification as well.

## Changelog
:cl:
del: Traitorous AIs can no longer doomsday.
/:cl:

### DIFF
--- a/code/game/objects/effects/countdown.dm
+++ b/code/game/objects/effects/countdown.dm
@@ -132,7 +132,7 @@
 	else if(T.cooldown)
 		var/seconds_left = max(0, (T.cooldown_timer - world.time) / 10)
 		return "[round(seconds_left)]"
-
+/*
 /obj/effect/countdown/doomsday
 	name = "doomsday countdown"
 
@@ -142,7 +142,7 @@
 		return
 	else if(DD.timing)
 		return "<div align='center' valign='middle' style='position:relative; top:0px; left:0px'>[DD.seconds_remaining()]</div>"
-
+*/
 /obj/effect/countdown/anomaly
 	name = "anomaly countdown"
 
--- a/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
+++ b/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
@@ -4,7 +4,7 @@
 GLOBAL_LIST_INIT(blacklisted_malf_machines, typecacheof(list(
 		/obj/machinery/field/containment,
 		/obj/machinery/power/supermatter_crystal,
-		/obj/machinery/doomsday_device,
+		// /obj/machinery/doomsday_device,
 		/obj/machinery/nuclearbomb,
 		/obj/machinery/nuclearbomb/selfdestruct,
 		/obj/machinery/nuclearbomb/syndicate,
@@ -233,7 +233,7 @@ GLOBAL_LIST_INIT(blacklisted_malf_machines, typecacheof(list(
 /datum/AI_Module/large //Big, powerful stuff that can only be used once.
 /datum/AI_Module/small //Weak, usually localized stuff with multiple uses.
 
-
+/* no fun allowed.
 //Doomsday Device: Starts the self-destruct timer. It can only be stopped by killing the AI completely.
 /datum/AI_Module/large/nuke_station
 	module_name = "Doomsday Device"
@@ -408,7 +408,7 @@ GLOBAL_LIST_INIT(blacklisted_malf_machines, typecacheof(list(
 		L.dust()
 	to_chat(world, "<B>The AI cleansed the station of life with the doomsday device!</B>")
 	SSticker.force_ending = 1
-
+*/
 
 //AI Turret Upgrade: Increases the health and damage of all turrets.
 /datum/AI_Module/large/upgrade_turrets
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -75,7 +75,7 @@
 	var/call_bot_cooldown = 0		//time of next call bot command
 	var/apc_override = FALSE		//hack for letting the AI use its APC even when visionless
 	var/nuking = FALSE
-	var/obj/machinery/doomsday_device/doomsday_device
+	//var/obj/machinery/doomsday_device/doomsday_device
 
 	var/mob/camera/aiEye/eyeobj
 	var/sprint = 10
--- a/code/modules/mob/living/silicon/ai/death.dm
+++ b/code/modules/mob/living/silicon/ai/death.dm
@@ -49,8 +49,9 @@
 		for(var/obj/item/pinpointer/nuke/P in GLOB.pinpointer_list)
 			P.switch_mode_to(TRACK_NUKE_DISK) //Party's over, back to work, everyone
 			P.alert = FALSE
-
+	/*
 	if(doomsday_device)
 		doomsday_device.timing = FALSE
 		SSshuttle.clearHostileEnvironment(doomsday_device)
 		qdel(doomsday_device)
+	*/


DeltaFire15 (changes requested):
This changes absolutely nothing about the destructive capabilities and threat potential of AIs as they quite literally have *control over the entire station.*
A being that has remote access to nearly everything will obviously, due to the inherent nature of this fact, always be a major threat, so uh, why take away their biggest ace / Plan-B ability, especially considering how easily a doomsdaying AI can be neutralized *in most scenarios.*
If carded AIs doomsdaying while being carried is an issue, which I can see being one, block doomsday from being used by carded AIs instead.

YakumoChen:
Agree with delta. Doomsday is the ace in the hole. A good malf can already decimate the station and this doesn't do anything to reduce the threat except needlessly remove an ability that is signature to the antagonist.

BlueWildrose:
Hope you're all aware of my points made about how there's players who don't think the AI should just immediately be killed as soon as they're discovered to be malf. This is part of why I've created the suggestion that the AI be toned down to begin with - it's an opening step towards what could be.

With that said, the fact that this was a malf AI ability that was basically the end goal of a malf AI, passed onto the traitor AI that doesn't really serve much purpose except to just end the round early, let me emphasize that "round ending" part.

I've seen traitor AIs who will intentionally try to end the round on purpose to simply end the round. They were not discovered at all - or if they were discovered, the message was not spread out widely enough. Whether or not they completed their objectives of killing a target, they would set it off. One out of two times, their target would manage to evacuate successfully if they do set off doomsday - it gives their target plenty of warning and they can take that opportunity to run off to lavaland, or run off to space. I've noted a few rounds like this where the AI would just run doomsday for the sake of doomsday. To say that parts of the playerbase is incapable of patience is an understatement, some even going as far as a blob spawning in to end a green shift early by spawning near the tesla, and just destroying the containment for the tesla.

And yes, the AI can just plasma flood as we're all aware and that could very well just get the shuttle called, but at least there's a chance for round continuity compared to the nuke. The malf AI has other means of escaping people who are out for their blood besides setting off a round ending timer, which lies mainly in their shunting ability, but if they're smart enough, they can keep their cyborgs alive from the kill button and use them to their advantage as well.

With that said, this PR isn't necessarily about balance, but about what the players of Citadel perceive what the AI should be, and whether or not a player should have that responsibility as a malf AI to begin with as it could be misused for less self-fulfilling situations in terms of their objective.

YakumoChen:
Player reaction to a malf really has nothing to do with Doomsday. If people are bum rushing the AI at the slightest sign that's more of a player issue and probably validhunting.

Remember it's a roleplaying game and not everyone is required to treat a malf AI like Blob.

BlueWildrose:
Justified bumrushing, considering not even bloodsuckers would immediately have access to such station-destructive power that the AI has. Until they manage to sire the right people for specific jobs or they gain access themselves, they wouldn't have instant access to plasma flooding for instance. Sure, it's true that the AI likely will just stay quiet if nobody calls them out, but there's also everything I mentioned above - doomsdaying 4noraisin.

And I never suggested people _had_ to go and kill something. Even in blob events people will take the opportunity to fly off into space or lavaland. Or just cryo. But if anyone is consciously aware of what the AI could do, they probably would want to do something about it in one way or the other.

To put it simply:
• A normal traitor, or changeling goes loud: Usually only those confronting or getting in the way of the antagonist is at risk.
• The traitor AI goes loud: Everyone is very much at risk.

YakumoChen:
We also have general rules against murderbone and round ending now which means the AI has to greatly justify activating doomsday. Arguably people bumrushing the AI to valid it only forces its hand.

BlueWildrose:
There actually isn't any rules about **round-ending specifically** as of currently. Explicitly, there isn't any specific policy for traitor AIs, just for antagonists in general. The spirit of the murderboning rule otherwise is to prevent people from being locked out of the round for an hour after their sensors-stripped corpse is just left to rot in maintenance for no reason. Doomsdaying will end the round regardless of what happens to the crew. Admins will have ambiguous interpretations of how they feel about activating doomsday as a result, and thus there isn't really a cemented area of how admins should really be approaching that situation, which is why many malf AIs before have gotten away with it.

Which is admittedly a problem with the rules moreso than anything, but I reckon someone is hopefully double checking that as of the existence of this PR. It's either that changes and is cemented in the rules listing, and admins are willing to enforce it, or just remove the doomsday entirely.

ForrestWick:
Leaving the confused emoji as my way of saying I have no strong feelings one way or the other regarding this idea, thought I will say it should give malf some other alterations, both in the form of new powers and some other changes regarding balance to make it less god mode (literally)

shellspeed1:
It might be better to instead of completely removing it turn doomsday into a power unlocked by having the equivalent of a hijack objective. Just my two cents.

silicons:
@YakumoChen the mere existence of a malf ai is an probable existential threat to the station

validhunting or no validhunting a solution needs to be found for this, likely by making it so taking out the ai doesn't always equate needing to keep it dead or out of the round. arguably it's barely even validhunting for people to want to neutralize a malf ai. you can tracking implant and take a traitor's gear, you can kill a ling if it goes rogue, but by the time you find out an ai isn't as nice as it claims your round is potentially already over.

the reason i do not want the doomsday module removed is because the ai itself having its remote access to the station's systems is alone a threat, doomsday or no doomsday

a malf ai can force a round end in 10 minutes or at the very least destabilize the station to the point of the brink of a shuttle. it doesn't need a doomsday to do so, and "playing nice" is not an option for the crew 95% of the time, as at that point you're just waiting for it to gather power to kill you or take over the station.

DeltaFire15:
Current maintainer consensus is to close this PR, considering the issue of AI previously having been able to doomsday while carded is now resolved, aswell as aforementioned reasons.

PR #13803, opened 2020-12-17
Prevents traitor AI from doomsdaying while carded

BlueWildrose:
## About The Pull Request

This prevents the malf AI from activating the doomsday device while they're inside a card.

There's code that actually deactivates doomsday if the traitor AI gets carded while doomsday is happening anyways, which implies that this was supposed to be intentional to begin with.

## Why It's Good For The Game

Please see #13800.

I created this PR in case the verdict is that the change that traitor AIs need to have doomsday removed is declined, as mentioned in #13800. If doomsday doesn't get removed, at least make it so the malf AI can't collaborate with funny teleporting bluespace tomato traitors to make it near impossible for everyone to kill them in the limited time they have.

## Changelog
:cl:
balance: The traitor AI can no longer activate the doomsday device while carded.
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
+++ b/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
@@ -262,7 +262,7 @@ GLOBAL_LIST_INIT(blacklisted_malf_machines, typecacheof(list(
 	if (owner_AI.shunted)
 		return //prevent AI from activating doomsday while shunted.
 	if (istype(owner.loc, /obj/item/aicard))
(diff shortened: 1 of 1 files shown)


DeltaFire15 (changes requested):
I recommend also checking for them being in a card between line 331 and 332, as with this code they could start the sequence, get carded before it actually enables the nuke, and it'd still work while in the card.
Good PR otherwise.

BlueWildrose:
yup, that would definitely be overlooked. Fixed.

DeltaFire15 (approved):

PR #13809, opened 2020-12-18
Makes gear harness cover nothing

Putnam3145:
## About The Pull Request

I did a PR for this earlier not realizing that the default is actually MORE covering than what it had. This is what I meant. I don't expect people to misbehave too badly when it comes to this.

## Why It's Good For The Game

My immersion!

## Changelog
:cl:
tweak: Gear harness now covers nothing.
/:cl:

### DIFF
--- a/code/modules/clothing/under/miscellaneous.dm
+++ b/code/modules/clothing/under/miscellaneous.dm
@@ -168,8 +168,15 @@
 	desc = "A simple, inconspicuous harness replacement for a jumpsuit."
 	icon_state = "gear_harness"
 	item_state = "gear_harness"
-	can_adjust = FALSE
-	body_parts_covered = NONE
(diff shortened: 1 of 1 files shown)


BlueWildrose:
Besides exposing genitals, what does this change?

I think it would be interesting to make the gear harness adjustable and have the alt-clicked variant be exposing genitals, and the non-altclicked variant as it was before.

R3dtail:
+1 to Bluewildrose's idea.  all the other equipment is alt+toggleable to expose, this shouldn't be much different

Hatterhat:
this might also affect wound protection (or a lack thereof) but if you're wearing a gear harness you're probably not expecting to get into fights

Putnam3145:
Done.

PR #13811, opened 2020-12-18
Prevents borg hypos from injecting if it would OD the target while on help intent

BlackMajor:
Figured I'd implement some form of accidental harm prevention much like we did to borg surgery.
As the name implies, prevents cyborg hyposprays from injecting if it means overdosing the target, unless they're on a non-help intent.

Edit: Also renames borghydro.dm to borghypo.dm

:cl:
tweak: Cyborg hypospray no longer injects if it means OD'ing while on help intent.
/:cl:


DeltaFire15:
Uhhhh seems kinda hugboxy imo.. Aswell as taking away any and all ambigusity from rogue borgs intentionally ODing people.
Maybe make it a borg hypo upgrade instead?

BlueWildrose:
Oh don't worry, borgs can still mess up by injecting slimepeople with anti-toxin.

(although they can just OD the slime on anti-toxin to undo their mistake, because anti-tox OD just deals toxin damage)

R3dtail:
Hugboxy yes but not all forms of hugbox are bad.

PR #13814, opened 2020-12-19
Fixes twin sword sheaths icon/item state from being blank sprites

Xantholne:
Gives item state and icon state so they can actually be clicked on in inventory and seen on mobs

Uses the base katana sheath sprite when worn since well, that's literally what goes in it so of course you'd see one when worn.

Also adds missing sheathing/unsheathing sound that all of the other rapier and sabre sheaths have since it didn't make sense that they would but this would not

:cl:
fix: Twin Sword Sheaths have an equipment icon and icon when worn now and make a sound when sheathed/unsheathed
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -869,37 +869,37 @@
 	var/obj/item/storage/belt/sabre/twin/B = S.parent
 	if(istype(B))
 		playsound(B, 'sound/items/unsheath.ogg', 25, 1)
-	..()
+	. = ..()
 
 /obj/item/melee/smith/twohand/katana/on_enter_storage(datum/component/storage/S)
 	var/obj/item/storage/belt/sabre/twin/B = S.parent
 	if(istype(B))
 		playsound(B, 'sound/items/sheath.ogg', 25, 1)
-	..()
+	. = ..()
 
 /obj/item/melee/smith/wakizashi/on_exit_storage(datum/component/storage/S)
 	var/obj/item/storage/belt/sabre/twin/B = S.parent
 	if(istype(B))
 		playsound(B, 'sound/items/unsheath.ogg', 25, 1)
-	..()
+	. = ..()
 
 /obj/item/melee/smith/wakizashi/on_enter_storage(datum/component/storage/S)
 	var/obj/item/storage/belt/sabre/twin/B = S.parent
 	if(istype(B))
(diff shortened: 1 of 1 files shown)


code/game/objects/items/storage/belt.dm
```diff
@@ -865,6 +865,42 @@
...
+	var/obj/item/storage/belt/sabre/twin/B = S.parent
+	if(istype(B))
+		playsound(B, 'sound/items/unsheath.ogg', 25, 1)
+	..()
+
+/obj/item/melee/bokken/on_enter_storage(datum/component/storage/S)
+	var/obj/item/storage/belt/sabre/twin/B = S.parent
+	if(istype(B))
+		playsound(B, 'sound/items/sheath.ogg', 25, 1)
+	..()
```
DeltaFire15:
```suggestion
	. = ..()

/obj/item/melee/smith/twohand/katana/on_enter_storage(datum/component/storage/S)
	var/obj/item/storage/belt/sabre/twin/B = S.parent
	if(istype(B))
		playsound(B, 'sound/items/sheath.ogg', 25, 1)
	. = ..()

/obj/item/melee/smith/wakizashi/on_exit_storage(datum/component/storage/S)
	var/obj/item/storage/belt/sabre/twin/B = S.parent
	if(istype(B))
		playsound(B, 'sound/items/unsheath.ogg', 25, 1)
	. = ..()

/obj/item/melee/smith/wakizashi/on_enter_storage(datum/component/storage/S)
	var/obj/item/storage/belt/sabre/twin/B = S.parent
	if(istype(B))
		playsound(B, 'sound/items/sheath.ogg', 25, 1)
	. = ..()

/obj/item/melee/bokken/on_exit_storage(datum/component/storage/S)
	var/obj/item/storage/belt/sabre/twin/B = S.parent
	if(istype(B))
		playsound(B, 'sound/items/unsheath.ogg', 25, 1)
	. = ..()

/obj/item/melee/bokken/on_enter_storage(datum/component/storage/S)
	var/obj/item/storage/belt/sabre/twin/B = S.parent
	if(istype(B))
		playsound(B, 'sound/items/sheath.ogg', 25, 1)
	. = ..()
```

DeltaFire15 (changes requested):
Just for safety, use . = ..() instead of ..() in this case, or alternatively: return ..()

code/game/objects/items/storage/belt.dm
Xantholne: Added

PR #13815, opened 2020-12-19
Removes slime blood to jelly conversion

YakumoChen:
## About The Pull Request
Removes the slime blood + toxin chem reaction

## Why It's Good For The Game

Chemistry has a lot of very overpowered chems that require slime jelly to make. Slime jelly, being originally gated by Xenobio getting done, is no longer difficult, or even inconvenient to make.... as long as you're playing a slimeperson.

Slime jelly itself is an extremely deadly toxin and probably shouldn't be trivially obtainable by needling your resident googirl.

I'm not quite sure what the reason for having the reaction in the first place was, considering slime jelly was removed from slimepeople blood for what I imagine to be the same reason.

## Changelog
:cl: Yakumo Chen
balance: Slime Jelly is no longer obtainable from slimepeople. Go ask Xenobio
/:cl:


lolman360:
finally

timothyteakettle:
i was going to do this myself but it seems you beat me to it, the alternative is to just nerf all the chems because they're stupid and don't have overdoses

DeltaFire15 (approved):
As much as I dislike this change as a player, this seems like a good change from a balance perspective due to how effective jelly-based chemicals are.

PR #13817, opened 2020-12-19
Adds a antag species blacklist and applies it to species it makes sense to have (#3620)

Arturlang:
## About The Pull Request
Blocks nonorganic and buggy races from being changelings and bloodsuckers, and vampires too, since double vampires do not make sense at all

## Changelog
:cl:
tweak: Nonorganic species cant be bloodsuckers/changeling, and a few more that dont make sense or work well.
/:cl:

### DIFF
--- a/code/game/gamemodes/bloodsucker/bloodsucker.dm
+++ b/code/game/gamemodes/bloodsucker/bloodsucker.dm
@@ -32,8 +32,7 @@
 	reroll_friendly = FALSE
 	enemy_minimum_age = 7
 	round_ends_with_antag_death = FALSE
-
-
+	restricted_species = list("plasmaman", "shadow", "skeleton", "synthliz", "synth", "vampire", "ipc", "dullahan", "android")
 	announce_span = "danger"
 	announce_text = "Filthy, bloodsucking vampires are crawling around disguised as crewmembers!\n\
 	<span class='danger'>Bloodsuckers</span>: The crew are cattle, while you are both shepherd and slaughterhouse.\n\
--- a/code/game/gamemodes/bloodsucker/bloodsucker.dm
+++ b/code/game/gamemodes/bloodsucker/bloodsucker.dm
@@ -26,13 +26,13 @@
 	false_report_weight = 1
 	restricted_jobs = list("AI","Cyborg")
 	protected_jobs = list("Chaplain", "Security Officer", "Warden", "Detective", "Head of Security", "Captain", "Head of Personnel", "Chief Engineer", "Chief Medical Officer", "Research Director", "Quartermaster")
+	restricted_species = list(SPECIES_SYNTH_LIZARD, SPECIES_SYNTH, SPECIES_ANDROID, SPECIES_DULLAHAN, SPECIES_PLASMAMAN, SPECIES_SKELETON, SPECIES_VAMPIRE, SPECIES_VAMPIRE_WEAK)
 	required_players = 20
 	required_enemies = 2
 	recommended_enemies = 4
 	reroll_friendly = FALSE
 	enemy_minimum_age = 7
 	round_ends_with_antag_death = FALSE
-	restricted_species = list("plasmaman", "shadow", "skeleton", "synthliz", "synth", "vampire", "ipc", "dullahan", "android")
 	announce_span = "danger"
 	announce_text = "Filthy, bloodsucking vampires are crawling around disguised as crewmembers!\n\
 	<span class='danger'>Bloodsuckers</span>: The crew are cattle, while you are both shepherd and slaughterhouse.\n\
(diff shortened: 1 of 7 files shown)


R3dtail:
IPCs: Bloodsucking module, hypnotic modules, and the ling worm could be using the IPC's body like a hermit crab would. 
Bugs: Uh, plenty of bugs suck blood, there's not much of a reason why they can't be a changeling either
Double vampires is just king vampire if we're following checker rules.

Species backlist is...pretty meh.  Then it becomes "Oh there's no way they can be X thing" .  Which is a very lame metashield.  Especially when IPCs and bugs are popular

MrJWhit:
Yeah, I'm not sure why it's seemingly arbitrarily applied, you can be creative for being a bloodsucker.

timothyteakettle:
it outright does not work with certain species, which is why the blacklist is being applied

R3dtail:
If it's due to bugs , it sounds like those bugs should be fixed  and/or given priority to fixing  since this is , more or less, a complete bandaid solution.   i get why it exists, but i've not seen these species being lings and bloodsuckers be abused in any significant way.  From what I was told it's because an IPC seen processing blood is a tell, well, that's part of the many many tells bloodsuckers have, very high/low blood, appearing dead on monitors unless they use an ability, among others.  Bloodsuckers are VERY obvious once you know what you're looking for.  Changelings  are, also , pretty obvious unless they specifically are going for as stealthy as possible.  Granted I'm not a amaintainer or much of a cooder so what do i know.

Arturlang:
It's not exactly a bandaid, the biggest reason this was done is because of full body prosthetics, which are IPCs, are not healed by bloodsucker regeneration, nor does it make sense to for changeling revives to do so, if we want to make it properly work, we would have to add a lot of special cases, maybe even make whole NEW variants of both antags JUST for IPC's. And for vampires, the reason they were added to the list, because double vampires are painfull to play as, they get DOUBLE blood drain, and yes that could be fixed too, but it all requires a lot of fixes for each race, sometimes in deeper parts of the code that might break a lot of other things. Dullahans for example were added to the bloodsucker list due to not having proper mouths on their "body", and if i even wanted to allow them to be bloodsuckers, I would have to figure out SOMEHOW to let the item head suck blood? That would require a whole new set of features just for that, something that is HIGHLY unlikely and only likely to happen in a very small time of the year, yes i probably might do it eventually, but this feature is to deal with the biggest issues right now, maybe for good if nobody is willing to fix the other ones, and for IPCs, them having blood even caueses wounds relating bugs, for example them bleeding forever. And plasmamaen can't really suck blood properly, as they will start to fucking burn without helmets, and it would make no sense either for them to be able to do it with, nor do they normally have blood either. These species all have good reasons to not be bloodsuckers, And will require SIGNIFICANT workarounds for them to properly work.

deathride58:
Species should **ONLY** be restricted from specific antag types if the species mechanics are inherently incompatible with the antag's mechanics (for example, dullahans and plasmamen for bloodsuckers, since they are mechanically incapable of biting). There is no valid reason at all to restrict other species from them (such as restricting skeletons from vampires, or IPCs from changelings), as the only thing it will do, is open the door to obnoxious levels of metagaming as a result of the inherent trust that those species would gain from the antag blacklist (lings aboard and no sign of traitors? With this PR in it's current state, ipcs and synths automatically become infinitely more trustworthy than both security and command combined. This in turn leads to validhunters having more reason to play IPCs, as performing actions that would otherwise make them extremely suspicious when lings are aboard, suddenly become perfectly acceptable due to their inherent trust, which leads to an environment that goes directly against the game's central theme of paranoia; it'll more or less turn ling rounds into straight-up TDM), while also locking out players of those species from playing those antag types.


>"we might even have to make whole new variants just for ipcs!!!"

lmao, no. You don't. If you want 'em to make sense, literally all you have to do is make the flavortext change depending on whether or not the user started the round as a synthetic species (changing organic-themed ling messages to focus more around nanites in the case of synth lings, for instance), and leave the mechanics as they are (while also fixing oversights that would otherwise prevent those mechanics from working as intended in those species' contexts). You don't need "SIGNIFICANT WORKAROUNDS!!!!!!" at all to make 'em work.

DeltaFire15:
Yeeah, while it literally makes zero sense from a logical perspective to have actual robot changelings / fleshmend repair robotic limbs, from a balance-perspective, this is absolutely required.
I'll probably make a PR soonish that allows fleshmend / bloodsucker healing to affect synthetic (aka hybrid, not fully robotic due to e.g. augmentations) bodyparts.

DeltaFire15:
After that PR is merged, robots don't need to be blacklisted from either.

Arturlang:
@DeltaFire15 Removed robots from the blacklist

silicons:
i'd like an admin message to be fired when someone is denied an antag rol ldue to this, and for them to be notified, and obviously logged
furthermore, please add an auto species swap (can just do list associations if you need it) as if a vampire rolls bloodsucker or a dullahan rolls ling and so on and so forth imo they should just revert to their normal species. it can result in metagame but it's better than hard blacklists. 
@Arturlang

Arturlang:
Will do

Arturlang:
Gonna work on this a bit more

PR #13818, opened 2020-12-19
Further gates x-ray and thermal implants in science; rebalances the x-ray laser

ghost:
This PR aims to further gate thermal and x-ray implants behind illegal tech and its own expensive node, as well as tweaking the x-ray laser gun to be much more fair to fight against while still being somewhat reliable for the darker situations on the station

Making thermals and x-ray take drastically longer to acquire is good for the game since this means antagonists can feel a bit more safe actually doing their duties or even existing on station without half the crew finding them before they can even begin. It should also help set up dastardly ambushes now that the chances of being seen from a screen length away by Billy Bob the assistant are lower due to said time gate. The current cost I think isn't too out there considering exotic ammunition is 25000. Given that these benefit all of the crew and are readily printed from medical/robotics, the extra cost seems quite worth it. Though the upgrade path might need a little work

The change to x-ray lasers likewise provides a more balanced approach to combat and antagonists. In exchange for not being able to fire through walls and giving your target little to no options for counterplay, the x-ray laser is now a side grade to the laser gun. While up front it deals 3 less damage, it has an armor pierce value of 20 _and_ the ability to apply 150 rads per shot

## Changelog
:cl:
add: Added new node to unlock thermal and x-ray implants for science
del: Removed thermals and x-ray from the combat cybernetics node to its own node
tweak: Tweaked the x-ray laser to have armour piercing, instead of wall piercing
balance: Balanced the x-ray laser to be a bit more fair to fight against; slightly more rads and damage, but no wall pierce
/:cl:

### DIFF
--- a/code/__HELPERS/shell.dm
+++ b/code/__HELPERS/shell.dm
@@ -13,6 +13,7 @@
 	var/err_file = ""
 	var/static/list/interpreters = list("[MS_WINDOWS]" = "cmd /c", "[UNIX]" = "sh -c")
 	var/interpreter = interpreters["[world.system_type]"]
+	log_subsystem("SHELL", "Executing command [command]")
 	if(interpreter)
 		for(var/seo_id in shelleo_ids)
 			if(!shelleo_ids[seo_id])
@@ -36,6 +37,7 @@
 	else
 		CRASH("Operating System: [world.system_type] not supported") // If you encounter this error, you are encouraged to update this proc with support for the new operating system
 	. = list(errorcode, stdout, stderr)
+	log_subsystem("SHELL", "Finished execution: [errorcode], [stdout], [stderr]")
 #undef SHELLEO_NAME
 #undef SHELLEO_ERR
 #undef SHELLEO_OUT
--- a/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
+++ b/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
@@ -261,6 +261,8 @@ GLOBAL_LIST_INIT(blacklisted_malf_machines, typecacheof(list(
 		return //prevent the AI from activating an already active doomsday
 	if (owner_AI.shunted)
 		return //prevent AI from activating doomsday while shunted.
+	if (istype(owner.loc, /obj/item/aicard))
+		return //prevent AI from activating doomsday while carded. If the AI gets carded after doomsdaying, there's already code to stop it then.
 	active = TRUE
 	set_us_up_the_bomb(owner)
 
@@ -326,6 +328,10 @@ GLOBAL_LIST_INIT(blacklisted_malf_machines, typecacheof(list(
 	owner.playsound_local(owner, 'sound/misc/server-ready.ogg', 50, 0)
 	sleep(30)
 	if(QDELETED(owner) || owner.stat == DEAD)
+		return
+	if(istype(owner.loc, /obj/item/aicard))
+		to_chat(owner, "<span class='boldnotice'>Error: Signal transmission failed. Reason: Lost connection to network.</span>")
+		to_chat(owner, "<span class='warning'>You can't activate the doomsday device while inside an intelliCard!</span>")
 		return
 	priority_announce("Hostile runtimes detected in all station systems, please deactivate your AI to prevent possible damage to its morality core.", "Anomaly Alert", "aimalf")
 	set_security_level("delta")
--- a/code/modules/atmospherics/machinery/datum_pipeline.dm
+++ b/code/modules/atmospherics/machinery/datum_pipeline.dm
@@ -206,7 +206,7 @@
 	. = other_airs + air
 	if(null in .)
 		stack_trace("[src]([REF(src)]) has one or more null gas mixtures, which may cause bugs. Null mixtures will not be considered in reconcile_air().")
-		return listclearnulls(.)
+		listclearnulls(.)
 
 /datum/pipeline/proc/empty()
 	for(var/datum/gas_mixture/GM in get_all_connected_airs())
--- a/code/modules/projectiles/ammunition/energy/laser.dm
+++ b/code/modules/projectiles/ammunition/energy/laser.dm
@@ -77,7 +77,7 @@
 
 /obj/item/ammo_casing/energy/xray
 	projectile_type = /obj/item/projectile/beam/xray
-	e_cost = 50
+	e_cost = 100
 	fire_sound = 'sound/weapons/laser3.ogg'
 
 /obj/item/ammo_casing/energy/mindflayer
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -69,10 +69,11 @@
 /obj/item/projectile/beam/xray
 	name = "\improper X-ray beam"
 	icon_state = "xray"
-	damage = 15
-	irradiate = 300
+	damage = 17
+	armour_penetration = 20
+	irradiate = 150
 	range = 15
-	pass_flags = PASSTABLE | PASSGLASS | PASSGRILLE | PASSCLOSEDTURF
+	pass_flags = PASSTABLE | PASSGLASS | PASSGRILLE
 
 	impact_effect_type = /obj/effect/temp_visual/impact_effect/green_laser
 	light_color = LIGHT_COLOR_GREEN
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -74,7 +74,6 @@
 	irradiate = 150
 	range = 15
 	pass_flags = PASSTABLE | PASSGLASS | PASSGRILLE
-
 	impact_effect_type = /obj/effect/temp_visual/impact_effect/green_laser
 	light_color = LIGHT_COLOR_GREEN
 	tracer_type = /obj/effect/projectile/tracer/xray
--- a/code/modules/research/techweb/nodes/medical_nodes.dm
+++ b/code/modules/research/techweb/nodes/medical_nodes.dm
@@ -104,8 +104,16 @@
 	display_name = "Combat Cybernetic Implants"
 	description = "Military grade combat implants to improve performance."
 	prereq_ids = list("adv_cyber_implants","weaponry","NVGtech","high_efficiency")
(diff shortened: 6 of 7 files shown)


YakumoChen:
Kind of a dumb PR given xray laser was literally just nerfed, feels knee jerk since it wasn't able to be "nerfed enough" I feel.

Robotic eyes are already very niche and this just makes them go from "very rarely used" to "never". They already possess significant downsides like EMP vulnerability and the fact surgery requires 2 people and high cost to produce.

This also wont lower the availability if thermals considering how common it is to see it from genetics, making the point moot.

ghost:
> 
> 
> Kind of a dumb PR given xray laser was literally just nerfed, feels knee jerk since it wasn't able to be "nerfed enough" I feel.
> 
> Robotic eyes are already very niche and this just makes them go from "very rarely used" to "never". They already possess significant downsides like EMP vulnerability and the fact surgery requires 2 people and high cost to produce.
> 
> This also wont lower the availability if thermals considering how common it is to see it from genetics, making the point moot.

In most cases this would be a buff to x-ray. In exchange for losing the wall piercing you gain more damage, more rads and more damage compared to a normal laser when faced with armor. For example, given 60% resistance a normal laser is going to do 8 damage but this, despite doing 17 normally, has a 20% pierce lowering the protection value to 40% and thus it dealing 10.2. This gives it a much better defined niche as something used to deal with organics that have armor against that weapon type while also providing a bit more of balanced gameplay for antags that simply cannot retaliate from attacks behind 1+ walls

The gating of the implants serves as one part as thermals will be PR'd to be replaced by I think a lesser version of devil sight? I believe it was a +2 to view range, instead of being able to see beings through walls. So this PR combined with a future one coming up will mean that thermals and x-ray will be much more rare in general, but still obtainable if there's enough dedication

YakumoChen:
Disagree. It's definitely not a buff, wall pierce is the wholesale reason you'd print an xray laser. I dont really feel your version sets it apart significantly enough anyway, as everything else in the lathe is pretty unique.
In fact I'd be leaning to argue one should always assume someone worth printing a laser for has armor anyhow. It leans toward straight upgrade.

YakumoChen:
I'd agree with removing the genetics powers but I'd argue that implants already have very steep costs and hiding them behind illegal tech is analogous to removing them more or less completely.

ghost:
> 
> 
> Disagree. It's definitely not a buff, wall pierce is the wholesale reason you'd print an xray laser. I dont really feel your version sets it apart significantly enough anyway, as everything else in the lathe is pretty unique.
> In fact I'd be leaning to argue one should always assume someone worth printing a laser for has armor anyhow. It leans toward straight upgrade.

Armor piercing itself is a reason to print it when you're dealing with someone that has a high amount of armor, which often isn't the case. Most antags you face either have little or negative resistances to the damage type, meaning you're better off using a stock laser that also had no cost associated with printing or acquiring. As far as I know, there's not really any other laser that crew can get their hands on with armor piercing _or_ rads attached to it. Unless the decloner does genetics and rads

> 
> 
> I'd agree with removing the genetics powers but I'd argue that implants already have very steep costs and hiding them behind illegal tech is analogous to removing them more or less completely.

Implants have far too few costs for how extremely powerful they are. For how game changing the ballistics can be (357) it's rightfully locked behind steep and hard costs. So it would follow that something that can provide vast benefit to all crew, while being printed in two different department would likewise, incur an equal or higher cost. In this case, due to the availability of it once researched, it would incur a higher cost than being able to print exotic ammo from the seclathe

YakumoChen:
> pre nerf thermal/x-ray locked behind weapons tech that has a significant and aggressive tech points costs, unlikely to be researched in the first hour without rushing 
>very high mineral cost, including diamonds for xray
>requires a second, trustworthy player willing to install it on you, requires fair time expenditure that leaves you completely helpless
>leaves you cripplingly vulnerable to EMP

There's a reason you almost never see implant work being done on people except arguably to meme / the roboticist is very bored. I'd like to see how all of this translates to "far too few costs" yet you want to lock it even further behind tech that is not likely to ever see completion except in all but the slowest calmest rounds.

Illegal ammo, the example you used, is a weapons tech that sees almost zero use traditionally in game. The guns supporting them are incredibly rare or antag only, meaning antags are likely the only ones to benefit, and most of that ammo is available in the auto lathe regardless. It provides near zero benefit to crew. The only illegal techs that I ever see used are the energy crossbow and the blue space item that let's you vanish from reality and even those are pretty heavily niche.

YakumoChen:
IArmor piercing itself is a reason to print it when you're dealing with someone that has a high amount of armor, which often isn't the case. Most antags you face either have little or negative resistances to the damage type, meaning you're better off using a stock laser that also had no cost associated with printing or acquiring. As far as I know, there's not really any other laser that crew can get their hands on with armor piercing or rads attached to it. Unless the decloner does genetics and rads

The only antags I think of that live up to this are blob and xenos. And those benefit better from mass arming, you'd be right. Most other dangerous antags worth gunning down have armor or are likely to acquire it and for those I see no reason to ever want to print anything but an armor piercing gun. Worse to worst, you down them in only one more shot anyway or cripple them via rads. Outside of the two use cases that see mass cargo ordering you have no reason to not use the xray.

DeltaFire15 (changes requested):
Imo xray lasers are fine as they are for now after they just got nerfed (half capacity, less rads).

I'm in on the locking xray eyes behind illegal tech but 
a) Please leave thermals in the normal combat implant tech, considering they have enough of a tradeoff for their effect (needs someone to actually implant it, aswell as severe susceptibility to EMPs which are generally pretty easy to acquire.), whereas xray eyes are full-on xray after all.
If you really want to nerf thermal eyes anyways,  make thermals a active action that only works for some time with a comperatively long cooldown (like, one minute or two or something), and maaybe make their eyes glow while it's active / have them have something that indicates that person does have thermals.

b) The price for this new node is too high, it already needs a bunch of points to get illegals plus combat implants plus unregulated bluespace, aswell as having access to illegal tech unlocking things  in the first place. Please reduce the tech point price, by alot.

ghost:
> 
> 
> Imo xray lasers are fine as they are for now after they just got nerfed (half capacity, less rads).
> 
> I'm in on the locking xray eyes behind illegal tech but
> a) Please leave thermals in the normal combat implant tech, considering they have enough of a tradeoff for their effect (needs someone to actually implant it, aswell as severe susceptibility to EMPs which are generally pretty easy to acquire.), whereas xray eyes are full-on xray after all.
> If you really want to nerf thermal eyes anyways, make thermals a active action that only works for some time with a comperatively long cooldown (like, one minute or two or something), and maaybe make their eyes glow while it's active / have them have something that indicates that person does have thermals.
> 
> b) The price for this new node is too high, it already needs a bunch of points to get illegals plus combat implants plus unregulated bluespace, aswell as having access to illegal tech unlocking things in the first place. Please reduce the tech point price, by alot.

X-ray lasers are really not fine. They got nerfed because they needed it badly, but it was only a bandaid considering it was a 20 shot wall piercer that did 300 rads/shot. The PR never touched their core issue, which _is_ being able to pierce walls. That's frankly absolutely insane to be able to do as it offers no way to defend or fight back against it. That's not to mention you can print many of them once researched and fit them in your bag. There's also being able to dual-wield for that juicy double shot.

The issue of needing someone to implant them is a minor one at best. Especially considering you can always ask a head which is known to rarely if ever be converted as an antag, nor is it much of a time gate itself since the procedure is akin to experimental organ surgery so, while it's not on the level of mass printing, you can do the surgery within a minute on a person which means security can easily gear up and snuff out any attempts at stealth or, in the case of some antags, being able to hide their existence until it's easily noticeable. 
EMP itself is a vulnerability, but that's also true for the person using it too. More often than not you can easily scrounge it up, but by the time you've realized you need one, you're often already exposed due to the ease of getting these implants. What's more, there's a few antags that either have no way to actually EMP or EMPing would cause massive damage/blindness to themself due to flashes burning out or doors becoming shocked.

At best if it's an active it'd need a very short duration with a very long cooldown if it were to stay around at combat implants due to how quickly a game changing implant can be researched

I can lower the cost of the illegal node though

DeltaFire15:
I stand by what I said in this case.
Xray lasers: should be fine for now, especially when options to see through walls become more rare.
Thermal eyes: Don't fully lock them, give them a active ability instead that only allows their usage for short bursts of time instead of always on.
Xray eyes: Sure, lock them behind a tech like you did, though with lower tech cost than right now.

ghost:
> 
> 
> I stand by what I said in this case.
> Xray lasers: should be fine for now, especially when options to see through walls become more rare.
> Thermal eyes: Don't fully lock them, give them a active ability instead that only allows their usage for short bursts of time instead of always on.
> Xray eyes: Sure, lock them behind a tech like you did, though with lower tech cost than right now.

Then at the very least, what about gating the lasers behind illegals and pushing thermals into its own slot above combat? Not requiring illegals, but still requiring at least a bit _more_ effort for them alongside making them an active ability. That way they're still accessible but they're not easily gotten alongside x-ray lasers, while the lasers themselves require a lot more work, but will at least remain the same.

And how does 7500 sound for the x-ray vision node?

YakumoChen:
Sometimes I wish maintainers would just close PRs when OP keeps trying to weasel more nerfs into something when the guy who merges PRs tells you No repeatedly.

OP is just trying to make a second attempt to kill Xray lasers because the first PR doing so didn't work.

DeltaFire15:
> > I stand by what I said in this case.
> > Xray lasers: should be fine for now, especially when options to see through walls become more rare.
> > Thermal eyes: Don't fully lock them, give them a active ability instead that only allows their usage for short bursts of time instead of always on.
> > Xray eyes: Sure, lock them behind a tech like you did, though with lower tech cost than right now.
> 
> Then at the very least, what about gating the lasers behind illegals and pushing thermals into its own slot above combat? Not requiring illegals, but still requiring at least a bit _more_ effort for them alongside making them an active ability. That way they're still accessible but they're not easily gotten alongside x-ray lasers, while the lasers themselves require a lot more work, but will at least remain the same.
> 
> And how does 7500 sound for the x-ray vision node?

7500 seems fine for a xray vision node.

And no, a illegal-tech lock is not required for xray  lasers, considering they strongly lose effectiveness without xray/ thermal vision, which is already being nerfed / made more difficult to acquire by this PR.

Thermals being in the combat implant tech should already be fine, no need to create another tech just for them that comes after that one.

DeltaFire15:
Marking this PR as stale due to requested changes & there being no recent commits.

DeltaFire15:
Closing this PR due to staleness.

PR #13836, opened 2020-12-22
refactors statpanel a bit to send topics to client instead of item which bypasses problematic /Topic overrides entirely

silicons:
rejoice no more broken fucking clicks

now when do we get dragdrop supported again 🥴


DeltaFire15 (approved):
Thank you silicons very cool.

PR #13837, opened 2020-12-22
sprinting now defaults to toggle rather than hold (because 90% of our players probably never check the keybinding preferences and from personal experience, holding down shift to move is awful)

silicons:
## About The Pull Request

title

## Why It's Good For The Game

title

## Changelog
:cl:
tweak: anyone new to the server is lucky enough to have their sprint default to toggle instead of hold
/:cl:


DeltaFire15 (approved):
yeah ok

PR #13838, opened 2020-12-22
removes deathgasping on death if already faking death as it's real easy to spot changelings in stasis otherwise

silicons:
## Why It's Good For The Game

makes no sense something playing dead seizes up and falls limp again on actually reaching the threshold

## Changelog
:cl:
fix: things in DEATHCOMA do not deathgasp on death
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/death.dm
+++ b/code/modules/mob/living/carbon/death.dm
@@ -5,7 +5,7 @@
 	silent = FALSE
 	losebreath = 0
 
-	if(!gibbed)
+	if(!gibbed && !HAS_TRAIT(src, TRAIT_DEATHCOMA))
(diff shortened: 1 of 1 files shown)


kiwedespars:
they also seize up and fall lifeless like twice when they go into stasis, and again when they die which is imo a much more concrete tell

DeltaFire15:
Yeah, tested locally, with this PR they still deathgasp twice when activating the ability, it'd be nice if you fixed that aswell.

silicons:
can i fix that later i have no clue why that happens

DeltaFire15:
Fiine

DeltaFire15 (approved):

PR #13843, opened 2020-12-22
allows varediting of admin stealthname

silicons:
## Why It's Good For The Game

this REALLY isn't but let's pretend that this will be used for fun and not just for OOC shitposting


DeltaFire15 (approved):
Bet

PR #13847, opened 2020-12-22
fixes not being able to emphasize +s+ingle letters in chat

silicons:


DeltaFire15 (approved):
Surely this doesn't break anything

PR #13856, opened 2020-12-23
Actually made gear harness properly adjustable

Putnam3145:
## About The Pull Request

the code is bad and this fixes my earlier fix. This one I actually tested properly instead of just assuming the code is sane.

## Why It's Good For The Game

i like non-error'd things.

## Changelog
:cl:
fix: gear harness alt-click is now sane
code: rolldown() and toggle_jumpsuit_adjust() now no longer mix behavior-that-should-be-overridden and behavior-that-shouldn't-be-overridden in ways that make no sense.
/:cl:

### DIFF
--- a/code/modules/clothing/under/miscellaneous.dm
+++ b/code/modules/clothing/under/miscellaneous.dm
@@ -178,7 +178,7 @@
 	else
 		to_chat(usr, "<span class='notice'>Your gear harness is no longer covering anything.</span>")
 		body_parts_covered = NONE
-	return !body_parts_covered
+	return TRUE
 
 /obj/item/clothing/under/misc/durathread
 	name = "durathread jumpsuit"
--- a/code/modules/clothing/under/miscellaneous.dm
+++ b/code/modules/clothing/under/miscellaneous.dm
@@ -171,19 +171,14 @@
 	can_adjust = TRUE
 	body_parts_covered = CHEST|GROIN
 
-/obj/item/clothing/under/misc/gear_harness/rolldown()
-	if(!can_use(usr))
-		return
+/obj/item/clothing/under/misc/gear_harness/toggle_jumpsuit_adjust()
 	if(!body_parts_covered)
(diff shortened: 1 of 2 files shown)


DeltaFire15 (changes requested):
Please do not override rolldown(), modify toggle_jumpsuit_adjust() instead.

Reason for that is rolldown() containing a bunch of checks, like if you are even allowed to do it, in the parent proc.

Putnam3145:
> 
> 
> Please do not override rolldown(), modify toggle_jumpsuit_adjust() instead.
> 
> Reason for that is rolldown() containing a bunch of checks, like if you are even allowed to do it, in the parent proc.

it also includes a bunch of hardcoded nonsense that doesn't make sense in this context in the first place because it has a fuckton of assumptions of what alt-clicking is supposed to mean in the first place and the fact that I was using it is (as you can plainly see) the reason it was breaking in the first place.

Putnam3145:
@DeltaFire15 alright, I've now reworked rolldown() and toggle_jumpsuit_adjust() so that you can actually customize jumpsuit adjusting instead of clearly being written with the assumption that neither would ever be overridden, as they were. I don't know why I didn't do this in the first place.

code/modules/clothing/under/miscellaneous.dm
```diff
@@ -172,11 +172,13 @@
...
-	if(adjusted)
-		body_parts_covered = NONE
-	else
+	if(!body_parts_covered)
+		to_chat(usr, "<span class='notice'>Your gear harness is now covering your chest and groin.</span>")
 		body_parts_covered = CHEST|GROIN
+	else
+		to_chat(usr, "<span class='notice'>Your gear harness is no longer covering anything.</span>")
+		body_parts_covered = NONE
+	return !body_parts_covered
```
DeltaFire15: This line should probably always return true considering the return value is used to see if the body needs to be updated.

Putnam3145: Also done.

PR #13858, opened 2020-12-23
Replaces Thermals in Genetics for Enhanced Vision

keronshb:
## About The Pull Request

This replaces Thermals in Genetics for Enhanced Vision.  Enhanced Vision increases your tile view by +2 tiles.  Yes it's already hugboxed to not work with either Thermals/Xray from Genetics or robotic implants.  And it's only +2 tiles of view, not that different from how widescreen already does this.

A big shoutout to Augments, Slime, Artur, and Saliith for all of the help.

## Why It's Good For The Game

Having thermals be so easily accessible as they are right now doesn't really help the intrigue portion of the game.  Enhanced Vision should give a decent benefit to ranged combat, but not overwhelmingly so like Thermals or XRay.

## Changelog
:cl:
add: Adds Enhanced Vision to the Genetics console
add: Adds a Enhanced Vision injector
add: Adds an Anti Enhanced Vision injector
tweak: Prevents Thermals from being found via the genetics console.
balance: Prevents overlap between Enhanced Vision and any type of xray/thermals, blinds the user.
/:cl:


DeltaFire15:
Resetting view size to default every time update_vision is called will probably mess with other things that change view size, like sniper rifles, shuttle docking consoles, etc.

lolman360:
honestly unconditional increased view range is also pretty op
this is one of those cases where improve dont remove doesnt apply

keronshb:
> 
> 
> Resetting view size to default every time update_vision is called will probably mess with other things that change view size, like sniper rifles, shuttle docking consoles, etc.

Hmm.  That could be an issue then. What would you replace this with then?  It seems like update_vision was the way to go for the other eyes.

keronshb:
> 
> 
> honestly unconditional increased view range is also pretty op
> this is one of those cases where improve dont remove doesnt apply

I figure this would cause the most issues for ships like the Nukie or Pirate ship because of turrets, but people already """Blind""" fire on them anyway without missing.

If it seems to be too much of an issue for maints I could also make it incompatible with night vision.

DeltaFire15:
Haven't seen anything too bad happen with the thermals mutation that would warrant removal in recent times, so, closing this for now.

PR #13867, opened 2020-12-25
Makes zeolites far, far easier to make

Putnam3145:
## About The Pull Request

They honestly do not need to be a fuck-you explody chem if you try making them at any reasonable pace. They don't need to be one of the hardest fermi chems *period*, even harder than powergame-ass SDGF or eigenstasium. And they don't particularly need to have a reliance on miners, either. Them being a fermi chem at all is a reasonable barrier.

## Why It's Good For The Game

Because radiation is godawful to deal with anyway, come on, it's not even a particularly well-designed mechanic.

## Changelog
:cl:
balance: Zeolites now only generate 1/5 the heat when reacting and don't require a catalyst.
/:cl:

### DIFF
--- a/modular_citadel/code/modules/reagents/chemistry/recipes/fermi.dm
+++ b/modular_citadel/code/modules/reagents/chemistry/recipes/fermi.dm
@@ -593,7 +593,6 @@
 	id = /datum/reagent/fermi/zeolites
 	results = list(/datum/reagent/fermi/zeolites = 5) //We make a lot! - But it's now somewhat dangerous, and needs a bit of gold to catalyze the reaction
 	required_reagents = list(/datum/reagent/medicine/potass_iodide = 1, /datum/reagent/aluminium = 1, /datum/reagent/silicon = 1, /datum/reagent/oxygen = 1)
-	required_catalysts = list(/datum/reagent/gold = 5)
 	//FermiChem vars:
 	OptimalTempMin 	= 500
 	OptimalTempMax 	= 750
@@ -604,7 +603,7 @@
 	//CatalystFact 	= 0
 	CurveSharpT 	= 1.5
 	CurveSharppH 	= 3
-	ThermicConstant = 5
+	ThermicConstant = 1
 	HIonRelease 	= -0.15
 	RateUpLim 		= 4
 	PurityMin 		= 0.5 //Good luck!


DeltaFire15:
What exactly is the difficulty here? I just tested them on local and managed to make big quantities of them in a pretty short time, with tier one machinery (120u, purity ~0.85 because I let the ph move out of the zone for a bit too long). Gold can be very easily procured, as miners are both very unlikely to die before they deliver some gold, and there is public mining available. And, in situations where there is no time for that, there is always the vault which has some gold in it, suitable for such emergencies.

Hartmann-von-Berlichingen:
There is no other way to deal with contamination the way zeolite does, and radiation contamination is not rare, therefore it should not be that hard to make in my opinion. Thing is when I asked Lilly to make a chem for contams I only needed it for contam only, but she added normal radiation removal feature (and a strong one) too! you two know better but if it's ok with you I have a suggestion, I wouldn't mind zeolite to only remove contaminations alone (not normal radiation), since we already have good chems to deal with normal radiation (Prussian blue, 3500 for a deluxe rad crate from cargo I always order, no recipe tho). Then it'd be reasonable for everyone to have zeolite an easy to make chem ya?

DeltaFire15:
> There is no other way to deal with contamination the way zeolite does, and radiation contamination is not rare, therefore it should not be that hard to make in my opinion. Thing is when I asked Lilly to make a chem for contams I only needed it for contam only, but she added normal radiation removal feature (and a strong one) too! you two know better but if it's ok with you I have a suggestion, I wouldn't mind zeolite to only remove contaminations alone (not normal radiation), since we already have good chems to deal with normal radiation (Prussian blue, 3500 for a deluxe rad crate from cargo I always order, no recipe tho). Then it'd be reasonable for everyone to have zeolite an easy to make chem ya?

Firstly, that doesn't answer my question, what *is* so difficult about making it, considering none of its properties are particularaly scary (the only thing that might be a nuisance the first time making the chem is that it requires a ph lower than the combined phs of its ingredients, means you need to apply a bit of buffer before starting or use another chem with better ph level as a ghetto buffer, but thats all really.)
Secondly, 5u for decontamming each means that without any techs (like bigger beakers, better reaction management via upgraded heaters, etc) you can decontam >20 items per batch, which is only required for high amounts of contam (showers slowly decontam too which works very fine for most stuff that doesn't go into the thousands of contam, and suit storage units also purge contamination & radiation from contents in emergencies.)

Hartmann-von-Berlichingen:
Your zeolite is far too hard to make for a chem that is the only one used against contam, your answer.
(if its too good to be easily made, make it less good, just give us something normal to make for contams, not this hard to make)
SSU cannot decontam lots of objects, it deals burn damage which is not a big deal and won't kill if there are meds around.
Shower as I said, is too slow and if contam is high (usually) it contams other objects around the shower.
again, for a chem that is the only one of its kind, this recipe is hard. If bica was the only way to deal with brute, you wouldn't make it this hard to make, you could add a better secondary chem for brute, like sali, and make that hard to make no problem.

Hartmann-von-Berlichingen:
also three things, regardless of what we see as hard or easy to make, note that 

1. every round which we had contams I've never seen one single shift for crew to remove contaminations, usually contaminated area is ignored if possible or shuttle is called.

2. before zeolite was nerfed (super super easy too make, I also agree) some medicalcrew made it just incase, and I myself used it a lot to decontam objects-people. however after it was changed, Ive asked multiple people of medical and not a single one happened to make the new zeolite as it was "too hard to make now, not worth it"

3. I knew ssu and shower can help with contams, but if they were viable options I wouldn't ask for zeolite. (tho this is personal preference I didn't ask many people about this)

Chiirno:
Zeolites currently are the only way to get rid of contamination on structures (Windows, lights, machinery, etc).
As it currently stands if a contamination leak happens and the chemist doesn't know how to make fermichems, it's an easy shuttle call.
I want to see janitors being sent to the SM to clean radiation. They do have a biosuit after all.

silicons:
zeolites should be easy or a water reaction should be added
furthermore regarding ease of use: it's not just the fact it's hard to use it's the fact that fucking up quite literally fucks up the entire station but at this point it's a fermichem systematic issue more than a zeolite issue

@Chiirno unless someone changed stuff biosuits don't block rads

at the very least water should be made to reduce (not immediately clear instantly for all levels) radiation on react, making spraying stuff down with a fire extinguisher very viable. perhaps even a pentetic effect where it reduces more the higher it is.

PR #13883, opened 2020-12-27
Set Pose from RP server is now a thing (in reality this just modifies existing temporary flavor text to be better but no one here understands enough code so I'll spare the explanation)

silicons:
## About The Pull Request

Temporary flavor text, as set pose, will now be fully visible in examine and have its own verb

## Why It's Good For The Game

Allows you to set a short (1024 char max, no line break) description of character poses/whatnot for ARPEE purposes.

## Changelog
:cl:
add: set-pose has been added
tweak: temporary flavor text renamed to set pose, fully visible in examine
/:cl:


DeltaFire15 (approved):
*Seems* fine and heard no complaints regarding it in the time it was testmerged.

PR #13885, opened 2020-12-27
Makes space heaters worth using

jack-avery:
## About The Pull Request

Space heaters now heat in a 3x3 area centered on the heater. As a compromise for this, their heating power is reduced to 1/4.

## Why It's Good For The Game

Space heaters are literally worthless right now.

Tested using a Pyro Anomaly. Works fine. Not sure about the performance impact with this method. Bestow wisdom butnam

## Changelog
:cl:
tweak: space heater heating range and power
/:cl:

### DIFF
--- a/code/game/machinery/spaceheater.dm
+++ b/code/game/machinery/spaceheater.dm
@@ -18,7 +18,7 @@
 	var/mode = HEATER_MODE_STANDBY
 	var/setMode = "auto" // Anything other than "heat" or "cool" is considered auto.
 	var/targetTemperature = T20C
-	var/heatingPower = 40000
+	var/heatingPower = 10000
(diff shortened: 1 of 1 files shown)


YakumoChen:
why the fuck doesn't heat spread in atmos so areas even out is the better question

Putnam3145:
> why the fuck doesn't heat spread in atmos so areas even out is the better question

because LINDA is broken and always will be.

anyway this makes space heaters 2.25x as powerful but eh whatever

jack-avery:
> anyway this makes space heaters 2.25x as powerful but eh whatever

it's a buff not a change in distribution. it also consumes power 2.25x as quickly

PR #13890, opened 2020-12-27
ablative parry armwraps

silicons:
**Sprites by Gladwyn/Catuai!** ^w^

## About The Pull Request

6 TC item
highly efficiency parry, but hard to perfectly parry melee
reflects parried projectiles
no cooldown on successful parry
does not lock gun usage or attacking even while parrying

cannot be purchased by nukeops because that would be broken

## Why It's Good For The Game

more fun traitor items that totally won't need rebalancing in the first day

## Changelog
:cl:
add: syndicate ablative armwraps have been added.
/:cl:


DeltaFire15:
Kinda seems extremely strong with the 'you can attack while parrying' *combined with* 'yeah if you fail the parry you don't even get a stagger'

silicons:
@DeltaFire15 removed no-stagger in favor for same stagger as cooldown (2.25) seconds, and removed the reflection

regardless i think this should be testmerged first because uhhh parry balance is kinda woozy rn

Linzolle (approved):

PR #13893, opened 2020-12-27
Nerfs meth and morphine to only remove damage slowdown rather than all slowdown (read this carefully, as this affects lings)

silicons:
## About The Pull Request

**This will affect changelings, as their adrenals currently use a meth subtype.**

## Why It's Good For The Game

This firmly cements their role as painkillers rather than total slowdown removers.
They're currently too powerful with the current movespeed balance of the codebase.
I originally was going to make bolas and certain other things not ignored but then I started wondering why I even did this in the first place.

## Changelog
:cl:
tweak: Meth and changeling adrenals no longer ignore all slowdowns, rather damage slowdowns.
/:cl:


YakumoChen:
thank fuck

DeltaFire15 (approved):

PR #13911, opened 2020-12-29
salicyclic acid

jack-avery:
## About The Pull Request

salicyclic acid

## Why It's Good For The Game

salicylic acid

## Changelog
:cl:
spellcheck: salicylic acid
/:cl:


lolman360:
it's not real salicylic acid it's saliCYCLIC acid because it makes the brute go away instead of continuing the cycle of suffering

jack-avery:
> it's not real salicylic acid it's saliCYCLIC acid because it makes the brute go away instead of continuing the cycle of suffering

No

PR #13924, opened 2020-12-30
Allows mass sticking money into your id

SandPoot:
## About The Pull Request
Title.

## Why It's Good For The Game
Faster money taking, unless you still want it with do afters.

## Changelog
:cl:
tweak: You can attack a pile of money on the floor with your id to put it all in quickly.
/:cl:

### DIFF
--- a/code/modules/mining/ores_coins.dm
+++ b/code/modules/mining/ores_coins.dm
@@ -397,7 +397,7 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
 		else
 			to_chat(user, "<span class='warning'>You need one length of cable to attach a string to the coin!</span>")
 			return
-	if(istype(W, /obj/item/card/id))
+	else if(istype(W, /obj/item/card/id))
 		var/obj/item/card/id/ID = W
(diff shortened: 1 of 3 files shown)


code/modules/mining/ores_coins.dm
```diff
@@ -397,6 +397,17 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
 		else
 			to_chat(user, "<span class='warning'>You need one length of cable to attach a string to the coin!</span>")
 			return
+	if(istype(W, /obj/item/card/id))
```
DeltaFire15:
```suggestion
	else if(istype(W, /obj/item/card/id))
```

DeltaFire15 (changes requested):
Looks alright, though probably put a else if here to preserve the else at the end working as before.

code/modules/mining/ores_coins.dm
SandPoot: yea, sounds about right

DeltaFire15 (approved):

PR #13925, opened 2020-12-30
Lowers max power req for Power Sink to explode in about 8.5 minutes and lowers the explosion size

keronshb:
## About The Pull Request

This lowers the power requirement for a sink to explode from 1+e010 to 3.5+e008.  This should explode roughly 8 and 1/2 minutes.  It was tested in an environment where the network had a consistent 1.21MW.  It gives people enough time to find it while still allowing it to blow.

## Why It's Good For The Game

The sink was never exploding for the whole round even with a whopping 7MW in the grid.  This should:

1) Allow the sink to actually do its job of blowing up

2) Not be terrible for the rest of the station if it's never found (IE draining power forever).

## Changelog
:cl:
balance: Max power from 1+e010 to 3.5+e008
balance: lowers explosion size to 3,7,14,28
/:cl:

### DIFF
--- a/code/game/objects/items/devices/powersink.dm
+++ b/code/game/objects/items/devices/powersink.dm
@@ -18,7 +18,7 @@ GLOBAL_LIST_EMPTY(power_sinks)
 	custom_materials = list(/datum/material/iron=750)
 	var/drain_rate = 1600000	// amount of power to drain per tick
 	var/power_drained = 0 		// has drained this much power
-	var/max_power = 1e10		// maximum power that can be drained before exploding
+	var/max_power = 1.25e008		// maximum power that can be drained before exploding
 	var/mode = 0		// 0 = off, 1=clamped (off), 2=operating
 	var/admins_warned = FALSE // stop spam, only warn the admins once that we are about to boom
 
--- a/code/game/objects/items/devices/powersink.dm
+++ b/code/game/objects/items/devices/powersink.dm
@@ -18,7 +18,7 @@ GLOBAL_LIST_EMPTY(power_sinks)
 	custom_materials = list(/datum/material/iron=750)
 	var/drain_rate = 1600000	// amount of power to drain per tick
 	var/power_drained = 0 		// has drained this much power
-	var/max_power = 1.25e008		// maximum power that can be drained before exploding
+	var/max_power = 3.5e008		// maximum power that can be drained before exploding
 	var/mode = 0		// 0 = off, 1=clamped (off), 2=operating
 	var/admins_warned = FALSE // stop spam, only warn the admins once that we are about to boom
 
--- a/code/game/objects/items/devices/powersink.dm
+++ b/code/game/objects/items/devices/powersink.dm
@@ -155,5 +155,5 @@ GLOBAL_LIST_EMPTY(power_sinks)
 
 	if(power_drained >= max_power)
 		STOP_PROCESSING(SSobj, src)
-		explosion(src.loc, 4,8,16,32)
+		explosion(src.loc, 3,7,14,28)
 		qdel(src)


YakumoChen:
Lower the TC cost if you're going to lower the potential round impact. 4 minutes is nothing and generally you want to keep the station depowered for longer. Make it 2GJ instead of 20 as it was before.

YakumoChen:
Keep in mind that it's intended purpose is to _depower the station_, NOT be just a delayed bomb.

YakumoChen:
Screenshot_20201230-193330_Discord

keronshb:
> 
> 
> Lower the TC cost if you're going to lower the potential round impact. 4 minutes is nothing and generally you want to keep the station depowered for longer. Make it 2GJ instead of 20 as it was before.

I had originally had it at around 2e008, but that was around five minutes+ and people are gonna find the sink by then.  4 minutes is even stretching it.

It's purpose is to depower the station,  but it's also completely missing the bomb impact, which is what people buy it for also.

Kevin also had said to lower the bomb impact at this time.

YakumoChen:
The bomb impact is arguably supposed to deter people from just trying to out power the drain, the purpose of it is to depower the station for an appreciable amount of time. 4 minutes is only twice as long as a syndiebomb, the APCs will barely have time to drain.

keronshb:
> 
> 
> The bomb impact is arguably supposed to deter people from just trying to out power the drain, the purpose of it is to depower the station for an appreciable amount of time. 4 minutes is only twice as long as a syndiebomb, the APCs will barely have time to drain.

It blows up in 4 minutes with a constant 1.21mw.  The power network usually doesn't have that much unless all the SMES are upgraded and all power sources are set up.

YakumoChen:
On a different note I have never seen anyone buy it for the explosion explicitly, its primary reason for buying has always been to deny power.

I feel like instead of an explosion it should start overloading APCs after a power threshold is crossed and the minimum drain is met (like the objective number) and make APCs shoot lightning like goon. It would deter overpowering it and not just put the power drain on a timer.

Considering it has to be on a wire its impossible to not find it eventually with a t ray unless someone is actively moving it (something I've done as a gimmick before that would no longer apply if the bomb has a shitty threshold)

YakumoChen:
If you're going to leave it a bomb on a short fuse give traitors a way to tell how much power it has stored (like a percentage until it explodes) to reward them for guarding it and let them use it like a makeshift syndicate bomb.

keronshb:
I changed it to 3.5e+008.  1e+009 is 28 minutes, that's too long.

This gives it more than enough ample time to drain all the apcs on engine + solars + turbine, even with the SMES' upgraded.

The main goals of this were to

1) Have the job description actually work (says it explodes, takes more than a whole round to explode)

2) Signify the traitor did their objective.

On top of what was already mentioned.

YakumoChen:
28 minutes is absolutely long enough. Draining the APCs isnt simply the goal, its keeping the power off. As soon as you disconnect the sink, everything returns to full functionality within a couple ticks. 

Keep in mind this affects nukeops and other roles that would want to use the sink to easily break into places or shoot someone while the lights are off.

YakumoChen:
> Have the job description actually work (says it explodes, takes more than a whole round to explode)

The name of the item is "power sink" not "power grid bomb"

> Signify the traitor did their objective.

We should be balancing it around it's primary function (draining power), not around an objective that uses it.

YakumoChen:
3.5e+008 is only 3 minutes and 36 seconds with the maximum power drain, by the way.

1 GigaJoule adds up to 10 minutes, 20 seconds at the maximum power drain before it explodes, for reference.

PR #13937, opened 2020-12-31
Bokken and Bokken Accessories: Rebalance Edition

necromanceranne:
## About The Pull Request

Makes Bokken require crafting via the crafting menu, and are a little more difficult to craft. Not too bad, ideally requires a botanist. Or someone with a cotten plant and some Left-4-Zed.

Bokken are now slightly stronger than before to compensate for the added hurdle to obtainment.

Adds wakizashi bokken, which now use the old bokken stats but are normal sized, so not as good for catching people, but shouldn't stamina drain you as much.

Bokken and crafted katanas now show up in your twin sheath scabbards, as they now have overlays. However, they should no longer fit two katanas, only the proper pairing or, unfortunately, two wakizashi-sized swords. I honestly can't be arsed to make this work better right now.

Fixes a runtime with twin scabbards trying to populate with a null value.

## Why It's Good For The Game

Bokken are neato but I think they shouldn't be stackcrafted.

It was weird katanas had a paired wakizashi but bokken don't. Since weight on weapons has actual value, it felt fitting to have a bokken wakizashi.

I wanna come back at some point when I'm not sleep deprived and attempt to maybe update bokken sprites, iunno.

## Changelog
:cl:
add: Wakizashi bokken!
tweak: Bokken are now crafted with the craft menu. They're slightly more difficult to make, with bokken requiring durathread.
tweak: Twin scabbards now have overlays for the weapons it fits. You now can see what swords you have. Twin scabbards have weight limits however, so no double katanas.
balance: Bokken are slightly stronger. Wakizashi bokken have the bokken stats but are normal sized.
fix: A runtime with twin scabbards.
/:cl:


timothyteakettle:
durathread to make a 13 force stamina weapon
.. really?

keronshb:
I don't think this is warranted at all.

necromanceranne:
Eh it's not but I didnt wanna make the full sized one too much stronger. Basically the wakizashi is as good a weapon as the old bokken, so it's basically just for the sake of getting an additional +1/2 damage boost. The wakizashi shouldn't be meaningfully more difficult to acquire than the current bokken. I just didn't have a better material to distinguish them to justify the full-sized bokken as an upgrade, so durathread is it.

necromanceranne:
Oh wait I didn't mean to push this too early, I'm actually redoing the pr since so much shit changed.

Fuck, this pr is full of spoilers!!

PR #13949, opened 2021-01-01
clamps cyborg sprint to 10 tiles/second and fixes them going methspeed

silicons:


PokerfaceCactus:
Cyborg sprint speed is now the same as walk speed. if this is the intended max speed suggest just disabling sprint altogether as now you just waste massive amounts of energy trying to sprint but not actually going any faster. 
(sprinting ups your passive power use from like 1 to 100. it drains you like crazy.)

PR #13951, opened 2021-01-02
Max groin cannon size increased before being punished for erecting a massive mayo-blaster to -max roundstart config- inches instead of 20

BlueWildrose:
## About The Pull Request

What the title says. The meat sticks can be a whole two feet long as per current server config instead of slightly below a two-footer before you suffer the consequences of slowdown and blood loss.

## Why It's Good For The Game

The server config was changed recently so that you can have roundstart 24 inch thunderswords instead of 20 inch yogurt flingers. If there are characters that are created with this maximum roundstart packaging in mind, they will likely throb out of the round before too long unless this is also changed. Meaning chemistry won't be pestered over flesh flute shrinking requests being erected up when players make this mistake, because the tan bananas will be more tolerable at that size.

It also likely will be appreciated by a certain giant red panda who people keep making vore jokes over.

## Changelog
:cl:
tweak: You can now have a max-roundstart-dicksize-config inch long johnson before you start suffering blood loss and slowdowns instead of a 20 inch one.
/:cl:

### DIFF
--- a/code/modules/arousal/organs/penis.dm
+++ b/code/modules/arousal/organs/penis.dm
@@ -45,9 +45,9 @@
 			new_size = 1
 		if(7 to 11) //If large
 			new_size = 2
-		if(12 to 20) //If massive
+		if(12 to 24) //If massive
 			new_size = 3
-		if(21 to 34) //If massive and due for large effects
+		if(25 to 34) //If massive and due for large effects
 			new_size = 3
 			enlargement = TRUE
 		if(35 to INFINITY) //If comical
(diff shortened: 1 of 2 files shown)


silicons:
@BlueWildrose use config please

aka if(size to config_get max size), no change

BlueWildrose:
Good idea, I'll thrust in and look into that. Probably would have been the better option to rub in.

BlueWildrose:
There we go, it's now set up for in case these custard launchers maximum sizes are altered yet again.

PR #13984, opened 2021-01-06
fixes removing toner from photocopiers (this is an untested fix but it probably works because it compiled)

silicons:


DeltaFire15 (approved):
Tested it and it does work.

PR #13985, opened 2021-01-06
Changes Sleeping Carp's deflect to Combat Mode.

keronshb:
## About The Pull Request

This lets sleeping carp deflect while in combat mode rather than throw mode.

## Why It's Good For The Game

Throw mode disrupts combos and eats an attack if you forget to turn it off.  Also if someone throws a bola at you or any other object, you'll catch it and disable throw mode... which also disables the reflect.

## Changelog
:cl:
refactor: Changes Scarp Deflect to Combat mode instead of Throw mode
/:cl:

### DIFF
--- a/code/datums/martial/sleeping_carp.dm
+++ b/code/datums/martial/sleeping_carp.dm
@@ -126,7 +126,7 @@
 		return BULLET_ACT_HIT
 	if(!isturf(A.loc)) //NO MOTHERFLIPPIN MECHS!
 		return BULLET_ACT_HIT
-	if(A.in_throw_mode)
+	if(SEND_SIGNAL(A, COMSIG_COMBAT_MODE_CHECK, COMBAT_MODE_ACTIVE))
 		A.visible_message("<span class='danger'>[A] effortlessly swats the projectile aside! They can deflect projectile with their bare hands!</span>", "<span class='userdanger'>You deflect the projectile!</span>")
 		playsound(get_turf(A), pick('sound/weapons/bulletflyby.ogg', 'sound/weapons/bulletflyby2.ogg', 'sound/weapons/bulletflyby3.ogg'), 75, TRUE)
 		P.firer = A


DeltaFire15:
Pretty sure not being able to block projectiles & attack simultaniously was a conscious design choice. (which I support)
Also combat mode does not influence speed unless you move towards the direction opposing the one you are facing (aka, backwards), which does have a slight slowdown.

keronshb:
> 
> 
> Pretty sure not being able to block projectiles & attack simultaniously was a conscious design choice. (which I support)
> Also combat mode does not influence speed unless you move towards the direction opposing the one you are facing (aka, backwards), which does have a slight slowdown.

Oh oops, I swore there was a slowdown for sprinting + combat mode if you're running away.  Guess not!

Anyway Kevinz talked about putting it to combat mode, so was just following through with that.

necromanceranne:
It's meant to be difficult to combo. I readded deflection because people wanted it in my original iteration on tg. Had my way, it wouldn't have it at all.

I think given the resistances and defenses of scarp, you would be better off axing the invisible armor and doing this instead.

dzahlus:
im gonna be honest with you, stunbaton > scarp. This is needed, atleast you don't get cucked by projectiles while trying to stunjoust sec.

keronshb:
> 
> 
> It's meant to be difficult to combo. I readded deflection because people wanted it in my original iteration on tg. Had my way, it wouldn't have it at all.
> 
> I think given the resistances and defenses of scarp, you would be better off axing the invisible armor and doing this instead.

IIRC the main difference between TG and Cit is that Cit has a HEAVY focus on stamina based combat - which scarp (still) doesn't do well against at all after the slight tweaks I gave it (despite its description being meant for engagement/anti stun).  Yeah you absolutely don't want this invincible thing that everyone picks, but when you need to engage, your options become limited to an extent of either needing slips, parrying, or EMP to deal with batons (and to fight exclusively in maint to prevent jousting).  Which the new parry on scarp IS good don't get me wrong.

And then when you need to combat mode help yourself up after getting smacked by a baton, you've interrupted your combo.  Unless you're in maint, you're not gonna get TWO clicks on someone while you're laying down.  If you need throw mode on to prevent shots, you're already setting yourself up for failure because if you forget to turn it off, it eats a click.  Or if someone throws something into your hand (which is funny considering Scarp deflects stuff), it goes off.

Not to mention that old SCarp's (really dumb) sleep grab combo is gone, which was one of its strongest points.  That's fine that it's gone.  But realize that needing to get two hits on someone to either A) Knock them Down or B) Knock them away is complete AIDS if they have a baton since their one hit > your two hits.  Especially when Krav can just do it in one hit also.  And keeping all this in mind + needing to use throw mode to deflect.. I hope you can see where this starts to become an issue.

Besides all someone needs to do really is one baton hit > shoot anyway since SCarp doesn't deflect on the ground.

The tl/dr is that this is just a SMALL QoL buff to Scarp.  Will it help deal with Batons any more?  No.  Will it become less shit to deal with a Sec officer?  Slightly.

necromanceranne:
It's not a small change at all. It fundamentally changes the way you engage sleeping carp users. There is never a downtime to their defenses if they have combat mode up, which is the opposite of what I wanted.

If batons are so grossly oppressive that the martial art meant to fight the fucking things can't compete even in ideal situations, you are looking at the wrong area.

This changes generally nothing anyway if people baton and shoot, unless you revert the nerf that stopped them deflecting on the floor.

keronshb:
> 
> 
> It's not a small change at all. It fundamentally changes the way you engage sleeping carp users. There is never a downtime to their defenses if they have combat mode up, which is the opposite of what I wanted.
> 
> If batons are so grossly oppressive that the martial art meant to fight the fucking things can't compete even in ideal situations, you are looking at the wrong area.
> 
> This changes generally nothing anyway if people baton and shoot, unless you revert the nerf that stopped them deflecting on the floor.

You need to turn off Combat mode to regain stamina.  Punching takes stamina, parrying takes stamina, shoving up takes stamina.  Not having throw mode on more often means they're more susceptible to Bolas.

Rising Bass is considered the better martial art for a few reasons (inb4 muh whataboutism):

1 Auto Deflect
2 Harm Harm for knockdown, compare that to harm grab and or harm disarm needed for SCarp
3 AUTO CATCHING ITEMS like Bolas
4 Being able to steal the person's held item
5 Forced movement

Yes again, moving deflect to combat mode doesn't help VS Batons, but it DOES help their survivability a bit more and does make it more worth more 17 tc now.  Though honestly the bundle with the Bo Staff is probably better for them anyway in terms of blocking attacks.

I don't want to make them knockdown proof, that's dumb.  They still can get bola'd if throw mode isn't on.  They still can get flashbanged and they still can get set on fire.  They still need a Sleeper/Medibot/Pills to heal rather than grabbing a burn patch/suture.

The whole point ISN'T to make them a One Man Army, the point is to justify their strength (a close range, martial artist that you shouldn't engage with guns).

DeltaFire15:
In my opinion, this is not needed at all due to the fact of 'you can't block ranged and attack at the same time' being the tradeoff for 'yeah you have a ton of physiology buffs (60% less brute & stamdamage, 40% less burn, etc)', aswell as them have a pretty effective parry too. So, no, not at all weaker then bass. Unless you rely on the dodging ranged ability, in which case, yes, of course the martial art about dodging is better for you than the unga bunga carp one.
Either way, I'll wait on @silicons 's opinion before closing this.

TheObserver-sys:
Personally, if scarp is going to get reflect on combat mode _and R-bass should also probably get this treatment too, no free rides_, then every reflect should eat away at your stamina bar, to make the conscious decision to reflect actually be more meaningful than "disable stamina regen for 5 seconds and then keep running".  You should have to weigh the options of taking a few shots and relying upon your increased durability, or reflecting everything and not being able to escape too easily without other ways of regaining stamina.

The two martial arts could also do with being turned into bundles as well, specifically because you can get free rides into stamina regensville if you get a good discount, which is very sketchy to me, now that fistwraps also grant parrying power and enable wounding. I dunno about you, but it gets old seeing someone with adrenals, mask, and carp using fist wraps to just ignore literally everything, like a goddamn mario star.

silicons:
I agree with @TheObserver-sys 

I'd rather have a full rebalance on this than buff it more. Most of the time, yes, scarp is a bit weak, but I'd rather not turn that into ye old carp near-unstoppable rampage.

PR #13994, opened 2021-01-08
hugboxes admin twitchplays clown car to not explode on death to avoid killing people (by default)

silicons:
bruh


Enzo-Leon:
lmao

DeltaFire15 (approved):

PR #14024, opened 2021-01-10
Suit Storage Unit changes

SandPoot:
## About The Pull Request
This pr modifies suit storage units to be like current tg's except suit cells and i added a special slot for shoes/boots.

## Why It's Good For The Game
Makes the suit storage unit generally better to use, also the radials help a lot.

## Changelog
:cl:
add: Suit storage units now use a radial menu.
add: Suit storage units have been modified to have both shoes and generic storage slots.
tweak: Suit storage units now put the items in your hand, rather than on the ground.
/:cl:


DeltaFire15:
From testing this on local, this doesn't exactly seem like it improves them, just makes interacting with them more difficult (having to cycle through radial menus instead of having all things in one conscise menu.)

silicons:
sorry not a fan of radial over tgui talk to me if you disagree

PR #13973, opened 2021-01-05
completely removes penis/breast enlargement effects due to it being completely unmaintained and to be frank awful

silicons:
yeah the funny's over 

- no one's maintaining it, it's deadweight on the codebase and doesn't add anything interesting to the game imo
- the effects are awful to deal with and i just had someone tell me they died of having a normal sized dick
- there's no reason at all to punish people for this outside of "funny" but honestly if someone runs around the halls with this that's an admin issue not a code issue

PR #14026, opened 2021-01-11
Skidaddle skidoodle, your 12+ inch dick no longer looks like a noodley elephant trunk

BlueWildrose:
## About The Pull Request

Apparently retrieving from the config directly regarding penis length isn't really possible and I would have to do something more elaborate to get that to work, which probably isn't worth it for genital code. I probably should have playtested that before submitting that last PR.

I'm going back to normal numbers for the switch statement, especially since we're removing the "debuffs" associated with having a huge dick as per #13973.

This fixes the penis size appearance being overly huge on the sprite when it shouldn't be like that for a 12 inch donger. That's reserved for dongers bigger than 36 inches.

## Why It's Good For The Game
no more funny dick.

This will likely merge-conflict with #13973 but I'm tired of waiting on someone to merge it so I'm posting my PR now.

## Changelog
:cl:
fix: Fixes noodle size appearance for 12+ inch members.
/:cl:


BlueWildrose:
Ah, the other one got merged. Merge conflict resolved.

PR #14029, opened 2021-01-11
Ports TG code for wormhole projector and gravity gun.

HeroWithYay:
## About The Pull Request
Primarily this will make these weapons require anomaly cores instead of firing pins, slightly modified to require any anomaly core instead of specific types to be more in line with citadel anomaly core behavior.

## Why It's Good For The Game
It's odd they required firing pins to begin with considering they were originally designed with anomaly cores in mind and this should make them actually see a bit of play during rounds while keeping them hard enough to get that they aren't easy to abuse, being weaker alternatives for cores compared to reactive armor or a phazon. 

## Changelog
:cl:
tweak: Wormhole Projector and Gravity Gun now require anomaly cores to function instead of firing pins.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -177,7 +177,7 @@
 
 /obj/item/gun/energy/gravity_gun
 	name = "one-point gravitational manipulator"
-	desc = "An experimental, multi-mode device that fires bolts of Zero-Point Energy, causing local distortions in gravity. Requires a gravitational anomaly core to function."
+	desc = "An experimental, multi-mode device that fires bolts of Zero-Point Energy, causing local distortions in gravity. Requires an anomaly core to function."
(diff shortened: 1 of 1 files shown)


DeltaFire15:
Can't exactly make up my mind on this. @silicons opinions?

silicons:
@DeltaFire15 in support, i hate pins

but just for the record `originally designed with anomaly cores in mind` this was a recent balancing thing, when i first made those anomaly cores weren't exactly a thing used for anything and the current tg anomaly core balance system wasn't in play.

HeroWithYay:
> @DeltaFire15 in support, i hate pins
> 
> but just for the record `originally designed with anomaly cores in mind` this was a recent balancing thing, when i first made those anomaly cores weren't exactly a thing used for anything and the current tg anomaly core balance system wasn't in play.

The guns share names with cores so I kinda assumed. Oof.

PR #14033, opened 2021-01-12
Adds Color Mates to every station (except Snaxi)

BlueWildrose:
## About The Pull Request

This adds Color Mates to every station _that matters,_ of a certain amount.

I'm not going to go into detail of where every single Colormate has been placed on each station, but the basic jist is, wherever a Clothesmate is in the room, a Colormate also now exists in the same room. Unless that Clothesmate is inside Genetics, which two stations happen to have. Which means you're likely going to be seeing these colormates in dorms or in locker rooms.

No objects have been deleted to make way for these Colormates - not even the potted plants. Some objects have been relocated however. Like said potted plants.

also apparently a minor fix to the HoS's pet carp on Kilo.

_(edit: Yes I may be bashing Snaxi out of silliness but the real reason I ignored Snaxi is because I believed another mapper was working on Snaxi right now and I felt it wasn't worth making changes when they're making changes to it right now.)_

edit 2: I figured one in the ghost cafe would also be handy, since I know people like to experiment with outfitting in there. So it's been added in there too.

## Why It's Good For The Game

Might as well make good use of this new feature that we've been given and save science the time of having to print a colormate circuit every round.

## Changelog
:cl:
add: Color Mates have been added to all stations (except Snaxi). Enjoy coloring your attire without having to bug science!
/:cl:

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -29620,10 +29620,10 @@
 /turf/open/floor/plasteel/white,
 /area/science/robotics/lab)
 "bsU" = (
-/mob/living/simple_animal/hostile/retaliate/ghost,
 /obj/machinery/atmospherics/components/unary/vent_pump/on{
 	dir = 4
 	},
+/mob/living/simple_animal/hostile/retaliate/ghost,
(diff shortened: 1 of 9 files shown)


DeltaFire15 (approved):
Sure I guess.

PR #14037, opened 2021-01-12
Nanogel for robotic limbs

DeltaFire15:
## About The Pull Request
Adds nanogel to the game, available from robotics and medvendors.
It is used on robotic limbs that passed their threshold, and have their damage level currently at the minimum damage you can get them to via external repairs. When applied, which is pretty fast when done on someone else, *incredibly* slow if used on yourself (basically, the using it on self is only in because antags generally should be able to handle themselves), it will repair the internal damage of that limb, allowing further repairs without the need for surgery (Though, surgery is probably still a bit faster if *all* your limbs passed the threshold).

Currently *only* available from vendors, might make it printable at advanced medtech.

~~[Pending localtesting], should work though.~~
Tested, seems to work fine.

Nanogel sprite is from Eris (https://github.com/discordia-space/CEV-Eris/blob/master/icons/obj/stack/items.dmi), bless them.

## Why It's Good For The Game
Synths being forced to have surgery when sufficiently damaged was ill received in some cases, especially combat roles which run into the problem more frequently. This alleviates that issue somewhat, while still encouraging seeking assistance instead of just using it on oneself (besides it not being available everywhere)

## Changelog
:cl:
add: Nanogel, available at medical and robotics, which fixes internal damage in sufficiently repaired robotic limbs.
/:cl:


ghost:
cool and good

PR #14051, opened 2021-01-14
disallows statue to move_crush stuff

silicons:
when i made move force, people being able to ram literally anything off their hinges at 20 tiles per second wasn't the intention.

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/statue.dm
+++ b/code/modules/mob/living/simple_animal/hostile/statue.dm
@@ -49,6 +49,7 @@
 	move_force = MOVE_FORCE_EXTREMELY_STRONG
 	move_resist = MOVE_FORCE_EXTREMELY_STRONG
 	pull_force = MOVE_FORCE_EXTREMELY_STRONG
+	var/can_move_crush = FALSE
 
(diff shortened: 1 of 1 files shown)


MrJWhit:
saltpr :v)

BlueWildrose:
Oh, so that's the noise I was hearing during that wizard round.

I heard a shitton of banging while I was working on genetics.

DeltaFire15:
Don't statues need this to be able to get through locked doors? And, they generally only exist on wiz rounds, so a bit more wackyness probably isn't that horrible.

BlueWildrose:
> 
> 
> Don't statues need this to be able to get through locked doors? And, they generally only exist on wiz rounds, so a bit more wackyness probably isn't that horrible.

There's at least one known away mission with a statue in it. The Snow Cabin area has a modified statue that appears as a snowman named Frosty. Granted, this statue is ordinarily not played by players on the other hand.

DeltaFire15:
Haven't seen any issues happening due to them having move crush, especially considering they usually only appear during wizard rounds (or well, actually sentient statues, that means. Therefore, closing this PR.

PR #14056, opened 2021-01-15
disables rotatium due to it being horifically bugged and unmaintained

silicons:
read files changed


code/modules/reagents/chemistry/reagents/toxin_reagents.dm
```diff
@@ -783,6 +783,8 @@
 	value = REAGENT_VALUE_VERY_RARE
 
 /datum/reagent/toxin/rotatium/on_mob_life(mob/living/carbon/M)
+	return		// until fixed - the rotations never stop
```
DeltaFire15:
```suggestion
	return ..()		// until fixed - the rotations never stop
```

DeltaFire15 (changes requested):
Needs ..() or it'll not metabolize out of their body (due to reagent removal being handled in the parent proc)

PR #14057, opened 2021-01-15
removes stealthmin orbit block

silicons:
## Why It's Good For The Game

i have never ever seen a good use of this and it's more annoying than not

## Changelog
no


DeltaFire15:
Uhh should probably keep it for stealthminned ghosts since those get made invisible afterall? Though I agree, not really needed when living.

silicons:
@DeltaFire15 done :+1:

DeltaFire15 (approved):

PR #14073, opened 2021-01-16
chemical bear trap

lolman360:
## About The Pull Request

Chemical bear trap. Does less damage than the regular bear trap but instantly injects 15u of chemicals if the target is injectable.

Piercing and bluespace ones exist, made from piercing and bluespace syringes.

## Why It's Good For The Game

ok honestly it might not be but beartraps are criminally underused

## Changelog
:cl:
add: Chemical beartrap, for when getting your leg caught wasn't already bad enough.
/:cl:


MrJWhit:
That crafting requirement seems way too low, it doesn't even have a tool that's needed to make it.

lolman360:
yes it's literally a syringe on a beartrap

also crafting menu bloat bad

MrJWhit:
You're... adding to the crafting menu bloat.

lolman360:
this doesn't use the crafting menu at all

silicons:
memes aside i'm really not sure why we need chemical boobytraps

silicons (changes requested):
admin only it.

SmArtKar:
Too OP. Beartraps are like superior energy bolas and you're making them even deadlier.

silicons:
reopen when fixed

PR #14075, opened 2021-01-17
Removes Hive's Blessing and Xeno Queen Corpse from the Alien Ruin

Tupinambis:
## About The Pull Request

Tin.

## Why It's Good For The Game

It's been a long time coming, and it seems to do a lot more harm than good, from causing balance issues with the strong knockdowns, to people soft griefing with it. It also seems like people are obnoxious about it. A lot of people dislike it, and as the person that originally added the ruin and Hive's blessing, I think it's time they go.

## Changelog
:cl:
del: Hive's blessing and the xeno queen corpse have been removed from the lavaland alien ruin.
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_alien_nest.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_alien_nest.dmm
@@ -218,8 +218,8 @@
 "aY" = (
 /obj/structure/alien/weeds,
 /obj/effect/decal/cleanable/blood/gibs,
-/mob/living/simple_animal/hostile/alien/drone,
 /obj/structure/alien/weeds/node,
+/mob/living/simple_animal/hostile/alien/drone,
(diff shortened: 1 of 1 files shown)


shellspeed1:
The corpse of the queen should really be left but it needs to have the eggsac removed from it.

timothyteakettle:
keep the corpse, remove its egg sac, replace hive's blessing with some funny xeno-related item that isn't going to turn people into aliens

keronshb:
By removing the corpse that's a hit to science/medical.  Even if it was RNG in the first place, experimental dissection of a queens corpse yields a decent amount of points.

shellspeed1:
Possible alternative what if we replaced the queen with other xenomorph subtypes like the praetorian or sentinels?

YakumoChen:
Reminder being a feature creator is not a valid reason for removal

Tupinambis:
It would seem that the overall reception this PR has gotten implies the issue is not as significant as I initially had assumed. Will be closing it.

PR #14083, opened 2021-01-18
Removes the queen, and increases xeno variety in lavaland xenomorph ruin.

shellspeed1:
## About The Pull Request
~~This PR is an alternative to https://github.com/Citadel-Station-13/Citadel-Station-13/pull/14075. It works to fix some of the worst features of the lavaland xenomorph hive while still making it interesting.~~ This PR aims to remove the eggsac from the lavaland queen primarily by removing the queen and replacing it with a mix of other xenomorph subtypes. The hive blessing has been readded as it has been deemed to be fine for now.

## Why It's Good For The Game
~~This PR works to maintain the integrity of the lavaland xenomorph hive while also removing both the eggsac and hives blessing from the ruin. These two items in particular are rather problematic based on what people have talked about.~~ The hive blessing syringe has been argued over again and has been deemed fine for now and has been added back.. The queen is however still potentially a problem with the eggsac.

## Changelog
:cl:
add: All xenomorph types have been added as corpses for mapping purposes
balance: The dead xenomorphs in the lavaland xenomorph hive now have more variety.
/:cl:

PR #14088, opened 2021-01-18
Turns hive blessing into a traitor item and reworks it to turn people into a xenomorph larvae.

shellspeed1:
## About The Pull Request
First things first, this should not be merged without https://github.com/Citadel-Station-13/Citadel-Station-13/pull/14075 or https://github.com/Citadel-Station-13/Citadel-Station-13/pull/14083. The reason for this is hive's blessing needs to be removed from the lavaland ruin first. Otherwise this PR does exactly what is says on the tin, it adds the Hive blessing syringe as a 24 tc role and pop locked traitor item. The TC value isn't set in stone although it will not be under 20. 

## Why It's Good For The Game
Instead of removing what is a fun item, why don't we turn it into a traitor item for people to use.

## Changelog
:cl:
add: The Hive blessing has been added as a 24 tc item for members of the science department. It requires 30 players minimum to use.
balance: Xenomorph transformation disease no longer turns people into a hunter and now turns them into a larvae so they can evolve into the other xenomorph species.
/:cl:

### DIFF
--- a/code/datums/diseases/transformation.dm
+++ b/code/datums/diseases/transformation.dm
@@ -200,7 +200,7 @@
 	stage3	= list("<span class='danger'>Your throat feels very scratchy.</span>", "Your skin feels tight.", "<span class='danger'>You can feel something move...inside.</span>")
 	stage4	= list("<span class='danger'>Your skin feels very tight.</span>", "<span class='danger'>Your blood boils!</span>", "<span class='danger'>You can feel... something...inside you.</span>")
 	stage5	= list("<span class='danger'>Your skin feels as if it's about to burst off!</span>")
-	new_form = /mob/living/carbon/alien/humanoid/hunter
+	new_form = /mob/living/carbon/alien/larva
 	bantype = ROLE_ALIEN
 
 /datum/disease/transformation/xeno/stage_act()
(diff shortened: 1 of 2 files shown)


MrJWhit:
Yeah I disagree on adding this for a roundstart traitor scientist, considering they can get 12 monkies roundstart, and plenty more if someone spends a little bit of time doing xenobio.

MacaroniCritter:
I agree with this because I really like getting my shit kicked in by simple mobs.

shellspeed1:
I mean xenos aren't simple mobs. They are actually counted as complex ones since you can do surgery and all that on them.

DeltaFire15:
Gonna say no to this in its current state.
Considering you can just stock up on monkeys before injecting yourself and immediately roundend the station afterwards, this seems far too strong to be a traitor item, even if it were at romerol price.
If you have any ideas to resolve that problem, be my guest. DNM till then.

Tupinambis:
> Gonna say no to this in its current state.
> Considering you can just stock up on monkeys before injecting yourself and immediately roundend the station afterwards, this seems far too strong to be a traitor item, even if it were at romerol price.
> If you have any ideas to resolve that problem, be my guest. DNM till then.

Delta, Hive's' blessing only turns you into a sentinel or runner. You literally cannot become a queen with it. I'd argue 24 is way too high, this is closer to the gorilla mutation in terms of power.

EDIT: Turns out this PR changes that, so yeah don't let people become larva that's a terrible idea.

shellspeed1:
I can fix the monkey problem but it would mean a significant nerf to xenomorphs. So A.

PR #14091, opened 2021-01-18
[s] UM

silicons:


shellspeed1:
Oh come on, this was funny.

TheObserver-sys:
Yes, but how long until someone **else** finds out, eh? What's funny now is maliciously abused later, at least in this case.

shellspeed1:
Fair, anyone that looks at the github knows this is a thing now at least.

PR #14108, opened 2021-01-20
A plethora of changes to clothing damage (gonna reopen with a new PR when I find time to actually make improvements on this)

BlueWildrose:
## About The Pull Request
A plethora of changes to clothing damage and about clothing damage.

Change 1:
Shredded clothing no longer drops. It's just completely useless and doesn't provide any armor bonuses anymore of any kind.

Change 2:
Suit sensors will get shorted out when clothing gets shredded, rather than getting damaged. Suit sensors can't be repaired until the shredded clothing gets repaired first.

Change 3:
Clothing will have "damage resistance" based on how many parts an article of clothing has - or lack thereof. The multiplier would still be 10 for any standard jumpsuit, but a jumpskirt, which doesn't have legs at all will have a multiplier of 6. This means it would take more hits to destroy a part of a jumpskirt than it would a jumpsuit. In other words, a sort-of buff to those type of clothes that don't cover everything. If someone was eswording with the intention to destroy suit sensors for instance, they would have to focus on less limbs to target overall but it'd take more hits to tear apart.

Change 4:
It wasn't really made clear that you _could_ repair suit sensors. Examine text has been adjusted, and two new tips of the round have been added to reflect this.

## Why It's Good For The Game
**Why I changed clothes dropping:**
I've never been a fan of a collection of equipment just dropping because someone attacked you with something. Pretty much why I also hate the shrink ray. It's a pointless hassle and I don't think anyone really liked such. When your uniform gets dropped, so does your belt, your pockets, and your ID. And in the heat of battle, you probably won't notice your clothes fell off until you try to access someplace without your ID.

**Why I changed the suit sensor setup:**
With the current setup for clothing damage, it generally takes one or two e-sword hits to just destroy suit sensors, or two laser hits from an e-gun, because suit sensors shorting out ordinarily triggers upon destroying one limb of the clothing, rather than all. This pretty much means it's completely pointless for an unarmored greyshirt to have suit sensors on to begin with, because funny traitor in maintenance could just obliterate those sensors with one or two e-sword hits or two e-gun-from-armory hits. This makes suit sensors important again. The traitor could "sensor-strip" someone by just cutting apart their clothes limb by limb until it becomes "shredded" under this new setup.

**Why I changed clothing durability based on how many "limbs" clothes have:**
I don't think anyone on this server wants to sacrifice personal style in favor of whatever meta is enforced; or to be punished because they chose to wear a skirt over a jumpsuit. This is a form of balance that I had put into consideration, even if it may be unrealistic or whatever you want to call it. _If your argument is going to be realism, then I will make a PR that makes being in a vacuum without protection collapse your lungs and likely kill you almost very quickly, and make the air just instantly get sucked out of a room if windows near space break. If you don't like the idea of that, then good. If  you do in fact like that idea, go play Baystation._

**Why I added tips and examine text:**
The game is damn complicated enough already. Educate people. I was told nobody even knew that they could repair sensors. I didn't even know until I looked at the code.

I dunno why I'm arguing for this part.

## Changelog
:cl:
balance: Clothing no longer drops when shredded. It just becomes useless.
balance: Suit sensors will short out when the clothes become shredded, not damaged now.
balance: If you're wearing clothes that cover less limbs than others, they are considered more durable in those limbs in how much damage they can get before getting destroyed.
/:cl:

### DIFF
--- a/code/__DEFINES/_flags/obj_flags.dm
+++ b/code/__DEFINES/_flags/obj_flags.dm
@@ -17,8 +17,8 @@
 
 /// Integrity defines for clothing (not flags but close enough)
 #define CLOTHING_PRISTINE	0 // We have no damage on the clothing
-#define CLOTHING_DAMAGED	1 // There's some damage on the clothing but it still has at least one functioning bodypart and can be equipped
-#define CLOTHING_SHREDDED	2 // The clothing is useless and cannot be equipped unless repaired first
+#define CLOTHING_DAMAGED	1 // There's some damage on the clothing but it still has at least one functioning bodypart
+#define CLOTHING_SHREDDED	2 // The clothing is near useless and has their sensors broken
 
 // If you add new ones, be sure to add them to /obj/Initialize as well for complete mapping support
 
--- a/code/__HELPERS/type2type.dm
+++ b/code/__HELPERS/type2type.dm
@@ -410,6 +410,31 @@
 
 	return covered_parts
 
+//Turns a Body_parts_covered bitfield into a count of limbs covered.
+/proc/body_parts_covered2count(bpc)
+	var/covered_parts_count = 0
+	if(!bpc)
+		return 0
+	if(bpc & HEAD)
+		covered_parts_count++
+	if(bpc & CHEST || bpc & GROIN)
+		covered_parts_count++
+	if(bpc & ARMS || bpc & HANDS)
+		covered_parts_count += 2
+	else
+		if(bpc & ARM_LEFT || bpc & HAND_LEFT)
+			covered_parts_count++
+		if(bpc & ARM_RIGHT || bpc & HAND_RIGHT)
+			covered_parts_count++
+	if(bpc & LEGS || bpc & FEET)
+		covered_parts_count += 2
+	else
+		if(bpc & LEG_LEFT || bpc & FOOT_LEFT)
+			covered_parts_count++
+		if(bpc & LEG_RIGHT || bpc & FOOT_RIGHT)
+			covered_parts_count++
+	return covered_parts_count
+
 /proc/slot2body_zone(slot)
 	switch(slot)
 		if(SLOT_BACK, SLOT_WEAR_SUIT, SLOT_W_UNIFORM, SLOT_BELT, SLOT_WEAR_ID)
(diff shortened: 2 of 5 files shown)


keronshb:
I feel like if these go through, portable suit sensors should go so people would have to pay attention to the console.

ghost:
the funny damage console works off of suit sensors you bird ass
rage

YakumoChen:
Not really a fan of clothing retaining the ability to hold gear when shredded. Seems like a needless buff and should obstanately make upkeeping your clothes more meaningful as well as inventory management. Can't bag your dropped toolbelt and ID because you decided to put 6 guns in your backpack? Too bad.

BlueWildrose:
Honestly, when you word it like that, this change would benefit traitors more than security, because security is more likely to carry the laser guns that are able to do damage to clothing to begin with. Shoot up a traitor enough with lasers, and they would have dropped their belt. It took two shots to each limb to destroy it clothing-wise.

The problem here primarily lies in the fact of, was the fact that jumpsuit damage existing _fun_ to begin with? Understandably there's got to be balance involved, and there has to be fun involved. Balance-wise, clothing dropping will be ambiguous between antag and non-antag parties, but personally, I didn't consider losing the jumpsuit _fun_, and a good number of people will probably agree with me on this.

BlueWildrose:
As for what Keron said, as discussed in development_discussion_main on Discord, I, or someone else can likely consider a system where the suit sensors have their own health, and that based on how much damage it receives, the maximum sensors you can have get reduced. For instance, if it reaches 66% health it'll be unable to broadcast your location on suit sensors, but people will know that you got hurt, etc. Combine this with clothing damage, and shredded clothing can provide a guarantee that your suit sensors WILL get destroyed. But I won't be putting that in this PR if so - that can come later.

YakumoChen:
Not all mechanics are meant to be """fun""". Losing your clothes is supposed to a punishment for rolling around in fire or not upkeeping your clothing. The "fun" is adapting to changing / disadvantageous situations.

The fact that your belt and bag don't also burn is the main "it wouldn't be fun then" factor behind this, something I pushed on TG a long time ago.

Removing clothing dropping your stuff would just be a massive nerf to fire in general as your items can't burn while attached to your clothing, I believe.

silicons:
i'm not sure buffing suit sensors is a good idea

ghost:
>people that don't know how crucial wound armor is and haven't seen the difference between shooting a naked person with buckshot and a person with even the tiniest amount of armor

BlueWildrose:
Gonna reopen with a new PR when I find time to actually make improvements on this.

PR #14111, opened 2021-01-20
[READY]Gives the revenant TK

Arturlang:
## About The Pull Request
Let's the revenant clickdrag objects to throw them, with various items having special actions upon being done so.

## Why It's Good For The Game
Hopefully this should make the revenant a much more active antagonist, as if there are no bodies or anything at the start the revenant can be really fucked and unable to do anything, I'm hoping this fixes that somewhat, letting them atleast fuck with people by throwing stuff at people

## Changelog
:cl:
add: Revenants can now clickdrag to throw stuff at people, with some items doing various things at the same time.
/:cl:


timothyteakettle:
i don't like the idea of a revenant being able to just hide and shotgun people using guns/weapons laying around, hell even hiding and just jabbing people with glass shards is annoying (which by using tk on a coffee vendor, then smashing the coffee using tk, you'll have plenty of)

MacaroniCritter:
epic funny revbuff plus one thankies

Arturlang:
@timothyteakettle youll be able to kill it while its doing TK, as it gets exposed like all the other abilities do, plus guns dont trigger because gun code is bad

timothyteakettle:
uhhh in that case yeah sure i guess

silicons:
i'm honestly not sure i want to take the plunge of buffing revenant when it's still designed around hiding the entire round
if we want that (which, we do), we should just get it fully changed to something akin to wraith, t b h.

lolman360:
honestly making rev invis a toggled ability that drains essence could work? and then give them a form of scrying orb ability so they can still scout ahead before going in

just spitballing idk

edit: the main problem is 'invisible untargetable thing goes around killing everyone' tbh so by making them more vulnerable we can make them more active

Arturlang:
@silicons Yes but who is going to put in the effort to rework a antag? I'm thinking this is a good middle-ground. Don't forget the fact that everty time the revenant tries to do this they'll reveal themselves, given with the low cooldown this should result in the revenant exposing it way more

ghost:
i'd recommend making it have a very short reveal (1-2s or so) if they're within line of sight and also there are no barriers between them and the TK'd object and target

and then give it a longer reveal (4-8 seconds) if they're within line of sight but there are barriers (ie windows) between them and the tk'd object or them and the target

this would add more depth to the tk mechanic by making it so if you decide to play more risky you might get away with taking no damage at all if the target doesn't react fast enough, but if you try to play it safer then you might get hit anyways if the target has lasers or tools to get to you. food for thought.

Arturlang:
@kappa-sama I can try but no promises

Arturlang:
Did as suggested, though hesitant to give too long cooldowns for each, If this leads to be too powerfull, we can change it later anyways

PR #14140, opened 2021-01-26
fixes chat highlighting not working (literally) half the time

silicons:
## About The Pull Request

so I don't actually know why this happens but,
having the global flag there made the filter function trim out every other word in the list.

this fixes that, although the fact that i don't know WHY this works other than "it works in the REPL" is kinda scary
tested through node.js REPL

## Changelog
:cl:
fix: chat highlighting no longer drops half your entered words.
/:cl:


Putnam3145:
the reason it works is cause of the +--it matches 1 or more of the given characters, i.e. it doesn't actually need to check the string more than once. no idea why *adding* it breaks but *having* it was always unnecessary

Linzolle (approved):
i mean i guess if it works?

PR #14144, opened 2021-01-26
advanced mob subtype part 1

timothyteakettle:
## About The Pull Request
this will allow advanced mobs to exist, which are just, complicated mobs, for mapping and admin purposes

part 1 contains these things:
- [x] allow modified loot through the use of a json table of path and then a set of variable names and corresponding values, the main use of this is for things like modified force/sprite weapons, and to drop custom id cards for custom doors
- [x] add support for sets of vars to be applied to the mob at different hp thresholds including sprite and armor changes (allows mobs to have multiple phases, so to speak)
- [x] add support for mobs to take the appearance of a dummy wearing any clothing as specified by a list and a special icon state that will apply the look using an overlay

these things are planned for part 2:
- [ ] add weighted attacks saved in json with cooldowns, armour changes, if the mob should be anchored during it, temporary sprite changes, these should also be conditionable by phase ids so certain attacks only appear at hp thresholds, they can also have set cooldowns
- [ ] add support for projectile attacks that take vars for a projectile, including number of projectiles, sequence/burst, and angle-between-each assuming it's a burst and not a sequence (a sequence is like one after another, a burst is like a shotgun), and time-between-each-projectile
- [ ] add speech support for all attacks, so they can speak when attacking, or changing phases

these things are planned for part 3:
- [ ] add minion attack which can only be done if the specific minion doesn't surpass a certain limit, and takes yet another json list to edit the minions variables in turn, so that you can summon weaker/stronger minions
- [ ] add melee attack effects including reagent transfer, lifesteal, fire, knockback
- [ ] add passive effects including regeneration and gravity (like the gravity anomaly it'll pull you towards it)
- [ ] add special effects such as timestop moves, turf replacement (turning random tiles into lava or chasms)

## Why It's Good For The Game
aids mappers and admins to create more complex mobs

## Changelog
:cl:
add: adds an advanced mob subtype that allows fine-grain editing of its properties, including support for mappers
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/advanced_mob.dm
+++ b/code/modules/mob/living/simple_animal/advanced_mob.dm
@@ -51,7 +51,7 @@
 			change_phase(closest_phase)
 
 // change the phase to a given one, this is not a safe proc, it should only ever be called by phase_check
-/mob/living/simple_animal/hostile/advanced/proc/change_phase(var/phase)
+/mob/living/simple_animal/hostile/advanced/proc/change_phase(phase)
 	current_phase = phase
 	var/list/corresponding_phase_list = phases[phase]
 	var/datum/datum_self = src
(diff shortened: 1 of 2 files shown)


timothyteakettle:
this pr isn't dying or dead it's just taking a long time to do

timothyteakettle:
this will probably be part 1, with part 2 being custom attacks and part 3 being custom effects

PR #14159, opened 2021-01-28
Fixes Alien Royals being able to ventcrawl

jack-avery:
## About The Pull Request

Alien Royals can ventcrawl. This fixes that. Oops, my bad.

Tested via spawning a Queen, Praetorian, and Drone. Praetorian and Queen could not ventcrawl. Drone could.

## Why It's Good For The Game

Royals are too fat to ventcrawl. Stop that.

## Changelog
:cl:
fix: alien royals can no longer ventcrawl
/:cl:


code/modules/mob/living/carbon/alien/alien.dm
```diff
@@ -38,7 +38,8 @@
 
 	create_internal_organs()
 
-	AddElement(/datum/element/ventcrawling, given_tier = VENTCRAWLER_ALWAYS)
+	if(!(name == "alien queen"))
+		AddElement(/datum/element/ventcrawling, given_tier = VENTCRAWLER_ALWAYS)
```
DeltaFire15: Not a great way to do this, better add a 'can ventcrawl' var to xenos that it checks for before adding the component, and set it to false for ones that shouldn't have it.
Also praetorians should also be incapable of ventcrawling.

DeltaFire15 (changes requested):

DeltaFire15:
@raspy-on-osu Please address these requested changes because the alternative to this PR would be to revert the entirity of the ventcrawling changes and I'd rather not continue to have ventcrawling high-tier xenos any longer.

DeltaFire15 (approved):

PR #14172, opened 2021-01-30
Cell Charger Buffs

keronshb:
## About The Pull Request

This buffs the regular cell chargers to be in line with that of Weapon Chargers.  This makes charging bluespace cells a bit less painful.

## Why It's Good For The Game

I'd say that everyone was starting to use a stunprod/teleprod with a weapon charger to charge bluespace cells, instead of making actual cell chargers for them.  This is just a nice QOL update.

## Changelog
:cl:
balance: Makes cell chargers, charge faster.
/:cl:


code/game/machinery/cell_charger.dm
```diff
@@ -122,17 +123,19 @@
...
-	charging.give(charge_rate)	//this is 2558, efficient batteries exist
+	if(charging)
+		var/obj/item/stock_parts/cell/C = charging.get_cell()
+		if(C)
+			if(C.charge < C.maxcharge)
+				C.give(C.chargerate * recharge_coeff)
+				use_power(250 * recharge_coeff)
+			update_icon()
 
 	update_icon()
```
SandPoot, line 140: Calling `update_icon` twice

keronshb: Oh thanks!  Which one should I get rid of?

PR #14174, opened 2021-01-30
Meathook Adjustments

keronshb:
## About The Pull Request

Minor adjustments to the meathook to lessen the knockdown so they don't drop items and makes the damage a bit less ridiculous.  This way it's still a good utility tool (amazing speed, range, grabbing), but not as a primary offensive tool.

## Why It's Good For The Game

It's better to have something like the long-ranged meathook a bit more balanced, even if it is lavaland loot.

## Changelog
:cl:
balance: Force and damage > 15 from 18/25
balance: Knockdown put down to 5 from 30
balance: Armor pen down to 10 from 100.
/:cl:


shellspeed1:
Question, why are we leaving it as lavaland loot when this could be a traitor item for chefs?

dzahlus:
> Question, why are we leaving it as lavaland loot when this could be a traitor item for chefs?

because miners are cringe

PR #14182, opened 2021-02-01
i hate xenobiology players

silicons:


silicons:
this should work

Linzolle (dismissed):

code/modules/research/xenobiology/crossbreeding/_status_effects.dm
```diff
@@ -884,7 +888,8 @@
 /datum/status_effect/stabilized/oil/tick()
 	if(owner.stat == DEAD)
 		explosion(get_turf(owner),1,2,4,flame_range = 5)
-		owner.remove_status_effect(/datum/status_effect/stabilized/oil)
+		qdel(linked_extract)
+		qdel(src)
```
DeltaFire15: qdeling linked extract will already delete this status effect since thats done in the crossbreed Destroy().
Also return without ..() to prevent doing things after being qdeld
```suggestion
		qdel(linked_extract)
		return
```
Other than that it's good, tested it on local and *seems* to work.

DeltaFire15 (changes requested):

DeltaFire15 (approved):

PR #14192, opened 2021-02-04
Ports *pain emote from CM13

dzahlus:
## About The Pull Request
What it says on the tin, since CM13 is open source now I can do this PR. .txt file with credits to CM13 included because they don't use github

## Why It's Good For The Game
It's a great emote that could be used in the future for example for getting wounded, also great for comedic effect. 
Slipped? *pain
your girl runs to dorms with someone else? *pain
antags running it down with bombs? *pain
lost your limbs and limped to medbay? *pain

## Changelog
:cl:
add: Added new emote
soundadd: added a new emote sound
/:cl:


moltoretardo:
A shame it'll probably get used more for erping, but let's just think about the comedic effects instead.

PR #14195, opened 2021-02-05
Donator Item - NT Choker

LetterJay:


Linzolle:
it looks like the icons are using the mob icons for both the mob icons as well as the item itself in the inventory?

PR #14202, opened 2021-02-07
Adds nanogel to the robodrobe

DeltaFire15:
## About The Pull Request
Apparently, there's two kinds of robotics vendors, and the kind that I added nanogel to already isn't used anywhere. Whoops!

## Why It's Good For The Game
Adds nanogel to the vendor it was supposed to be in already.

## Changelog
:cl:
tweak: Added nanogel to the robodrobe.
/:cl:


zeroisthebiggay:
have you tried just adding the robotics vendor that remains totally unused to already preexisting maps

PR #14187, opened 2021-02-03
We hate powersinks

keronshb:
## About The Pull Request

Ok jokes aside, this adds a loud beep to the powersink and increases the light it gives off a bit more.  In turn the power consumption was raised from 2mj to 2.5mj.

## Why It's Good For The Game

Not being able to find a power sink for most of the round is awful.  People get it for free as an objective.  And with a recent change it takes 33+ minutes to explode which is insane.

## Changelog
:cl:
add: Adds a constant beep to the power sink as long as its on
balance: increased the light from 5 to 8 for powersinks
balance: increased the power consumption from 2mj to 2.5mj to make up for the beep and extra light.
/:cl:

PR #14204, opened 2021-02-08
Removes the Power Sink Objective

keronshb:
## About The Pull Request

As an alternative to #14187 and this removes the power sink objective which will:

1) Prevent someone from getting a free power sink

2) Also prevent someone from rerolling just to have a free power sink or two.

3) While it's less variety, no one likes this objective.

I'd love to have something replace this instead of just outright removal of an objective, but enough is enough.

## Why It's Good For The Game

The three reasons above pretty much cover it.  If a power sink is well hidden, especially at round start, it causes a lot of agony for everyone and it ends up being an annoying objective that either no one wants to do or no one wants to deal with.

## Changelog
:cl:
del: Comments out power sink objective.
/:cl:


Tupinambis:
it's also impossible to tell if you've even completed the objective and it's way to inconsistent. Sucks to deal with, good riddance.

dzahlus:
would like a different objective in turn but thats another issue +1 from me

PR #14207, opened 2021-02-08
Multilingual fix? (Has since been tested.)

TyrianTyrell:
## About The Pull Request

I used "LANGUAGE_MIND," and I think that restricts it to only understanding the language and not actually being able to speak it. Can we testmerge this to see if it fixes it?

## Why It's Good For The Game

I think it'll fix a problem.

## Changelog
:cl:
code: added a define for multilingual granted languages, and changed the multilingual trait to use it.
/:cl:


silicons:
might want to test this on local.

TyrianTyrell:
Will do so.

TyrianTyrell:
Okay, I tested it on local. I believe it will work, unless the files I altered have been committed to already.

PR #14211, opened 2021-02-08
pickpocketing is now logged outside of hrefs

timothyteakettle:
## About The Pull Request
why was it not logged before

## Why It's Good For The Game
shouldn't we be able to know if someone pickpocketed someone without going through href logs

## Changelog
:cl:
admin: pickpocketing is now logged using log_combat
/:cl:


SandPoot:
I'm wondering how bad it would be if pickpocketing was it's own category, or at least subcategory...

PR #14237, opened 2021-02-13
adds sanity checks to heretic blade shatters

silicons:
good one gamers


DeltaFire15:
Oh oops uhh just rename check_flags of that action to required_mobility_flags and make the isavailable check the ..() return value, while doing the same for Trigger()

Linzolle (dismissed):

Linzolle (changes requested):
see above. required_mobility_flags instead of check_flags to fix issue

Linzolle (approved):

PR #14246, opened 2021-02-14
Multilingual conditions fix.

TyrianTyrell:
## About The Pull Request

"can_speak_vocal" wasn't properly set up for the signed language. I fixed that.

## Why It's Good For The Game

It's a bugfix.

## Changelog
:cl:
fix: fixed the signed language so that you can actually use it, and that it's unusable when it's meant to be.
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -332,7 +332,7 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 	if(HAS_TRAIT(src, TRAIT_MUTE) && get_selected_language() != /datum/language/signlanguage)
 		return 0
 			
-	if (get_bodypart(BODY_ZONE_R_ARM).is_disabled() && get_bodypart(BODY_ZONE_L_ARM).is_disabled() && get_selected_language() == /datum/language/signlanguage)
+	if (src.get_bodypart(BODY_ZONE_R_ARM).is_disabled() && src.get_bodypart(BODY_ZONE_L_ARM).is_disabled() && get_selected_language() == /datum/language/signlanguage)
 		return 0
 
 	if(is_muzzled())
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -329,13 +329,11 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 	return 1
 
 /mob/living/proc/can_speak_vocal(message) //Check AFTER handling of xeno and ling channels
-	if(HAS_TRAIT(src, TRAIT_MUTE))
-		if (get_selected_language() != /datum/language/signlanguage)
-			return 0
+	if(HAS_TRAIT(src, TRAIT_MUTE) && get_selected_language() != /datum/language/signlanguage)
+		return 0
 			
-	if (get_bodypart(BODY_ZONE_R_ARM).is_disabled() && get_bodypart(BODY_ZONE_L_ARM).is_disabled())
-		if (get_selected_language() == /datum/language/signlanguage)
-			return 0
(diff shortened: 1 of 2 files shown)


TyrianTyrell:
Also, this has already been tested locally. I used a fresh install of the master branch for testing.

code/modules/mob/living/say.dm
```diff
@@ -329,7 +329,12 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 	return 1
 
 /mob/living/proc/can_speak_vocal(message) //Check AFTER handling of xeno and ling channels
-	if(HAS_TRAIT(src, TRAIT_MUTE) && get_message_language(message) != /datum/language/signlanguage)
+	var/obj/item/bodypart/leftarm = get_bodypart(BODY_ZONE_L_ARM)
+	var/obj/item/bodypart/rightarm = get_bodypart(BODY_ZONE_R_ARM)
+	if(HAS_TRAIT(src, TRAIT_MUTE) && get_selected_language() != /datum/language/signlanguage)
+		return 0
+			
+	if (rightarm.is_disabled() && leftarm.is_disabled() && get_selected_language() == /datum/language/signlanguage)
```
Linzolle: Is this intentionally written that _both_ arms are required to be disabled to be unable to use sign language?

TyrianTyrell: Yes- because if only one arm is broken, presumably you could still sign with the other. Though I get what you're getting at- perhaps I should make it appear "slurred" if one arm is broken.

Linzolle: this is good enough for now then

Linzolle: this should actually probably check if leftarm and rightarm are not null and return early if they are, since it's possible if the mob has no arm for get_bodypart to return null.

TyrianTyrell: Fair enough. Will add that.

code/modules/mob/living/say.dm
```diff
@@ -329,7 +329,12 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
...
 
 /mob/living/proc/can_speak_vocal(message) //Check AFTER handling of xeno and ling channels
-	if(HAS_TRAIT(src, TRAIT_MUTE) && get_message_language(message) != /datum/language/signlanguage)
+	var/obj/item/bodypart/leftarm = get_bodypart(BODY_ZONE_L_ARM)
+	var/obj/item/bodypart/rightarm = get_bodypart(BODY_ZONE_R_ARM)
+	if(HAS_TRAIT(src, TRAIT_MUTE) && get_selected_language() != /datum/language/signlanguage)
+		return 0
+	
+	if ((rightarm == null || rightarm.is_disabled()) && (leftarm == null ||leftarm.is_disabled()) && get_selected_language() == /datum/language/signlanguage)
 		return 0
```
Linzolle: doing it this way will still runtime error because its still trying to run the function on rightarm and leftarm even if they're null. Something more like this might work better:
```suggestion
	if (get_selected_language() == /datum/language/signlanguage)
		var/left_disabled = FALSE
		var/right_disabled = FALSE
		if (istype(leftarm)) // Need to check if the arms exist first before checking if they are disabled or else it will runtime
			if (leftarm.is_disabled())
				left_disabled = TRUE
		else
			left_disabled = TRUE
		if (istype(rightarm))
			if (rightarm.is_disabled())
				right_disabled = TRUE
		else
			right_disabled = TRUE
		if (left_disabled && right_disabled) // We want this to only return false if both arms are either missing or disabled since you could technically sign one-handed.
			return 0
```
(Can someone else please look at this to see if it makes sense or if there's a cleaner way to do it.)

Linzolle: in fact you could even move the assigning of the leftarm and rightarm underneath the first check if you did it like this. I _think_ it will work but its kind of ugly

Linzolle (approved):

PR #14251, opened 2021-02-15
Massive synthflesh buff

Chiirno:
## About The Pull Request

Changes the amount of synthflesh needed to unhusk from 101 and above -> 100 and above.

## Why It's Good For The Game

Makes it easier for doctors to get the right ratios for splashing.

## Changelog
:cl:
tweak: Synthflesh now unhusks with 100u instead of requiring 101u.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -499,7 +499,7 @@
 			SEND_SIGNAL(M, COMSIG_ADD_MOOD_EVENT, "painful_medicine", /datum/mood_event/painful_medicine)
 			var/vol = reac_volume + M.reagents.get_reagent_amount(/datum/reagent/medicine/synthflesh)
 			//Has to be at less than THRESHOLD_UNHUSK burn damage and have 100 synthflesh before unhusking. Corpses dont metabolize.
-			if(HAS_TRAIT_FROM(M, TRAIT_HUSK, "burn") && M.getFireLoss() < THRESHOLD_UNHUSK && (vol > 100))
+			if(HAS_TRAIT_FROM(M, TRAIT_HUSK, "burn") && M.getFireLoss() < THRESHOLD_UNHUSK && (vol > 99))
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -499,7 +499,7 @@
 			SEND_SIGNAL(M, COMSIG_ADD_MOOD_EVENT, "painful_medicine", /datum/mood_event/painful_medicine)
 			var/vol = reac_volume + M.reagents.get_reagent_amount(/datum/reagent/medicine/synthflesh)
 			//Has to be at less than THRESHOLD_UNHUSK burn damage and have 100 synthflesh before unhusking. Corpses dont metabolize.
-			if(HAS_TRAIT_FROM(M, TRAIT_HUSK, "burn") && M.getFireLoss() < THRESHOLD_UNHUSK && (vol > 100))
+			if(HAS_TRAIT_FROM(M, TRAIT_HUSK, "burn") && M.getFireLoss() < THRESHOLD_UNHUSK && (vol > 99))
```
SandPoot:
```suggestion
			if(HAS_TRAIT_FROM(M, TRAIT_HUSK, "burn") && M.getFireLoss() < THRESHOLD_UNHUSK && (vol >= 100))
```
"Greater than or equal to" looks like a nicer way to do it.

Chiirno: I don't want to mess with floating points.
You take a beaker and pour 100u in but its actually 99.999998u and when you splash it doesn't take effect.
I might have misheard, but I remember a friend telling me they were having such problems.
So I want to check with them before changing it. Unless you have reasons to believe floating point chemistry wouldn't be an issue?

SandPoot: right

Chiirno:
Seems like the problem arose at numbers even above 101u, so it's likely not due to floating point. I'll put in the >= for now.

Linzolle (approved):

PR #14262, opened 2021-02-17
stops people using Message All on PDAs when their cartridge doesn't allow it

timothyteakettle:
## About The Pull Request
stops people using Message All on PDAs when their cartridge doesn't allow it

## Why It's Good For The Game
stops people using Message All on PDAs when their cartridge doesn't allow it

## Changelog
:cl:
fix: stops people using Message All on PDAs when their cartridge doesn't allow it
/:cl:

### DIFF
--- a/code/game/objects/items/devices/PDA/PDA.dm
+++ b/code/game/objects/items/devices/PDA/PDA.dm
@@ -665,7 +665,8 @@ GLOBAL_LIST_EMPTY(PDAs)
 				create_message(U, locate(href_list["target"]))
 
 			if("MessageAll")
-				send_to_all(U)
+				if(cartridge && cartridge.spam_enabled)
+					send_to_all(U)
 
 			if("toggle_block")
 				toggle_blocking(usr, href_list["target"])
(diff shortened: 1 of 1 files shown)


Arturlang:
pr Info not accurate, spam enabled are more than just te lawyer's

code/game/objects/items/devices/PDA/PDA.dm
```diff
@@ -665,7 +665,8 @@ GLOBAL_LIST_EMPTY(PDAs)
 				create_message(U, locate(href_list["target"]))
 
 			if("MessageAll")
-				send_to_all(U)
+				if(cartridge && cartridge.spam_enabled)
```
LetterN:
```suggestion
				if(cartridge?.spam_enabled)
```

LetterN (changes requested):

code/game/objects/items/devices/PDA/PDA.dm
timothyteakettle: thanks

PR #14266, opened 2021-02-19
Replaces carbon xeno drone with a simplemob sentinal in the beno cube

MrJWhit:
## About The Pull Request

Replaces carbon xeno drone with a simplemob sentinal in the beno cube

## Why It's Good For The Game

It's clearly not the intended design of the cube, spawning in a brainless carbon mob, based on the "run!" part of the description. Plus, a drone can easily turn into a queen, which isn't ideal for something that's only accessible through silver slimes.

## Changelog
:cl:
tweak: Xeno cube now makes an aggressive sentinal instead of a brainless drone
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/food/snacks_meat.dm
+++ b/code/modules/food_and_drinks/food/snacks_meat.dm
@@ -432,7 +432,7 @@
 	dried_being = /mob/living/carbon/monkey
 
 /obj/item/reagent_containers/food/snacks/cube/beno
-	name = "alien drone cube"
+	name = "alien sentinel cube"
 	desc = "Just add water and run!"
 	tastes = list("the jungle" = 1, "acid" = 1)
 	dried_being = /mob/living/simple_animal/hostile/alien/sentinel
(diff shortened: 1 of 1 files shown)


RogueSteampunker:
Considering there's only been one reported person who's used it for brain stuff, and the fact that making it a simplemob completely destroys what purpose the cube would have as a means of getting xeno organs for cargo and surgery points for rnd, I'm gonna have to disagree.

Turning it into a simplemob only accomplishes in adding another mindless hostile mob in a role where hostile mobs far better than it already exist. All this does is punish the majority over one or two players the adminbase had failed to properly reprimand until recently.

Solaris-Shade:
LINE 435 name = "alien drone cube"

If it spawns an alien sentinel, wouldn't the name need to be: name = "alien sentinel cube"    ?

silicons:
```
	///list of things spawned at mob's loc when it dies.
	var/list/loot = list()
	///causes mob to be deleted on death, useful for mobs that spawn lootable corpses.
	var/del_on_death = FALSE
```
my recommendation is making it spawn a varedited (see: subtype) sentinel with slightly higher stats, and dropping a carbon sentinel body on death. this preserves old functionality without just giving people a free catatonic DRONE (see: evolving into queen, structures, etc), while keeping the item true to its name and frankly, intended function.

MrJWhit:
Fair enough, that's updated. 

A drone causes eggsack balance concerns, and, I honestly couldn't find the alien drone simplemob.

MrJWhit:
Looks good to me.

PR #14267, opened 2021-02-19
allows all non queen/empress SIMPLE MOB xenos to spawn from gold cores

silicons:
## Why It's Good For The Game

i'm frankly not sure why this is disabled
they don't drop organs, they don't have the insane carbon tackles, their strongest thing is neurospit which is.. eh, not really the worst in the world when it comes to how powerful something is and trying to balance around it.
let's try this again.

## Changelog
:cl:
balance: gold cores can spawn simplemob xenos again.
/:cl:


BlueWildrose:
This also means the chef can get xeno meat guaranteed each round if that's wanted for whatever reason.

Good.

Tupinambis:
This also means that there is actually a gold core people will play now.

Probably just for ERP but you know, at least it's something.

PR #14268, opened 2021-02-19
alternatively, replaces the cube sentinel spawn with an actual simplemob sentinel that drops a carbon sentinel body on death so you still get to run this to the ground without being able to be a literal alien queen 

silicons:
## Why It's Good For The Game

preserves some of the old functionality while making the cube not entirely harmless
xenobio is the only place that can get this with RNG and they have a billion ways to end a round. 
far as i'm concerned this is the right amount of anti-fun (see: no more free drones to brain transplant with the capacity of being a queen) without completely gutting the item.

closes #14266 

## Changelog
:cl:
balance: xeno cube makes hostile xenos now, and drops a sentinel instead of a drone.
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/food/snacks_meat.dm
+++ b/code/modules/food_and_drinks/food/snacks_meat.dm
@@ -435,7 +435,7 @@
 	name = "alien drone cube"
 	desc = "Just add water and run!"
 	tastes = list("the jungle" = 1, "acid" = 1)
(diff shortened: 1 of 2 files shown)


shellspeed1:
better idea do 50 50 for hunter and sentinel on death.

DeltaFire15 (approved):

PR #14273, opened 2021-02-19
chaplain vampires won't burn in their own chapel now

timothyteakettle:
## About The Pull Request
they literally burn to death in the chapel if they sign up as chaplain

## Why It's Good For The Game
all species should be able to go all jobs

## Changelog
:cl:
add: vampires no longer burn in the chapel if they signed up as the chaplain
/:cl:


timothyteakettle:
the alternative here would be to remove the chapel burning from roundstart vampires, if someone really wants to do a chaplain-vampire gimmick (which is also fine, personally)

timothyteakettle:
removal is infact gay, and so i think this is like a million times better

PR #14274, opened 2021-02-20
lets ghost cafe residents disguise themselves as any mob or object

timothyteakettle:
## About The Pull Request
title
it's visuals only, it replaces your sprite
it's activated through an action button (using the button a second time removes the disguise)
it asks you if you're searching an object or mob, then proceeds to ask you for a search term, then displays the results (searches are done through paths, i.e. if you type alien it'll show you all paths with 'alien' in them)
and yes, it's been tested

## Why It's Good For The Game
freedom in ghost cafe

## Changelog
:cl:
add: ghost cafe residents can now disguise themselves as any mob or object
/:cl:

### DIFF
--- a/code/game/objects/structures/ghost_role_spawners.dm
+++ b/code/game/objects/structures/ghost_role_spawners.dm
@@ -668,6 +668,34 @@
 	icon_icon = 'icons/mob/actions/actions_changeling.dmi'
 	background_icon_state = "bg_mime"
 	var/currently_disguised = FALSE
+	var/static/list/mob_blacklist = typecacheof(list(
+		/mob/living/simple_animal/pet,
+		/mob/living/simple_animal/hostile/retaliate/goose,
+		/mob/living/simple_animal/hostile/poison,
+		/mob/living/simple_animal/hostile/retaliate/goat,
+		/mob/living/simple_animal/cow,
+		/mob/living/simple_animal/chick,
+		/mob/living/simple_animal/chicken,
+		/mob/living/simple_animal/kiwi,
+		/mob/living/simple_animal/babyKiwi,
+		/mob/living/simple_animal/deer,
+		/mob/living/simple_animal/parrot,
+		/mob/living/simple_animal/hostile/lizard,
+		/mob/living/simple_animal/crab,
+		/mob/living/simple_animal/cockroach,
+		/mob/living/simple_animal/butterfly,
+		/mob/living/simple_animal/mouse,
+		/mob/living/simple_animal/sloth,
+		/mob/living/simple_animal/opossum,
+		/mob/living/simple_animal/hostile/bear,
+		/mob/living/simple_animal/hostile/asteroid/polarbear,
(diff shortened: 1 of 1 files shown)


BlueWildrose:
finally, I can ERP with someone pretending to be a chair.

Add agent IDs to ghost and you'll likely make things mostly perfect.

timothyteakettle:
yeah you can change name at the mirror, forgot about ID, will look into giving them agent IDs soon

silicons:
OK OKOK OK HOLD ON 
To be entirely blunt why would we merge this given our latest spiel of animalfucking drama.

timothyteakettle:
good point (i entirely forgot about that as i was just thinking 'wouldnt it be cool if people could turn into xenos and other cool things in the cafe')

solution: i could just make it a whitelist of mobs you can turn into, instead

timothyteakettle:
alright it now blacklists all real-life animals, i think i've covered them all but someone else should also go through the list to make sure i didn't miss anything

silicons:
i'm fine with this now but saying it now if more real-life or meant to be real life replica mobs get added i'm immediately adding them to the blacklist and any whining about it will fall on deaf ears because we're not having a repeat.

timothyteakettle:
yeah the point of this pr wasn't to let people do that, preferably they can add them under /mob/living/simple_animal/pet so that they're automatically blacklisted as that's where they /should/ go

code/game/objects/structures/ghost_role_spawners.dm
```diff
@@ -662,6 +662,72 @@
...
+		/mob/living/carbon/monkey,
+		/mob/living/simple_animal/hostile/gorilla,
+		/mob/living/carbon/alien/larva,
+		/mob/living/simple_animal/hostile/retaliate/frog
+	))
+
+
+/datum/action/disguise/Trigger()
+	var/mob/living/carbon/human/H = owner
+	if(!currently_disguised)
```
LetterN, line 702: please stop making paragraphs of code, and have it early return

code/game/objects/structures/ghost_role_spawners.dm
```diff
@@ -662,6 +662,72 @@
...
+		var/user_object_type = input(H, "Disguising as OBJECT or MOB?") as null|anything in list("OBJECT", "MOB")
```
LetterN, line 703: `/obj/screen` :)

LetterN (changes requested):

code/game/objects/structures/ghost_role_spawners.dm
timothyteakettle: ill get round to this, bad habit from teaching self to code

code/game/objects/structures/ghost_role_spawners.dm
timothyteakettle: uh, will this make any difference considering the icon will just be empty, just like lots and lots of existing objects

PR #14276, opened 2021-02-20
adds a neutral quirk 'food critic' which lets you see food quality on examine

timothyteakettle:
## About The Pull Request
for anyone who isn't aware food has quality which scales off things like machine upgrades, family knife, cutting your meat in the kitchen (or using an upgraded gibber), high potency ingredients, etc

it's just some nice flavour stuff to be able to see the exact quality i guess.

(and yes, this is tested.)

## Changelog
:cl:
add: adds a neutral quirk 'food critic' which lets you see food quality on examine
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -218,6 +218,8 @@
 #define TRAIT_LIVING_NO_DENSITY			"living_no_density"
 /// forces us to not render our overlays
 #define TRAIT_HUMAN_NO_RENDER			"human_no_render"
+/// allows you to see food quality by examining
+#define TRAIT_FOOD_CRITIC		"food_critic"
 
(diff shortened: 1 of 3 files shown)


DeltaFire15:
Should probably be a 1 point quirk considering it simply gives you an advantage and there is zero reason not to take the quirk, as opposed to all other 0 point quirks.

timothyteakettle:
it's not really an advantage which is why i put it as 0 points, you can tell the quality of the food by just biting into something you wouldn't normally like, and if you like it, hey presto it's above 90

Linzolle:
alternatively you could make this a trait that cooks start with perhaps

timothyteakettle:
not a bad idea i think i'll do that

PR #14285, opened 2021-02-21
makes tcg cards and binders small and thus easier to store for persistence

ghost:
## About The Pull Request
this is literally a two line value change for a half broken feature so i can't be bothered to test

## Why It's Good For The Game
i want to carry my binder without having to get rid of gamer gear

## Changelog
:cl:
tweak: cards and card binders are now small-class items
/:cl:

### DIFF
--- a/code/modules/tcg/cards.dm
+++ b/code/modules/tcg/cards.dm
@@ -66,6 +66,7 @@
 	desc = "A flipped TCG-branded card."
 	icon_state = "cardback"
 	icon = 'icons/obj/tcg/pack_1.dmi'
+	w_class = WEIGHT_CLASS_SMALL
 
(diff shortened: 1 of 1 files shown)


code/modules/tcg/cards.dm
```diff
@@ -66,6 +66,7 @@
 	desc = "A flipped TCG-branded card."
 	icon_state = "cardback"
 	icon = 'icons/obj/tcg/pack_1.dmi'
+	w_class = WEIGHT_CLASS_SMALL
```
SmArtKar: Cards already have 	w_class = WEIGHT_CLASS_TINY

ghost: ok i just realized the var is just below this because i speedread

PR #14292, opened 2021-02-23
Added an entirely new wetness system

Putnam3145:
## About The Pull Request

Reworks water reagent, water vapor to add water/ice to a turf instead of just sorta generically making it slippery. This water will freeze, melt, evaporate and condense realistically, depending on the temperature and pressure of the surrounding environment.

## Why It's Good For The Game

I was asked to make saunas work. Some people were under the impression that they used to work, but they didn't. This is what is required to get them to work.

## Changelog
:cl:
add: A full rework of wetness for water/ice
tweak: Water vapor's reaction with turfs now wets, instead of just making the water vapor disappear
/:cl:

### DIFF
--- a/code/datums/components/wetness.dm
+++ b/code/datums/components/wetness.dm
@@ -1,6 +1,6 @@
 #define ICE_SPECIFIC_HEAT 40 // joules/mole-kelvin
 #define WATER_SPECIFIC_HEAT 75
-#define WATER_ENTHALPY_OF_VAPORIZATION		40000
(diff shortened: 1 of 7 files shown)


YakumoChen:
Make sure this doesn't cause any kind of weird interactions with slimes and whatever else doesn't like being wet.

timothyteakettle:
slimes being 'wet' isnt a thing, the reagent reaction on mobs that are a type of slime makes them directly take damage

PR #14301, opened 2021-02-24
towering tendrils - a slight addition to the minimum spawn system

Hatterhat:
## About The Pull Request
the minimum spawn subsystem now drops Towering Necropolis Tendrils, which are basically tendrils but:
- unique signal: "Unnerving Signal" instead of "Eerie Signal"
- intended to be indestructible (max_integrity 1e31, resistance_flags INDESTRUCTIBLE)
- spawn a new mob every 20 seconds instead of every 30

other than that basically nothing else
## Why It's Good For The Game
three less chances for lords of gunk to smash the funny heehoo tendrils and get quirky tendril loot
## Changelog
:cl:
add: The tendrils placed by the minimum spawn subsystem are now indestructible "towering necropolis tendrils".
/:cl:

### DIFF
--- a/code/controllers/subsystem/minimum_spawns.dm
+++ b/code/controllers/subsystem/minimum_spawns.dm
@@ -11,9 +11,9 @@ SUBSYSTEM_DEF(min_spawns)
 	var/list/valid_mining_turfs_2 = list() // snaxi underground turfs
 
 GLOBAL_LIST_INIT(minimum_lavaland_spawns, list(
-	/obj/structure/spawner/lavaland,
-	/obj/structure/spawner/lavaland/goliath,
-	/obj/structure/spawner/lavaland/legion,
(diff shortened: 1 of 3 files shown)


DeltaFire15:
I do not think this is needed. Coordinating with other miners for which tendrils to kill, aswell as the results of not doing so isn't exactly an issue that requires coding out.

Hatterhat:
it's literally three slightly different tendrils just strewn around replacing the basic three tendrils tossed about by the minimum spawn system and it's also an idea that i kind of like, to emphasize the nature of lavaland being a remarkably scuffed place

i also wanted to make the tendrils, like, 2x scale or whatever to really make them distinct but then i couldnt find the vars for it

silicons:
why is this good for the game?

Hatterhat:
iunno i guess it was an interesting concept to me

PR #14307, opened 2021-02-24
Adds a new event: supermatter surge

Putnam3145:
## About The Pull Request

Adds an event that makes the supermatter power up a good deal for a short time. That's it. Only happens if there is a supermatter and it has been turned on, otherwise it won't try.

## Why It's Good For The Game

Mostly just trying to curb the anomaly spam. Plus, I think a tiny wrench thrown into SM operations is fun.

## Changelog
:cl:
add: "Supermatter surge" event, which might cause problems if the supermatter is not sufficiently cooled (i.e. the setup is messed up in some way)
/:cl:


dzahlus:
oh boy i cant wait for incompetent engies to blow up the engine once again! cool PR though unironically

PR #14319, opened 2021-02-26
enforces SHOULD_NOT_SLEEP on Life(), fixes some more process() sleeps that might be causing controller lockups

silicons:
/process() enforcement too soon (tm) but we'd need to rewrite the entire chemistry system so 🥴


Linzolle (approved):

DeltaFire15:
Merge conflicting.

PR #14323, opened 2021-02-26
Hierophant staff no longer deals 30 damage

dzahlus:
## About The Pull Request
Hierophant melee does 15 damage 50AP, now does 15 on all attacks

## Why It's Good For The Game
We all know the hierophant does 15 damage and 100AP right? Wrong, it actually has 50AP and deals 30 damage on any attack that isn't melee, let's bring it in line with what people think it is. Mining power gamers no longer get an e-sword bullethell minigame

## Changelog
:cl:
balance: rebalanced hierophant STAFF to do 15 damage on all attacks
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -1282,8 +1282,8 @@
 			if(isliving(target) && chaser_timer <= world.time) //living and chasers off cooldown? fire one!
 				chaser_timer = world.time + chaser_cooldown
 				var/obj/effect/temp_visual/hierophant/chaser/C = new(get_turf(user), user, target, chaser_speed, friendly_fire_check)
-				C.damage = 30
-				C.monster_damage_boost = FALSE
+				C.damage = 15
+				C.monster_damage_boost = TRUE
 				log_combat(user, target, "fired a chaser at", src)
 			else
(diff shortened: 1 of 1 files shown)


HeroWithYay:
Excuse me sir, my brain is not very high quality.

Is this a nerf to the boss or a nerf to the item?

dzahlus:
> Excuse me sir, my brain is not very high quality.
> 
> Is this a nerf to the boss or a nerf to the item?

The hierophant STAFF, the item, sorry I wasn't clear in the title on opening the PR

silicons:
i'm iffy on a hard nerf to the only item that takes more brainpower to use out of all the gamer loot megafauna drop

why aren't we dropping AP instead and damage by a lower amount if needed?

dzahlus:
> i'm iffy on a hard nerf to the only item that takes more brainpower to use out of all the gamer loot megafauna drop
> 
> why aren't we dropping AP instead and damage by a lower amount if needed?

15 damage 50 AP is really good, considering it does burn damage. However the reason I put everything to 15 instead of 30 it was before is because oh god oh fuck 30 damage bullethell minigame with chaser spam. hrnngngh legs never have armor anyways so its 30 damage for free regardless of reducing AP. The only real part when the AP applies is to miners because they have leg armor, or if you melee which already does 15 damage regardless.

Issue #13650, opened 2020-11-01
Pacifist Zombie Cant Hurt


TripleShades:
Issue reported from  Round ID: 24902 (Citadel Station)

Reporting client version: 513.1536

When turned into a zombie (via romerol if that makes any difference) as a pacifist, you are unable to use your claw attack to harm anyone.
I was able to do so later, after cloning and being re-turned, but I have a feeling it erased my pacifist nature somehow doing so.

PR #14327, opened 2021-02-26
fixes pacifist zombies

ghost:
## About The Pull Request

Fixes Infectious Zombies not being able to attack living beings if the character has the "Pacifist" quirk. Relevant Issue #13650
Also adds a way to handle blacklisted quirks for each species, saving the ones that got removed to be restored when you transform from one species to the other. In case a similar bug ever occurs again.
Any optimization and feedback is appreciated this is my first time doing this.

## Why It's Good For The Game
Mean zombies NEED to hurt to be fun

## Changelog
:cl:
fix: fixed infectious zombies not being able to attack if host was pacifist
add: adds a way for species to have blacklisted quirks, the removal, and restoration of said quirks upon species changes
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -82,7 +82,7 @@
 	SEND_SIGNAL(user, COMSIG_MOB_ITEM_ATTACK, M, user)
 	if(item_flags & NOBLUDGEON)
 		return
-	if(force && damtype != STAMINA && HAS_TRAIT(user, TRAIT_PACIFISM) && !is_species(user, /datum/species/zombie/infectious))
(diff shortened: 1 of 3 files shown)


ghost:
i hope this is not "hacky code"
Object Oriented is one thing. Trying to understand DM is a whole another can of worms

timothyteakettle:
the ideal way to solve this would be to not check on every attack if they're not a zombie, but instead, just remove their pacifism when they become a zombie

ghost:
hm... that also means having to restore that when they are cured, if they get cured... time to tackle that

ghost:
got a small question though, "ROUNDSTART_TRAIT" source has comment saying "cannot be removed without admin intervention"
what does that mean

PR #14332, opened 2021-02-27
Readds exhibitionist as a trait

Putnam3145:
## About The Pull Request

Adds an entirely new implementation of exhibitionist. Previously this was a preference toggle (??) which made it so that every mob near you (?!) increased your arousal if you had exposed bits. Instead, it now has a *probability* to increase your arousal if you have exposed bits *and someone who has the ability to see them examines you*.

## Why It's Good For The Game

Personally I think it's kind of an iffy implementation that might cause weirds, but it's all logged so at worst it's not completely horrible. Besides that, I always felt kinda bad for removing it.

## Changelog
:cl:
add: Exbihitionism, as a trait
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -125,7 +125,7 @@
 							exhib_happened = TRUE
 						for(var/g in genits) // don't think this is O(n^2) and n is never more than... 4? anyway
 							var/obj/item/organ/genital/G = g
-							to_chat(M, "<span class='userlove'>[G.arousal_verb]!</span>")
+							to_chat(src, "<span class='userlove'>[G.arousal_verb]!</span>")
 				. += "[dicc.desc]"
 	if(CHECK_BITFIELD(user.client?.prefs.cit_toggles, VORE_EXAMINE))
 		var/cursed_stuff = attempt_vr(src,"examine_bellies",args) //vore Code
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -114,8 +114,18 @@
 
 	//CIT CHANGES START HERE - adds genital details to examine text
 	if(LAZYLEN(internal_organs) && CHECK_BITFIELD(user.client?.prefs.cit_toggles, GENITAL_EXAMINE))
+		var/exhib_happened = FALSE
 		for(var/obj/item/organ/genital/dicc in internal_organs)
 			if(istype(dicc) && dicc.is_exposed())
+				if(HAS_TRAIT(src, TRAIT_EXHIB))
+					var/list/genits = src.adjust_arousal(5, "exhibitionism")
+					if(length(genits))
+						if(!exhib_happened)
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -114,8 +114,18 @@
...
 			if(istype(dicc) && dicc.is_exposed())
+				if(HAS_TRAIT(src, TRAIT_EXHIB))
+					var/list/genits = src.adjust_arousal(5, "exhibitionism")
+					if(length(genits))
+						if(!exhib_happened)
+							to_chat(src,"<span class='love'>You're getting some attention...</span>")
+							exhib_happened = TRUE
+						for(var/g in genits) // don't think this is O(n^2) and n is never more than... 4? anyway
+							var/obj/item/organ/genital/G = g
+							to_chat(M, "<span class='userlove'>[G.arousal_verb]!</span>")
```
SandPoot:
```suggestion
							to_chat(src, "<span class='userlove'>[G.arousal_verb]!</span>")
```

moltoretardo:
I don't think giving people more reasons to be naked in public a good idea.

ghost:
> I don't think giving people more reasons to be naked in public a good idea.

I don't think giving people more reasons to be naked in public a good idea.

BlueWildrose:
An arousal quirk doesn't really mean people are going to actively go naked in public with this enabled. All it does is add flavor and mechanically gives your character arousal when you're examined while naked, that's all.

If nothing else this might encourage open nude rooms, which isn't really a problem if it's clearly labeled as such.

ghost:
adding perks that benefit you for doing a particular action will encourage people to perform that particular action

as such adding a perk that gives you cool sex messages for breaking the rules will encourage people to break the rules

moltoretardo:
Low RP behaviours shouldn't be encouraged via perks. It's a pretty good rule 9 breaking bait imo.

BlueWildrose:
Yeah alright. What Putnam can do is do something similar to the quirk, Reality Disassociation Disorder (or something along those lines) that gives your character constant hallucinations: add a big fat huge text policy that pops up whenever the character joins in the round.

If they start being a creep despite that, that's entirely on them from that point forward.

silicons:
@Putnam3145 can you make sure the user is exceedingly informed that this is not a reason to break rules on being a creep and that admins will get on their case if they use this as a reason to run around naked? i'd rather not have to deal with that again.

DeltaFire15:
Yeeeah I don't like this at all and have been putting off this PR for too long.
According to Putnam, the only reason for this is because they felt bad about removing it before.
I do not see this as enough of a reason to add this specific trait which does nothing save for encouraging questionable stuff.
Given that the only use for this are ERP-specific scenarios (anything else being a big fucking yikes), while making room for some really discomforting stuff, I see no reason to (re)add this to the game. This falls under the 'yeah ok if you are into that feel free to RP it *with people who oocly consented to it as to not cause issues with rule 9*, but as an actual trait it'd be a yikes' category.

Therefore, closing this PR.

PR #14337, opened 2021-02-28
nerfs cqc (probably)

kiwedespars:
## About The Pull Request

nerfs cqc by lowering the hardstun on the aggressive pin, because it's a force disarm and a forced aggrograb anyways, nerfing chokehold into a mute rather than a 30 second sleep, and making it so you can't hardstun baton users by blocking them with throwmode.

## Why It's Good For The Game

citcombat is moving away from hardstuns, cqc is one of the strongest things in the game, and it doesn't take a brain to use.

it's still incredibly strong, this change just makes it a bit more bearable rather than a instawin.

## Changelog
:cl:
aggressive pin hardstun 10 seconds -> 1 second
chokehold now mutes instead of knocking people out for **30** seconds
blocking things no longer hardstuns the person who gets hit for 8 seconds, just blocks the attack.
/:cl:


code/datums/martial/cqc.dm
```diff
@@ -97,7 +97,7 @@
 		D.visible_message("<span class='warning'>[A] locks [D] into a restraining position!</span>", \
 							"<span class='userdanger'>[A] locks you into a restraining position!</span>")
 		D.apply_damage(damage, STAMINA)
-		D.Stun(100)
+		D.Stun(10)
```
necromanceranne, line 100: I think the reason this exists is because otherwise, CQC is still at the mercy of RNG grabs. Of course, the core issue with Restrain is that it is a combo, and because it is a combo, even if the initial grab fails and they break free, the second grab, which completes the restrain combo, will always succeed and stun.

This isn't necessarily a problem with CQC specifically, but actually a problem with grabs being extremely all-or-nothing.

code/datums/martial/cqc.dm
```diff
@@ -196,7 +196,8 @@
 		log_combat(A, D, "knocked out (Chokehold)(CQC)")
 		D.visible_message("<span class='danger'>[A] puts [D] into a chokehold!</span>", \
 							"<span class='userdanger'>[A] puts you into a chokehold!</span>")
-		D.SetSleeping(400)
+		if(D.silent <= 10)
```
necromanceranne, line 199: Not entirely sure what this accomplishes. You can get krav maga gloves and achieve basically this without the fuss. You should consider some damage on top of this. Possibly stamina damage. Though in all honesty you probably would be better off killing someone expediently with your other moves.

necromanceranne (commented):
CQC has an absolutely absurd amount of jank in it that needs to get ripped out. Number tweaks don't do much here. You can simply bypass this nerf with the ability to slam and then headkick someone asleep, all the while killing the target in the process. Another consideration is that CQC isn't meant to really be fair on the person being attacked either. If you're fighting a CQC user in a joust, you're already making a terrible mistake.

If the goal was to remove hard stuns all together, you should kill CQC off in it's entirety instead of pruning it into worse state, since you'll quickly find how little is beneath the surface for this martial art. It's rework or nothing, I feel.

timothyteakettle:
sad because this removes chef's ability to non-lethally remove people from the kitchen, consider making the chef's version just do stamina damage

dzahlus:
From personal experience CQC suffers from the same jank as old SCarp did where you could just walk out of the aggrograb. Rework or nothin' is what I personally say. Since it's nukie only it has to be powerful. I say make the funny stun at the end of a combo, not the initiation of it for a band aid fix. As for the cook give em krav maga instead.

edit: though in the end the problem with martial arts is that stunbatons will always be superior :). The only martial art that doesn't suffer from this is rising bass because it has a one click disarm. Same with krav maga.

code/datums/martial/cqc.dm
kiwedespars: mute and they're already in a really bad position in a microstun + aggrograb, enough to where you could probably take them out on that, a little bit of stamdamage would probably be good though.

code/datums/martial/cqc.dm
kiwedespars: admittedly yeah, it still exists because rng grabs are cruel and usually will just end up irritate whoever's using the martial art, this gives it a little bit more of a base to work with while still giving someone the chance to escape shortly after. combo-stunning is goofy but i think this should make it far better at fighting single targets, rather than being able to chain-stun multiple people in quick succession, and keep them there for 10s per person.

kiwedespars:
> You can simply bypass this nerf with the ability to slam and then headkick someone asleep, all the while killing the target in the process. Another consideration is that CQC isn't meant to really be fair on the person being attacked either.

I'm aware of that, and thinking it over I just think adding _some_ counterplay to CQC is a lot better and gives reason to actually bother with other things in CQC other than just grabspamming and having throwmode on to become invulnerable to all forms of melee. If this forces people to do something other than sprint at people, spamming grab 

> sad because this removes chef's ability to non-lethally remove people from the kitchen, consider making the chef's version just do stamina damage

it also removes their ability to zero-risk take down any carbon in the kitchen. i think they can still remove people from the kitchen though, considering the stun is a second long and it's still a hardstun aggrograb.

> From personal experience CQC suffers from the same jank as old SCarp did where you could just walk out of the aggrograb. Rework or nothin' is what I personally say. Since it's nukie only it has to be powerful

Having two timers for resisting, and one being available from just spamming WASD is partially the reason why aggrograbs seem ineffective/instantly easy to break out of. Reworking CQC itself is something I don't trust myself to do, and this should just make it a bit more bearable to fight chefs (who will end your round if you come into their kitchen as a station-side antag with close to zero counterplay other than "don't") and nukies, who can become insanely durable with it due to stunbatons being ineffective, elite suit giving a good amount of armor, adrenals covering movespeed, and an e-shield that'll either make them immune to energy projectiles, or a regular shield for ballistics (which wouldn't do much anyways, unless it's a combat shotgun/smg with ap rounds)

timothyteakettle:
aside from agreeing with what necromancer said, uh, these nerfs are way too hard for something that's mainly used by nuke ops, and the chef in the kitchen, this pr shouldn't even exist because of the upcoming combat rework

ghost:
saying that development shouldn't happen because someone might be doing development in the future is sus

dzahlus:
AHAHAHAH WHY DOES IT GIVE YOU A 4 SECOND HARDSTUN FOR BEING CLICKED ON IN THROWMODE HOLY FUCK MERGE THIS

PR #14353, opened 2021-03-02
MWS-01 Changes

dzahlus:
## About The Pull Request
Makes the MWS-01 alternative hosgun a viable choice by giving it 2 magazines and more mictobatteries

## Why It's Good For The Game
MWS-01 was hilariously bad compared to the hosgun, now it starts off with two micro battery holders (the magazine) and one more of each laser and disabler microbattery (3 for both now instead of 2) as well as a single taser battery (if you dont put 1 taser and 2 laser batteries inside for TASE LASE you don't deserve this gun). Disabler battery now has 10 shots instead of 6+1 like any other egun.

## Changelog
:cl:
add: Added taser microbattery for MWS-01
tweak: tweaked MWS-01 beacondrop to have more batteries
balance: rebalanced MWS-01 disabler battery to fire 10 shots
soundadd: added unique sound to the MWS-01
spellcheck: fixed Modula Weapons System to "Modular Weapon System"
/:cl:


code/modules/projectiles/boxes_magazines/external/rechargable.dm
```diff
@@ -114,3 +115,9 @@
 	type_color = "#d084d6"
 	type_name = "<span class='ion'>ION</span>"
 	projectile_type = /obj/item/projectile/ion
+
+/obj/item/ammo_casing/mws_batt/taser
+	name = "'MWS' microbattery - TASER"
+	type_color = "#e5ff00"
+	type_name = "<span class='stun'>TASE</span>"
+	projectile_type = /obj/item/projectile/energy/electrode
```
zeroisthebiggay: is this not the electrode projectile that knocks people down and was removed from the HoS gun because it was the only ranged knockdown security had which made it blatantly overpowered compared to Literally Everything Else in the arsenal
pretty sure the balanced one is /obj/item/projectile/energy/electrode/security but the security electrode is just a slowdown so why would you ever use that over the disabler cartridge unless you totally knew what this was and wanted to make it Substantially Better than the multiphase

zeroisthebiggay (changes requested):

code/modules/projectiles/boxes_magazines/external/rechargable.dm
dzahlus: I totally didnt know what I was doing but the reason is literally tase > lase, I'll swap it out with the not funny electrode

PrivateJoe09:
Honestly, if I had to choose, I'd much rather have this sidearm and keep the WT than getting a energy shotgun.

PR #14361, opened 2021-03-04
Makes hyper-nob, nitryl easier to make

Putnam3145:
## About The Pull Request

Nitryl and hyper-nob require baffling amounts of temperature, essentially having fusion as a prerequisite. I'd rather this... not be the case. This does not change the reactions any, they still have the same effects, they're just easier to meet the conditions for.

Note that this means the supermatter can now spontaneously make nitryl if it gets spicy enough and it's N2/N2O cooled. Very fun.

EDIT: ALSO FIXES A BUG THAT'S BEEN IN NITRYL SINCE WE **ADDED EXTOOLS** BUT I NEVER NOTICED BECAUSE **NOBODY MAKES NITRYL LOL**

## Why It's Good For The Game

Nitryl is not... made. Ever. Specifically because it requires, I shit you not, 149,260 kelvins for a reaction which has a baseline temperature of... 37315. Hyper-nob is similarly ridiculous, using 2 gigajoules of heat but for some reason requiring *5,000,000* kelvins to start.

## Changelog
:cl:
fix: nitryl now consumes oxygen/nitrogen instead of generating them
balance: Hyper-nob and nitryl are easier to make.
/:cl:

### DIFF
--- a/code/modules/atmospherics/gasmixtures/reactions.dm
+++ b/code/modules/atmospherics/gasmixtures/reactions.dm
@@ -358,7 +358,7 @@
 		/datum/gas/oxygen = 20,
 		/datum/gas/nitrogen = 20,
 		/datum/gas/nitrous_oxide = 5,
-		"TEMP" = FIRE_MINIMUM_TEMPERATURE_TO_EXIST*400
+		"TEMP" = FIRE_MINIMUM_TEMPERATURE_TO_EXIST*25
 	)
 
 /datum/gas_reaction/nitrylformation/react(datum/gas_mixture/air)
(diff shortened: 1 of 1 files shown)


GrayRachnid:
what's the new temps

Putnam3145:
> 
> 
> what's the new temps

9328.75 for nitryl and no specific temp for hyper-nob, you just need enough energy, which is equivalent to pressure/volume so, like, uh, gimme a sec... using the rule-of-thumb that 1 mole is 1 kPa at 300 kelvins in 2500 liters (this is only 0.3% off), you need at least 5 moles trit and 10 moles nitrogen at a pressure of at least 26,667 kPa, which you can get either by having enough temperature OR by having enough moles, though that assumes the 2 : 1 ratio of nitrogen to trit

PR #14373, opened 2021-03-07
Makes has_tail have a simpler check, probably fixing tail intertwining

Putnam3145:
## About The Pull Request

See title. Tail intertwining isn't working cause only mammalians (?) actually pass the check. This replaces it with a watered-down version of the can_wag_tail check, which I know works.

## Why It's Good For The Game

fetish content

## Changelog
:cl:
fix: Made tail intertwining work.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -264,10 +264,10 @@
 	return FALSE
 
 /mob/living/carbon/human/has_tail()
-	if(!dna || !dna.species)
+	if(!dna?.species?.tail_type)
 		return ..()
-	var/list/L = dna.species.mutant_bodyparts		// caches list because i refuse to type it out and because performance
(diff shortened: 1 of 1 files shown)


timothyteakettle:
what i said in the discord: you need to check if the value is not null and not "None", your current new proc doesn't do this and so will return "None" when people have no tail, which in turn makes it think they have a tail because it didn't return null

Putnam3145:
> 
> 
> what i said in the discord: you need to check if the value is not null and not "None", your current new proc doesn't do this and so will return "None" when people have no tail, which in turn makes it think they have a tail because it didn't return null

my character has a tail and her tail is "None", so i think there might be a misunderstanding somewhere with how the system actually works

timothyteakettle:
reminder change this to features not mutant bodyparts as what was talked about in the discord

PR #14405, opened 2021-03-10
Makes vampries dust when they can't pay the blood cost, not before

Arturlang:
## About The Pull Request
 Makes vampire the race only get dusted if they're cant pay the blood toll, plus makes dusting drop their items on the ground. Plus removes the span in their statpanel thing which are no longer needed
Makes vampires have the nomarrow trait, just like bloodsuckers do, so they don't suffer from low blood effects, but they'll still die once blood reaches zero

## Why It's Good For The Game
The current amount can be a bit sudden, and unexpected, let people have the chance to save vampires, and it's rather fucky to completely have items destroyed due to a vampire dying.

## Changelog
:cl:
tweak: Vampires no longer suffer from low blood effects, and sets their species blood loss to 0.5, due to them not benefitting from the blood regeneration that they used to from handle_blood
tweak:Vampires the race only get dusted if they cant pay their blood toll, which is 0.5, instead of 125 or so, and no longer delete their inventories upon dusting due to lack of blood
fix: Vampire statpanel no longer shows spans unnecesary
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/vampire.dm
+++ b/code/modules/mob/living/carbon/human/species_types/vampire.dm
@@ -53,7 +53,7 @@
 		C.adjustOxyLoss(-4)
 		C.adjustCloneLoss(-4)
 		return
-	if(C.blood_volume < 0.75)
+	if(C.blood_volume > 0.75)
 		C.blood_volume -= 0.75 //Will take roughly 19.5 minutes to die from standard blood volume, roughly 83 minutes to die from max blood volume.
 	else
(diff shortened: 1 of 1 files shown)


DeltaFire15:
I dunno I think it's pretty okay for them to get dusted if they have less bloodvolume than is the threshold to survive (something around like ~20-30% iirc), considering being lower than that volume just instakills people anyways.

Arturlang:
@DeltaFire15 it's mostly so medical can TRY to save them, instead of them almost always dusting in their care like now

PR #14409, opened 2021-03-10
Alternative Energy Bow Sec Rework

keronshb:
## About The Pull Request

Alternative to #14407 

This increases the cooldown from 2 > 4 seconds

Removes the knockdown

Lessens the eye effect

Removes the slur effect

Reduces stam damage from 60 > 30

Increases mat cost and exchanges silver for diamond

## Why It's Good For The Game

Instead of the nuclear option of removing Sec EBows, this balances them for being mass printable and easily hidden in Bluespace Body Bags/Being dragged/put in BOH

## Changelog
:cl:
balance: 2 > 4 second firing time
balance: 60 > 30 stamina damage
balance: 50000 Metal and 15000 diamond cost instead of 5k metal and 1.5k silver
balance: No knockdown
balance: Reduced eye effect (so you can still tell you were hit)
balance: Removed the slur effect
/:cl:


silicons:
this makes them shittier, bulky disabler-lasers with a combined dps of 12, and that's counting both stam and tox.
stop making prs while you're upset about a round and genuinely think about what the problem about it is
i'd actually be onboard to replace or remove them wholesale if the pr wasn't made at the exact round you died to one

not to mention being catgirl-tier in balancing with the mat costs

DeltaFire15:
.. Thats literally as much of a material cost as an *entire* Gygax, if not more depending on stock parts.

necromanceranne:
Yo man I getcha, I hate it too. But I ask this.

Why not delete it?

Tupinambis:
While I support nerfing them in some way this is less a nerf and more a tactical nuke. 

I'd personally just increase the cooldown and mat costs somewhat (not 10 times good lord) so it's a high-risk high reward version of a disabler or something to that effect.

PR #14427, opened 2021-03-13
BoH volume nerf

YakumoChen:
## About The Pull Request
The volume of bags of holding has been reduced to match exactly double the volume of a regular backpack

## Why It's Good For The Game
Arguably the biggest strength of Holding is that it can store large items like hardsuits, lavaland staves, and big guns like pump shotguns. However the current volume is also nearly triple a typical backpack as well. 

Bag of holding's storage currently feels nearly endless, the amount of stuff you can store feels excessive and the Bag of Holding I do not feel has enough downsides to outweigh the frankly obscene amount of stuff it can hold. For how relatively easy it is to get one and how readily they are given out, I think it deserves to be toned down.

The only main downside (it fucks with teleports) isn't very strong. Teleporting is not common and can in fact be used to the user's advantage to evade pursuit, making it in some cases a boon rather than a downside.

This tones it down a fair bit while keeping it still a substantial upgrade to a backpack.

## Changelog
:cl:
balance: Bag of Holding volume is now double a backpack instead of almost triple.
/:cl:


silicons:
stating the facts on how big it is, is not a valid "why is it good for the game"

YakumoChen:
@silicons 
# Why It's Good For The Game
I want Skylar Lineman to be forced to hold 6 less guns at any given time

also I added some reasons to the OP

TheObserver-sys:
I'm not sold on this. At this point, why not also just port over the "you need a bluespace anomaly core." bit too so that they're turned from nice item to "Oh, you found hilberts, then?"

YakumoChen:
In almost all use cases, BoH is a straight upgrade unless you are one of the 2 roles in the game who are expected to use do_teleport more than once every 5 rounds (the captain and RD). This makes it less of a straight upgrade.

 Is there any counter or reason the Bag of Holding needs to hold _20_ normal size items rather than _14_?

Are any jobs significantly impacted by reduction of storage space?

RogueSteampunker:
> In almost all use cases, BoH is a straight upgrade unless you are one of the 2 roles in the game who are expected to use do_teleport more than once every 5 rounds (the captain and RD). This makes it less of a straight upgrade.
> 
> Is there any counter or reason the Bag of Holding needs to hold _20_ normal size items rather than _14_?
> 
> Are any jobs significantly impacted by reduction of storage space?

Well of course it's a straight upgrade, it's a late-tier techwebs node that requires a heavy toll of resources just to print a few of them.

It genuinely just feels like there's no real point behind any of this change, besides some vague notion that it's "too big".

Tupinambis:
I'm a strong believer in the idea that most of science late-round power game stuff is obnoxious and bad for both gameplay and balance, but that's a matter of how science and techfabs work rather than an issue specifically with bags of holding. 

Nerfing bags of holding won't fundamentally fix anything.

silicons:
yeah this is just not needed

PR #14415, opened 2021-03-11
Adjusts Thermals in Genetics and Thermals/Xray in techwebs.  Makes Thermals/Space Adapt Mutually Exclusive

keronshb:
## About The Pull Request

This adjusts thermals in genetics to:

1) Be POSSIBLY more difficult to obtain (from 18 > 24 difficulty).  Which means it's possible to have up to 24 blocks.

2) Be unusable with Space Adaptation (mutually exclusive)

3) 25 > 40 Instability

This also adjusts thermals in techwebs to:

1) ONLY be printable from mechfabs

2) 2500 > 7500 point cost

3) Removes XRay from Combat Implants and puts them into Illegal Combat Implants

4) 6s > 12s print time

## Why It's Good For The Game

This isn't the full edit I wanted to do as I wanted to get some sort of layering effect to work with thermals, but there's an increasing push to get these smaller changes out now.  So I'm putting these forward while I work on another alternative.

Thermals have become a rampant issue again where either genetics hand them out with Space Adapt like candy to sec when it's unwarranted, or even the entire crew.  This isn't the nuclear option of completely removing it from genetics, but at least balances it out so it can't be used with Space Adapt and Sec is going to have to be mindful about using it with flashbangs.

Also XRays are entirely behind illegal tech because we're not really suppose to have crew side XRays.  I didn't remove them at the risk of the option being too nuclear.

## Changelog
:cl:
add: Adds Illegal Combat Implants for XRays, 10000 points and requires Illegal Tech
balance: 2500 > 7500 cost for Combat Implants
balance: Removed Thermals and XRays from Medilathes.  They're in Mechfabs only
balance: Increased time to print from 6s > 12s
balance: 25 > 40 Genetic instability for thermals in genetics
balance: Makes Thermals and Space Adapt Mutually exclusive
balance: -2 flash protection for both versions of thermals
balance: 18 > 24 difficulty
balance: removed see in dark for thermal robo eyes only
/:cl:

PR #14430, opened 2021-03-13
Adjusts Thermals in Genetics; Moves X-ray implant to illegal tech.

YakumoChen:
## About The Pull Request

closes  #14415

Makes some minor nerfs to Thermal Vision from genetics
It is a recipe now. Recipe is Near-sighted + Stimmed. 
because it's a recipe you can no longer have it activated for free. Instability raised to 40.
Stimmed increased slightly in difficulty to match what Thermals was at.
X-ray vision mutation (not obtainable) had instability raised to 50, for parity.

Xray eyes is now illegal tech. No other changes.

## Why It's Good For The Game

#14415 has no idea what it's doing and just shotgunned a bunch of numbers to nuke vision techs because salt.

28 difficulty means *approximately 28 missing blocks* or *up to 14 missing pairs*. Obviously that makes Thermal vision more or less impossible to obtain from genetics. 71 instability is also not well considered and you can just bypass it by having Thermal rolled in your gene pool. Keron has me blocked so I can't exactly point this out.

This is a lot more sane and also means you can't cheese the instability of thermals by just rolling it in your natural genes. 

I could probably let the other PR just close because it's clearly untested code, but thermals is kinda gay so it could probably use a light nerf.

## Changelog
:cl:
balance: Genetics - Thermal vision is a recipe instead of a natural gene now. Nearsighted+Stimmed. Other minor nerfs, Thermal is 40 instability.
balance: Research - X-ray eyes are now an illegal tech.
/:cl:


timothyteakettle:
the changes here seem fine but overall alternatives are being discussed which would mean a change like this would be better off being delayed until talks on them have finished

timothyteakettle:
in response to the difficulty comment made in this PR, please see my comment about how difficulty functions in the other PR which you reference, it's a bit different to how it is expected to behave

YakumoChen:
18 is the current default for Thermals. From playing genetics myself, it typically rolls between 3-5 missing pairs, with a couple other mutations sharing similar difficulty. 

3 pairs is about the maximum you can brute force, going through every combination until one works.

To get to three you either have to be lucky, use Jokers and hope you are researching the right gene, or manually log other combinations of the same gene and slap them together. All of these can be very time consuming, and only sometimes very fast if you get lucky.

I dont feel anything more than maybe 20 difficulty is reasonable for gameplay balance. TG has not changed the difficulty of any mutations to make them harder, making 18 the upper reasonable bound I would expect it to be generally considered "hard but not impossible"

The-Sun-In-Splendour:
Delete genetics & X-ray

timothyteakettle:
I was more just wanting to suggest that you change the description as it's misleading, difficulty 28 does not mean 28 missing pairs, it means 28 blocks are replaced to be missing, but because they can replace missing blocks, it's 'up to' 28, not always 28 (and for all 28 to be unique would be incredibly rare)

DeltaFire15 (approved):
After discussion, looks good and should make the mutation somewhat more difficult to acquire, especially considering it can no longer be a 'free' mutation.

Behind the xray change aswell since why'd you ever print thermal eyes if xray ones exist. The illegal node lock should make them more rare than before (given, it wasn't exactly that common before, but not rare either)

DeltaFire15:
Merging this in favor of #14415

PR #14435, opened 2021-03-14
organic repair surgery can be done on a cyborg torso

TheEaterOfBeans:
## About The Pull Request

Title says it all.

## Why It's Good For The Game

Sometimes people need their torsos replaced due to infection, resulting in someone who has a cyborg torso but is otherwise organic.

## Changelog
:cl:
tweak: you can repair organic limbs on a person with a cyborg torso
/:cl:

### DIFF
--- a/code/modules/surgery/healing.dm
+++ b/code/modules/surgery/healing.dm
@@ -8,7 +8,7 @@
 
 	target_mobtypes = list(/mob/living/carbon/human, /mob/living/carbon/monkey)
 	possible_locs = list(BODY_ZONE_CHEST)
-	requires_bodypart_type = 0
+	requires_bodypart_type = NONE
 	replaced_by = /datum/surgery
 	ignore_clothes = TRUE
 	var/healing_step_type
--- a/code/modules/surgery/healing.dm
+++ b/code/modules/surgery/healing.dm
@@ -21,6 +21,12 @@
 					healing_step_type, //hehe cheeky
 					/datum/surgery_step/close)
 
+/datum/surgery/healing/can_start(mob/user, mob/living/carbon/target, obj/item/tool)
+	for(var/obj/item/bodypart/B in target.bodyparts)
+		if(B.is_organic_limb(TRUE))
+			return ..()
+
+
 /datum/surgery_step/heal
 	name = "repair body"
 	implements = list(TOOL_HEMOSTAT = 100, TOOL_SCREWDRIVER = 65, /obj/item/pen = 55)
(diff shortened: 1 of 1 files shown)


code/modules/surgery/healing.dm
```diff
@@ -8,7 +8,7 @@
 
 	target_mobtypes = list(/mob/living/carbon/human, /mob/living/carbon/monkey)
 	possible_locs = list(BODY_ZONE_CHEST)
-	requires_bodypart_type = BODYPART_ORGANIC
+	requires_bodypart_type = 0
```
silicons: NONE instead of 0, please!

DeltaFire15 (changes requested):
To prevent some issues (mainly, confusion for people who see both surgery types pop up on people with robot limbs, even if a full robot), please add an check akin to repair surgery for robots, just with the robot limb check replaced with a check for if there's at least one organic limb.
 (reference: ../code/modules/surgery/robot_healing.dm, proc: /datum/surgery/robot_healing/can_start)

TheEaterOfBeans:
> To prevent some issues (mainly, confusion for people who see both surgery types pop up on people with robot limbs, even if a full robot), please add an check akin to repair surgery for robots, just with the robot limb check replaced with a check for if it's an organic limb.
> (reference: ../code/modules/surgery/robot_healing.dm, proc: /datum/surgery/robot_healing/can_start)

...how would I check if OTHER limbs are organic? This is for people with mechanical torsos and an otherwise organic body.

DeltaFire15:
> > To prevent some issues (mainly, confusion for people who see both surgery types pop up on people with robot limbs, even if a full robot), please add an check akin to repair surgery for robots, just with the robot limb check replaced with a check for if it's an organic limb.
> > (reference: ../code/modules/surgery/robot_healing.dm, proc: /datum/surgery/robot_healing/can_start)
> 
> ...how would I check if OTHER limbs are organic? This is for people with mechanical torsos and an otherwise organic body.

Look at the reference I mentioned, it checks for if at least one limb is robotic and only does the other cheks / allows it if that'd be a case. here it'd basically just be the inverted version, aka checking for if at least one bodypart is overganic (ignoring hybrid bodyparts, so with the TRUE arg for the is_organic proc

The code part here is:
(of course with different intendation, look at the reference for that)
>/datum/surgery/robot_healing/can_start(mob/user, mob/living/carbon/target, obj/item/tool)
>    for(var/obj/item/bodypart/B in target.bodyparts)
>		if(B.is_robotic_limb())
>			return ..()

Copypaste it into a reasonable place in the file, with the correct path (so the normal heal surgery's instead of robot_healing), change the is_robotic_limb() to is_organic_limb(TRUE), and it should work

DeltaFire15 (approved):

PR #14436, opened 2021-03-14
Buffed supermatter surge to make it up to 50x as strong as before.

Putnam3145:
## About The Pull Request

Turns out it was undertuned... a lot. Live testing showed it doesn't even catch fire at the previous equivalent of a class *200* supermatter surge, when it only went up to *9*. Now it goes up to the previous equivalent of a class 400.

## Why It's Good For The Game

An event that does literally nothing is no fun at all.

## Changelog
:cl:
balance: Buffed supermatter surge massively.
/:cl:

### DIFF
--- a/code/modules/events/supermatter_surge.dm
+++ b/code/modules/events/supermatter_surge.dm
@@ -13,11 +13,27 @@
 	var/power = 2000
 
 /datum/round_event/supermatter_surge/setup()
-	power = rand(200,4000)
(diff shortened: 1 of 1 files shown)


DeltaFire15:
.. You are aware that it goes into tesla delam if it has > 5000 power, right? Just that it doesn't immediately fully convert matter power to power.

PR #14456, opened 2021-03-17
Added a "no antags please" pref

Putnam3145:
## About The Pull Request

Just in case you're not feeling it.

## Why It's Good For The Game

Because having to manually disable every antag role sucks.

## Changelog
:cl:
add: "Antag" role that can be toggled to disable all antags
/:cl:


timothyteakettle:
based so we don't have to keep going through every set of buttons each time

PR #14458, opened 2021-03-17
Buffs the Warden's Particle defender to something that isn't pure dumpster trash

GrayRachnid:
## About The Pull Request
This increases the stamina damage per shot from 13->15. The defender shoots 6 per and has a range of 6 tiles. Does this make it any better at range? No, it's still absolute shit at range. However, this puts it on the same level as rubbershot, which also deals 15 damage and has 6 pellets. This pr also rounds up the laser energy cost. It now has 8 disabler shots and 4 laser shots.

Pros: The particle defender is now not dog shit and you can carry it.
Cons: The particle defender now has a 2 shot stamcrit(previously 3 at full stam) and you can do this to 4 people if your pointblank skills are legendary.

## Why It's Good For The Game
The particle defender is so doo doo right now that it's more worth your time to grab a disabler or an energy from the armory compared to it. I know the warden already gets two very overpowering toys but if we're going to keep the particle defender then it ought to be at least on par with rubbershot.

## Changelog
:cl:
balance: buffed particle defender disabler shots from 13->15 stamina (15*6=90)
balance: buffed particle defender laser ammo to 4 shots
/:cl:

### DIFF
--- a/modular_citadel/code/modules/projectiles/guns/pumpenergy.dm
+++ b/modular_citadel/code/modules/projectiles/guns/pumpenergy.dm
@@ -189,5 +189,5 @@
 	name = "disabling blast"
 	icon_state = "disablerslug"
 	color = null
-	stamina = 16
+	stamina = 15
 	range = 6
--- a/modular_citadel/code/modules/projectiles/guns/pumpenergy.dm
+++ b/modular_citadel/code/modules/projectiles/guns/pumpenergy.dm
@@ -107,7 +107,7 @@
 
 /obj/item/stock_parts/cell/pumpaction	//nice number to achieve the amount of shots wanted
 	name = "pump action particle blaster power supply"
-	maxcharge = 1500
+	maxcharge = 1200
 
 //PUMP ACTION DISABLER
 
--- a/modular_citadel/code/modules/projectiles/guns/pumpenergy.dm
+++ b/modular_citadel/code/modules/projectiles/guns/pumpenergy.dm
@@ -107,7 +107,7 @@
 
 /obj/item/stock_parts/cell/pumpaction	//nice number to achieve the amount of shots wanted
 	name = "pump action particle blaster power supply"
-	maxcharge = 1200
+	maxcharge = 1500
 
 //PUMP ACTION DISABLER
 
@@ -153,7 +153,7 @@
 
 /obj/item/ammo_casing/energy/laser/pump
 	projectile_type = /obj/item/projectile/beam/pump
-	e_cost = 350
+	e_cost = 300
 	select_name = "kill"
 	pellets = 6
 	variance = 15
@@ -189,5 +189,5 @@
 	name = "disabling blast"
 	icon_state = "disablerslug"
 	color = null
-	stamina = 13
+	stamina = 16
 	range = 6


keronshb:
This was lowered for a reason because Wardens would take this out of the brig to valid hunt.

I don't think this buff is warranted at all.

GrayRachnid:
> 
> 
> This was lowered for a reason because Wardens would take this out of the brig to valid hunt.
> 
> I don't think this buff is warranted at all.

This was not 'changed before'. The particle defender before Zero changed it used to be a taser with knockdown and all the other taser effects. Ever since it was changed to a disabler, it was never touched and has stayed this way for almost a year now. It has been terrible for a year and it's about time it gets a facelift. Yes, this does make it a 2-hit KO on PB but at the end of the day it's still a disabler shotgun that is outshadowed by a riot shotgun with rubbershot.

keronshb:
> 
> 
> > This was lowered for a reason because Wardens would take this out of the brig to valid hunt.
> > I don't think this buff is warranted at all.
> 
> This was not 'changed before'. The particle defender before Zero changed it used to be a taser with knockdown and all the other taser effects. Ever since it was changed to a disabler, it was never touched and has stayed this way for almost a year now. It has been terrible for a year and it's about time it gets a facelift. Yes, this does make it a 2-hit KO on PB but at the end of the day it's still a disabler shotgun that is outshadowed by a riot shotgun with rubbershot.

Either the charge gets upped or the disabler gets upped, not both, or this becomes too strong of a weapon.

Mind you it can be fired with one hand IIRC so you don't want to have another shotgun+shield meta with the Particle Defender.

GrayRachnid:
Shields do not have passive block ever since blocking was added.

keronshb:
> 
> 
> Shields do not have passive block ever since blocking was added.

You can just hold F even with it on offhand.

GrayRachnid:
I'm of the firm belief that if anybody can make blocking work with all its current downsides and stamina draws then it shouldn't be punished.

I see your concern and understand why you think this weapon shouldn't have 10 disabler shots though. After further math crunching, i've discovered that the taser similarly has a number of expected targets it can take down and i've updated the defender to match it.

keronshb:
> 
> 
> I'm of the firm belief that if anybody can make blocking work with all its current downsides and stamina draws then it shouldn't be punished.
> 
> I see your concern and understand why you think this weapon shouldn't have 10 disabler shots though. After further math crunching, i've discovered that the taser similarly has a number of expected targets it can take down and i've updated the defender to match it.

Yeah I think this should be fine then.  I'm not a maintainer so I don't really have much pull but it just seemed a tad bit overtuned with both the shots + stam upgrade.

zeroisthebiggay:
> > This was lowered for a reason because Wardens would take this out of the brig to valid hunt.
> > I don't think this buff is warranted at all.
> 
> This was not 'changed before'. The particle defender before Zero changed it used to be a taser with knockdown and all the other taser effects. Ever since it was changed to a disabler, it was never touched and has stayed this way for almost a year now. It has been terrible for a year and it's about time it gets a facelift. Yes, this does make it a 2-hit KO on PB but at the end of the day it's still a disabler shotgun that is outshadowed by a riot shotgun with rubbershot.

i specifically kept it from being two shots because the warden already has a shotgun that can dumpster someone in two to three shots pb with beanbags or lethally with slugs
wardens have the most powerful roundstart ballistic weapon and the particle defender has no reason to even exist other than citadel snowflake code
i'm for giving it more ammo to bring it more in line but giving the warden more power to add to the already deadly combo of instant disarms and floors with krav and unmatched dps with the combat shotty is just too much for a single role
i'd consider this fair if krav and the combat shotty already weren't both the most busted things in the security arsenal

grr secbalance

GrayRachnid:
>I know the warden already gets two very overpowering toys but if we're going to keep the particle defender then it ought to be at least on par with rubbershot.

I know but the defender is just pure ass right now :pensive:

ghost:
Unless I'm wrong, the taser also does _energy damage_ which often has the lowest values for protection compared to ballistics. Then again I'm also assume rubbers/beanbags are affected by ballistics and not energy armor

keronshb:
> 
> 
> Unless I'm wrong, the taser also does _energy damage_ which often has the lowest values for protection compared to ballistics. Then again I'm also assume rubbers/beanbags are affected by ballistics and not energy armor

The disabler function on the Particle Defender is only a disabler AFAIK.  Only the laser function should be doing any damage,

ghost:
> 
> 
> > Unless I'm wrong, the taser also does _energy damage_ which often has the lowest values for protection compared to ballistics. Then again I'm also assume rubbers/beanbags are affected by ballistics and not energy armor
> 
> The disabler function on the Particle Defender is only a disabler AFAIK. Only the laser function should be doing any damage,

I worded that poorly, but I mean the disabler function deals energy damage, which is stamina? I think. I know batons deal energy damage instead of melee separately from their harm intent

necromanceranne:
> This was lowered for a reason because Wardens would take this out of the brig to valid hunt.
> 
> I don't think this buff is warranted at all.

If you didn't want wardens using it, you should have removed it. Not nerfed to be the obviously weaker option.

At 15 damage per shot, that is equal to rubber shot. Seems pretty fine to me.

PR #14467, opened 2021-03-18
IC Chat filter

LetterN:
## Changelog
:cl:
admin: Adds an ic chat filter (you can still bypass this with force enabled)
/:cl:


LetterN:
oh there's already a pr for it

PR #14476, opened 2021-03-19
Makes station time tick at 1 second for every second.. rather than 12 seconds for each second.

silicons:
## Why It's Good For The Game

This was done to force a full day/night cycle but in reality it's confusing and obnoxious and if we really want that I can make night shift lighting toggle every 30 minutes instead.

## Changelog
:cl:
tweak: Station time now ticks in a realistic manner and is reliable again.
/:cl:


DeltaFire15:
Yeeah uh it's kinda wacky to have a complete shift during day / night (even if reasonable), so probably change that.

TheEaterOfBeans:
the day/night cycle should be per-shift (one shift is day, next one is night, next one is day, etc.)

PR #14492, opened 2021-03-22
rebalances most available cybernetic limbs, axes damage reduction, makes upgraded limbs quite beefy/expensive (waiting on combat rework)

necromanceranne:
## About The Pull Request

The majority of the cyberlimbs were all over the shop, numbers wise, so this tries to bring them more in line with some of their intent.

**1) Flat damage reduction is removed**

There was a balance change to prevent augmented people from being spaceproof, and that spaceproofedness was due to cyberlimbs simply reducing any incoming brute/burn by a flat amount, not ACTUAL space immunity.

The current numbers are so tiny as to be pretty much pointless, so instead of trying to work out what a reasonable number might be. So in the bin it goes.

**2) Alters the health values of the limbs based on tier, as well as their damage thresholds**

Surplus limbs have slightly higher health, but much worse thresholds for damage caps. As a result, getting hit a few times too hard in those limbs will permanently cap you at half limb health. In addition, surplus limbs aren't resilient against stamina damage anymore. A disabler shot WILL take out your limb without armor.

Normal cyber limbs are unchanged for health/thresholding, but the head and chest have been brought into line with the rest of the body. Why these were left untouched is entirely beyond me, but there you go.

Upgraded limbs are now significantly more material hungry and a little deeper in the tree, but are much beefier than a standard limb, with much higher thresholding, max health and more forgiving damage-to-core health ratio (which is an effective damage reduction just fyi). It does, however, cap at a much higher value than the other limbs if you hit the damage threshold, which can become a significant liability. 

The upgraded torso and head, however, are strictly an upgrade on basic cybernetic torsos and heads, with a hefty easy heal threshold and very small minimum damage point.

## Why It's Good For The Game

Cybernetics are kind of a mess and it might have something to do with mixed intentions with a **reinforced** _surplus_ limb still being harder to damage, yet not more durable.

## Changelog
:cl:
balance: The tiers for cybernetic limbs have been largely overhauled.
balance: Surplus are slightly more durable but break far more easily, and more vulnerable to stamina damage.
balance: Cyber torsos and heads no longer get oddly disproportionate values to their limbs.
balance: Upgraded cyber limbs are very hard to destroy, but once they break, they require specialized help.
balance: Upgraded cyber torsos and heads are much easier to maintain and even MORE difficult to permanently damage than their limbs. (mostly because they're part of core health and don't have inherent damage reductions as a result)
balance: Upgraded prosthetics require more tech and a lot, lot more materials to make.
/:cl:

### DIFF
--- a/code/modules/cargo/exports/organs_robotics.dm
+++ b/code/modules/cargo/exports/organs_robotics.dm
@@ -136,10 +136,10 @@
 	unit_name = "robotic limb replacement"
 	export_types = list(/obj/item/bodypart/l_arm/robot/surplus, /obj/item/bodypart/r_arm/robot/surplus, /obj/item/bodypart/l_leg/robot/surplus, /obj/item/bodypart/r_leg/robot/surplus)
 
-/datum/export/robotics/surplus_upgraded
+/datum/export/robotics/upgraded
 	cost = 80
 	unit_name = "upgraded robotic limb replacement"
-	export_types = list(/obj/item/bodypart/l_arm/robot/surplus_upgraded, /obj/item/bodypart/r_arm/robot/surplus_upgraded, /obj/item/bodypart/l_leg/robot/surplus_upgraded, /obj/item/bodypart/r_leg/robot/surplus_upgraded)
+	export_types = list(/obj/item/bodypart/l_arm/robot/upgraded, /obj/item/bodypart/r_arm/robot/upgraded, /obj/item/bodypart/l_leg/robot/upgraded, /obj/item/bodypart/r_leg/robot/upgraded, /obj/item/bodypart/chest/robot/upgraded, /obj/item/bodypart/head/robot/upgraded)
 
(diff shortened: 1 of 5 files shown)


The-Sun-In-Splendour:
Are cyborg limbs unchanged?

necromanceranne:
They have no longer have damage reduction. Lemme just clarify that more.

DeltaFire15:
.. Why exactly are we *nerfing base borg bodyparts*?

necromanceranne:
Because upgrades exist? If you mean the loss of damage reduction, a 1/2 point damage reduction per instance of brute damage isn't exactly doing much. But if you think the base limbs could use some more improvements, I'll happily do something to make them a bit more useful.

Anyway, this pr might have to wait on kevs combat rework because I hadn't realized he was changing limb health values so Im gonna draft this one.

necromanceranne:
Gonna close till after combat rework. Less on the tracker.

PR #14494, opened 2021-03-23
Ports #50904 and #54163 from /tg/, makes one small concession on my part; Scarp Nerf basically

necromanceranne:
## About The Pull Request

Ports both https://github.com/tgstation/tgstation/pull/50904 and https://github.com/tgstation/tgstation/pull/54163, removals of Sleeping Carp's defenses and to its immunity to piercing, as well as removing the Gnashing Teeth crit.

The former is basically me agreeing with Ryll that my initial implementation on the physiology mods was, by all accounts, a massive mistake. It's invisible armor that gets significantly worse when you stack on top of it actual armor, resulting in an incredibly tanky opponent to have to fight, who can also deal a significant amount of damage. 

The only real weapon against them at that point was batons, and as I said in my baton PR, SCarp would need to be weakened from being a damage sponge if a change like that was to make sense, since armor piercing doesn't respect physiology mods. I've had a chat with kev, and I wanna later look at giving SCarp an active block, but hopefully its parry is enough for now.

The latter port was because I added an antag check. This is also generally more of a useful trait.

The concession to this is that I've added the change to scarp's Gnashing Teeth that removes the random crit, but grants it wounding power, with an addition by me. Wounding power is determined by species stun threshold, and the wounds themselves are done as slashing attacks for cutting wounds. ~~Because you're biting with your fingers.~~ Because you're karate chopping peoples heads off, duh. On top of this, Sleeping Carp deals both Brute AND Stamina damage on harm punches and Gnashing Teeth. This reserves Keelhaul for openings but also ensures you can follow up every attack with more stamina damage to bring an opponent into submission quickly.

In addition, I'm using the 12 TC price tag. You know why? I already got to GotNS so ScarpGloves isn't possible without contractor/help from soneone else. We get the cooler price tag.

## Why It's Good For The Game

I basically made a really bad choice with adding the physiology mods. There are better ways of doing it, and I'll get to those at some point. This version of scarp isn't necessarily too much worse. There are a lot of citadel specific problems that are keeping it from the same potential it has over on tg (namely, wallstuns are completely dead on cit), but for damage output its still doing pretty well.

Maybe the price drop will be enough, but if not, I'll come back as necessary.

## Changelog
:cl: Ryll/Shaps, NecromancerAnne
del: Ports tgstation/tgstation#50904: Sleeping carp no longer grants outright damage resistances to all forms of damage. Its price has been decreased to 12TC
balance: Ports tgstation/tgstation#54163: Sleeping carp users are no longer immune to piercing, which means they can once again use most medical supplies, but will trip on glass and such like normal people. In return, they are now somewhat more resistant to suffering wounds.
balance: Sleeping Carp now deals additional stamina damage on normal harm punches as well Gnashing Teeth.
/:cl:


keronshb:
There seems like 0 reason to buy this anymore other than the 20tc bundle with the funny stick.  There was already no reason to buy it because batons really messed its day up and the fact you needed throw mode on constantly to deal with threats (meaning people can just force Bolas into your hand to turn it off).

necromanceranne:
The active block stuff might take a bit so hopefully that solves any issues with getting wasted by batons.

keronshb:
> 
> 
> The active block stuff might take a bit so hopefully that solves any issues with getting wasted by batons.

Stam Armor not working aside, no one takes Sleeping Carp anymore even WITH the armor + piercing immunity because it's just not that good.  Even at 12TC I'd argue it's not worth it unless you get some sort of ungodly luck to mix with with GOTNS.

Even damage output.. it sucks as it is right now.  Spamming Harm Harm Harm Harm is not nearly as good as GOTNS or a standard holopara (which costs less and has a worse tradeoff via dusting).  Anyone who's used Sleeping Carp in recent times can tell you how much it's not worth it.

Some people even bought it for the bostaff with blockchance, and even then it's not that great since you don't have enough room to buy glue with it and it takes longer to stun with it than a baton.

As of right now even with the baton + combat rework, this would just kill SCarp even moreso than it has before.

necromanceranne:
I suppose that might be my fault actually, I may have went a little low on overall damage when I changed martial arts to use species values. I'll throw a bone here and add a little more on the damage.

necromanceranne:
by the by, the reason GotNS is so much better is pretty simple. I never intended it for traitors. It was meant for nukies. Someone later on gave it to traitors, so I guess thats just how it is now. I can't keep up with endless fucking powercreep if people keep doing that shit.

necromanceranne:
Actually, I'm gonna close this one and give it another shot. I think I might try for some more complicated changes.

PR #14495, opened 2021-03-23
is it really an augmentation if you're using a synthflesh-grown limb? no, probably not, but what if you could

Hatterhat:
## About The Pull Request
allows limbgrower printed synthetic limbs to be used in augmentation surgeries, which is probably a good idea for catastrophic burns on chest or head (unless you just want to chuck their ass in cryo for a while or just duct tape a shitty prosthetic chest or head on them)
## Why It's Good For The Game
"uhh, limb's fucked mate" handling options:
a. replace with prosthetic
b. shove into cryo, if it's set
this pr adds a variant of a, "replace with synthflesh limb", in case you want to keep your fleshy bits fleshy but don't want to wait in cryo or w/e. or maybe you want to replace your perfectly good metal torso with a fleshy one? up to you i guess
## Changelog
:cl:
tweak: Limb grower-printed synthetic limbs can now be used to "augment" (or, more accurately, replace) limbs. Mainly the torso, I guess? Unless you wanted to replace your robot hand with a fleshy one. ...That's a weird thought, actually.
/:cl:

### DIFF
--- a/code/game/machinery/limbgrower.dm
+++ b/code/game/machinery/limbgrower.dm
@@ -178,6 +178,7 @@
 	limb.update_icon_dropped()
 	limb.name = "\improper synthetic [lowertext(selected.name)] [limb.name]"
 	limb.desc = "A synthetic [selected_category] limb that will morph on its first use in surgery. This one is for the [parse_zone(limb.body_zone)]."
+	limb.forcereplace = TRUE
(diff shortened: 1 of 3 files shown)


The-Sun-In-Splendour:
...Can't you just do tend wounds? Seems a bit useless IMO.

PR #14502, opened 2021-03-23
Experimental - adds javascript autoping. anyone who doesn't send a ping for 0.5 seconds has their input halted, so people who are lagging don't fall into lava/chasms/supermatter/whatever.

silicons:
This may be laggy on highpop, probably testmerge!


silicons:
autopings break reconnect as it'll cause byond to try to reconnect to a closed socket due to it never stopping
we could fix it by mkaing reocnnects shut it off but it isn't worth it, as then hard reboots won't reconnect
closing, until i have time to find a way to hook this into SSserver_maint

PR #14512, opened 2021-03-25
Gives the Blackmarket Uplink a description

R3dtail:
Gives the item a description and actually gives somewhat of an idea of what it does and how to use it

## About The Pull Request

Gives the Blackmarket Uplink a description  that gives at least some idea of what the item actually does.

## Why It's Good For The Game

Flavortext and actually helps players understand both what it takes to make the item, and how the item works  without having to code dive too much

## Changelog
:cl:
add: Added a description to the black market uplink
/:cl:

### DIFF
--- a/code/modules/cargo/blackmarket/blackmarket_uplink.dm
+++ b/code/modules/cargo/blackmarket/blackmarket_uplink.dm
@@ -1,5 +1,6 @@
 /obj/item/blackmarket_uplink
 	name = "Black Market Uplink"
+	desc = "A cobbled together mishmash of a subspace amplifier, a radio, and an analyzer bound together with some cables.  Somehow, this pinnacle of hodge-podge engineering is able to access the black market.  Such a sordid place has a variable inventory in limited stock at inflated prices...but at least you won't have to dirty your hands making the items yourself once you have this device.  Remember, no refunds, and once you buy the item it's up to you to be able to retrieve it from the drop-off point." 
 	icon = 'icons/obj/blackmarket.dmi'
 	icon_state = "uplink"
 	// UI variables.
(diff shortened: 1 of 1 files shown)


code/modules/cargo/blackmarket/blackmarket_uplink.dm
```diff
@@ -1,5 +1,6 @@
 /obj/item/blackmarket_uplink
 	name = "Black Market Uplink"
+	desc = "A cobbled together mishmash of a subspace amplifier, a radio, and an analyzer bound together with some cables.  Somehow, this pinnacle of hodge-podge engineering is able to access the black market.  Such a sordid place has a variable inventory in limited stock at inflated prices...but at least you won't have to dirty your hands making the items yourself once you have this device.  Remember, no refunds, and once you buy the item it's up to you to be able to retrieve it from the drop-off point."
```
Hatterhat:
```suggestion
	desc = "A mishmash of a subspace amplifier, a radio, and an analyzer.  Somehow able to access the black market, with a variable inventory in limited stock at inflated prices. No refunds, customer responsible for pick-ups." 
```
with all due respect my good man, less is more

Hatterhat (changes requested):

code/modules/cargo/blackmarket/blackmarket_uplink.dm
DeltaFire15: mhm, does sound a bit better compact

R3dtail: I never said I was good with words  but yeah that sounds better.  Thankies.

PR #14518, opened 2021-03-25
Shields: Changes, additional flags, bug fixes, sprites

necromanceranne:
## About The Pull Request

Adds several flags to help specify exactly what a shield is resilient against, what it is weak against, what it can block, and what it cannot block. Currently most all shields are identical, including several shields that shouldn't be doing anything at all (like fake Roman shields).

Take note: The armor values on those shields do not do anything for the resilience of the shield while blocking, just when it gets directly smacked. Like on the floor.

### **Shield Flags**
**SHIELD_KINETIC_WEAK** means that a shield takes more damage from any projectile that cannot be reflected (anything deswords cannot bounce away), while **SHIELD_KINETIC_STRONG** means that it takes less. It also determines how well it absorbs brute from melee attacks in the same pattern.

**SHIELD_ENERGY_WEAK** means that a shield takes more damage from any projectile that can be reflected, while **SHIELD_ENERGY_STRONG** means that it takes less. It also determines how well it absorbs burn from melee attacks in the same pattern.

**SHIELD_DISABLER_DISRUPTOR** means this shield takes massive damage from stamina damage projectiles (note, not the stamina var, it has to be flagged as stamina damage as its base damage, like disabler shots and not beanbag slugs).

**SHIELD_NO_RANGED** means a shield cannot block range attacks.
**SHIELD_NO_MELEE** means a shield cannot block melee attacks. 

### **Fixes**

Tower Shields can block once again, and have integrity to match their apparent strength. 

Fake Roman Shields are not just riot shields in disguise, and can no longer block ranged attacks and shatter apart easily, nor can they shield bash. I kept their ability to block melee only so you can do your roman legionnaire re-enactments with a little bit of realism.

Bucklers no longer has RNG blocking, and is also vulnerable to fire damage and lasers.

Stamina damage (with a flagged exception), toxin damage, clone damage and oxygen damage no longer do damage to shields, so you no longer can shatter shields with disablers.

Ports part of https://github.com/tgstation/tgstation/pull/42721 (which we already had but not the change I brought over) to make that pesky strobe shield no longer a flash subtype, just a shield with an embedded flash. Should fix the various issues with the strobe shield.

### **Features**

**Kinetic Shield**
Replaces the bullet shield. The kinetic shield is resilient against brute attacks and bullets, but breaks easily against burn attacks and energy projectiles.

bullet shield

**Ablative Shield**
Replaces the laser shield. The ablative shield is resilient against burn attacks and energy projectiles, but breaks apart against brute attacks and bullets.

ablative shield

**Holoshield**
The shield implant has been changed to a hardlight shield. This shield can take quite a beating from brute damage, but disablers destroy it extremely quickly.

## Why It's Good For The Game

Shields were all kinds of fucked.

## Changelog
:cl:
add: Replaces the useless bullet and laser shields with new Kinetic and Ablative shields, which do as they advertise.
add: Replaces the shield implants shield with a hardlight shield that can take large amounts of brute damage, but disintegrates when shot with disablers.
fix: Fixes Fake Roman Shields being able to be used as normal riot shields.
fix: Fixes nonlethal/non-physical damage types destroying shields. (stamina, toxins, oxygen, clone, brain)
fix: Fixes tower shields not doing anything, while also giving them intergrity to match their advertised durability.
fix: Uses additional flags to determine what kind effects work well and what works poorly against various shields.
/:cl:


GrayRachnid:
PLEASE FIX THE STROBE SHIELD PLEASE PLEASE PLEASE

necromanceranne:
What broke exactly?

GrayRachnid:
1) It has an invisible onmob sprite
2) It does not behave like a shield because it is considered a flash with an RNG block
3) It still has an RNG block

necromanceranne:
That's fucked up I'll see what I can do.

necromanceranne:
Should be fixed.

moltoretardo:
Farewell RNG blocking bucklers.

PR #14509, opened 2021-03-24
Adjusts blackmarket uplink

R3dtail:
Removes bone spears from the uplink, and makes EMP grenades rarer and more expensive

## About The Pull Request

Removes bonespears from the blackmarket uplink.  Additionally, makes EMP grenades more expensive and rarer to get
## Why It's Good For The Game

It rubs me the wrong way to be able to just buy a weapon with 2 reach for a paltry amount of credits without having to go to lavaland or otherwise interact with a miner.  Sure, the arguement could be made about it being marginally more difficult to get the parts to make the blackmarket uplink in general, but uh..still making the PR anwyay for it to be discussed.  

As for the EMP grenades, I fully understand that it's fairly easy to make enormous EMP grenades as well with the simple ingredients, but again just being able to buy the stuff outright for a paltry amount of credits rubs me the wrong way too.  So, I figured make it "rarer" in the uplink and make it more expensive.  Additionally, makes fewer show up.  There was a good bit of discussion in admin-chat over the item so I figured I'd try and make the changes myself.  

## Changelog
:cl:

tweak: Altered price, probability, and available stock of EMP grenades in the blackmarket uplink.
balance: Removed the bonespear from the blackmarket uplink, and made EMP grenades harder to get from the same item.

/:cl:

PR #14519, opened 2021-03-25
Black Market - Some tweaks, additions and comments

ghost:
## About The Pull Request
This is not only about the tweaks but also the general direction I think the Black Market can take.
So alright, what was done:

- Each category with the exception of "Weapons" will have two guaranteed products in stock that are nearly harmless or fun;
- There was a change to prices, stock, and also how prices are calculated (now prices are calculated all in multiples of 5);
- Ported some items from /tg/'s Black Market, like the Clear PDA;
- Added some items of my own, like the Meson Goggles, a Random Drink item, and the Internals Box;
- Weapons were barely touched because I am afraid of adding some stuff to it that are WAY too dangerous, tho maybe, that would make rounds funnier.

## Why It's Good For The Game
Here is my thought process:

- Having some consistency will keep players coming back to it, maybe even using it more often than not because of the new drink option which was my best idea so far;
- Some items got their stock raised because they were just harmless, so no reason to not have plenty. Also because, in my mind, the Uplink is not simply a single store selling stolen crap, it's something like Space Alibaba/eBay;
- The ports and new items are just because the list was too short!

## Changelog
:cl:
tweak: With the exception of weapons, every category now has 2 items that are always available
tweak: Prices are multiples of 5, raised, rounded, and SOME things lowered
add: Added Earmuffs
add: Added Random Drink
add: Added Clear PDA
add: Added Internals Box
add: Added Meson Goggles
add: Added Smoke Grenade
/:cl:


ghost:
ok kids, never assume your code will pass tests just because the changes are simple. they are never simple

ghost:
oops hang on hold the phone. #14509 and #14512, just read those two. Yeah good they both are great, I can deal with conflicts later if need be

PR #14526, opened 2021-03-27
what if i just gave the x01 an rmb taser while also giving it, like, way more disablers

Hatterhat:
## About The Pull Request
see title
+ x01 gets sec-taser tier electrode on combat mode rmb
+ x01 now fires hos disablers, which use 25 energy instead of a regular disabler's 40
(the above has the effect of granting the hos gun **40!!** max disablers, if you don't use literally any other firemode)
## Why It's Good For The Game
dzahl is a hack
## Changelog
:cl:
balance: The X-01 now has incredibly scuffed power efficiency for spitting disablers down a hallway while also getting an auxiliary taser.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/energy/stun.dm
+++ b/code/modules/projectiles/ammunition/energy/stun.dm
@@ -30,3 +30,6 @@
 
 /obj/item/ammo_casing/energy/disabler/secborg
 	e_cost = 50
+
+/obj/item/ammo_casing/energy/disabler/hos
+	e_cost = 25 // if the big iron can get a silly amount of disablers, why can't i?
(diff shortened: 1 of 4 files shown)


The-Sun-In-Splendour:
Delete the particle defender

LordSaladin:
disabler goes brrr

DeltaFire15:
Not really on board with the timeline of
HoS revolver gets added
Revolver gets buffed because it was seen as too weak
Now the X01 is getting buffed because the Revolver is 'better'

Especially considering as opposed to the Revolver, the Multiphase gun can use all its charge on one single ammo type aswell as it being able to switch between them very easily, making it more flexible than the Revolver.

Leaving this to the other maintainers though.

DeltaFire15:
Since apparently noone else feels like looking at it, guess I'll go ahead with my judgement and close it.

PR #14545, opened 2021-03-31
renders storage objects independently per user

silicons:
prevents the ghost override bug, with some costs:
- normal ui works fine
- volumetric ui is forced to use add_overlays. this means whenever the item itself changes, it won't reflect in the ui. i can't use vis_contents as that'd make it subject to being transformed.
- i'm working on a way to do volumetric ui without add_overlay. might have to do a virtual holder.

jk it's fully working now testmerge this


DeltaFire15 (approved):
Nobody reported any issues with this TMd and it *seems* fine, so, time to merge this

PR #14546, opened 2021-03-31
Ports bluespace pipes from yogs

Putnam3145:
## About The Pull Request

Adds bluespace pipes, which can connect two pipenets at any range, including across unconnected Z levels, as long as they have the same ID. They have to be researched and printed from protolathes and researched from the Bluespace Travel tech node. They cost 1000 gold, 750 diamond, 250 uranium and 2000 bluespace crystal, and you need two of 'em to do anything, obviously.

## Why It's Good For The Game

More toys to play with. I've got some funnies in mind already, actually, long-range space cooling and such.

## Changelog
:cl:
add: Bluespace pipes, which can teleport gas over long distances
/:cl:


silicons:
you are evil

Putnam3145:
WHY is that

SandPoot:
REEEE

DeltaFire15:
Ah yes I can't wait for a remote plasmaflood via connecting one of these pipes to distro somewhere in maint and the other on a gateway map with plasma or simillar source of bad gas
Seems like a rather questionable thing over all.

BlueWildrose:
At first I thought this was going to be about disposal pipes which would be pretty horrifying...

_**But this is also pretty awful to think about.**_

How does it respond to something ventcrawling through it? Or, does it simply not do it?

dzahlus:
hhhhh time to hide a passive vent inside the plasma miner for portable plasmafloods lets goooooooooooooo

Putnam3145:
the thing is that all of these wonderful sabotage ideas are already trivially executed without bluespace pipes. i could permanently fuck distro up in a way that would literally require admin intervention to fix with only passive vents, hell, I actually *do* this regularly as a "fortification" effort, I just happen to do it with breathable air instead of, say, nitrous oxide

hiding away a sabotage space in maint is already possible with overstuffed cans, hell, you can immediately destroy the evidence, the single worst new sabotage that you can do with this that I can think of is portable scrubber stuff and you can... again, already do this pretty easily *without* bluespace pipes, it's just slightly harder to find, but finding it doesn't matter when it takes two minutes to completely, irrevocably fuck up the round, which atmos can already do

what i'm saying is that this just makes some fun stuff more convenient and, yes, that includes sabotage, but sabotage is already so piss-easy that this doesn't actually *change* anything

shellspeed1:
I'll take it just for easier Multi Z pipe stuff on multi Z stations.

silicons:
fine

PR #14562, opened 2021-04-03
Buffs the genetic mutation Autotomy

BlueWildrose:
## About The Pull Request

This changes Autotomy a bit.
- Instability reduced from 30 to 20
- Delimbing is now harmless instead of dealing damage to the chest area, making you suffer a mood loss, and making you bleed.

## Why It's Good For The Game

Why is it that we have mutations that seem like they're fun but aren't that worthwhile in getting? Voluntarily popping your limbs at best, could free you from handcuffs but it obviously comes at a cost of reducing your maximum health and well, leaving you with one less arm. I figure that'd be punishment enough, set aside the fact that Autotomy is also random in which limb it pops off, so if someone's trying to free themselves this way they'll be taking a gamble on if they lose an arm or a leg. They obviously may not be getting that limb back without external help unless they were a slimeperson or had Ligament Hook surgery. Autotomy is very situational, while other genes like Space Adaptation and Telekinesis have a lot more times where they're useful compared to when they're not.

This also will encourage the Ligament Hook experimental surgery to be more intriguing, because there's also the fact that you can't put a limb back on without help from someone else otherwise.

And yes, combined with being a slimeperson it has a lot of potential considering they can regrow limbs by themselves at the cost of nutrition and blood. Did you know you can attach slimeperson limbs to any other species?

I might potentially see if Tongue Spike/Chem Spike can be made more intriguing later on, but it'll also mean I'll need to rebalance that potentially because it does manage to at least have a worthwhile benefit to losing your tongue.

## Changelog
:cl:
balance: The genetics mutation Autotomy has been buffed to be 20 instability instead of 30, and harmless in delimbing.
/:cl:

### DIFF
--- a/code/datums/mutations/actions.dm
+++ b/code/datums/mutations/actions.dm
@@ -283,7 +283,7 @@
 	desc = "Allows a creature to voluntary discard a random appendage."
 	quality = POSITIVE
 	text_gain_indication = "<span class='notice'>Your joints feel loose.</span>"
-	instability = 30
+	instability = 20
 	power = /obj/effect/proc_holder/spell/self/self_amputation
 
 	energy_coeff = 1
@@ -316,7 +316,7 @@
 		return
 
 	var/obj/item/bodypart/BP = pick(parts)
-	BP.dismember()
+	BP.dismember(harmless=TRUE)
 
 //spider webs
 /datum/mutation/human/webbing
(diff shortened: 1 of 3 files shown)


BlueWildrose:
Oh, the linters are complaining because I added a new arg to the dismember proc, and the overrides aren't receiving that proc. It doesn't _need_ those args, does it?

It probably does, I'll fix that.

necromanceranne:
It's a joke mutation, its not really meant to be useful. It definitely shouldn't be doing damage and likely this was an oversight, but usefulness is debatable unless you're a slime.

Or a ling.

Linzolle (approved):

PR #14563, opened 2021-04-03
wooden planks sell for 8 now to stop the funny cargo lumbermill memes

timothyteakettle:
## About The Pull Request
profit from a log crate is now 200 instead of 1150

## Why It's Good For The Game
stops another cargo meme

## Changelog
:cl:
balance: wooden planks sell for 8 instead of 15 each
/:cl:

PR #14564, opened 2021-04-03
kills the 50 towercap crate, closes #14563

Hatterhat:
## About The Pull Request
closes #14563 by deleting the 50 log crate
## Why It's Good For The Game
interdepartmental cooperation

## Changelog
:cl:
balance: After a sudden crash in the tower-cap log slash wooden plank economy, NanoTrasen has decided to stop selling tower-cap logs to Cargo.
/:cl:

### DIFF
--- a/code/modules/cargo/packs/materials.dm
+++ b/code/modules/cargo/packs/materials.dm
@@ -69,18 +69,6 @@
 	cost = 600
 	contains = list(/obj/item/rcd_ammo/large)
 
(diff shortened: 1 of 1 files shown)


DeltaFire15 (approved):
Yeeeah we have had an actual plank crate since ages but why'd anyone order that if you could just order a ton of logs instead to create infinite money.
Lets not have cargo get free money via throwing ordered stuff into a grinder and selling that.

PR #14569, opened 2021-04-04
ghost cafe 'disguise' now removes the offset of the object you disguise as

timothyteakettle:
## About The Pull Request
title

so people can disguise as narsie and stuff without being weirdly offset

**untested (testing it currently)**

## Why It's Good For The Game
someone asked me to do it and i can't see why not, let the people erp as ratvar/narsie if they really want i guess

## Changelog
:cl:
tweak: ghost cafe 'disguise' now removes the offset of the object you disguise as
/:cl:


timothyteakettle:
okay so the issue with this isn't what i thought it is, the pixel x/y of nar'sie doesnt stop it being offset weirdly, going to look at some other solution

PR #14575, opened 2021-04-06
[Donator] Toy Gun

LetterJay:
Toy gun for a donator


AzericTraveller:
Hello! I'm the donator that wanted this, if it means anything.

DeltaFire15:
Fails CI due to not having an empty lastline file ending in custom_items.dm

Also needs to not be able to fire riot darts, as to not make it an item with actual gameplay benefit.

DeltaFire15 (changes requested):
See above comment.

PR #14578, opened 2021-04-06
LOUDER DELAMS

ghost:
## About The Pull Request

This is more something I had a problem with, but when thinking about it, it made sense to bring it up in a PR. The SM messages broadcasted over common are no different from a cyborg or any other machine stating something. While that could be fine when going on the engineering channels, a common broadcast alerting to something dangerous like that, SPECIALLY the countdown, should be "louder" and therefore easier to notice. I ain't dyslexic or anything but even I had a hard time seeing shit was going down in my peripheral vision.
Tldr.: Integrity % broadcast will show in bold, but still same font size; Countdown will be just like Command Headsets on the LOUD setting (might be a little too much). REMINDER THAT ONLY APPLIES TO COMMON BROADCAST. The engineering alerts will remain the same.

## Why It's Good For The Game

Makes it easier to see when crystal integrity/delam countdown in case you are distracted by anything else, or have difficulty reading the chat.

## Changelog
:cl:
tweak: Super Matter Crystal integrity messages over common now will be bold.
tweak: Super Matter Crystal Countdown messages over common will now be larger and bold.
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -390,7 +390,7 @@ GLOBAL_DATUM(main_supermatter_engine, /obj/machinery/power/supermatter_crystal)
 			speaking = "[DisplayTimeText(i, TRUE)] remain before causality stabilization."
 		else
 			speaking = "[i*0.1]..."
-		radio.talk_into(src, speaking, common_channel)
+		radio.talk_into(src, speaking, common_channel, list(SPAN_COMMAND))	// IT GOT WORSE, LOUD TIME
 		sleep(10)
 
 	explode()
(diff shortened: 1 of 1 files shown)


The-Sun-In-Splendour:
Nah bro just give it the narsie font trust me on this one

shellspeed1:
Why use Narsie when you can use Ratvar

ghost:
alright how about a 1/255 chance that it will be one of the cult fonts/blob

DeltaFire15:
People missing the 
30..
20..
10..
5.4.3.2.1.
 countdown is obviously just a feature

Anyway, joke aside, not exactly opposed to this but not inclined either, so, leaving this to the other maintainers to decide.

PR #14579, opened 2021-04-07
you can now pointedly point stunbatons at people

silicons:


Linzolle:
If we're going to use the baton rework PR, wouldn't this make you point the baton every time you use the alt attack?

DeltaFire15:
@silicons how does this interact with the rightclick altattack?
May want to look into that, since worst-case it'd override it, best case it'd make you point at people if you miss them (which also seems kinda not-so-good)

silicons:
@DeltaFire15 @Linzolle  believe it'd be the latter but i haven't tested
hm

PR #14581, opened 2021-04-07
pirate payoff minimum from 20k to 1k

Hatterhat:
## About The Pull Request
see title
## Why It's Good For The Game
nobody plays cargo and i don't think this was a thing beforehand
## Changelog
:cl:
balance: Space pirates are now slightly more aware of how much money the station has, and will demand payment accordingly. (No more 20k minimum payouts and basically-confirmed three midround skeletons.)
/:cl:

### DIFF
--- a/code/modules/events/pirates.dm
+++ b/code/modules/events/pirates.dm
@@ -21,7 +21,6 @@
 	startWhen = 60 //2 minutes to answer
 	var/datum/comm_message/threat_msg
 	var/payoff = 0
-	var/payoff_min = 20000
 	var/paid_off = FALSE
 	var/pirate_type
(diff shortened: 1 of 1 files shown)


The-Sun-In-Splendour:
Pirate sleeper agent making pirates basically confirmed everytime they hit

DeltaFire15 (changes requested):
Mhh I'd say just change the minimum to something like 500 or 1k credits which should be affordable, but not completely free, but yeah 20k seems like just a tiiny bit too much, with how "often" we have actual moneymaking cargo.

DeltaFire15 (approved):

PR #14596, opened 2021-04-09
makes a config option off by default

timothyteakettle:
## About The Pull Request
title its the one that asks your DoB if you're a new player who isnt registered to the bunker

i'd put it in the title but people would just take it out of context when it's the default value, not the current value, and its just so people can have people connect to their server without having a db

## Why It's Good For The Game
kevin told me to

## Changelog
:cl:
config: makes AGE_VERIFICATION option off by default
/:cl:

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -382,7 +382,7 @@ NOTIFY_NEW_PLAYER_ACCOUNT_AGE 1
 #PANIC_SERVER_NAME [Put the name here]
 
 ##Automated age verification, comment this out to not ask new users if they are 18+
-AGE_VERIFICATION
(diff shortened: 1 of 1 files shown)


tiramisuapimancer:
I recently had issues with turning the option in the config off not even doing anything in the first place, people were unable to play even if the option was off, and I did triple check that not only it was off but that the game was definitely reading the config file i was editing. I only managed to make my server playable by manually editing it to return true instead of false so that the check didn't functionally do anything beyond printing a message to admin logs, which I edited to be appropriate to the situation 
image

DeltaFire15 (approved):

PR #14604, opened 2021-04-13
Machinery&stuff Hardsync

LetterN:
## Changelog
:cl:
code: APC Sync, it's better code now?
code: Area sync. Power management should be better now?
refactor: Icon update system
/:cl:


Putnam3145:
can i just hard no to all hardsyncs from now on

anyway, doors can't be welded, huge scrubbers have the wrong icon, analyzers can't be used on canisters anymore, in general clicking on machines seems broken, which is kinda 99% of what machines are

EDIT: windoor'd tables are unplaceable apparently

DeltaFire15:
I do not think we need heat / pressure caps on atmos stuff. Has never been an issue here.

TripleShades:
Bolted and emergency access door lights are much dimmer then normal, while door open lights are much brighter.
Might be seeing things, but I also believe all light sources are much brighter as well (bulbs, lamps), and the dimming of lights for flavor throughout the shift may also be broken, or not dimming them nearly as much anymore as a result.

Putnam3145:
> 
> 
> Bolted and emergency access door lights are much dimmer then normal, while door open lights are much brighter.
> Might be seeing things, but I also believe all light sources are much brighter as well (bulbs, lamps), and the dimming of lights for flavor throughout the shift may also be broken, or not dimming them nearly as much anymore as a result.

I've been having this problem lately... on a local server, which is definitely not incorporating this testmerge. Might be other screwiness.

PR #14615, opened 2021-04-15
that said let's fix radiation not going away when it should at least i guess.

silicons:


code/datums/components/radioactive.dm
```diff
@@ -52,7 +52,7 @@
 		return
 	strength -= strength / hl3_release_date
 	if(strength <= RAD_BACKGROUND_RADIATION)
-		return PROCESS_KILL
+		qdel(src)
```
LetterN, line 55: pkill it still?

silicons: in theory this is auto

DeltaFire15: yeeah qdel should stop it.

DeltaFire15 (approved):

PR #14628, opened 2021-04-18
No more Suicidal Parrots! (Ready)

WanderingFox95:
## About The Pull Request
So, the stabilized Golden slime extract spawns a random simple animal from the comprehensive list of parrots and maybe a dog or a cat sometimes. ...Jokes aside, the amount of parrots in the code is enormous and that is a bit of a problem...
The extract summons a familiar while the extract is held and deletes it when the extract is dropped.
Parrots are scripted to steal small, held items. When they steal the extract, they delete themselves and the extract with them.
In order to mitigate this, I moved them to the hostile spawns instead, making the stabilized gold extract unable to summon them now. Arguably, kleptomania isn't exactly "friendly", now is it?

edit: I removed them from gold extracts completely for now.

## Why It's Good For The Game
Xenobiologists won't randomly lose their work due to self-destructive animal behaviour anymore.

## Changelog
:cl:
tweak: removed parrots from gold extract spawns altogether.

### DIFF
--- a/code/modules/mob/living/simple_animal/parrot.dm
+++ b/code/modules/mob/living/simple_animal/parrot.dm
@@ -64,7 +64,6 @@
 	friendly_verb_simple = "groom"
 	mob_size = MOB_SIZE_SMALL
 	movement_type = FLYING
-	gold_core_spawnable = HOSTILE_SPAWN
 
 	var/parrot_damage_upper = 10
 	var/parrot_state = PARROT_WANDER //Hunt for a perch when created
--- a/code/modules/mob/living/simple_animal/parrot.dm
+++ b/code/modules/mob/living/simple_animal/parrot.dm
@@ -64,7 +64,7 @@
 	friendly_verb_simple = "groom"
 	mob_size = MOB_SIZE_SMALL
 	movement_type = FLYING
-	gold_core_spawnable = FRIENDLY_SPAWN
+	gold_core_spawnable = HOSTILE_SPAWN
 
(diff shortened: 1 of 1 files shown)


The-Sun-In-Splendour:
Just remove them from the spawn entirely. You're watering down the hostile spawn with a simplemob that does not hurt anyone unless you hurt it first (and it's kinda shit at hitting)

silicons (dismissed):
what @The-Sun-In-Splendour said

WanderingFox95:
I mean, there were a few ways to go about it,
making gold extracts uninteresting for parrots,
making them drop items before getting qdeleted...
this is just the easiest. 
(I just wasn't sure how many people actually liked parrots.
I figured with how many kinds there are it would be a lot)

BlueWildrose:
I mean, there was my suggestion that parrots just drop the items they're carrying if they get qdel'd, which I feel would be the best way of doing it personally. They'll still steal the gold stabilized extract from your hands, but at least you won't lose that extract when they go poof.

WanderingFox95:
Eh, parrots do take up way too many slots, in my opinion. Variety is good and all but there are maybe 3 cats, 3 dogs and over a dozen parrots. Anyway, removal of the line removes them from gold extracts altogether.

BlueWildrose:
Yeah that's understandable.

Maybe I'll make a unique gold slime reaction one day that makes it spawn only parrots if you inject it with salt. I say salt in particular because crackers are salty.

DeltaFire15 (approved):

PR #14630, opened 2021-04-19
sawing off a lever-action rifle makes it WEAPON_MEDIUM (allowing one-handed firing)

Hatterhat:
## About The Pull Request
see title
## Why It's Good For The Game
well if you're gonna saw it off to lose a round of capacity and make it normal size you might as well gain some sort of benefit to it
(note: i should probably make the capacity suffer a bit more if you do, though...)
## Changelog
:cl:
balance: Sawing-off a lever-action rifle now lets you fire it one-handed. This may not have been wise.
/:cl:


DeltaFire15:
I don't think their ammo cap being reduced by one is enough of a tradeoff to the fact this makes you able to dualwield them / use other weaponry / equipment in your other hand while firing.

PR #14636, opened 2021-04-21
Inconsistency Fixes, 10MM Rifle tweaks.

TheObserver-sys:
## About The Pull Request
Did you know? If you scanned a multitool on the cargo console, you can just.... purchase sec grade armor, guerilla gloves, and a (admittedly awful) 10 MM Rifle without needing to break the crate open. 

Did you also know that you could buy a double pack of combat boots, bullet proof vests and helmets, and the selfsame gloves up above without needing to break the crate as well?

I didn't until I started fiddling about with cargo. This is awfully silly.

So, what this PR is doing:
Relocking these crates behind sec access
Allowing the 10MM Surplus Rifle to be mounted on armor (for some reason, it was set to item_weight_huge instead of bulky), and allows it to fire faster than once a year, taking the delay from 30 to 10. This can be tweaked if it becomes a problem, but it's still an eternity in a heated fight.
Fixes the Dumdum crate from being labeled as Match mistakenly.
Replaces the non armored stuff in the surplus crate with their armored version, as it's mean to be actual service wear, according to the description, anyway.
Fixes a spelling error in said description as well.

## Why It's Good For The Game
Crates should, if nothing else, have a consistent locking policy, unless they are emagged. I can tell you with a fact, these aren't good enough to be emag worthy, so I have elected to fix them up. Gangs is also not in rotation anymore, so the gun we have held over from it should probably be updated to be more usable by the common station folk who decide to buy the partisan crate.

## Changelog
:cl:
fix: Restores the access lock on crates that should have them, given the goods inside.
fix: Makes the 10MM Surplus Rifle a less awful thing to use.
fix: replaces unarmored things with their armored versions.
/:cl:


TheObserver-sys:
Hey, even more of a fun fact, did you know that you can just get a mosin out of the crate with a gun, at a 1 in 5 chance? This is without needing sec or to break the crate open!

This too, is ready for a merge.

PR #14644, opened 2021-04-23
adds kinetic crusher safety override kits to the uplink

silicons:
## About The Pull Request

random horrible idea i cooked up 

## Why It's Good For The Game

creative and hopefully not too overpowered way (MARK DETONATION APPLIES TO TROPHIES STILL!!) for traitors to do the funny stabbing. you're usually better off with an esword if you really want to melee but then again this isn't factoring into how someone with this probably has megafauna loot (as well as the fact that an emag is a bit more multi use than an esword)

## Changelog
:cl:
balance: kinetic crushers can now be emagged to use on people. around 0.25x damage multiplier to crusher detonation bonuses so no hitting people with 100 force backstabs.
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -387,9 +387,8 @@
 	var/obj/item/kinetic_crusher/hammer_synced
 
 /datum/status_effect/crusher_mark/on_creation(mob/living/new_owner, obj/item/kinetic_crusher/new_hammer_synced)
-	. = ..()
-	if(.)
-		hammer_synced = new_hammer_synced
+	hammer_synced = new_hammer_synced
+	return ..()
 
 /datum/status_effect/crusher_mark/on_apply()
 	. = ..()
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -393,7 +393,7 @@
 
 /datum/status_effect/crusher_mark/on_apply()
 	. = ..()
-	if(owner.mob_size >= MOB_SIZE_LARGE)
+	if(hammer_synced? hammer_synced.can_mark(owner) : (owner.mob_size >= MOB_SIZE_LARGE))
 		marked_underlay = mutable_appearance('icons/effects/effects.dmi', "shield2")
 		marked_underlay.pixel_x = -owner.pixel_x
 		marked_underlay.pixel_y = -owner.pixel_y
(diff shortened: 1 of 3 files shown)


necromanceranne:
Is it really a good idea for this to be added to the emag and not it's own item, like a trophy?

silicons:
@necromanceranne i **could** make a traitor item that does this for 2-4 tc or something, if you think that's better

necromanceranne:
Maybe 5, just so it matches the normalized pressure mod cost.

silicons:
doing nec's thing soon

DeltaFire15:
Disliking this in its current state considering glaives are a kinetic crusher subtype and would be affected by this aswell (plus the fact the PKA mod needs a lot of its upgrade mod capacity, while crushers don't have one at all)

ghost:
> 
> 
> Disliking this in its current state considering glaives are a kinetic crusher subtype and would be affected by this aswell (plus the fact the PKA mod needs a lot of its upgrade mod capacity, while crushers don't have one at all)

How would this be an issue? The kinetic glaive not only does less base damage, but the only positive it has over the crusher is the parry, which if you miss, puts you on an insanely long cooldown before you can even attack again. If you miss the parry then you get punished harder than any other weapon.

And for the other part I'm assuming you're referring to it being emag. If not, then crushers likewise can only fit so many trophies at once. I think the max is 4? Meaning if this takes a slot, then that's only 3 options; 2 since legion is mandatory for fluid fighting

ForrestWick:
Hey, Kev, remember when Skyrat did what was basically the same thing as this in the form of buying a crusher that can just damage normal people? I can verify from firsthand accounts and my own witnessing of it that it was an absolute disaster, even with the dumb combat rules at the time of it happening until it's inevitable revert.

ghost:
> 
> 
> Hey, Kev, remember when Skyrat did what was basically the same thing as this in the form of buying a crusher that can just damage normal people? I can verify from firsthand accounts and my own witnessing of it that it was an absolute disaster, even with the dumb combat rules at the time of it happening until it's inevitable revert.
> 
> Things to note: Skyrat's version was an uplink-added weapon that you bought for 15tc, instead of what I remember being an 8tc emag on an item you can get super easily

It wouldn't even be worth it at 15TC honestly. Megafauna loot is easy to beat now that VoG can't hardstun and the meathook isn't some insane 100 AP weapon that knocks down. A crusher can't parry and if it's a glaive, then parrying might save you if you can time everything right. Otherwise you're a sitting duck until the cooldown expires. Even with HECK/Ashdrake, they're countered by lasers/bullets respectively

PR #14653, opened 2021-04-24
ports rp's marking system

timothyteakettle:
## About The Pull Request
title

i say "port" but i am literally having to just write this myself because the codebases are that different please send help

if you don't know what it is:
you can apply markings limb by limb, apply MULTIPLE markings to a limb, colour them individually, and control how they layer over each other

also it removes lizard markings by color matrixing them and making them subtypes of mam markings because this is how it should have been done in the first place, migration for this is done

the main issue with the migration is i think how i convert colours might be wrong and i might need to still colour with sets of 3 values regardless of if it has 1 2 or 3 matrixed sections to ensure that the right piece of data is used, ill have a think about it

## Why It's Good For The Game
unnecessary code bloat that should really be refactored further due to how inefficient it is
oh also literally everyone wants this it's the coolest thing rp has in terms of customization

## Changelog
:cl:
add: ports rp's marking system
/:cl:


IgiariValkyr:
Praise be unto you

silicons:
based

timothyteakettle:
progress update:
image

SandPoot:
Yes, i am doing constant check-outs to see how this is going because i love it so, up/down arrows inverted

IgiariValkyr:
testmerge when

TripleShades:
Current bugs:
Markings are completely solid on slimes
Adding new markings and then trying to color them doesn't work
Synthlizard markings in specific are using the wrong color mapping, or they used to and are now using the right ones(?)
Only Anthromorph X species can use the new markings system

IgiariValkyr:
Other bug: Sometimes you can't just open the appearance tab at all.

Figured out the circumstances: Existing characters can open the appearance tab. Newly created ones can not, regardless of species.

Poojawa:
This is amusing to me because I did the initial system, tried to port it to virgo and just screamed at how horridly out dated all of their code was, but their coders took over the idea and did their implementation. Then later I did the dismemberment-compatible work and color matrix, and now it's getting boosted with a unified overall system.

Very good work.

Also if the matrixing isn't playing nice, I did leave the original RGB compatibility in there as well, you'll probably more-appreciate being able to set a single RGB hex code and just adding one or two extras depending on the sprite itself, rather than Matrix color blocks entirely.

TripleShades:
Bug: Sometimes (perhaps per specific marking? ex. only the chest of Eevee or only the head of Red Panda) the color box wont update from white to reflect the chosen color but the marking will change color properly.
Bug: New markings always start pure white instead of taking from your chosen secondary and tertiary, forcing you to manually set them.

timothyteakettle:
fixed the majority of issues now, and added an 'All' option for adding markings

PR #14658, opened 2021-04-25
Updates tgui and build backend, fixes the build.bat for those with latest eslint (?)

silicons:
## About The Pull Request

All I know is this fixed it and doesn't seem to be causing issues so.

And yes, this ports in a bunch of interfaces we aren't using, but imo it doesn't hurt.


SandPoot:
Please check out the stuff that still need to use citadel's version.

tgui/packages/tgui/interfaces/CrewManifest.js
```diff
@@ -0,0 +1,97 @@
...
+import { useBackend } from "../backend";
+import { Icon, Section, Table, Tooltip } from "../components";
+import { Window } from "../layouts";
+
+const commandJobs = [
+  "Head of Personnel",
+  "Head of Security",
+  "Chief Engineer",
+  "Research Director",
+  "Chief Medical Officer",
```
SandPoot, line 11: Someone's missing

SandPoot: Actually, don't mind, all looks good for me, if someone's going to touch it later, they can change it.

silicons:
#14681

PR #14660, opened 2021-04-26
Shotgun Shell rebalance

dzahlus:
## About The Pull Request
Reduces raw damage of slugs and gives them AP as compensation
Increases raw damage of buck and slightly increases damage falloff

## Why It's Good For The Game
Slugs are just straight up the strongest ranged non antag weapon, kind of fucked up when it's so good nothing else gets used, this gives it AP so it's a viable alternative for armored targets instead of just better than everything. With buckshot now outdamaging slugs, before buckshot did 45 damage if every single pellet hit point blank, now it does 51 because rounding and has -10 AP.

## Changelog
:cl:
balance: rebalanced buckshot and and slugs to be more competitive to eachother
balance: Buckshot damage increase from 7.5 to 8.5 totaling 51 damage point blank -10 AP
balance: Slug damage decreased from 50 to 40, 30 AP given as compensation
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets/shotgun.dm
+++ b/code/modules/projectiles/projectile/bullets/shotgun.dm
@@ -1,6 +1,7 @@
 /obj/item/projectile/bullet/shotgun_slug
 	name = "12g shotgun slug"
-	damage = 50
+	damage = 30
+	armour_penetration = 30
 	sharpness = SHARP_POINTY
 	wound_bonus = 0
 
--- a/code/modules/projectiles/projectile/bullets/shotgun.dm
+++ b/code/modules/projectiles/projectile/bullets/shotgun.dm
@@ -87,12 +87,13 @@
 	return BULLET_ACT_HIT
 
 /obj/item/projectile/bullet/pellet
-	var/tile_dropoff = 0.45
+	var/tile_dropoff = 0.5
 	var/tile_dropoff_s = 1.25
 
 /obj/item/projectile/bullet/pellet/shotgun_buckshot
 	name = "buckshot pellet"
-	damage = 7.5
+	damage = 8.5
+	armour_penetration = -10
 	wound_bonus = 5
 	bare_wound_bonus = 5
 	wound_falloff_tile = -2.5 // low damage + additional dropoff will already curb wounding potential anything past point blank
--- a/code/modules/projectiles/projectile/bullets/shotgun.dm
+++ b/code/modules/projectiles/projectile/bullets/shotgun.dm
@@ -1,6 +1,6 @@
 /obj/item/projectile/bullet/shotgun_slug
 	name = "12g shotgun slug"
-	damage = 30
+	damage = 40
 	armour_penetration = 30
 	sharpness = SHARP_POINTY
 	wound_bonus = 0


silicons:
>30

dzahlus:
> > 30

You and I both know slugs are overpowered, and judging from the reactions of OOC to the opening of this PR everyone agrees, 30 damage seems low, its a 4 hit crit, but you have to remember that it now has 30 AP. I'm willing to bump it to 35, for a 3 hit crit to most targets, remind you that armor is kind of rare and a bullet armor level over 3 is even rarer.

keronshb:
> 
> 
> > 30

As a comparison, .38 lethals do 25 damage and fire 6 similarly long range shots with -20 wound bonus.  Currently Slugs can do the same but fire off 6, 50 damage long range shots which don't have a negative wound bonus.

They both have speedloaders (speedloaders for the .38 and stripperclips for the shotgun).  So I don't think 30 really is that far off the mark.

DeltaFire15:
Most shotguns require both hands to use in combat and / or need to be racked after every shot, making them not exactly the same as the det's revolver which can be used in tandem with other weaponry and does not require racking.

Anyways, onto the actual balancing of the PR:
I'm personally not really opposed to making slugs a more anti-armor weapon, though unsure if the base damage should be reduced by that much.. (For reference: This currently makes them worse than current slugs at anything below ~60 armor rating) Then again, 50 damage on a single limb with one hit is pretty wack.
Don't think the buckshot stuff is needed though.

dzahlus:
> Most shotguns require both hands to use in combat and / or need to be racked after every shot, making them not exactly the same as the det's revolver which can be used in tandem with other weaponry and does not require racking.
> 
> Anyways, onto the actual balancing of the PR:
> I'm personally not really opposed to making slugs a more anti-armor weapon, though unsure if the base damage should be reduced by that much.. (For reference: This currently makes them worse than current slugs at anything below ~60 armor rating) Then again, 50 damage on a single limb with one hit is pretty wack.
> Don't think the buckshot stuff is needed though.

I put the damage at 40 now, my issue was that 2 slugs put you to crit, aswell as funny disembowel because of the extreme damage. As for buckshot nobody is using it, theres no reason to use buckshot when slugs do 50 damage and have better range. Mind you you need to poin blank hit every single pellet to do 51 damage with this PR, right now its 45 damage which is underwhelming in my opinion for hitting that sweet PB.

silicons:
after further thought i'm onboard with trying this...
... after combat balance pr is merged.

the hard truth is that even after that a lot of antagonists will be straight up bullshit to fight with current state of security gear (warops, ninjas, etc).
even after it'll only be slightly better.

Poojawa:
round and round the balance machine goes

DeltaFire15:
@silicons current opinion on this? Would rather get this out of the backlog

Change in general seems reasonable, slugs being more AP-oriented (more effective damage than before at 21+ armor) whilst buckshot is shortrange scatter weak(er) against armor.

DeltaFire15:
Grrr I have heard nothing from anyone since I last commented so I guess I'm just closing this because the current state seems okay.

PR #14674, opened 2021-05-02
[ATMOS CHANGE TESTMERGE THIS OH GOD] adds multiz_accessible_levels infinite loop guard, reworks atmos adjacent turfs proc 

silicons:
- atmos adjacent proc no longer scans multiz and is probably more efficient in general
- get_multiz_accessible_levels is now guarded against infinite loops

the gains of multiz scans are heavily outweighed by the following:

because we don't have a realistic canatmospass that won't be horrific in performance for diagonals (or god forbid, multiz diagonals) that isn't going to be slow as sin, 
each diagonal check on the same zlevel requires 2 steps. when doing 4 diagonals, you're doing a best case of 4 checks to a worst case of 8 checks.

each up/down diagonal check is just as expensive. you have 4 + 4 = 8 of these. this brings access count up to 16 + 8 = 24
now, for the CORNERS of a 3x3x3 blocks, you are now doing: 3 possible first steps, 2 possible second steps, and 1 possible third steps.
3 + 3 * 2 + 3 * 1 = 12 possible checks. 12 * 8 corners = !!48!! checks, **bringing the access count up to 72.**

it's not worth it.
it just isn't.

@Putnam3145 please check my code.

### DIFF
--- a/code/modules/atmospherics/environmental/LINDA_system.dm
+++ b/code/modules/atmospherics/environmental/LINDA_system.dm
@@ -44,7 +44,7 @@
 	return FALSE
 
 /turf/proc/ImmediateCalculateAdjacentTurfs()
-	var/canpass = CANATMOSPASS(src, src) 
+	var/canpass = CANATMOSPASS(src, src)
 	var/canvpass = CANVERTICALATMOSPASS(src, src)
 	for(var/direction in GLOB.cardinals_multiz)
 		var/turf/T = get_step_multiz(src, direction)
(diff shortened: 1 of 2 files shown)


Putnam3145:
Uhh, to be clear, reading the code I don't see multi-Z atmos support... at all? cause multi-Z atmos totally works right now, and we even use it for... one or two maps. Of course, if it doesn't work, that's not a terrible loss, just requires some minor reworking, but yeah.

PR #14443, opened 2021-03-15
Another dreaded baton rework pr; batons knockdown and slow on rightclick (with a cooldown), otherwise stagger, plus more

necromanceranne:
## About The Pull Request

A reworked idea from something I cooked up a while back here https://github.com/tgstation/tgstation/pull/50863 but with iterations I was going to apply once things were in a stable place. Cit actually has a very similar concept already, but quite a bit more oppressive than tg's current batons by taking the worst elements of both and cranking it up a ton.

Stun Batons no longer knock people down on left click, or disarm on right click. 

Stun Batons now knockdown and apply the tased effect (the weak one, so just a stamina degen and slow and not turning off combat mode) on rightclick, but once every few seconds depending on the type of baton. Normal batons have a cooldown of 5 seconds, while stunprods have a cooldown of 7 seconds.

Stun Batons otherwise apply a stagger effect on hit, preventing sprinting. This lasts as long as the tased effect. The effect lasts either 5 seconds for normal stun batons, or 3 seconds for stun prods.  For half these durations, right clicking with disarm intent will also automatically disarm an item from the attacked.

You can't juggle to a new baton to reapply a knockdown. You need to wait until the previous baton would be off cooldown. Borrowing from this pr basically https://github.com/tgstation/tgstation/pull/49774

Stun Batons respect chest armor for their damage, allowing security (or a well armored ne'er-do-well) to reduce incoming damage from a baton. However, the cell max charge also grants the baton improved armor penetration. For every 1000 charge, the baton gains 1 armor pen on the stun attack. Therefore, roundstart batons begin with 15 armor pen. So hitting into a security officer will probably still do a lot of damage to them, but not as much as current. Once you get bluespace cells, armor becomes less viable against the baton. (this likely needs sprite changes to make it more visually discernible)

## Why It's Good For The Game

Batons are fairly goddamn strong as weapons and don't leave the person on the receiving end with many options besides wishing they had a baton themselves. 

It's important they are still usable for detainment and sticking to people, so most of the changes are meant to focus on that aspect. Staggering leaves someone incapable of just disappearing from your screen, and rightclick makes someone far less able to retaliate against you effectively while sapping their speed and stamina.

I am going to atomize a few other changes that likely need to get made alongside this once this gets some discussion. _cough scarp nerf cough_

[...]

PR #14683, opened 2021-05-04
Off Balance Disarms only if shoved into a wall or person.

keronshb:
## About The Pull Request

This changes Off Balance to only disarm if they were shoved into a wall or a person WHILE they're affected by Off Balance.

## Why It's Good For The Game

Letting the crew disarm anyone instantly with two shoves is too good for what it's worth.  It's like everyone has a free and better baton or shield bash.  This way they need to be knocked into something while off balance for them to lose grip.

## Changelog
:cl:
balance: Made it so Off Balance only disarms if they're shoved into a wall or person.
balance: Reduced Off Balance time to 2 seconds.
/:cl:


The-Sun-In-Splendour:
Salt pr?

silicons:
mmmmmm don't like this post, making shovies useless in melee without guns involved especially with sprint being removed soon is pretty awful lol.

necromanceranne:
I made this change, and I actually was of two minds about it. I didn't want security to completely lose the potential of being able to disarm people after a baton hit, so maybe it'd be a good idea to change the off-balance effect to improve the standard disarm probability too. In addition, I felt the mechanic was so finicky (requiring two shoves normally without a baton) and difficult to pull off and so *deliberate* that anyone could see it coming long enough to stow a weapon before a second shove occurred that it wouldn't be too overpowering if we really don't want knockdowns disarming.

It makes sense for what security does for them to be able to get weapons out of peoples hands with their gear, but maybe this was a little too extreme and there are better ways to accomplish this.

Suffice to say, no, it wasn't unintentional, I just didn't clean up a comment. You can check the original pr that changed this effect to verify that. I forgot to update the body, but it's in the changelog, which is the player facing element and the bit that matters. https://github.com/Citadel-Station-13/Citadel-Station-13/pull/14443

ghost:
> 
> 
> I made this change, and I actually was of two minds about it. I didn't want security to completely lose the potential of being able to disarm people after a baton hit, so maybe it'd be a good idea to change the off-balance effect to improve the standard disarm probability too. In addition, I felt the mechanic was so finicky (requiring two shoves normally without a baton) and difficult to pull off and so _deliberate_ that anyone could see it coming long enough to stow a weapon before a second shove occurred that it wouldn't be too overpowering if we really don't want knockdowns disarming.
> 
> It makes sense for what security does for them to be able to get weapons out of peoples hands with their gear, but maybe this was a little too extreme and there are better ways to accomplish this.
> 
> Suffice to say, no, it wasn't unintentional, I just didn't clean up a comment. You can check the original pr that changed this effect to verify that. I forgot to update the body, but it's in the changelog, which is the player facing element and the bit that matters. #14443

Shoving disarm is absolutely awful with melee weapons, because it just takes 2 consecutive pushes and then haha, there goes your weapon.

Besides, security already has the means to disarm consistently with the right click on the baton, followed with a regular disarm. In its current state this buffs the crew's power to absurdly broken levels with the amount of reliability they have in disarming someone with nothing but their hands. I can see it being acceptable for guns, since if you let a person get close that's on you, but melee is melee and not all weapons can be swapped/stashed or doing so would mean you make yourself a sitting duck just because you got lightly shoved once, to avoid being disarmed

necromanceranne:
Yeah, I think in hindsight I'm not particularly impressed with the change and I wouldn't disagree with a revert.

DeltaFire15:
> Besides, security already has the means to disarm consistently with the right click on the baton, followed with a regular disarm.

This is actually related to the same thing, the baton applies to same debuff as the shove does, so this nerfs the baton-related disarm aswell.


Anyways, an idea would be to replace the unconditional disarm if caused by that status with a "always disarm gun, disarm anything if you shove them into a solid object instead"

ghost:
> 
> 
> > Besides, security already has the means to disarm consistently with the right click on the baton, followed with a regular disarm.
> 
> This is actually related to the same thing, the baton applies to same debuff as the shove does, so this nerfs the baton-related disarm aswell.
> 
> Anyways, an idea would be to replace the unconditional disarm if caused by that status with a "always disarm gun, disarm anything if you shove them into a solid object instead"

If that's the case then it'd be better to create a separate debuff that's applied by shoving specifically. But being pushed into walls to disarm any weapon is likewise not great. Especially since it encroaches upon shield bashing, which has a lengthy cooldown whether you hit or not. If it disarmed just ranged weapons then that'd be fine, but not melee

silicons:
do what deltafire said tbh

silicons:
and do NOT have it, if you do, that shoving someone who's down into a wall/person/table/whatever doesn't disarm (aka don't make it only disarm while they're up, meaning to disarm you'd have to shake them up)
that would be very silly and nonsensical.

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -2046,11 +2046,15 @@ GLOBAL_LIST_EMPTY(roundstart_race_names)
...
-					"<span class='danger'>You drop \the [target_held_item]!!</span>", null, COMBAT_MESSAGE_RANGE)
-				append_message += ", causing them to drop [target_held_item]"
+		else if(target.has_status_effect(STATUS_EFFECT_OFF_BALANCE))
+			if(target_held_item)
+				if(shove_blocked)
+					if (target.buckled)
+						return
+					else if(target.dropItemToGround(target_held_item))
+						target.visible_message("<span class='danger'>[target.name] drops \the [target_held_item]!!</span>",
+							"<span class='danger'>You drop \the [target_held_item]!!</span>", null, COMBAT_MESSAGE_RANGE)
```
silicons, line 2056: no extra !

code/__DEFINES/combat.dm
```diff
@@ -158,7 +158,7 @@
 /// how long they're staggered for
 #define SHOVE_STAGGER_DURATION 35
 /// how long they're off balance for
-#define SHOVE_OFFBALANCE_DURATION 30
```
silicons, line 161: @DeltaFire15 thoughts?
i'm against the change.

keronshb: Two seconds feels a lot like 3-4.  The window is bigger than it lets on to be.  When it's at 3 seconds it feels like 4-5.  I could change it back to 3 if we really need it, but even allowing everyone to disarm anything by getting shoved into a wall/person is still really goofy.

code/modules/mob/living/carbon/human/species.dm
keronshb: I'll fix this when the other change is ready too.

PR #14688, opened 2021-05-05
Makes tooltips good and creates a base for adding more.

SandPoot:
## About The Pull Request
Title, this pr will make it so tooltips can be added onto everything if set-up, by default it already adds it to humans and robots, it also solves the problem of trying to drag something but tooltips being on your way.

Example of moving something and not being blocked by tooltips:
i99l2DWgbs
Example of human with tooltips:
image
Example of borg with tooltips:
image
image
Example of something var-edited to have tooltips enabled, using default examine:
image

## Why It's Good For The Game
It is very good.

## Changelog
:cl:
tweak: Tooltips immediately go away on clicking, so you can drag stuff where the tooltip used to be.
fix: Fixed Cyborg examines adding an extra weird line.
code: Everything can be set to have tooltips, and even coded to have neat tooltips.
add: Makes it so humans and borgs already have tooltips.
/:cl:

### DIFF
--- a/code/modules/tooltip/tooltip.dm
+++ b/code/modules/tooltip/tooltip.dm
@@ -122,9 +122,6 @@ Notes:
 		if(user.client && user.client.tooltips)
 			user.client.tooltips.hide()
 
-/atom/movable
-	var/tooltips = FALSE
-
 /atom/movable/proc/get_tooltip_data()
 	return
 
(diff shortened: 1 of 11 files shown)


code/modules/tooltip/tooltip.dm
```diff
@@ -122,4 +122,27 @@ Notes:
 		if(user.client && user.client.tooltips)
 			user.client.tooltips.hide()
 
-
+/atom/movable
```
LetterN: move this var to atom movable

SandPoot: what file?

SandPoot:
image

silicons:
:heart:

PR #14711, opened 2021-05-10
removes gateway visuals because byond being in charge of not stack overflowing is too much to ask 

silicons:
sigh


DeltaFire15 (approved):
sigh

PR #14734, opened 2021-05-18
(READY) Ports Space Dragon Midround Event.

keronshb:
## About The Pull Request

This ports the Space Dragon + all reworks to Citadel for midround events.

Special thanks to @qweq12yt for a lot of help on this.

Requires #14742 

This also adds a few more things, like the owners original body gibbing when being turned into a Space Dragon with the traitor panel and some cit changes since we don't have rightclick usage for things.

## Why It's Good For The Game

We need more midround variety.  This adds that.

Space Dragon is a round ending threat and our current pool of midrounds are a bit stretched thin from some weight changes + pop changes.  Since this is another one added to the pool, it helps keeps midrounds more spiced up.

## Changelog
:cl:
add: Adds the Space Dragon midround event
soundadd: Space Dragon sounds
imageadd: Space Dragon + effects
code: Added Spacewalk trait
code: Gibs the original owner if they are turned into a Space Dragon with the traitor panel
admin: logging for Space Dragon creation
/:cl:


code/modules/events/space_dragon.dm
```diff
@@ -0,0 +1,45 @@
+/datum/round_event_control/space_dragon
+	name = "Spawn Space Dragon"
+	typepath = /datum/round_event/ghost_role/space_dragon
+	weight = 8
+	max_occurrences = 1
+	min_players = 20
```
zeroisthebiggay, line 6: you're putting a lot of faith into a four hundred health antag that can permanently hide corpses, move freely in space, heal, and hardstun people plus summon a tide of carp for just twenty people on >citadel

zeroisthebiggay, line 6: blob has 

```
earliest_start = 40 MINUTES
min_players = 35
```

and this is arguably stronger than blob with how it can permanently hide corpses until its death plus end the round
I'm saying blob should be treated as the baseline for roundending side antags and it definitely shouldn't be allowed to happen before the midround time mark
especially on >citadel

zeroisthebiggay (changes requested):

code/modules/events/space_dragon.dm
keronshb: The thing is with blob, it can lose relatively easy with only 3 people fighting it.  35 pop is still too high for blob, especially now adays when we have even less.

zeroisthebiggay: > The thing is with blob, it can lose relatively easy with only 3 people fighting it. 35 pop is still too high for blob, especially now adays when we have even less.

account for the rate of coomers per players
twenty players might have at most two competents and likely no security

kiwilover8:
given the sheer amount of danger and threat this antag presents, being near blob or at a minimum Xenos level threat on the station, and much harder on its own to kill than a Xeno queen, the population minimum on this antag needs to be raised above the smaller number it's currently at, a good number from a quick asking of people who've played with it think 30-35 and a minimum round time requirement would be a good idea. I'd suggest about 30 minutes. Using Xenos as a basis would also be a realistic idea.

PR #14735, opened 2021-05-18
Igiari Donator Item fix

IgiariValkyr:
It's a uniform damnit, not a suit. I never said make it a suit.

### DIFF
--- a/modular_citadel/code/modules/client/loadout/__donator.dm
+++ b/modular_citadel/code/modules/client/loadout/__donator.dm
@@ -551,6 +551,6 @@
 
 /datum/gear/donator/smaiden
 	name = "Shrine Maiden"
-	slot = SLOT_WEAR_SUIT
+	slot = SLOT_W_UNIFORM
 	path = /obj/item/clothing/suit/smaiden
 	ckeywhitelist = list("ultimarifox")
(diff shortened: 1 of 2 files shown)


modular_citadel/code/modules/client/loadout/__donator.dm
```diff
@@ -551,6 +551,6 @@
 
 /datum/gear/donator/smaiden
 	name = "Shrine Maiden"
```
zeroisthebiggay:
```suggestion
	name = "shrine maiden outfit"
```
'shrine maiden' isn't a name and doesn't need capitalization
lack of object definition makes this show up as
>_This is **Mother Fucker**!_
>_She is wearing a **Shrine Maiden**_

as opposed to

>_This is **Mother Fucker**!_
>_She is wearing a **shrine maiden outfit**_

IgiariValkyr: notice that I only changed one thing. So... blame Jay for that. I'll fix it up.

Linzolle:
This will break the outfit unless you go into modular_citadel/code/modules/custom_loadout/custom_items.dm to change the item to be a uniform. "/obj/item/clothing/under/smaiden" does not exist and will create an error. You should also change the name of the actual item itself to "shrine maiden outfit" since the part you changed it not the name of the actual item itself, but its name that is displayed on the loadout screen.

PR #14759, opened 2021-05-26
Changes the cost of the holoparasite injector and the ricochet eyepatch

Arturlang:
## About The Pull Request
Changes the cost of the holoparasite injector and the ricochet eyepatch, and including the operative price variation.
Holoparasites for traitors had cost 15 crystals, now 12, and operatives it had cost 15, and now 8.
The eyepatch now costs 4 from 8.

## Why It's Good For The Game
Because it's more of a bad thing for them due to having really heavy armor, and the holoparasite who can't wear their armor, and anyways neither of them almost ever use holoparasites due to the player aspect and them using most of their crystals.
And the eyepatch is a really funny gimmick thing to get, but it's just not worth it for the minor effect of sometimes ricocheting, so it was never thought to be really good enough to make it worth the large cost

## Changelog
:cl:
balance: Holoparasites for traitors now cost 12 crystals, for operatives 8, the ricochet eyepath traitor item now 4.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items/uplink_dangerous.dm
+++ b/code/modules/uplink/uplink_items/uplink_dangerous.dm
@@ -189,7 +189,7 @@
 	desc = "Though capable of near sorcerous feats via use of hardlight holograms and nanomachines, they require an \
 			organic host as a home base and source of fuel. Holoparasites come in various types and share damage with their host."
 	item = /obj/item/storage/box/syndie_kit/guardian
-	cost = 15
+	cost = 12
 	limited_stock = 1 // you can only have one holopara apparently?
 	refundable = TRUE
 	cant_discount = TRUE
@@ -204,7 +204,7 @@
 	desc = "Though capable of near sorcerous feats via use of hardlight holograms and nanomachines, they require an \
 			organic host as a home base and source of fuel. Holoparasites come in various types and share damage with their host."
 	item = /obj/item/storage/box/syndie_kit/nukieguardian
-	cost = 15
+	cost = 8
 	refundable = TRUE
 	surplus = 50
 	refund_path = /obj/item/guardiancreator/tech/choose/nukie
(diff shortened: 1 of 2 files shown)


Arturlang:
Please discuss what you think about the changes, no, really, feedback is great, it's why I made this PR

code/modules/uplink/uplink_items/uplink_dangerous.dm
```diff
@@ -204,7 +204,7 @@
 	desc = "Though capable of near sorcerous feats via use of hardlight holograms and nanomachines, they require an \
 			organic host as a home base and source of fuel. Holoparasites come in various types and share damage with their host."
 	item = /obj/item/storage/box/syndie_kit/nukieguardian
-	cost = 15
+	cost = 8
```
zeroisthebiggay, line 207: "and operatives it had cost the same amount, and now 12."
🤔

Arturlang: HAD cost, yes

timothyteakettle:
the eyepatch is incredibly powerful, it ALWAYS ricochets, you just need to fire at an angle to what you hit iirc, if it's not always ricocheting then there's a bug in ricochet code

Arturlang:
@timothyteakettle I think that's just due to the fact hta window's just don't ricochet, which you'd think is a thing that'd make it really likely to ricochet? If someone's willing to change that i'd be willing to remove the change, but it's not appropiate for a purely balance PR

code/modules/uplink/uplink_items/uplink_dangerous.dm
zeroisthebiggay: I am against giving this to ops for so cheap because it's a really powerful tool
support stands (which everyone uses anyways) let the op teleport back to the ship on demand and can heal for free out of crit and while out of combat
the one and only reason I am against this is because support is such a powerful tool and letting nukies have it would basically make it a must buy during warops when there is more than likely enough ghosts to actually accept it
that and chaos and charger make excellent bullet shields with their exceptionally high burn and brute damage respectively plus chaos' innate ability to make you immune to fire ticks

PR #14765, opened 2021-05-27
[Draft] [Testers needed] Tries to fix the cults not being able to do jack shit on station

ghost:
## About The Pull Request

This draft is so people can test this for me because today i'll have some stuff to do, mainly @keronshb
It SHOULD fix the problem cults have since the Port from some weeks ago

Check the edited files, and read the commit comment. Should things stay broken on your side, then something weird happened to my local repo. 
Otherwise if you can attest the fix is working, then my course of action would be just sync that one file i changed directly with /tg/'s, it's just easier than manually change everything
A good thing to do would also be sync both cults, the entirety of area and turf code. but one thing at a time

For now test this, see if Cults are playable like this, if they are, good i'll make this a PR or let Keron copy it on their pr

## Why It's Good For The Game

I mean the antag better work if it's in rotation right?

## Changelog
:cl:
fix: fixed a few things
/:cl:

### DIFF
--- a/code/game/area/Space_Station_13_areas.dm
+++ b/code/game/area/Space_Station_13_areas.dm
@@ -37,7 +37,7 @@ NOTE: there are two lists of areas in the end of this file: centcom and station 
 /area/ai_monitored/turret_protected/AIsatextAP
 	name = "AI Sat Ext"
 	icon_state = "storage"
-	
+
 /area/arrival
 	requires_power = FALSE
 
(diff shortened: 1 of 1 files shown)


Linzolle:
CULT_PERMITTED_1 is looking to be deprecated as the only check that actually uses it will likely be removed soon, so you don't really need to add that flags_1 line. Also, instead of adding VALID_TERRITORY to the area_flags variable, you can instead delete that line since it is redundant.

ghost:
> Also, instead of adding VALID_TERRITORY to the area_flags variable, you can instead delete that line since it is redundant.

deleting that we return to the previous predicament of cults being broken though. I don't see where it's redundant when it's existence ensures proper gameplay (unless you mean that i should put it on area/ instead of area/maintenance)

Linzolle:
> 
> 
> > Also, instead of adding VALID_TERRITORY to the area_flags variable, you can instead delete that line since it is redundant.
> 
> deleting that we return to the previous predicament of cults being broken though. I don't see where it's redundant when it's existence ensures proper gameplay (unless you mean that i should put it on area/ instead of area/maintenance)

/area already defines `var/area_flags = VALID_TERRITORY | BLOBS_ALLOWED | UNIQUE_AREA` in code/game/area/areas.dm
Writing `area_flags = VALID_TERRITORY | BLOBS_ALLOWED | UNIQUE_AREA` is redundant, the problem here is that it was redefined without VALID_TERRITORY, so if you just deleted that line it would default to what /area has.

ghost:
AH very well, then all the issues are just fixed by that, i didn't notice that

PR #14768, opened 2021-05-28
makes xenos immune to stamina damage, with something that might be potentially horrifically abusable but isn't really fixable in a clean way yet 

silicons:
## About The Pull Request

alien/larva bodyparts are now immune to stamina

## Why It's Good For The Game

fix
this might end badly, if someone ends up figuring out how to attach them to humans.

## Changelog
:cl:
fix: xenos are now truly immune to stamina damage.
/:cl:


BlackMajor:
Fix it later then lmao

PR #14770, opened 2021-05-28
Changes Space Adapt to a proper animated outline so it doesn't look awful.

keronshb:
## About The Pull Request

https://streamable.com/j4htim

## Why It's Good For The Game

The previous outline was too small and not noticeable, and it didn't account for digi legs.  Compared to the previous Space Adapt aura which was animated, it also looked like complete shit.

## Changelog
:cl:
tweak: Changes Space Adapt to a proper aura instead of an outline
/:cl:


zeroisthebiggay:
this looks infinitely worse

keronshb:
> 
> 
> this looks infinitely worse

It uses the in game outline instead of a static sprite.  The static sprite looks like absolute shit, you can't even see that someone has it:

image

It also doesn't account for taur or digi style sprites.

ghost:
It's nice to be able to actually see that someone has this, but how does this interact with someone who's bathed in rads?

keronshb:
> 
> 
> It's nice to be able to actually see that someone has this, but how does this interact with someone who's bathed in rads?

It accounts for it

image

Solaris-Shade:
I'm guessing there is some colors of yellow next to green for the space adapt and radiated outline. I have to guess because I'm red-green color blind, (deuteranopia).

keronshb:
> 
> 
> I'm guessing there is some colors of yellow next to green for the space adapt and radiated outline. I have to guess because I'm red-green color blind, (deuteranopia).

Yep if it's layered it's yellow on the inside and green on the outside.

ghost:
im actually unsure as to why a visual feedback for that is even needed...

keronshb:
> 
> 
> im actually unsure as to why a visual feedback for that is even needed...

It's a pretty strong power so it should have a tell (why thermals doesn't have one, I don't know).  For ex you could just throw someone into perma but not realize that they have Space Adapt if it didn't have a visual feedback and it means breaking out easier.

TomGarou:
The 'current' space adaptation is way to subtle, space adaptation needs to stick out like a sore thumb like a hardsuit for mechanical purposes, knowing who can follow you out into space and who can't be cryo healed ect... This is much nicer for folks with poor vision who need to see those 2 sprites on a moving player quickly. Shame it's not possible to make the 'subtle' one a client side option. 

Edit: just thought of an idea.
Maybe a good middle ground?... What about a new chromosome for genetics called 'subtle'?

PR #14776, opened 2021-05-30
New Challenger: The Cat Surgeon has appeared! (Ready!)

WanderingFox95:
## About The Pull Request

So, storytime. Before I came to Citadel I played on Bee for a bit and I recall more than just spiders, syndie troops and carp showing up. Among the assorted maint mobs, there was the cat surgeon. That same one spawning in the kitty ruin, yeah. But instead of just trying to hit you, as it does here, it would actually replace your tail and ears and try to turn you into a catperson - yes, regardless of if you already have a tail or ears. Then it would pacify you and heal you and leave you be and move on to the next. Now, that's hilarious and I really wonder why we didn't have it. I did HOWEVER change the pacifism. Instead of requiring surgery, it can be treated with neurine. Also, as a treat, it also shoots you with furranium. For that extra bit of cat. Oh nyo! 

OH also yes, I did test this locally, an event for it to spawn on the station included.

## Why It's Good For The Game

Scary things don't always have to be spiders, carps or syndies, why not some deranged psycho who wants to turn people into cats?

## Changelog
:cl:
tweak: An obsessed local has a few new tricks, now experiments on people for cat science.
add: A random event for the cat surgeon to invade the station. Listen for scary noises!
soundadd: Screaming Cat SFX, you know, for the mood.
/:cl:


Solaris-Shade:
How does this handle a target with a tuar body type?

WanderingFox95:
> 
> 
> How does this handle a target with a tuar body type?

Ah, yeah, about Taurs and Anthros in general... Anthros kinda don't have tails, codewise. So that one isn't cut off, but is "replaced" but not visually. Same goes for Taurs. But for humans, lizards and felinids, it should work properly.

TL;DR, Visually only the ears will be affected, codewise their "Tails" will be cattails.

WanderingFox95:
Now, I can't seem to trigger the event but according to @keronshb this one should work as an event that randomly spawns a cat surgeon on the station. Rejoice, it's not limited to a space ruin! It's very real and very scary!

code/modules/projectiles/projectile/bullets/dart_syringe.dm
```diff
@@ -38,6 +38,11 @@
 	reagents.add_reagent(/datum/reagent/foaming_agent, 5)
 	reagents.add_reagent(/datum/reagent/toxin/acid, 5)
 
+/obj/item/projectile/bullet/dart/catranq/Initialize()
+	. = ..()
+	reagents.add_reagent(/datum/reagent/toxin/sodium_thiopental, 5)
+	reagents.add_reagent(/datum/reagent/fermi/furranium, 10)
```
zeroisthebiggay: ono

WanderingFox95: Did you mean "Oh nyo?"

bunny232:
How does one fix the cat ears/tail if you say had a custom tail/ears as a human before since afaik tail/ear code is jank. Do you just replace the ears with normal human ears and remove the tail?

WanderingFox95:
> 
> 
> How does one fix the cat ears/tail if you say had a custom tail/ears as a human before since afaik tail/ear code is jank. Do you just replace the ears with normal human ears and remove the tail?

As with Taurs, any tails that don't "actually" exist in organ slots, are unaffected visually. Ears can be replaced as normal, tail should not actually change visually. The original ears and tails (if they exist as objects) are always dropped on the floor so they can be taken to med, like say, arms blown off by an explosion.

WanderingFox95:
OKAY, the event should work now.

kiwilover8:
this really needs a station announcement when it procs

PR #14787, opened 2021-06-02
ORV Uniform Revamp

WanderingFox95:
## About The Pull Request

After realizing that the uniforms I like to use are named ORV because they're inspired by The Orville I double-checked how they're colour-coded in the series and found that ours are kinda.. off. So I fixed that. Now Assistants are grey, as before, Service, Civilians and Cargo are also grey but with golden shoulder pads, MedSci is green, Engineering is orange, Security is red and Command in general is Blue. While command gets blue with metal shoulder pads and the captain gets blue with golden shoulder pads, each department head also has a suit option for their own department colour with golden shoulder pads. The difference may be a little harder to spot, but you got an ID to show your rank, ideally anyway. Also, the Sec uniforms now share the defence stats with all the regular sec uniforms too. Finally, I also took the time to recolour the matching hats. NOW, on to the images!

I spawned some dummies and dressed them up to ensure everything works. And yes, this works with plantigrade and digitigrade legs.
https://orville.fandom.com/wiki/Uniform#Standard_issue_uniform <- Colour Source
OH, Also, I made the badge on the chest the colour of the citadel logo. I felt that was a nice touch.

## Why It's Good For The Game

I guess it's a more accurate reference! It's also not a bad show at all, I checked, it's actually good. Who knows, maybe with this change people who aren't me will use the uniforms too. That'd be fun right?

## Changelog
:cl:
tweak: Split ORV Ops into Sec and Engineering and tweaked accessibility to make cargo part of the service jumpsuit instead of ops. I mean... They push crates and order stuff. What's "ops" about that?
tweak: Now that they're split this way, they're actually coloured the way they are in The Orville, which is what they're based on.
/:cl:


NopemanMcHalt:
Extremely Late to the party, but still:
They were color-coded after Star Trek: The Original Series at first. I didn't wanted 1:1 reference, hence the reason why they were called "Orvi-like" in the code comments.
Still, good job.

PR #14794, opened 2021-06-04
updates GLOB.flags to the new byond limit of 24 flags in one variable, up from 16.

silicons:
this is needed since this list is used for some stuff regarding bitfields. not having this means later flags are invisible.


code/__DEFINES/_flags/_flags.dm
```diff
@@ -17,7 +17,11 @@
 	} while(0);
 #define CHECK_MULTIPLE_BITFIELDS(flagvar, flags) (((flagvar) & (flags)) == (flags))
 
-GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768))
+GLOBAL_LIST_INIT(bitflags, list(
+	1<<0, 1<<1, 1<<2, 1<<3, 1<<4, 1<<5, 1<<6, 1<<7,
```
Putnam3145, line 21: i think you might actually have to bring this up to 16777216 manually

PR #14797, opened 2021-06-05
Delete docker_publish.yml

silicons:
We don't use this and it's failing our CI.


Linzolle (approved):

Linzolle:
Finally i can stop getting spam emails about nothing

PR #14798, opened 2021-06-05
Changes Sec eBows to focus on toxin damage and become modular

keronshb:
## About The Pull Request

There's been a fair amount of complaints about the ebow in sec.  Instead of outright removing it, this changes it up.

The rate of fire is increased to be the same speed as a KA
The bolt was changed to be more noticeable.
Knockdown and Stam Damage were removed.
It can now have a knife and light on it.
It's also compatible with range and cooldown KA modkits.

## Why It's Good For The Game

It's better than outright removing it, and it makes it focus more on the toxin damage aspect more than just free 12 tc items for all.

## Changelog
:cl:
tweak: Made the firing sound more unique
balance: 20 > 16 rate of fire
balance: removed knockdown + stam damage
balance: Allows bayonets and flashlights on it
balance: Is compatible with KA range and cooldown modkits
imageadd: Makes the bolt sprite bigger
/:cl:


kiwedespars:
look at you gutting the ebow.

while i'd say good riddance you're outright salt pr nerfing it into the fucking ground and leaving it at 12 tc, what a joke.

((reading it over again though you're actually gutting the security ebow and not the actual ebow, which is fine but also, up the damage maybe. 20 damage every 2-3 seconds is really fucking bad, and i still retain the right to call it a salt pr))

silicons:
> It's better than outright removing it

this nerfs it so hard that it isn't

keronshb:
I'm open to ideas.  I have no idea what to replace it with.  I'm tinkering with some ideas by letting it be upgraded with KA upgrades, but I'm not sure if that might be too far.

ghost:
> 
> 
> > It's better than outright removing it
> 
> this nerfs it so hard that it isn't

It's a lot better than the current state and there's not much that can be done to make the weapon unique. A weapon that can be mass produced and used by sec in numbers, without it being absolute aids to fight against. Security has their gear and numbers, so you're likely always gonna be up against multiple of these if one is whipped out

silicons:
yeah this is literally worse than removal

PR #14807, opened 2021-06-08
makes medihound sleeper toggle do something

Putnam3145:
## About The Pull Request

Turns out it just changes some minor flavor text to be less fetishy instead of actually doing what you might think it does.

Also makes it so that bodies can't be carried in sleepers at all, which... might be bad? I don't know? Pretty sure you can still buckle people, so.

## Why It's Good For The Game

i'm not sure i need to explain this. one might object on grounds of balance, but I'm not sure medihounds are supposed to be balanced in the first place

## Changelog
:cl:
tweak: "Vocarious Medihound Sleeper" toggle now actually makes you not able to be medihound sleeper'd
/:cl:

### DIFF
--- a/code/game/objects/items/devices/dogborg_sleeper.dm
+++ b/code/game/objects/items/devices/dogborg_sleeper.dm
@@ -81,7 +81,7 @@
 		to_chat(user, "<span class='warning'>The user has opted out of the use of your [src].")
 		return
 	var/voracious = TRUE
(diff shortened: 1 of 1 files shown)


SandPoot:
Consider using the...
```dm
	if(!CHECK_BITFIELD(target.vore_flags, DEVOURABLE))
		to_chat(user, "The target registers an error code. Unable to insert into [src].")
		return
```
or not, dunno.

silicons:
check bitfield is a cancer that i regret to this day

PR #14814, opened 2021-06-09
Delta Tweaks

TripleShades:
## About The Pull Request

Makes some small tweaks to various areas on Delta Station.
Lights were added in Atmos and to AI Sat's outer walkway, because those spots were notably dark.
Swapped out the gas cans from the Gas Miners for the floor labels Box has, since you dont need two clarification visuals.
Removed like, three wires from Security total, and rotated one, because it was extremely messy and overwired in those spots.

## Why It's Good For The Game

Being able to see is good. Clarity that doesn't look messy is good.

## Changelog
:cl:
add: Lights to AI Sat Walkways
add: Lights to Atmospherics
add: Floor labels to Atmospherics Gas Miner containment units
del: Gas canisters from Gas Miner containment units
del: Excessive wiring in Security and Detective's Office
/:cl:


zeroisthebiggay:
8mlrlXb
the chemist can no longer access the fridge without opening the windoor
old direction was intentional

TripleShades:
> 8mlrlXb
> the chemist can no longer access the fridge without opening the windoor
> old direction was intentional

Well shit, okay.
If I added the double windoor like other areas, would that be better? Or should I just revert that entirely.

zeroisthebiggay:
> Well shit, okay.
> If I added the double windoor like other areas, would that be better? Or should I just revert that entirely.

you should revert it unless you feel like redoing chemistry from scratch to prevent that quirk

TripleShades:
> > Well shit, okay.
> > If I added the double windoor like other areas, would that be better? Or should I just revert that entirely.
> 
> you should revert it unless you feel like redoing chemistry from scratch to prevent that quirk

Alright, I'll revert it for now, because I feel like redoing chemistry's layout is beyond the scope of this pr.

PR #14822, opened 2021-06-11
Solving inconsistencies regarding QM being command

ghost:
## About The Pull Request

Here's the problem:
- The rules barely *said* anything;
- I asked some admins they said "YES they are command";
- The code and map says "so-so".

So with that in mind, I went to #ss13-main on the discord server and talked to @silicons and some other folks, and we found common ground.

What have I done:
- QM office airlocks are now the same as command's (I'm willing to remove that for the "QM is not command" meme);
- Box station HoP supply console set to request only, matching other maps;
- Removed useless cargo access from the HoP (already lacked common cargo access, but had mining access, which was useless)
- QMs can make station announcements using the wall mounted request console (not actual change, they always could, but apparently it was not documented).

I suppose Kev will want to read this since it touches some server dynamics, and I'm open to talk about it on discord and clarify even more of it whenever I am awake.
REMINDER that these changes mean NOTHING if the rules are not updated to actually explain that the QM is command (which Silicons said it will, and I'm totally willing to help with that).

tldr: This PR aligns parts of the game to with our **Rules** and the **Way we play**.

## Why It's Good For The Game

This is not even about QM being a head or not, it's **consistency**. The rules are being updated but we played with the QM as a command for a GOOD while, and that doesn't seem to be changing either. The rules didn't said it clearly, and I had to call for admins every once in a while. Not only that, some things IN THE GAME ITSELF could point both in favor and against that decision.
Our Rules don't need to be 100% strict, and that's healthy! But come on now, at least the Chain of Command SHOULD, so we can avoid problems.

## Changelog
:cl:
tweak: QM office airlocks are now blue!
tweak: HoP supply consoles set to request only.
tweak: QMs can make station announcements using the wall mounted request console in their Office. (Documentation only)
del: HoP's cargo access was removed...
/:cl:


ghost:
uh oh i think i broke something

ghost:
i don't know why TGUI stuff is being updated like that somebody help

ghost:
nope it's still there
it's not like anything will break because of that anyway
i hope

PR #14828, opened 2021-06-11
fixes the crafting menu, adds an unit test to automatically yell at people who PR untested code that ended up breaking it for me (thanks drpanis)

silicons:
:)

Well, at least I learned how to use the TGUI debug server on main.


TyrianTyrell (commented):
You misspelled "category."

PR #14842, opened 2021-06-13
[Ready] Wheelchair Sprite Rollback + Motorized Hoverchairs

WanderingFox95:
## About The Pull Request

So, I don't know who did this and when but making wheelchairs hovering chairs is just weird. The scripts even still include that they should have an OVERLAY where the foreground wheel on the sprite went before (which was completely gone but is also restored by this). 

So, here's a theory/explanation:
Manual Wheelchairs: Reason for delayed moves, slower than walking.
Hovering Chair: Probably more fluid motion, about as fast as walking.

The hovering sprites we got seem like a sort of... halfassed change. The noises are still wheel noises. The examination text still refers to the chairs as having wheels. So... I plan on making the hovering chair a motorized one instead.
To prepare for this, I returned the original sprites and renamed the hoverchair sprite until I can get the motorized one tested properly and once I figure out a good sound for it too. Consider this atomization, I suppose.

Edit: Forget Atomization, I already include the motorized hoverchair in this now.

## Why It's Good For The Game
I guess "realism" qualifies? See the explanation above.

## Changelog
:cl:
imageadd: Returns the wheelchair sprites to having, you know, wheels?
add: A motorized wheelchair
addsound: Chairwhoosh. 
/:cl:


WanderingFox95:
So... the only thing that may need some fine-tuning is the movement speed.
As it is, the move delay compared to the normal wheelchair should be halved and that's it, really.
I don't know how the speed is on the server (cause it's configuration dependent), but it's definitely functional beyond that.

PR #14862, opened 2021-06-17
BUG FIX: Changeling Bone Gauntlets work again

keronshb:
## About The Pull Request

This removes code that prevented Bone Gaunlets from working.

## Why It's Good For The Game

Bugfixes good.

## Changelog
:cl:
fix: Ling Bone Gauntlets work again
/:cl:


DeltaFire15 (changes requested):
Instead of deleting this section preventing martial artists from using then, considering merely adding a "return ..()" below the check / its early return

keronshb:
> 
> 
> Instead of deleting this section preventing martial artists from using then, considering merely adding a "return ..()" below the check / its early return

I mean the only time you're going to see martial arts + ling is if a tot and ling work together, or the ling ends up succ'ing someone who hasn't spent a single TC.

Anyway I tried return ..() and that doesn't work at all.

code/modules/antagonists/changeling/powers/mutations.dm
```diff
@@ -683,7 +683,7 @@
 /obj/effect/proc_holder/changeling/gloves/gauntlets/sting_action(mob/living/user)
 	if(HAS_TRAIT(user, TRAIT_NOPUGILIST))
 		to_chat(user, "<span class='warning'>We would gain nothing by forming our fists into brute-force weapons when we are trained in precision martial arts!</span>")
-		return
```
DeltaFire15, line 686: Leave this return in too though so the if still has an effect.

keronshb: Done

DeltaFire15 (approved):

PR #14866, opened 2021-06-17
Lowers Traitor Adrenal poplock to 15

keronshb:
## About The Pull Request

Lings have a lower pop requirement than this and they get to use adrenals as much as they want, where this is a 3 and done type deal.

## Why It's Good For The Game

This seems like an odd design choice to lock adrenals to 25 pop when they're so limited.

## Changelog
:cl:
balance: 25 > 15 pop for adrenals
/:cl:


silicons:
🤔

keronshb:
> 
> 
> 🤔

You only get 3 uses.  If it needs to be at 25 it should be at least 5 uses.

moltoretardo:
Why would you need adrenals on low pop with probably no sec.

keronshb:
On second thought probably a bad idea.

PR #14867, opened 2021-06-17
wheely heelies finally have that downside people have been telling me about

kiwedespars:
## About The Pull Request

to my memory, every time someone talks about wheely-heelies being overpowered, they're always going "it has a downside though!" 10 stamina damage, a little bit of brute and a knockdown is not anywhere close to the same downside of a skateboard or even hoverboard, especially when you can toggle them on-and-off. 

also slowed it down to skateboard speeds since going as fast as a bike is goofy, especially since we removed sprint.

## Why It's Good For The Game

that downside people always tell me about is finally here

## Changelog
:cl:
add: a downside to wheely heelies ((made it actually detrimental))
/:cl:


HeroWithYay:
Yeah to be honest I was wondering how anyone ever managed to slip these into the game in their current form.

PR #14868, opened 2021-06-18
buildmode mapgen tool save [ready]

Arturlang:
## About The Pull Request
Let's admins save maps to their own computer via buildmode.
Ports https://github.com/BeeStation/BeeStation-Hornet/pull/1522

## Why It's Good For The Game
Some good things shouldnt always be lost to the ether.

## Changelog
:cl:
add: Buldmode mapgen save tool
/:cl:


Arturlang:
This works perfectly for exporting maps, format is compatible with strongdmm, in my testing and the cause for the delays I thought it broke due to the inbuilt uploading verbs failing to work, but that is not this pr's fault, as they were already there far before this

Linzolle (approved):

PR #14875, opened 2021-06-21
[Ready] Boxstation's Vault Balance

WanderingFox95:
## About The Pull Request

Okay so hear me out: Boxtation's Vault is in the very centre of the map, connected to the hallway, captain's office and the meeting room/bridge. The Vault itself has a camera with a proximity sensor, so silicons are notified when motion is detected. Objectives like stealing the station's self-destruct core or secret documents are much more difficult to do on Box than on maps like Delta or Pubby. Mind you I wasn't around when it was changed to be this way, I only heard it was moved there and the theatre itself was all-new when that happened. So in order to be more in line with the other maps, I started to experiment, chat and discuss with others. We reached the conclusion that the vault could be back in that hallway but close to command, much like it is on pubby and the gateway could be in its place by the bridge. That extra hallway between the captain's room and the staff room isn't exactly needed anymore then, either.  Now, to make sure there's still space between the vault and the station, I needed to make the theatre a little smaller. Since it was newly added when the first change happened, Mimes and clowns never use it (or anyone for that matter) and there was a lot of empty space, especially on the lower left side of it, I changed the whole thing to be more compact keeping the contents the same. Finally, I moved the Gateway over to where the vault was, so the little area in front of the new gateway is a little smaller, but still contains the same vendors, just fewer tables. Screenshots below.

The new Gateway Location:

The new Theatre and Vault Location:

This change is only a suggestion that came to be as a result of a long discussion, but I'm hoping people will approve.

## Why It's Good For The Game

Making difficulty depend less on the selected Map is a good thing, as is balance. It's still not super far away from the Bridge or Sec, much like on other maps, i.e. Delta or Pubby. Also, something feels very pleasant about the gateway being at the centre of the station. People wanna go there a lot anyway, so now it's a sort of hub location. 

## Changelog
:cl:
tweak: Tweaked the size of the theatre to make some space.
balance: Brings the Vault location more in line with other maps.
/:cl:


Tupinambis:
Hey look it is where the vault used to be before a series of poorly thought out changes to box were made upstream. Glad to see the vault return to its intended location.

WanderingFox95:
Right, so it wasn't the camera.

WanderingFox95:
So it was the OTHER Camera. Good.
No more merge conflicts and no more extra cameras in the theatre.
Here we go.

Solaris-Shade:
This change pleases me both from the view of a crew member and from an antag POV.
Crew: Still a fun chill zone to redecorate just out side of the "new" gateway location. (OLD BACK IN SOMEONE'S DAY) 
Antag: Multiple angles to try and get into the vault. (hack, emag, deconstruction, etc.)

WanderingFox95:
I need to redo this PR, one moment.

PR #14879, opened 2021-06-21
[Review/Testmerge] Icemoon Ruins but on Lavaland!

WanderingFox95:
## About The Pull Request

Well, here's another more experimental PR. I picked some of the Ruins that aren't strictly winter-themed from the ice moon ruins and gave them a Lavaland paint job. Things that were not included are, for example: The forbidden village, The Lavaland-on-the-Icemoon Ashdrake, the Wendigo and the Asteroid. Included Maps can be found below:

The Hot Springs (Note: This one is NOT cursed, I replaced the cursed water with plain water and removed the TCG card):

The Bathhouse (Those are silver walls, not frozen walls):

The ruined Engineering Outpost:

Lust (The card is the Lusty Xenomorph Maid, so I think it's okay to keep it there):

Wrath (I would have considered replacing that with a random tendril but I feel this is designed to HAVE an eldritch portal):

The Library (The only source of secret documents as far as I know. Those are always a bounty, so why not?)

Mining Outpost(Ice Moon BDM Trading Card removed), Ladder was now also removed and there's a BDM by the table.)

Note:  This needs testing as I've yet to see any BDM spawns that aren't the second one in the maps. Config/Blacklist related, perhaps? So that may need a Review or Testmerge for Balance and Testing reasons.

## Why It's Good For The Game

Variety is the spice of space. Adding more ruins to the pool seemed like a fun idea.

## Changelog
:cl:
add: Added new ruin maps: The Bathhouse, The Library, The Engineering outpost, The Hotsprings(un-cursed), Lust, Wrath and an alternate spawn for the BDM in the form of a mining outpost, based on the same Ruins on the Ice Moon.
removed: A lot of the fun items within the ice moon-based ruins that would break mining even more and trading cards.
balance: Yes, this includes that the lavaland version of the hot spring is literally just water and not cursed.
tweak: Also "fixed" that only one version of the BDM ruin ever spawns. Not sure it needed fixing but even locally hosted, only blooddrunk2.dmm would spawn. Since I added another spawn for the BDM, I fixed that too.
/:cl:


WanderingFox95:
Right so now that the check worked,
essentially after a lot of testing, it kept only spawning the ruin labelled blooddrunk2.dmm, which was WEIRD.
So I changed the odds with a probability function to actually randomize them. Decided to include it here since it is supposed to add ruins.

PR #14895, opened 2021-06-27
Adds a new survival pod, the penthouse pod.

shellspeed1:
## About The Pull Request
Does what it says on the can, adds a new survival pod which costs 15k mining points.
https://imgur.com/a/7JnEvNg

## Why It's Good For The Game
Ya know I was looking at what is supposed to be a luxury survival pod and I thought that it was kinda bad. Upon thinking about it further I realized I could make a new survival pod which encourages people to go mining with something actually comfortable and usable as the basis for a lavaland base. Beyond that can be handy for ghost roles that are stuck in space or for free golems. 

## Changelog
:cl:
add: Adds an extremely expensive survival pod to mining for people to work towards. Get to work cracking rocks today.
/:cl:

### DIFF
--- a/_maps/templates/shelter_5.dmm
+++ b/_maps/templates/shelter_5.dmm
@@ -1,0 +1,480 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"cc" = (
+/obj/machinery/vending/boozeomat,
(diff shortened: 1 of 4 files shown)


shellspeed1:
Before I forget it does have an RTG to power the pod itself as it does have an apc and all that. I'm not too worried about this being abused as it only outputs 10k and only one is received per 15k mining points.

PR #14896, opened 2021-06-27
Fixes shoddy math & assumptions causing APCs in sufficiently high-power-requiring rooms to work forever

Putnam3145:
## About The Pull Request

Yeah, as title. I forgot a multiplication on the right hand side of a >= and didn't even need the >= in the first place, since the rest of the function was dedicated to auto-setting power *anyway*. Very silly of me.

## Why It's Good For The Game

Best that loss of power be consequential *at all*.

## Changelog
:cl:
fix: APCs aren't infinite power anymore
/:cl:


DeltaFire15 (approved):
Oh so thats what was causing that. Fix looks good to me.

PR #14907, opened 2021-06-30
Auxmos update (TESTED); updates build scripts to use auxmos (UNTESTED)

Putnam3145:
## About The Pull Request

Makes TGS's build script auto-build auxmos instead of extools. This probably makes it work better on linux, haha.

Also makes generic fires work, whoops. This is mostly irrelevant, since it's not easy to get methane/methyl bromide anyway and those are the only generic fire fuels, but it's best stuff works even if hard to see.

## Why It's Good For The Game

Downstreams running linux are going to want this.

Generic fires working is good anyway.

## Changelog
:cl:
fix: Generic fire works now
server: Linux build script should now compile auxmos instead of extools
/:cl:


cadyn:
I really suggest you take a look at my #14908, but if you really wanna just update the PreCompile.sh yourself, you do need to get rid of the libauxmos.so that's already placed in the root directory of the repo, and also there's a couple additional packages that auxtools depends on: `apt-get install g++-multilib libc6-i386 libstdc++6:i386`

PR #14908, opened 2021-06-30
Make cit actually work on linux by updating scripts

cadyn:
## About The Pull Request
I made a few changes, added a new Build.sh which links to the linux build script from the main directory for easy access, made it so that the main linux build script (now /tools/build/build.sh) also compiles rust_g and auxmos, as well as installing the necessary dependencies as tested on three different distros. I also updated the tgs4 PreCompile.sh to compile auxmos instead of extools now.

## Why It's Good For The Game
I get that most of the people here probably don't care about linux, but I feel like it's used commonly enough that something like this is justifiable. Especially in the world of hosting where an operating system like Linux is often considered more secure, reliable, lightweight, and cheaper compared to Windows. If we can have linux support without interfering with existing windows support, why shouldn't we?

## Changelog
:cl:
server: Updated server scripts for proper linux support
/:cl:


cadyn:
Lovely, didn't notice that CI uses /tools/build/build. Gonna move my custom scripts to /tools/build/build.sh

silicons:
@Putnam3145

PR #14917, opened 2021-07-03
Pubby mirror for the captains office

MrJWhit:
## About The Pull Request

Adds a mirror in the captains bathroom.

## Why It's Good For The Game

Adds a mirror above the sink, so the captain can... I guess look at their face while washing their hands. Why do humans have mirrors above their sink?

## Changelog
:cl:
add: Adds a mirror above the sink in the captains bedroom in pubby
/:cl:


MrJWhit:
Fak it.

PR #14922, opened 2021-07-04
Removes causes_dirt_buildup_on_floor and replaces it with just making humans do that in their Move

Putnam3145:
## About The Pull Request

causes_dirt_buildup_on_floor caused a snowflake proc to be run if and only if that particular var was true, which it was if and only if the mob in question was a human. With that in mind, I just... made human movement do that.

## Why It's Good For The Game

Fewer vars is good.

## Changelog
:cl:
refactor: causes_dirt_buildup_on_floor is now just a thing humans do instead of a weird var only true for humans
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -98,7 +98,7 @@
 		return TRUE
 	return ..()
 
-/mob/living/carbon/human/proc/dirt_buildup(strength)
+/mob/living/carbon/human/proc/dirt_buildup(strength = 1)
 	if(!shoes || !(shoes.body_parts_covered & FEET))
 		return	// barefoot advantage
 	var/turf/open/T = loc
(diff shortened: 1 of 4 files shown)


silicons:
wai didn't i make this
FUCK

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -90,13 +90,31 @@
...
+	if(movement_type & GROUND)
+		dirt_buildup()
 
 /mob/living/carbon/human/Process_Spacemove(movement_dir = 0) //Temporary laziness thing. Will change to handles by species reee.
 	if(dna.species.space_move(src))
 		return TRUE
 	return ..()
 
-/mob/living/carbon/human/dirt_buildup(strength)
+/mob/living/carbon/human/proc/dirt_buildup(strength)
```
Linzolle: shouldn't this also have the strength = 1

Putnam3145: yes

HeroWithYay:
Isn't this likely to be reverted the moment anyone needs something that _isn't_ a human to cause dirt build up on floors?

PR #14924, opened 2021-07-04
handcuffs now require you to stand still to break out of again. Buffs bolas and energy snares breakout time because I forgot to do that when I made you able to move while breaking out of it 

silicons:
I'm really starting to regret blindly letting through every security salt nerf in history.


ghost:
I don't think it's wise to buff them to 5 seconds to resist out of. 5 seconds is a very long time and being able to move at walk speed during it only means you can't accidentally mistime your step and move a half second too early, cancelling the breakout

silicons:
@Lynxless legcuffs can still be broken out of with movement, i just accidentally did this to handcuffs too

PR #14937, opened 2021-07-09
Adds lucky/unlucky quirk, lucky/unlucky item affixes

Putnam3145:
oh god why

## About The Pull Request

Adds a "luck factor" to various rolls in the game, causing various rand() and prob()-type stuff to go toward or against your favor more often depending on your luck. This is only attainable via RPG loot and a couple quirks I added right now. They may or may not be actually far more powerful than I put them down as. This may or may not actually work. I changed too much code. Aa.

## Why It's Good For The Game

It's cool?? I guess?? It might be super bad for performance though??

## Changelog
:cl:
add: Luck
/:cl:


DeltaFire15:
Quite terrible idea due to the massive balance issues with having a trait that makes rng favor (or hate) you. No point cost would balance that much impact.

PR #14940, opened 2021-07-13
adds a neutral trash eating quirk

timothyteakettle:
## About The Pull Request
only objects under the trash object type are considered 'trash'
the quirk is neutral as the items don't affect gameplay it's just a flavor thing

it ties into the vore system by changing the consume text if you have it enabled to your current preferred vore verb and belly name and putting the item into the belly, otherwise it qdel's the item and shows regular text

at some point ill make 'trash' be a object flag

oh also you can't forcefeed garbage to people unless they have a pref for it enabled (because it uses their selected vore verb and belly name if applicable and some people might not want to see that or have it be seen?)

also trash is now an element and cigbutts count as trash

and trash is in the maint pool at 15 weight

## Why It's Good For The Game
let the garbage gremlins exist

## Changelog
:cl:
add: new quirk that allows you to eat trash
/:cl:

### DIFF
--- a/code/__DEFINES/citadel_defines.dm
+++ b/code/__DEFINES/citadel_defines.dm
@@ -101,6 +101,7 @@
 #define NO_AUTO_WAG			(1<<12)
 #define GENITAL_EXAMINE		(1<<13)
 #define VORE_EXAMINE		(1<<14)
+#define TRASH_FORCEFEED		(1<<15)
 #define TOGGLES_CITADEL 0
 
 //belly sound pref things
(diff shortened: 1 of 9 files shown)


BlueWildrose:
https://www.youtube.com/watch?v=mMAbqjjgEec

In all seriousness, I dunno why that it would need a pref flag, trash eating isn't necessarily creepy in the same sense as some erotic things, and you can literally eat crayons already anyways. Let's not mention the people who go out of their way to eat dead mice lying around in maintenance.

LordSaladin:
Trash sandwich

timothyteakettle:
it has a pref because if the person has a vore belly selected it'll use the belly name and verb

bunny232:
https://www.youtube.com/watch?v=wNTGYSG-SFk

PR #14941, opened 2021-07-14
Vore additions/update

SandPoot:
## About The Pull Request
This ports the vore panel from vorestation with some modifications to work here, adds smelling that people would love to have it alongside licking, takes a small fix for tgui_alerts and makes sure that your vore prefs get saved on clicking save prefs on the panel. Everything should be working and if it's not, tell me, also if something needs to go off, tell me too.

## Why It's Good For The Game
TGUI is good.

## Changelog
:cl:
add: Put back mob vore with a pref.
code: Taken some stuff from tg for tgui_alert.
refactor: Refactored a lot of code on vore panel.
/:cl:
CL8XBKcJoY
CKl0mZmMOY
image
image


html/changelog.html
```diff
@@ -590,13 +590,6 @@ <h3 class="author">zeroisthebiggay updated:</h3>
 			<ul class="changes bgimages16">
 				<li class="rscadd">vegas style bunny ears</li>
 			</ul>
-
```
BlueWildrose: wait why is the changelog being changed, leave that alone.

tgui/packages/tgui/interfaces/VorePanel.js
```diff
@@ -0,0 +1,605 @@
...
+            fluid
+            tooltip={"Be able to hear digestion sounds. "
+            + (digestion_sounds
+              ? "Click here to turn off digestion sounds."
+              : "Click here to turn on digestion sounds.")}
+            content={digestion_sounds ? "Digestion Sounds Enabled" : "Digestion Sounds Disabled"} />
+        </Flex.Item>
+        <Flex.Item basis="50%">
+          <Button
+            onClick={() => act("toggle_allow_spontaneous_tf")}
```
BlueWildrose: I dunno if we really need to show features that don't actually exist, unless you want to add something like that in time later. The question is, would this mechanical erotic feature intervene with intended game mechanics? That's the biggest concern when it comes to citcode.

BlueWildrose (commented):
Changelog shouldn't be changed at all, the bots do that.

see comment on VorePanel.js

tgui/packages/tgui/interfaces/VorePanel.js
SandPoot: I'm not 100% about this, it's kind of planned for whoever wishes to touch it, after all, a lot of transformations can already happen in the game.

Linzolle: If you're not going to implement it now, it doesn't need to be there. It can be added back in another PR.

SandPoot: fair nuff, solving commit coming in a bit.

SandPoot:
image
Updated preferences (only the bottom section)

Linzolle (approved):

PR #14944, opened 2021-07-15
Fix tiny oopsies

cadyn:
## About The Pull Request

Uhh, so, before I made a PR that enabled cit to work on linux, but clearly I didn't test the TGS script correctly because this tiny 2 character whoopsies made it through. Should be fine now.

## Why It's Good For The Game

Fixes precompile.sh so that it can actually work for linux now lol

## Changelog
:cl:
server: precompile.sh updated
/:cl:


cadyn:
Not sure why map diff died for me, pff

PR #14945, opened 2021-07-15
fixes photosynthesis stopping nutrition going past well fed from non-photosynthesis means

timothyteakettle:
## About The Pull Request
it's supposed to give you nutrition up to well-fed, not reduce it if you go past well-fed from other means

## Why It's Good For The Game
bug fix

## Changelog
:cl:
fix: fixes photosynthesis stopping nutrition going past well fed from non-photosynthesis means
/:cl:

### DIFF
--- a/code/datums/elements/photosynthesis.dm
+++ b/code/datums/elements/photosynthesis.dm
@@ -59,7 +59,7 @@
 			var/mob/living/L = AM
 			if(L.stat == DEAD)
 				continue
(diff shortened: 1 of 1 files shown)


TheEaterOfBeans:
ok, now give us a quirk that turns this bug into a feature + makes you get hungry faster

timothyteakettle:
yeah okay but i need to figure out if it'd be 1 or -1 or 0

PR #14947, opened 2021-07-16
Crafting menu cleaning up

Arturlang:
Fixes crafting buttons making more of themselves on relogs, and cleans some wierd as fuck shit.

## About The Pull Request
Removes a file that got left behind because TG renamed their file, a random d in the middle of nowhere in tracer.dm, removes the misc tab for the crafting menu because we already have Miscelanious, and puts the stuff that used to be in it to the now existing one, and fixes crafting menu buttons quietly self-replacing themselves on every login.

## Changelog
:cl:
fix: The crafting button should no longer silently make more copies of itself on reconects
code: There are no longer two copies of the crafting component.
code: There is no longer a rogue d in tracer.dm
tweak: Everything in Misc was moved to Miscelanious in the crafting menu, and Misc was nuked from orbit. Nobody will miss you.
/:cl:


Arturlang:
not properly tested, remind me to test it

PR #14955, opened 2021-07-19
What if monophobia took twice as long to start being a issue

bunny232:
## About The Pull Request
Monophobia stress gain is now .25 instead of .5 per tick. Aka twice as long

## Why It's Good For The Game
Maybe people will take this quirk now?

## Changelog
:cl:
balance: Monophobia now takes twice as long to stack stress
/:cl:


timothyteakettle:
A better solution would to just be outright remove the quirk, it encourages metabuddying, and if I recall correctly, only ever PR'd here because people wanted it on skyrat and they were a direct downstream at the time.

TheEaterOfBeans:
Maybe separate the quirk version and the acquired trauma version, and make the quirk version not as bad?
Or maybe that's a bad idea, since... it not being as bad might make too many people use it, and if too many people use it, it'll start causing problems.

TheEaterOfBeans:
> A better solution would to just be outright remove the quirk, it encourages metabuddying, and if I recall correctly, only ever PR'd here because people wanted it on skyrat and they were a direct downstream at the time.

it got PR'd solely because I wanted it for snowflake reasons, other people using it wasn't something I thought would happen (and... it didn't really happen, I've seen like... 2 to 5 people using it and only 1 of them at a time)

TripleShades:
Honestly, monophobia is a pretty neat challenge (especially if you go solo mining), but it does feel like a "Why would a company hire a person with this" sort of deal.
Also, the notifications break frequently on death, and sometimes not even on death, which makes keeping yourself alive kinda shitty feeling outside of dragging around a station pet.
A removal might just be better, like timothyteakettle said, or changing the quirk so it functions more on random with scaling intensity if you dont take medication for it instead of on being alone, like other quirks we have, since I've found that it mimics panic attacks pretty nicely otherwise.

bunny232:
as of right now it takes 4 minutes to reach 60 stress (high stress threshold, where heart attacks become a thing that can happen).
As it stands right now next to no one takes the quirk except those that want a challange. Also AFAIK there hasent been issues with metabuddying due to this quirk, and imo if people are already metabuddying with this quirk removing it won't stop them.

KrabSpider:
Metabuddying is a problem on just about any ERP server under the sun. I doubt buffing a quirk that *requires* you to go interact with someone or else you'll suffer heart attacks after 4 minutes will actually fix that. @timothyteakettle

DeltaFire15:
Considering this trait is **-3** points, it's very reasonable for it to be a very major annoyance. For reference, -3 is brain tumor / paraplegic tier of significance. Even currently, it's basically the least punishing of the -3 traits since you just need to be near at least one person or pet every now and then.

silicons:
Yeah I gotta agree with everyone else here, sorry.

PR #14957, opened 2021-07-20
Pubby Deco

TripleShades:
## About The Pull Request

Adds in a fountain in front of public mining on Pubby Station.
Also moves the public mining area set dressing around to look nicer?

image

## Why It's Good For The Game

Big Empty Space Bad

## Changelog
:cl:
add: Fountain area to public mining station-side
tweak: Moved around the tables and chairs and monitor at public mining station-side
/:cl:


Solaris-Shade:
One question, can I knock someone prone in the water and let them suffocate to death?

TripleShades:
> One question, can I knock someone prone in the water and let them suffocate to death?

Dunno! This is the same water tile as jungle, so if it works there, it works here.

PR #14975, opened 2021-07-31
Fixes certain skirty uniforms not having the FEMALE_UNIFORM_TOP fitted flag

BlueWildrose:
## About The Pull Request

Reminder to future clothing makers implementing clothing into the game: If the uniform has a skirt, make sure to add:
fitted = FEMALE_UNIFORM_TOP
to your outfit.

or maybe just make your uniform a subset of skirt so that we don't have this problem maybe.

this was a webedit change but I know it checks out.

If anyone has noticed this with any other outfits than what I changed, do inform me before this PR gets merged. I merely noticed this while I was customizing a character's loadout so I only went for loadout items that I spotted.

## Why It's Good For The Game

stupid annoying missing pixel at the crotch level due to an activated mask when wearing clothes meant to be a skirt that shouldn't be there. go away.

## Changelog
:cl:
fix: The black dress, pink tutu, the bathrobe, the kimonos, and the qipaos no longer have a missing pixel when wearing them with the feminine bodytype. They're also no longer adjustable (they have no sprite for the adjusted variant and therefore it would just make an error if someone did that.)
/:cl:

### DIFF
--- a/code/modules/clothing/under/costume.dm
+++ b/code/modules/clothing/under/costume.dm
@@ -280,23 +280,21 @@
 	item_state = "qipao"
 	body_parts_covered = CHEST|GROIN
 	can_adjust = FALSE
+	fitted = FEMALE_UNIFORM_TOP
 	mutantrace_variation = STYLE_DIGITIGRADE|STYLE_NO_ANTHRO_ICON
+	
 
 /obj/item/clothing/under/costume/qipao/white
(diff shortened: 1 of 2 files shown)


BlueWildrose:
after checking the history after making this PR I can confirm that @timothyteakettle is the one to blame for this. Everyone bully him.

fuckin' nerd.

PR #14981, opened 2021-08-04
Makes rod of asclepius usable for revival surgery

Putnam3145:
## About The Pull Request

Right now asclepius users are one-handed and can thus only use ghetto setups like stun batons for this surgery, even though the rod of asclepius is explicitly a surgery buffer, so this makes it so they can just use their dang rod

not changing the verb from "shock" because, honestly, i find the idea of this rod of ultimate pacifism and healing shooting a lightning bolt funny

## Why It's Good For The Game

Rod of asclepius is intended to buff surgery, so I'm buffing its surgery

## Changelog
:cl:
balance: Rod of asclepius can now be used for revival surgery
/:cl:

### DIFF
--- a/code/modules/surgery/advanced/revival.dm
+++ b/code/modules/surgery/advanced/revival.dm
@@ -25,7 +25,7 @@
 	return TRUE
 /datum/surgery_step/revive
 	name = "electrically stimulate brain"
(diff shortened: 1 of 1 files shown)


shellspeed1:
what if we just made it function as a defib?

Linzolle:
While the above is a pretty good idea, I'm just gonna merge this for now. Can see about adding that later if that's something you're interested in.

PR #14987, opened 2021-08-07
Reworks body temperature to actually resemble body temperature in any way

Putnam3145:
## About The Pull Request

Instead of body temperature being some weird messed-up delayed representative of the temperature of the environment, body temperature is now actually representing the core body temperature of an individual.

1. Endotherms will slowly heat up their environments, since they actually generate heat. Ectotherms will much more slowly move toward the temperature of the environment, and can now safely live on lavaland (!!).
2. Hypothermia and hyperthermia are now a suite of symptoms instead of raw burn damage. Burn damage still exists, but is caused by the environmental temperature, with no delay.
3. The temperature display is now actually meaningful. It'll show you as feeling warm if your core body temperature is low; unlike before, your core body temperature won't instantly adjust to the surrounding temperature, which was the cause of the paradox. When it says it's hot, it means that it's actually hot around you.
4. The pool now has an internal temperature that people in the pool will move toward via proper heat physics. Sorry, but not really. Ectotherms gain a lot from this.
5. The cryo has been nerfed, somewhat, I... think. Cryoxadone was buffed majorly, but the cryos take a lot more time to reduce body heat sufficiently, and heat their occupants back up after (if you don't let them heat back up, they'll die of heart failure due to hypothermia, so please do that). This one was kind of an accident, oh well.
6. Temperature guns might be a lot more dangerous now, didn't test: they'll cause hallucinations, confusion, vomiting due to causing hypothermia. Bodies can only change 0.25 kelvins per second.
7. Fevers and chills actually change your thermoregulation baseline, i.e. a fever actually just makes your body try to stay at a higher temperature, and *can* cause hyperthermia if it's dangerous (i.e. meets thresholds).

## Why It's Good For The Game

Remains to be seen. I just think it's neat.

## Changelog
:cl:
add: Body temperature
del: "Body temperature"
/:cl:


BlueWildrose:
Boy I can't wait for constant hallucinations as a slimeperson because atmos was leaking into space and therefore letting in space-cold and therefore having my vacuoles proceeding to freak the fuck out even moreso than usual.

Hatterhat:
if you do this consider adding space blankets and whatnot like /vg/station did for hypothermia simulation

edit: the thermal homeostasis regulator, also from /vg/, is also a thing which could help alleviate concerns about cryo. all it does is warm people back up. i just think it's funny

ghost:
While it's neat in idea, it doesn't sound like it's going to be enjoyable at all and simply add a bit of needless complexity with a lot of annoying debuffs thrown in. Tack on that the temperature gun being able to have these effects means you're taking an already absurdly power weapon and making it a must-have

keronshb:
There's a lot of things to consider here.

Fires will become even more deadlier, they're already super deadly because of heat + sweating.  Sometimes I feel it's super hard to cool down from sweating.

Temp Gun but that's a given.

People with disadvantages like Slimes and Cold Blooded people.

Lings and Cryo Stings.

Putnam3145:
Yes, there's likely to be a lot of bizarre knock-on effects and i'm pretty sure I am literally incapable of considering them all.

The most important is that *things that change body temperature no longer deal burn damage*. At all. The effects are completely separate now. The main intent is that body temperature changes are a totally separate health issue from burn damage, which is actually... manageable with tools already in the game, since we have medicines for this.

Cold-blooded people get some leniency on this and their baseline is room-temperature instead of 37C because they'd just die.

Slimes are identical to humans when it comes to hypothermia. They take more *burn damage from cold*, they're not actually weak to the cold.

PR #14992, opened 2021-08-08
Changelings now use a radial menu to select DNAs (tg port)

SandPoot:
This is a port from https://github.com/tgstation/tgstation/pull/55680.
## About The Pull Request
Changelings are now using a radial menu when choosing DNAs instead of a clumsy input menu, which affects transformation and transformation sting abilities. This is important so that changelings have an idea what exact disguise are they putting on.

"Drop" button in the menu is for Drop Flesh Disguise, which removes all disguise parts changeling have currently on.

Example image:

image

It also there's some minor tweaks: Changeling will use the good code for making fake clothes (GLOB.slots) and xenobio changeling does not get antag_moodlets (this will probably be asked to be removed).

I guess i could also make a tgui panel for this but please don't

## Why It's Good For The Game
Better DNA selection method.

## Changelog
:cl:
tweak: Changelings now use a radial menu to select DNAs.
/:cl:


SandPoot:
Currently flesh disguise broke.

PR #14994, opened 2021-08-09
[draft] rewrites character setup

silicons:
## About The Pull Request

completely datumizes character setup
everything is relegated to accessors

character listing is now on the side, switching characters is now easy as fuck
sections overhauled, antagonism and fetish content options can now be overridden to be per-character
migrations standardized and made into procs, etc

things like respawn penalties, and other persistent round tracking like admin mutes are now in /datum/persistent_client_variables
why they were in preferences in the first place is an abomination i will never understand

## Why It's Good For The Game

i don't want to have to explain why having a barely maintainable 5000 line file is bad for a furry erp server that literally lives through customization
the new system will be pretty painful to adapt to because well now you can't just snowflake everything but should be far more modular
everything needed is accessed through accessors including stuff like antag preferences, preferences itself has no direct variables outside of the master data-stores per "category collection"
exceptions include things like keybindings  where performance is key and modularity doesn't matter because there's so many accesses happening we really shouldn't put on the proc overhead

when this is finished i'm going to raise our character slot limit too because it won't be godawful and ugly to look at.

## Changelog
:cl: silicons
refactor: I rewrote character setup and saving. Oof.
/:cl:


silicons:
pending conversion to datumized tgui :joy:

PR #15000, opened 2021-08-11
[WIP] datumized overlay system

timothyteakettle:
## About The Pull Request
this is a gazillion times faster than the existing system because the idea is to only update what needs to be updated by storing overlays in finer categories inside datums

so far it supports limbs, eyes, and all clothing! (not inhands yet though!)

feel free to review it while it's being worked on incase i've many any weird mistakes
as of right now it works in parallel with the existing system, except i've commented out how it added overlays to stop it interfering, i'm slowly replacing one with the other as i go through

an example case: when you want to update your eye colour, currently you have to update your entire body, instead of just, y'know, removing the eye overlay, adding the new one, this new system optimizes cases like these

## Why It's Good For The Game
ridiculous level of optimization

## Changelog
:cl:
add: datumized overlay system
/:cl:


Iamgoofball:
congrats on #15000

timothyteakettle:
split some more stuff up into individual procs which makes it easier to follow for development purposes and also to measure the impact of the new system, it's slowly getting there, less than 100 errors!

timothyteakettle:
splitting into separate pr because there's a better way to optimise overlays

PR #15004, opened 2021-08-13
Admin buildmode repair block now fixes baseturfs instead of simply placing plasteel flooring on empty space

silicons:
Oof!


BlueWildrose:
yeah I was wondering why this happened, this is appreciated.

If this can connect atmos and disposal piping outside the region as well as well it'd be perfect

PR #15008, opened 2021-08-14
[Ready] Shark Tails, dodododododo

WanderingFox95:
## About The Pull Request

So, long story short, I like these sprites, they're better than the ones we have, let's get them. Easy, right? Right.
For comparison, this one's the one we got, the one that looks very much like a carp and not a shark:

And this is the replacement:

## Why It's Good For The Game

Look at them and tell me that you prefer the cheap space-carp knockoff tail, I dare you.

## Changelog
:cl:
imageadd: Better Shark Tails, dodododododo~
add: The old ones are now listed as carp tails.
/:cl:


zeroisthebiggay:
this actually looks worse
replacing cosmetics is cringe and this would be better as an alt so nobody gets upset

Solaris-Shade:
I'm all for more options and items added to the game. This gets my approval.

timothyteakettle:
I'm uh, not quite sure what you've done with the tails. In terms of code, you've added both a human and mam tail named carp tail, and no others meaning I am confused which is the old and new, 

Also it's best to either add a migration so people with the existing tail don't get changed, or to leave the current tail as it is, and rename the new one to be (Alt)

WanderingFox95:
> 
> I'm uh, not quite sure what you've done with the tails. In terms of code, you've added both a human and mam tail named carp tail, and no others meaning I am confused which is the old and new,

I made the old ones into the carp tails and replaced the shark tail with the new shark tails, check sprite diff details for more info.

> 
> Also it's best to either add a migration so people with the existing tail don't get changed, or to leave the current tail as it is, and rename the new one to be (Alt)

I would mostly want to add a migration if I knew how, as making the new one an alt when the old one doesn't look like much of a shark tail, but I also guess looks are highly subjective which would only be a problem if people actually using the tails in question comment. Big Space Sharks lead the Space Carps, that's why it made sense to put the more basic fishtail as carp tails. 

When I split rabbit ears into the floppy and straight ones, neither setting existed and it was a matter of one game til people re-set their things. One game with something you don't like as much isn't _that_ bad. Tails are hidden by most hardsuits and long coats anyway. You could also be having it on another character, notice and set it then. Further, if it's a slimeperson they can just manually change it. I don't think there's a lot of need for nitpicking here.

That being said, if you insist I won't have much of a choice but to change it.

Linzolle:
Why would you not just add a new one with (Alt)...? Seems way simpler and wouldn't change any preexisting characters.

WanderingFox95:
> 
> 
> Why would you not just add a new one with (Alt)...? Seems way simpler and wouldn't change any preexisting characters.

Because preexisting characters can be edited with two clicks and then there's literally no issue.
But okay, I moved some things around again.

PR #15009, opened 2021-08-14
Ship happens on OmegaStation.

EmeraldSundisk:
## About The Pull Request

A continuation of #14964, **this update provides OmegaStation with both a law office and courtroom.** You might be wondering how we found the space, but it should become rather obvious at first glance:
image

Nonetheless, it should suffice for when legal matters absolutely, positively CANNOT wait to be handled at CentCom. **Rest assured the station still has bathrooms, they've just been relocated to the starboard hall:**
image

**You'll also notice a new gateway accessible from the departures hall,** although this isn't exactly a station suited for an expedition. Nevertheless it's there...**as is a pool of all things.** Yes, you read that right:
image

**Unfortunately, the engineers that brought it in managed to destroy the maintenance garden,** although they insist the structural integrity of the surrounding area is wholly unaffected. Currently there are no plans to replace the maintenance garden, since its location is now occupied by a shipwreck. However, **an improvised replacement appears to have manifested in port quarter maintenance:**
image

It'll still likely be a PR or two before this station is ready for primetime, but with this update it's a step closer.

## Why It's Good For The Game

Admittedly I forgot to mention this last time but like the previous PR the goal is to make OmegaStation "fully functional" for all job roles.

## Changelog
:cl:
add: Adds a law office/courtroom to OmegaStation
add: Adds a gateway to OmegaStation
add: Adds a pool/maintenance bar to OmegaStation
del: Removes the original maintenance garden in OmegaStation
tweak: Relocates the bathrooms to the starboard hall
tweak: Modifies port quarter maintenance to include some affected items
/:cl:


Solaris-Shade:
I like how the pool ship came out, it has lots of character and charm about it. would be pretty neat to see what the ship would have looked like completed like.

The micro garden in the asteroid is a brilliant touch, really has that, made by a 'bored assistant' feel to it.

Only just a question about the water tile because in another PR, Friend found out the **/FUN/** way, does the water tile have atmosphere and not thinking it's a space tile?

Regardless I like the path OmegaStation is progressing to an up to date map that can be added to the rotation.

EmeraldSundisk:
> Only just a question about the water tile because in another PR, Friend found out the **/FUN/** way, does the water tile have atmosphere and not thinking it's a space tile?

It SHOULD be fine. I don't see a reason it wouldn't be, unless the puddle itself depressurizes that room.

PR #15014, opened 2021-08-16
Made supermatter respect heat capacity

Putnam3145:
## About The Pull Request

Supermatter was setting temperature directly before, which was dumb. This meant that the actual best supermatter setup was as little nitrogen as you could get, enough hyper-nob to saturate, and huge passive cooling with plasma or hyper-nob underneath all that. Now putting more coolant in will actually *do* something.

An unfortunate knock-on effect is that hyper-nob is now terrifyingly overpowered for the supermatter: even just 500 moles of it rotating around will make it nigh invincible. A much more exciting knock-on effect is that *plasma* supermatters are now very viable, due to its high heat capacity.

## Why It's Good For The Game

It's way more intuitive this way, honestly. Plus, supermatter making *more* heat with *more* coolant is kinda dumb.

## Changelog
:cl:
tweak: Supermatter adjusts heat instead of temperature
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -39,7 +39,8 @@
 #define DAMAGE_INCREASE_MULTIPLIER 0.25
 
 
-#define THERMAL_RELEASE_MODIFIER 5         //Higher == less heat released during reaction, not to be confused with the above values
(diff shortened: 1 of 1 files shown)


LemonInTheDark:
I will be porting this, at least the concept

PR #15019, opened 2021-08-17
[Ready] Blank Bottles

WanderingFox95:
## About The Pull Request

Ever been in that situation where you are a bartender and nobody's asking for any drinks but you still wanna mix drinks? You only got one shaker though and can't store drinks properly, the auto bottler only makes the same bottles that you can get from your vendor... and maybe you want to mix a lot of drinks and make a little menu for the bar - Today's special drink? 

I've found myself in that situation a lot and usually just grabbed beakers or bluespace beakers to do it, but I'm delivering a solution now, a much better one at that. Bottles!

Okay, two bottles and one pitcher. The bottles are ported from Beestation and the pitcher is custom-made. They suggested not to make the bottles available through the autolathe so I added them to the service lathe instead. Small bottles hold 50 units, large bottles hold 100 units and pitchers hold 200 units. They can be found in the bar vending machine in very limited quantities, more can be made in the "Equipment" tab of the service lathe. You can also label them with a hand labeller. ...well, you should, anyway.

## Why It's Good For The Game

Adding more flavour and useful utility based on something that should've probably been a thing before is never wrong.

## Changelog
:cl:
add: Empty bottles and pitchers are available to the bartender now.
imageadd: They even come with 10 different fillstates!
/:cl:


timothyteakettle:
absolutely based, you should be able to name them with a pen because they're a subtype of food containers, if not, i'd suggest adding the item flag

WanderingFox95:
> 
> 
> absolutely based, you should be able to name them with a pen because they're a subtype of food containers, if not, i'd suggest adding the item flag

Done!

HeroWithYay:
The chemistry meta is changing, now we tide into the service lathe for roundstart 200u beakers.

TripleShades:
Could always tie the larger ones to the bottling research node?

WanderingFox95:
> 
> 
> Could always tie the larger ones to the bottling research node?

I'll do some digging into how to make recipes disappear until researched.

That's literally why they're available from the start.
I had them set to be part of the first bottle research but the recipes showed up anyway.

WanderingFox95:
Update: I was wrong.
I just added it to the research node that was unlocked from roundstart. I'll fix that.

WanderingFox95:
To nullify the chances of chemistry abusing the bottles, I lowered the capacity of the pitcher to 120 units, the big bottles to 90 units and buffed the small bottles to 60 units.

PR #15028, opened 2021-08-20
[NEEDS TESTMERGE] Attempts an optimisation for power_change

MarinaGryphon:
## About The Pull Request
- Adds an alternate, hopefully faster list iteration method for machines in power_change
- Removes uses of INVOKE_ASYNC for power_change, since it's slower than normal proc calls and does nothing since it doesn't sleep
TODO: Diagnose what the fuck is making power_change so slow anyway.

## Why It's Good For The Game
Alternative to #15024 that won't break power usage, operating on the hypothesis that the slow thing here is the iteration.
I'm pretty sure it's machines, but this (probably) can't hurt.

## Changelog
:cl:
tweak: Attempts to optimise iteration in `area/proc/power_change()`
/:cl:

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -310,7 +310,7 @@ GLOBAL_LIST_EMPTY(teleportlocs)
 				A.power_light = FALSE
 				A.power_equip = FALSE
 				A.power_environ = FALSE
-			INVOKE_ASYNC(A, .proc/power_change)
+			A.power_change()
 	STOP_PROCESSING(SSobj, src)
 	return ..()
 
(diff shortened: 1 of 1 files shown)


code/game/area/areas.dm
```diff
@@ -310,7 +310,7 @@ GLOBAL_LIST_EMPTY(teleportlocs)
 				A.power_light = FALSE
 				A.power_equip = FALSE
 				A.power_environ = FALSE
-			INVOKE_ASYNC(A, .proc/power_change)
+			A.power_change()
```
LetterN, line 313: this is invoke async because iterating through the machines in an area is slow, and we do **not** want to clog the gc.

MarinaGryphon: this is incorrect, `INVOKE_ASYNC` just sets `waitfor = FALSE` on arbitrary procs, it isn't even remotely similar to a spawn. the name is very misleading. re-adding the `INVOKE_ASYNC` will do nothing but make calling `power_change` here slightly slower.

DeltaFire15 (commented):
Seems good. waitfor only helps vs. sleep calls, which power_change shouldn't do, thus it should be fine.
I'll uncomment the should_not_sleep in power_change and get that to work aswell, since it's a prerequisite for removing the invoke_async (and some things did have sleeps.. or well, one)

DeltaFire15 (changes requested):
Oh hey maintainer push is disabled for this PR, fun!. Guess I get to request changes on this 2.5 year old PR instead of just doing it myself as I did.

Please uncomment the SHOULD_NOT_SLEEP in machinery power_change() and / or throw one into area power_change, then fix the single unaccounted case in firelocks with INVOKE_ASYNC

DeltaFire15:
Actually given there's not been much of a commit history this year, I'll just PR it myself with this required change.

PR #15053, opened 2021-08-30
Adds carcinisoprojection jelly, a limb regrowth chem that is made from slime mutation toxin & blood

BlueWildrose:
## About The Pull Request

This adds a new reagent that regrows limbs, much like how slimepeople can regrow their limbs. It's made from the xenobiology exclusive slime mutation toxin, and blood. Essentially, if you're at that phase in xenobiology where you can turn people into slimepeople, you can make this chemical. Also much like the slimeperson limb regrowth ability, it will use blood to regrow the limbs upon consumption. It doesn't check for a minimum blood amount like the ability however, so any abusers should beware.

Overdosing gives a new special, nasty disease called "crab cancer", which must be cured by mutadone if it occurs. It deals nasty amounts of brute damage under a probability of 5 per tick, while causing flesh (read: meat and crab meat) to fall off of the victim of the disease. The more-human meat, much like how the gibber works, will have the victim's name on it. Could be used as a means to harvest meat from monkeys - there's always silver extracts if the xenobiologist doesn't care about meat otherwise.

## Why It's Good For The Game

Some people may want to regrow their limbs without having to change species.

This essentially is a chemical that removes the middleman with switching to a slime subspecies and back, and offers something that could be passed to the chemistry desk safely. It's balanced around what xenobiology currently is.

## Changelog
:cl:
add: Added carcinisoprojection jelly, a limb regrowth chemical made from xenobiology's slime mutation toxin & blood. Don't overdose!
/:cl:


TheEaterOfBeans:
time to chug carcinisoprojection jelly and bicaridine so I can convert to Sarkicism

bunny232:
I don't see the issue with this for it to be held for so long.

BlueWildrose:
The maintainers have been a bit busy to review the more potentially controversial PRs from what I've understood, but they can take their time.

BlueWildrose:
Looking over this though, maybe I put the overdose threshold a bit too high for giving crab cancer. Xenobiologists already have access to a sufficiently more powerful method of eliminating someone, and that's the advanced mutation toxin from the black slime extracts. Which, if you're already on the path of green slimes, that's only just one more evolution step for xenobiology to access before you have such a thing. You would only need a syringe gun with optional piercing rounds to inject it into others.

We'll see what the maintainers think when they come around for this. Eventually.

TheEaterOfBeans:
Crab cancer should have a small chance to spawn actual crabs.

Linzolle:
I don't know why this exactly needed further discussion or approval, but I personally am into it. Will probably merge this soon if nobody else says anything about it.

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -1799,3 +1799,47 @@
...
+	description = "Also known as \"limb regrowth jelly\", this crabby looking xenobiological jelly will rapidly regrow any missing limbs someone has at the cost of some of their blood. Do not overdose."
+	taste_description = "salty slime"
+	color = "#EF5428" //like cooked crab!
+	reagent_state = LIQUID
+	overdose_threshold = 65 //it takes more than one bluespace syringe to overdose someone with this given how nasty the OD is.
+	value = REAGENT_VALUE_RARE
+
+/datum/reagent/medicine/limb_regrowth/reaction_mob(mob/living/carbon/C, method=TOUCH, reac_volume)
+	. = ..()
+	if(method == TOUCH) //as funny as it would be to have this hurled at someone, nah.
```
DeltaFire15: Preserve parent proc check (it returning false should prevent all further actions) (don't commit this directly because tabs don't work well with this)
```suggestion
	. = ..()
	if(!.)
		return
	if(method == TOUCH) //as funny as it would be to have this hurled at someone, nah.
```

DeltaFire15 (changes requested):
One requested change, plus probably add a check for whether the mob is dead and prevent it from working if true (as reaction_mob does not care about a mob's state)

BlueWildrose:
I added a few extras like a mood debuff and a bounty inclusion so that should be everything.

code/datums/diseases/crabcancer.dm
```diff
@@ -0,0 +1,68 @@
...
+					for(var/i=1 to humanmeatamount)
+						var/obj/item/reagent_containers/food/snacks/meat/slab/newmeat = new meattype
+						newmeat.name = "[affected_mob.real_name] [newmeat.name]"
+						newmeat.forceMove(affected_mob.loc)
+				if(crabmeatamount)
+					for(var/i=1 to crabmeatamount)
+						new /obj/item/reagent_containers/food/snacks/meat/rawcrab(affected_mob.loc)
+				affected_mob.jitteriness += 3
+
+	return
```
DeltaFire15: No return needed here since it's already the end of the proc, should be good to merge other than that so I'll do that as soon as its gone.

DeltaFire15 (approved):

PR #15075, opened 2021-09-03
Fixes the "unreadied player gamemode votes don't count" not working

DeltaFire15:
## About The Pull Request
Title.
The check for readied players was only run if it the vote mode was "gamemode", yet the roundstart gamemode vote mode is "roundtype". This fixes that.

Tested on local and it works.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Unreadied player gamemode votes now actually get ignored (if the config for this is enabled).
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -68,7 +68,7 @@ SUBSYSTEM_DEF(vote)
 	//get the highest number of votes
 	var/greatest_votes = 0
 	var/total_votes = 0
-	if(mode == "gamemode" && CONFIG_GET(flag/must_be_readied_to_vote_gamemode))
+	if((mode == "gamemode" || mode == "roundtype") && CONFIG_GET(flag/must_be_readied_to_vote_gamemode))
 		for(var/mob/dead/new_player/P in GLOB.player_list)
 			if(P.ready != PLAYER_READY_TO_PLAY && voted[P.ckey])
 				choices[choices[voted[P.ckey]]]--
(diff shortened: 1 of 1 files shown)


keronshb:
Get ready for 24/7 extended if this gets in.

BlueWildrose:
> Get ready for 24/7 extended if this gets in.

Part of me wants to say you're right but the other part wants to say "cowards who want to valid the antag but don't want to be valided by the antag have no rights". This is really kind of needed. At the most optimal level, this means people will actually ready up and we could potentially get nukies. Even then, high pop means stronger midround antags and events.

Speaking as an admin, I've seen days where secret voting was overwhelming, days where secret/extended voting are 1:1, and days where it's been overwhelming extended, but once this gets merged and the server gets updated, we'll see where this leads.



Worst case scenario, I can see a bunch of shitheads immediately cryoing as soon as the round starts because they wanted to get a vote in but didn't want to actually play the game.

... Which probably would be something that we need to keep track of.

_sigh._

keronshb:
> 
> 
> > Get ready for 24/7 extended if this gets in.
> 
> Part of me wants to say you're right but the other part wants to say "cowards who want to valid the antag but don't want to be valided by the antag have no rights". This is really kind of needed. At the most optimal level, this means people will actually ready up and we could potentially get nukies. Even then, high pop means stronger midround antags and events.
> 
> Speaking as an admin, I've seen days where secret voting was overwhelming, days where secret/extended voting are 1:1, and days where it's been overwhelming extended, but once this gets merged and the server gets updated, we'll see where this leads.
> 
> Worst case scenario, I can see a bunch of shitheads immediately cryoing as soon as the round starts because they wanted to get a vote in but didn't want to actually play the game.
> 
> ... Which probably would be something that we need to keep track of.
> 
> _sigh._

I suppose the best way to see is if there's any kind of accessible logging for votes right now and then to compare it to votes after this is fixed.

My issue with this was always that this highly discourages against starting ghost roles.  IE if you do want to participate in a secret round but want to play a ghost antagonist role like a Nuke Op borg/Cayanne, a holopara, apprentice wizards/demons, or clock cult ghost roles.  I'm also not sure how this would flow with dynamic too because, IIRC, dynamic nukies is a ghost role and not a round start role.

PR #15084, opened 2021-09-06
Fixes Plague Rats + other buffs

keronshb:
## About The Pull Request

Fixes a myriad of issues with Plague Rats:

1. No more decals.  They were nice, but I want to cut down on lag.
2. No more corpses spawned per rat.
3. They can't spawn/transform off of station z
4. Fixes the explosive growth by putting a 10 second delay per rat
5. they're capped to 10 rats now instead of the config default.
6. To compensate their health is now 100 and the attack was slightly buffed
7. The miasma was nerfed to 3 mols a tick
8. If there's 200 mols on the tile they stop producing miasma
9. To prevent farming, they can ventcrawl through EVERYTHING.

## Why It's Good For The Game

While their explosive nature was great, it was a bit too good and caused the server to crash.

And upon new discovery, atmos was also fucky.  So all of this should fix plagues.

## Changelog
:cl:
fix: Plague Rats will no longer spawn thousands of dirt decals
fix: Plague Rats will no longer spawn thousands of corpses
fix: Plague Rats can't spawn or transform off station z anymore
fix: Plague Rats shouldn't explosively grow instantly
fix: Plague Rats won't spawn infinite miasma anymore.
balance: Plague Rats health is now 100.
balance: Plague Rats can now ventcrawl through everything to prevent farming.
/:cl:


necromanceranne:
I'm gonna be honest chief, introducing a creature that literally just exists to make atmos hell may as well be server sabotage given just how much of a problem atmos diffs cause.

code/modules/mob/living/simple_animal/hostile/plaguerat.dm
```diff
@@ -116,16 +116,18 @@
...
 			new /obj/effect/decal/cleanable/blood/(T)
 		if(7 to 18)
-			to_chat(owner, "<span class='notice'>A corpse rises from the ground. Best to leave it alone.</span>")
-			new /obj/effect/mob_spawn/human/corpse/assistant(T)
+			if(locate(/mob/living/carbon/human) in T)
+				to_chat(owner, "<span class='notice'>A corpse is blocking you from digging.</span>")
+				return
+			else
+				to_chat(owner, "<span class='notice'>A corpse rises from the ground. Best to leave it alone.</span>")
+				new /obj/effect/mob_spawn/human/corpse/assistant(T)
```
LetterN, line 130: this is still iffy, we don't want to have assistants within every tile

LetterN: still iffy, we do not want to have assistants in every single tile.

LetterN (changes requested):
Fix the \\/. Until then @silicons do not merge
```dm
	if(isopenturf(loc))
		var/turf/open/T = src.loc
		if(T.air)
			T.atmos_spawn_air("miasma=5;TEMP=293.15")
```

silicons:
merging since this is better than nothing

plague rats stay disabled until rework tho

PR #15085, opened 2021-09-06
Adds the "keymashing" language.

silicons:
This is most obviously adminspawn only.

Blame @kraso for reminding me.

(in reality I'm just testing the web dev environment)


BlueWildrose:
So this would mean miners with the babel book and curators would know this language?

Well then.

DeltaFire15:
Why is this good for the game?

PR #15089, opened 2021-09-07
Mask of Madness no longer wallhacks

keronshb:
## About The Pull Request

Yeah it shouldn't have ever done this in the first place.

## Why It's Good For The Game

It's probably (note: very) good to be able to see what's causing you to go bonkers.

## Changelog
:cl:
fix: Adds viewers for mask of madness so it doesn't wallhack
/:cl:


silicons:
ill admit i requested this salt pr
that said imo the mask is fine as long as it doesn't wallhack
slowing down everyone around you is **strong**, but counterable

slowing down everyone around you before they can even see you and react is uh, not that fun

PR #15092, opened 2021-09-07
allows autohydro trays to be manually filled

lolman360:
since botany rework became a thing this shouldn't be a problem anymore


Solaris-Shade:
YES!!!!!

PR #15102, opened 2021-09-08
allows custom taste text and color on the custom ice cream setting in the ice cream vat

timothyteakettle:
## About The Pull Request
title
see below:
image

## Why It's Good For The Game
so we can make ice cream with better taste text for the sake of flavour
haha
get it?
flavour?

## Changelog
:cl:
add: allows custom taste text and color on the custom ice cream setting in the ice cream vat
/:cl:


BlueWildrose:
_"You can taste cum."_

BlueWildrose:
Adding to my joke above, in all honesty, this probably should be adminlogged. Possibly into game.log or something for if someone somehow exploits this or someone is being a creep with this system, it can easily be traced back.

timothyteakettle:
not needed because href logs exist and the two links have their own unique tag in the href so they're easy to identify and ctrl f for

PR #15105, opened 2021-09-09
comedically controversial stunbaton """"buff""""

silicons:
## About The Pull Request

Raises stunprod to 30 stamina from 25
Lowers stunbaton to 35 stamina from 40.
**Lowers stunbaton right click taser effect to 30 from 50.**

**DoT removed from right click**

Instead of scaling armor penetration from 0 to 40 based on cell charge max / 1000 (meaning most of the time you have 10), stunbatons now have 100% armor penetration again with 35% penetration on stunprods.

## Why It's Good For The Game

Stunbatons are not nearly as some game ender melee weapon as they were on sprint days when the game was to turn on sprint and dodge someone's shots to close distance and hit them.
I think it's better to return to them being an invariable staple of security instead of one that gets better if you remember to upgrade the cell to 40k, and **one of the few things in security's arsenal (as well as anyone else who steals one) that is more or less unaffected by late round powergaming.**
Furthermore, without sprint, closing the distance in melee is far more of a risky proposition, and should give a good reward for being able to *stay* in melee range.
This actually nerfs their effect on unarmored targets, requiring 5 hits instead of 4 to down from full health (though right click is, actually, hilariously strong due to the 30 DoT **along with the normal damage that also applies** that might need to be nerfed at some point), but this means closing the distance with a heavily armed and armored target is once again viable, regardless of what baton cell you have.

The reason the right click is being nerfed is to make up for the fact that it is *also* full armor penetrating, **and you should probably be able to keep up the melee hits if you are in melee range anyways.**

## Changelog
:cl:
balance: Stunbatons and stunprods now have massively increased/full armor penetration in exchange for slightly less damage overall. Stunbaton right click now does normal damage + slow over 3 seconds instead of 50 DoT + slow over 5 seconds.
balance: Stunprods got buffed. Viva la revolution.
/:cl:

### DIFF
--- a/code/__DEFINES/status_effects.dm
+++ b/code/__DEFINES/status_effects.dm
@@ -65,6 +65,8 @@
 /// Blocks sprint
 #define STATUS_EFFECT_STAGGERED /datum/status_effect/staggered
 
+#define STATUS_EFFECT_TASED_WEAK_NODMG /datum/status_effect/electrode/no_damage     // no damage
+
 #define STATUS_EFFECT_TASED_WEAK /datum/status_effect/electrode				//not as crippling, just slows down
 
 #define STATUS_EFFECT_TASED /datum/status_effect/electrode/no_combat_mode //the affected has been tased, preventing fine muscle control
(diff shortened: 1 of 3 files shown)


dzahlus:
100% armor pen goodbye scarp lmao thanks kevin

silicons:
okay to clarify this since two people have asked already

This doesn't make it any more effective on scarp, because scarp uses physiology **damage mods** and is **not considered armor**.

silicons:
DoT removed.

keronshb:
The 100% armor pen is a bit of a weird choice.  It makes harm batons a bit too good and ignore things like the padding traitors get and I think nukie armor.

PR #15111, opened 2021-09-09
Updates auxmos not to crash

Putnam3145:
## About The Pull Request

Basically, there were some situations where auxmos would crash. Like, on purpose. I thought they were "this should never happen" situations, but ***NEVER TRUST BYOND.***

## Why It's Good For The Game

crashes are bad

## Changelog
:cl:
fix: Removed some crashes
/:cl:


silicons:
ily putnam <3

PR #15113, opened 2021-09-10
Stops the SM from blowing things up

keronshb:
## About The Pull Request

I just witnessed a SM delete tiles multiple times because of machinery and it exploding it and going from 11% integrity back to fucking 100% because it was spaced

Also now incorporates #15114 into this

## Why It's Good For The Game

The SM saving itself is a bit much.  Instead of giving engineers a job to fix it, they just let it fix itself and take care of the damage afterwards.  Currently as is, it's impossible to do any destructive kind of sabotage because it self corrects.

## Changelog
:cl:
balance: Removes zap obj damage and machinery explosion from the SM arcs
/:cl:


Soranayo:
Yeaaah, i can understand that salt PR and i agree, honestly, seeing the SM literally fixing itself up from 28% to 100% without any input is actually incredibly upsetting.

SandPoot:
Oh, so that's why there are holes in the setups.

silicons:
@Putnam3145

Putnam3145 (approved):

silicons:
we discussed with putnam this is fine

DeltaFire15:
Isn't the SM supposed to slowly continue delaminating when in a very-low-pressure environment? Why *is* it fixing itself when its chamber is breached?

keronshb:
> 
> 
> Isn't the SM supposed to slowly continue delaminating when in a very-low-pressure environment? Why _is_ it fixing itself when its chamber is breached?

It should yeah.  Which means anyone can just make a setup right now with a single tile cut in it.

PR #15129, opened 2021-09-13
if you recognize this is not the way to do it, don't do it (bandaid to polychromatic loadout lookup)

silicons:
Runtime.


timothyteakettle:
"if you recognize this is not the way to do it" there's no proper way to do it to my knowledge, i did ask about it at the time and got no reply

Putnam3145:
why does GetElement not work for this

PR #15138, opened 2021-09-13
salt pr to buff/nerf minebots by having their obedience be gained by emag but they also get buffed from it

timothyteakettle:
## About The Pull Request
sentient minebots can be emagged to gain hp, speed, damage, and the main thing: obedience

this comes at the cost of regular sentient minebots not being obedient to the miner because holy shit they have 125 max hp default, being able to churn an infinite amount of these out where they just die, and respawn in a new shell after blowing up a welding tank is stupid

the comments on this (assuming it gets comments and not just emojis) will probably be along the lines of:
'but killer tomato spam' (which will mass die in like 3 seconds to a single spaced tile)
'but gold core mobs' (i would nerf science if i could but i'd be mailed a pipe bomb)
'if you know its a salt pr why make it' (salt prs are not necessarily bad prs)
'you're just mad' (yes)

## Why It's Good For The Game
the funny minebot spam now requires tc but also results in stronger minebots which should be fairly balanced(?)

## Changelog
:cl:
balance: salt pr to buff/nerf minebots by having their obedience be gained by emag but they also get buffed from it
/:cl:

### DIFF
--- a/code/modules/mining/minebot.dm
+++ b/code/modules/mining/minebot.dm
@@ -222,6 +222,22 @@
 		else
 			SetOffenseBehavior()
 
+/mob/living/simple_animal/hostile/mining_drone/emag_act(mob/user)
+	if(client && user)
+		if(obj_flags & EMAGGED)
+			return
+		obj_flags |= EMAGGED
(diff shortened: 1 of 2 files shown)


Knightfall5:
Fastest salt PR in Citadel history. Obligatory "skill issue" too

timothyteakettle:
this should make little difference except instead of being able to spam large quantities of hostile bots, you're only able to spawn a smaller number of stronger bots

PR #15160, opened 2021-09-16
removes passkey from access circuits as it's not used anymore

timothyteakettle:
## About The Pull Request
title

this is nothing to do with regular usage of the circuit, copying id access, etc, you can still do that

tested on local to work with existing circuits, makes no change to them

## Why It's Good For The Game
its not used by anything due to an old pr making circuit stuff look at the assembly access id which is written to by the card reader, the passkey just exists letting people work out the cipherkey which is bad considering some other stuff relies on it

## Changelog
:cl:
fix: removes passkey from access circuits as its not used anymore
/:cl:


SandPoot:
I've been told that this allows you to open any and all airlocks, haven't tested it, but you might want to check it

PR #15164, opened 2021-09-16
saltpr to let janitors put their trash bags on their belts

buffyuwu:
## About The Pull Request

1 line webedit tier salt

adds trash bags as a possible addition to janitor belts like the rest of their supplies because why wasnt this already a thing?? trash bags were added 4 days after janibelts but weirdly never added to the list of things they could be stored into smh

## Changelog
:cl:
tweak: janibelt can now hold trash bags, too
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -594,7 +594,8 @@
 		/obj/item/paint/paint_remover,
 		/obj/item/assembly/mousetrap,
 		/obj/item/screwdriver,
-		/obj/item/stack/cable_coil
+		/obj/item/stack/cable_coil,
+		/obj/item/storage/bag/trash
 		))
 
 /obj/item/storage/belt/bandolier


TripleShades:
Isn't this a bit broken re: balance? Trash bags can hold quite a lot of things, being able to belt them is a huge upgrade.

Solaris-Shade:
I don't think adding Trash bags to the belt is a good choice.

First, the trash bag is a w_class = WEIGHT_CLASS_BULKY the same as a Jani belt.
Second a trash bag can hold 30 items of WEIGHT_CLASS_SMALL that are NOT brains, wallets, or Disky.
Not even mentioning the bluespace trash bags that hold 60 items.

image


Assuming that even half of references to WEIGHT_CLASS_SMALL are to individual items.... 
You looking at well over 100 different items in the game that can be put into a trash bag, in 30 or 60 chunks, times 6 for each slot.
You can fill up a backpack with belts
Or a bag of holding(s).

-1

BlueWildrose:
Let me point out that all the toolbelt items are small items.

This essentially means you could store the regular contents of a toolbelt in a trashbag and hold it around your belt, and hold all sorts of other items as well like sheets of metal and glass and other materials, medical gear, and other nonsense. An engineer might as well just haul trashbags on their belts instead of toolbelts.

Definitely a bad idea.

Also, there's a way to carry around trashbags without them being an inconvenience in your hand all the time - you just put them onto your janicart or janitor pimpmobile. Just saying that.

buffyuwu:
> Let me point out that all the toolbelt items are small items.
> 
> This essentially means you could store the regular contents of a toolbelt in a trashbag and hold it around your belt, and hold all sorts of other items as well like sheets of metal and glass and other materials, medical gear, and other nonsense. An engineer might as well just haul trashbags on their belts instead of toolbelts.
> 
> Definitely a bad idea.
> 
> Also, there's a way to carry around trashbags without them being an inconvenience in your hand all the time - you just put them onto your janicart or janitor pimpmobile. Just saying that.

valid point

PR #15165, opened 2021-09-17
xenobio regen core nerf (replaces aheal with legion core effect, 5s apply)

Hatterhat:
## About The Pull Request
5s apply time for regen cores because lol you can literally shit them out
reduction of heal from literal full heal to basically just legion core

## Why It's Good For The Game
i hate xenobio players and i have no stake in whatever round this happened in again

## Changelog
:cl:
balance: Xenobiology regen cores now have a 5s apply time and don't full heal.
/:cl:

PR #15167, opened 2021-09-17
Xenobio regenerative extract nerfing - Hold still for 5 seconds to use on others, 1 second for self, adds stackable disgust when used

BlueWildrose:
So I was going to do this eventually, but the recent PR had me decide to prioritize this. This is intended to be a better alternative to #15165... which doesn't break anything in the process as well. Which, Hatter's code currently does as of this PR.

## About The Pull Request

This does three things to all regenerative extracts:

- Requires a 5 second windup before it's used on others
- Requires a 1 second windup before it's used on self
- Adds 25 disgust to carbon mobs when used.

## Why It's Good For The Game

Sometime yesterday, a regenerative bluespace extract was used in an unconventional way outside of its intended use, and that was forcefully teleporting an antagonist off the shuttle, essentially denying them their greentext. Because of the instant use, there was no way for that player to counteract that extract. It was basically as overpowered as cultstuns, to put it simply - a one click instant win button, placed in the hands of a crew member.

There's also how abusable these extracts can be. They are a much more powerful healing item that could be used in the middle of combat. Having a one second requirement puts them on the level of sutures for the amount of downtime that's required to use it for healing.

Also with critique of the regen extracts, there was the complaint that they can be spammed. This is where disgust comes in.

It adds 25 disgust every time you use it. With enough time between usages, you can use another one safely. But given it's disgust, if you use too many at once, disgust can make you **vomit**.

Vomiting, as we should know from the nerf to the voice of god, _**hardstuns**_. The disgust factor is designed to punish regenerative abuse. One is perfectly fine, but using another one recently can be risky, and a third one can definitely make you have a chance to start vomiting. So if you're some bad guy that's getting chased around by sec and they're keeping you tightly choked with lethals, you better hope to have enough downtime when you manage to lose them to have that disgust lowered enough for another regenerative dose.

Why I consider this better than Hatter's proposal is that nerfing it to something that's pretty much weaker than legion cores is not only pretty much salt tier nerfing _(the commit messages you gave in your PR are especially pretty sus of your intent Hatter, just saying)_, it otherwise offers a better response to the critique it's been given while keeping it useful in many regards still. Let's try to hold back from doing kneejerk nerf PRs from now on, please.

## Changelog
:cl:
balance: Slime regenerative extracts now require five seconds of wait before they are used. They add 25 disgust when used.
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/regenerative.dm
+++ b/code/modules/research/xenobiology/crossbreeding/regenerative.dm
@@ -21,14 +21,16 @@ Regenerative extracts:
 			to_chat(user, "<span class='warning'>[src] will not work on the dead!</span>")
 			return TRUE //returning TRUE preemptively ends the attack chain and thus doesn't call afterattack, this is noteworthy for below as well
 		//inform the target that they're about to have a regenerative extract used on them
-		if(M != user)
+		if(M != user) //targeting someone else
 			M.visible_message("<span class='notice'>[user] readies [src], holding it steady near [M] and guiding it to the center of [M.p_their()] mass...</span>",
 				"<span class='notice'>[user] readies [src], holding it steady near you and guiding it to the center of your mass...</span>")
-		else
+			if(!do_after(user, 50, target = M)) //5 seconds
+				return TRUE
+		else //targeting self
 			M.visible_message("<span class='notice'>[user] readies [src], holding it steady near [user.p_them()]self and guiding it to the center of [user.p_their()] mass...</span>",
 				"<span class='notice'>You ready [src], holding it steady near you and guiding it to the center of your mass...</span>")
-		if(!do_after(user, 50, target = M)) //5 seconds
-			return TRUE
+			if(!do_after(user, 10, target = M)) //1 second
+				return TRUE
 		. = ..()
 	else
 		. = ..()
@@ -58,6 +60,7 @@ Regenerative extracts:
 	playsound(target, 'sound/effects/splat.ogg', 40, 1)
 	//warn receivers of the extract about the disgust if they're carbon, making it clear that the regenerative extract is causing this.
 	if(iscarbon(M))
+		var/obj/item/organ/stomach/S = M.getorganslot(ORGAN_SLOT_STOMACH) //for getting the stummy name
 		switch(new_disgust_level)
 			if(0 to DISGUST_LEVEL_GROSS)
 				to_chat(M,"<span class='warning'>While you recovered from [src], you feel a little nauseous.</span>")
@@ -66,7 +69,7 @@ Regenerative extracts:
 			if(DISGUST_LEVEL_VERYGROSS to DISGUST_LEVEL_DISGUSTED)
 				to_chat(M,"<span class='warning'>While you recovered from [src], you feel like you're about to vomit!</span>")
 			if(DISGUST_LEVEL_DISGUSTED to INFINITY)
-				to_chat(M,"<span class='userdanger'>You feel absolutely sick. Maybe you should lay off the regenerative extracts until your stomach settles!</span>")
+				to_chat(M,"<span class='userdanger'>You feel absolutely sick. Maybe you should lay off the regenerative extracts until your [(S ? S.name : "stomach")] settles!</span>")
 	qdel(src)
 
 /obj/item/slimecross/regenerative/grey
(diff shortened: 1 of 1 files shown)


DeltaFire15:
Disgust alone already seems like a pretty strong tradeoff for self-use. Would prefer if the apply delay did only apply for others like with e.g. pills.

Also codewise you can probably just put that check in the afterattack aswell if you keep in mind still rechecking proximity etc. after the doafter.

BlueWildrose:
> Disgust alone already seems like a pretty strong tradeoff for self-use. Would prefer if the apply delay did only apply for others like with e.g. pills.

I guess if another maintainer agrees to instant-usage with self use like pills, I'll go for that. I'm personally not against the idea, but I would like another confirmation so people don't go accuse me of being that "xenobio main powergamer trying to keep her powergame gear!!!" - it doesn't help that I've done nothing but xenobiology as of late when I actually play a role besides assistant.

But it arguably makes sense otherwise, just desiring that second opinion.

> Also codewise you can probably just put that check in the afterattack aswell if you keep in mind still rechecking proximity etc. after the doafter.

It plays the attack animation before afterattack, which just looks ugly, seeing someone boop the extract on someone before it's actually been used, so that's why I went for pre_attack instead.

DeltaFire15:
Preattack seems fair enough then.

On the delay front,  if zero is seems a bit low for a fullheal which is a understandable view, five seconds for a self apply definitely seems a bit wack.. If anything go with maybe 1 - 2 in that case.

keronshb:
You can achieve the same teleport effect with a teleprod too.

I'm not sure how I feel about the whole disgust thing at all.  Seems kind of like a huge downside for no super reason.

KrabSpider:
> I'm not sure how I feel about the whole disgust thing at all. Seems kind of like a huge downside for no super reason.
@keronshb 

Disgust scrambles movement, has a chance to induce vomiting, and reduces mood (or sanity, whichever). That was tacked to prevent back to back stacking of what is effectively an aheal. 🙃

By the way, this is for regenerative extracts as a collective, not just regenerative bluespace.

Hatterhat:
> So I was going to do this eventually, but the recent PR had me decide to prioritize this.

you're welcome btw

anyways: making the attack animation play is also a good tell, i think - visual indicators are always nice when you're busy Fucking Gaming
5s is alright, a delay reduction on self usage could be sorta? interesting i.e. 5s to 3.

regen bluespace nerf idea could be adding a trail/tracer back to the site

keronshb:
> 
> 
> > I'm not sure how I feel about the whole disgust thing at all. Seems kind of like a huge downside for no super reason.
> > @keronshb
> 
> Disgust scrambles movement, has a chance to induce vomiting, and reduces mood (or sanity, whichever). That was tacked to prevent back to back stacking of what is effectively an aheal. 🙃
> 
> By the way, this is for regenerative extracts as a collective, not just regenerative bluespace.

Oh wait this isn't for ALL cores?

Oops.

BlueWildrose:
> You can achieve the same teleport effect with a teleprod too.

They're a little different. For reference, regenerative bluespace extracts teleports you to where the bluespace adult slime was crossbread at. Which can straight up _take you straight back to the station_, or if a xenobiologist was feeling clever and nasty, straight up forcefully teleport you into the SM room, or just a shut off room with a remote-detonated grenade of some kind that the victim won't have much chance to escape from - also reinforcing that cultstun comparison I made in my PR. Teleprods, if you're lucky, gives you a chance to get back onto the shuttle when you're struck with it (as long as you don't have a BoH which teleports you further) and that's about it.

> I'm not sure how I feel about the whole disgust thing at all. Seems kind of like a huge downside for no super reason.

I've tested it and feels reasonable enough. Provided nobody's been adding disgust to you through other ways (which currently, there really isn't that many ways aside from self-induced via eating foods considered disgusting to your character) you can get away with about possibly two regens before you start running risks, and otherwise disgust is removed overtime. The alternative is giving a cooldown to using regeneratives, but I considered punishing for abusing too many regens at once to be a more interesting option.

Doing further research just now, it can also be mitigated with cybernetic stomachs, which leads to their own vulnerabilities. EMPs and all that. A possibly dangerous tradeoff.

zeroisthebiggay:
> disgust is removed overtime.

psicodine rapidly removes disgust and is publically available for premium at any large nanomed
disgust is thematic because you're smearing yourself with slime goop to actually use these things and unless you're literally any science player being smeared with the guts of a long dead singlecelled organism would probably be pretty icky and sticky i'd say

BlueWildrose:
If no other maintainers pitch in by tonight, I'll go ahead and reduce/remove the time for self-use as Delta suggested, since switching between branches is a pain in the neck and I want to work on something else without interruptions, so doing it now rather than later would be of interest to me.

BlueWildrose:
Changed to 1 second wait for self-use.

PR #15172, opened 2021-09-17
Massively reduced oxygen req for bonfire

Putnam3145:
## About The Pull Request

Bonfires required 13 moles of oxygen to light. This is... a lot, actually. Lavaland can go all the way down to 5, so sometimes ashwalkers couldn't get it. This sets it to 3 instead of 13.

## Why It's Good For The Game

13 is a huge amount anyway and this barely does anything balance-wise

## Changelog
:cl:
tweak: Bonfires need only 3 moles of oxy to light instead of 13
/:cl:


buffyuwu:
ty my love

PR #15173, opened 2021-09-18
adds PANIC_BUNKER_LIVING_VPN option (READ, THIS IS NOT A VPN BAN)

silicons:
tl;dr

adds an option to only bunker-block people with vpns

**THIS IS NOT A VPN BAN**. Hell, I play on a VPN (no i'm not a ban evader haha haha.. right...)
This means that if you **don't have the required number of minutes**, you will be bounced off the bunker if you have a VPN.
This means established players or those that access request (which they'll be redirected to if they bounce off) **may always play with a VPN.**


BlueWildrose:
This won't halt the tides from coming in, but it'll certainly keep them from coming back. Definitely needed.

timothyteakettle:
based

Issue #15159, opened 2021-09-16
Airlocks and other objects normally requiring a damage threshold to get damaged don't have one


BlueWildrose:
Issue reported from  Round ID: 28377 (Citadel Station)

Reporting client version: 514.1557

You can break open airlocks and some other objects (haven't tracked which exactly) with literally anything as long as there's a force value behind what you hit it with. This is probably not intended behavior. Or if it is, it's kind of a stupid behavior. Normally you have to use a high force weapon like an e-sword or a fireaxe.

## Reproduction:

Break open an airlock using a metal rod or something weak. You can do it. It works.

You can literally break open an airlock with a coin if you take the time to do it.

BlueWildrose:
Issue fixed

Issue #15078, opened 2021-09-04
airlocks have no armor


zeroisthebiggay:
## Round ID:

it does not matter

## Testmerges:

probably does not matter

## Reproduction:

take a lightbulb
break it
use it to break open an airlock and completely ignore the
#define AIRLOCK_DAMAGE_DEFLECTION_N  21  // Normal airlock damage deflection
#define AIRLOCK_DAMAGE_DEFLECTION_R  30  // Reinforced airlock damage deflection

BlueWildrose:
This got fixed

PR #15185, opened 2021-09-19
fixes airlock deflection

silicons:


BlueWildrose:
Add:
fixes #15078
fixes #15159

PR #15187, opened 2021-09-20
blacklists cryogenics from anomaly spawns

KrabSpider:
Blacklists cryo from anomaly spawns

## Why It's Good For The Game
I don't know. Maybe the one area with completely unreplaceable machinery short of admin intervention that is also integral for freeing up job slots shouldn't be one of the areas that have candidacy for spawning a pyroclastic anomaly that will inevitably burn the entire room down within up to 20 seconds and burn *everything* to ash.

## Changelog
:cl:
code: cryogenics ain't a candidate for anomaly spawns anymore. 
/:cl:

### DIFF
--- a/code/modules/events/anomaly.dm
+++ b/code/modules/events/anomaly.dm
@@ -24,7 +24,8 @@
 		/area/holodeck,
 		/area/shuttle,
 		/area/maintenance,
-		/area/science/test_area)
+		/area/science/test_area,
+		/area/commons/cryopod)
 		)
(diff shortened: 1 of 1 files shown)


buffyuwu:
as much fun as it was to bake alive when trying to peacefully leave the round, i thank you

BlueWildrose:
I was considering the idea of marking the chambers as INDESTRUCTIBLE but this is also a good solution

PR #15195, opened 2021-09-22
Weakens chemistry's revival chems

DeltaFire15:
## About The Pull Request
By all that is holy testmerge this first so I don't have to notice I used a < instead of a > somewhere after it's merged or something like that (well actually mostly to see about the balancing and possible oversights)

Currently, medical has three main ways to revive people.
First of which is, clone someone, accept the memory loss and time need of that, then fix them up when it finishes.
Second, fix someone up via surgery and simillar, replace or repair their damaged organs, then defib them and fix their other conditions (with the last sometimes done before defibbing), then worry about the speed of this all due to the five minute memory disorder time mark.

Or, thirdly, use strange reagent or synthtissue, which sidestep all of that and instantly revive an individual with their memory usually intact (strange reagent five minutes as usual), synthtissue *always* apparently)
This third way is very easy to achieve, requiring some very minor preperation (acquiring holy water for strange reagent, setting up a tissue culture or multiple and then just letting it sit for synthtissue) with very little drawback, making them the perfect tools in every single situation if available.
This PR attempts to keep the both of them usable, whilst limiting their superiority over all other methods.

Lets talk about strange reagent first.
This one, previously, required you to be at less than 100 damage brute / burn, in which case it healed all your organs, restored all your blood, healed some oxy and toxdamage and revived you.
Now instead, it heals your blood levels enough to not immediately die and repairs your heart a bit if it's failing,but leaves your other organs alone meaning they'll still require treatment, though the slight tox and oxyloss healing is unaffected. It will also always cause memory loss simillar to reviving somebody too late. Revival of non-carbons (aka, simplemobs) is unaffected.
**Further Updates:**
Strange reagent heals more organs, all failing organs get healed 5 points (which, while leaving them messed up, stops them from being fully failed). The heart is still healed 15 points.

Synthtissue, this one's.. special. See, when healing (and subsequently reviving) somebody dead via synthtissue, it's supposed to slowly cause damage influenced by how much it healed, aka, "borrow" the health temporarily. This unfortunately has been broken ever since synthtissue was initially added due to this borrowed health never being preserved from the reagent applier to the actual person, so it never actually took that into account. This fixes that and therefore for the first time causes it to have actual drawbacks to this method. Additionally, using synthtissue on somebody while they are still recovering from this will drastically worsen this condition. This component of synthtissue also now is locked to being applied via PATCH method (aka, hyposprays, patches, medsprayers, that kinda stuff), instead of being usable via beakersplashing. Like strange reagent, it will cause memory loss when being revived by it, regardless of time passed since death. All other parts of synthtissue (organ healing / replacement, growth speed, healing normally while alive, etc.) are untouched.
**Further Updates:**
Borrowed health gets capped at 250 max and also gradually flips its damage from initially pure clonedamage over to toxdamage, which is far easier to treat (fully flips after 45 cycles total)

Edit: Additionally modifies Hyposprays to use "PATCH" when on sprays mode as normal medsprays already do that. Feels like an inconsistancy between the two (and I expected hypos to already do that)

## Why It's Good For The Game
These two chems currently are the nonplusultra of reviving people, whilst being available from shift start (or close to it, for synthtissue), bypassing a major part of consideration when reviving people due to their sheer superiority in pretty much every aspect when compared to other methods provided they are available.
This weakens them (or, for synthtissue, mostly fixes something that has been broken since forever, so I'll just call it balance change anyways), but should still keep them useful, with their own drawbacks.

## Changelog
:cl:
balance: rebalanced synthtissue's and strange reagent's revival.
tweak: Hyposprays on spray mode now use "patch" as opposed to "touch" to mirror normal medsprays.
/:cl:

### DIFF
--- a/code/__DEFINES/reagents.dm
+++ b/code/__DEFINES/reagents.dm
@@ -63,6 +63,9 @@
 
 #define THRESHOLD_UNHUSK 50 // health threshold for synthflesh/rezadone to unhusk someone
 
+#define SYNTHTISSUE_BORROW_CAP 250	//The cap for synthtissue's borrowed health value when used on someone dead or already having borrowed health.
+#define SYNTHTISSUE_DAMAGE_FLIP_CYCLES 45	//After how many cycles the damage will be pure toxdamage as opposed to clonedamage like initially. Gradually changes during its cycles.
+
 //reagent bitflags, used for altering how they works
 #define REAGENT_DEAD_PROCESS		(1<<0)	//calls on_mob_dead() if present in a dead body
 #define REAGENT_DONOTSPLIT			(1<<1)	//Do not split the chem at all during processing
--- a/code/game/objects/items/defib.dm
+++ b/code/game/objects/items/defib.dm
@@ -423,7 +423,7 @@
 	if((!req_defib && grab_ghost) || (req_defib && defib.grab_ghost))
 		H.notify_ghost_cloning("Your heart is being defibrillated!")
 		H.grab_ghost() // Shove them back in their body.
-	else if(H.can_defib())
+	else if(H.can_revive())
 		H.notify_ghost_cloning("Your heart is being defibrillated. Re-enter your corpse if you want to be revived!", source = src)
 
 	do_help(H, user)
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -898,16 +898,16 @@
 	update_inv_handcuffed()
 	update_hud_handcuffed()
 
-/mob/living/carbon/proc/can_defib()
+/mob/living/carbon/proc/can_revive(ignore_timelimit = FALSE, maximum_brute_dam = MAX_REVIVE_BRUTE_DAMAGE, maximum_fire_dam = MAX_REVIVE_FIRE_DAMAGE, ignore_heart = FALSE)
 	var/tlimit = DEFIB_TIME_LIMIT * 10
 	var/obj/item/organ/heart = getorgan(/obj/item/organ/heart)
 	if(suiciding || hellbound || HAS_TRAIT(src, TRAIT_HUSK) || AmBloodsucker(src))
 		return
-	if((world.time - timeofdeath) > tlimit)
+	if(!ignore_timelimit && (world.time - timeofdeath) > tlimit)
 		return
 	if((getBruteLoss() >= MAX_REVIVE_BRUTE_DAMAGE) || (getFireLoss() >= MAX_REVIVE_FIRE_DAMAGE))
 		return
-	if(!heart || (heart.organ_flags & ORGAN_FAILING))
+	if(!ignore_heart && (!heart || (heart.organ_flags & ORGAN_FAILING)))
 		return
 	var/obj/item/organ/brain/BR = getorgan(/obj/item/organ/brain)
 	if(QDELETED(BR) || BR.brain_death || (BR.organ_flags & ORGAN_FAILING) || suiciding)
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -905,7 +905,7 @@
 		return
 	if(!ignore_timelimit && (world.time - timeofdeath) > tlimit)
 		return
-	if((getBruteLoss() >= MAX_REVIVE_BRUTE_DAMAGE) || (getFireLoss() >= MAX_REVIVE_FIRE_DAMAGE))
+	if((getBruteLoss() >= maximum_brute_dam) || (getFireLoss() >= maximum_fire_dam))
 		return
 	if(!ignore_heart && (!heart || (heart.organ_flags & ORGAN_FAILING)))
 		return
--- a/code/modules/reagents/chemistry/holder.dm
+++ b/code/modules/reagents/chemistry/holder.dm
@@ -199,6 +199,7 @@
 		var/transfer_amount = T.volume * part
 		if(preserve_data)
 			trans_data = copy_data(T)
+			post_copy_data(T)
 		transferred += "[T] - [transfer_amount]"
 
 		R.add_reagent(T.type, transfer_amount * multiplier, trans_data, chem_temp, T.purity, pH, no_react = TRUE, ignore_pH = TRUE) //we only handle reaction after every reagent has been transfered.
@@ -239,7 +240,8 @@
 		var/datum/reagent/T = reagent
 		var/copy_amount = T.volume * part
 		if(preserve_data)
-			trans_data = T.data
+			trans_data = copy_data(T)
+			post_copy_data(T)
 		R.add_reagent(T.type, copy_amount * multiplier, trans_data)
 
 	src.update_total()
@@ -266,7 +268,8 @@
 		var/datum/reagent/current_reagent = CR
 		if(current_reagent.type == reagent)
 			if(preserve_data)
-				trans_data = current_reagent.data
+				trans_data = copy_data(current_reagent)
+				post_copy_data(current_reagent)
 			R.add_reagent(current_reagent.type, amount, trans_data, chem_temp, current_reagent.purity, pH, no_react = TRUE)
 			remove_reagent(current_reagent.type, amount, 1)
 			if(log && amount > 0)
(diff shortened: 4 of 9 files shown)


shellspeed1:
I'm not sure I necessarily agree with the SR and synthtissue causing guaranteed memory loss. This is locking medbay into having to defib or use revival surgery to ensure memories if they want to recover that. SR in particular is definitely not just an instant revive due to having to treat them and you are essentially relegating it into being a worse revival option. I would say that it would overall be better to lock it to the same timer that we have for defib otherwise I'm fine with the other changes.

EvGRyder:
During my tests large colonies seem to dump massive amounts of brute damage on metab end regardless of whether purge chems were used or not. My suspicion is that borrowed health needs to be capped or properly prorated with metabolism rate so that high level colonies don't dump leftover borrowed health that hasn't been processed.

Chiirno:
From the tests I've made, any sufficiently large colony is going to dump a massive amount of brute damage on mob metab end regardless of whether a purge chem was used or not. More than likely it's dumping any remaining borrowed health, _which isn't dependent on metabolism rate, only on colony size_, once the chem is metabolized out. You'll likely need to cap damage from borrowed health or properly prorate metabolic rate so the borrowed health is reduced in time. Otherwise you'll be having corpses with no damage, revived on 10u synthtissue, properly have their toxins and clone damage addressed, die 5 minutes later with 900 brute damage because the colony was **too good.**

DeltaFire15:
> From the tests I've made, any sufficiently large colony is going to dump a massive amount of brute damage on mob metab end regardless of whether a purge chem was used or not. More than likely it's dumping any remaining borrowed health, _which isn't dependent on metabolism rate, only on colony size_, once the chem is metabolized out. You'll likely need to cap damage from borrowed health or properly prorate metabolic rate so the borrowed health is reduced in time. Otherwise you'll be having corpses with no damage, revived on 10u synthtissue, properly have their toxins and clone damage addressed, die 5 minutes later with 900 brute damage because the colony was **too good.**

Fair point, this is due to the way this 'borrowed health' gets added (which I didn't really touch conceptwise as of now). Aka, it goes directly of 'how potent is this synthtissue' as opposed to 'how much did it actually heal'.
 I'll be looking into replacing the former with the latter, whether it be 1:1 or somewhat scaled I'm not yet sure.
Also borrowed heal always decreases the same amount, whilst strength per u can be scaled infinitely.. Which, honestly is kinda an issue to deal with but I guess linking it to health actually healed would make it useful to add *more* synthtissue than needed for the pure heal so there's more time for it to metab and slowly take that damage instead of say, getting all at once (which funnily enough makes purge chems scary to use on people recovering from it, honestly an interaction I like)

DeltaFire15:
Updated borrowed health to go off healed damage (oxydamage healing for dead counts half due to how trivial it is and defibs fixing it) and doubled default reduction of the value since it deals 2 damage per tick, modified by metabolism as that changes process rate. Might still be somewhat slow total time wise but should technically be less likely to cause its big backfire.
Using purge chems will still very likely kill them.

TheObserver-sys:
If I recall, Strange Reagent already forces you to forget, as you are litterally being resurrected by funky magic chem. Every time I've been resurrected by it, I've gotten the spiel about my memories being shot..

Chiirno:
While not as badly the problem is still occurring on some edge cases.
image

Killed this NPC with 1k brute damage then removed all damage using VV.
After patching them with 1u of synthtissue grown to 999u colony size, they revived, took toxin and clone damage for 20 cycles ( accruing 20 clone and toxins damage in the process, exactly 1 per 0.05u metabolism cycle ) and then getting bumped to 75 brute, 95 clone, and 95 toxins damage.
image

While a colony of size 999u isn't exactly realistic (I max out at around 650u during a 2hr shift), I don't think this is intended behavior.
I'm also unable to figure out the reason this is happening.

On the other hand, I tried the same tests with 5u and after a while they stopped taking clone and toxin damage while still having synthtissue to metabolize. On metab end they accrued no additional damage, so that seems to be working fine. 

Giving them a pent injector after the 5u did screw them up pretty badly.
image

DeltaFire15:
> While not as badly the problem is still occurring on some edge cases.
> image
> 
> Killed this NPC with 1k brute damage then removed all damage using VV.
> After patching them with 1u of synthtissue grown to 999u colony size, they revived, took toxin and clone damage for 20 cycles ( accruing 20 clone and toxins damage in the process, exactly 1 per 0.05u metabolism cycle ) and then getting bumped to 75 brute, 95 clone, and 95 toxins damage.
> image
> 
> While a colony of size 999u isn't exactly realistic (I max out at around 650u during a 2hr shift), I don't think this is intended behavior.
> I'm also unable to figure out the reason this is happening.
> 
> On the other hand, I tried the same tests with 5u and after a while they stopped taking clone and toxin damage while still having synthtissue to metabolize. On metab end they accrued no additional damage, so that seems to be working fine.
> 
> Giving them a pent injector after the 5u did screw them up pretty badly.
> image

The actual action of reviving somebody via synthtissue either multiplies their borrowed health applied by this specific heal by 2 or sets it to 100, whichever is worse.
So yeah if you want to revive someone with little damage, it may be advisable to revive them normally, or add more synthtissue than needed as a 'buffer' for the metabolisation speed.

KrabSpider:
Good.

RogueSteampunker:
Yeah, to be quite honest, this makes using strange reagent and synthflesh entirely purposeless.

Organ damage is half the reason anyone uses strange reagent over defibs or the cloner. Coupled with toxins damage rapidly hitting someone as soon as they're revived due to rotten organs, and you have a chem who can't reliably be used outside of xenobio tomfuckery until you have experimental surgery to remove toxins and heal organs - except, by that point, you have revival surgery that does the exact same thing but with an organ heal.

It needs touched at, yeah. But this is just way too far with it.

KrabSpider:
Hey why not just add an actual mininum threshold to strange reagent, implement a cap to the organ healing so atleast there's **some** fucking point to using it.  Memories are just a huge fucking "whatever, I guess". 

Synthtissue should probably just be limited majorly since its just "powercreep, the fermichem". But thats not here, not there, and not over yonder. The borrowed health is a nice touch, I guess. No thoughts on that or anything else that was tweaked — that's a complete and utter "whatever". 🙂

This is pretty heavy handed. All in all, though.

DeltaFire15:
> Yeah, to be quite honest, this makes using strange reagent and synthflesh entirely purposeless.
> 
> Organ damage is half the reason anyone uses strange reagent over defibs or the cloner. Coupled with toxins damage rapidly hitting someone as soon as they're revived due to rotten organs, and you have a chem who can't reliably be used outside of xenobio tomfuckery until you have experimental surgery to remove toxins and heal organs - except, by that point, you have revival surgery that does the exact same thing but with an organ heal.
> 
> It needs touched at, yeah. But this is just way too far with it.

Strange reagent still has a stronger heart heal than revival, which combined with the fact the heart usually is the first organ to reach acute failure before any others, is fairly significant, especially considering it still allows revival even if the heart is missing or failing afterwards (though obviously with the caveats reviving somebody without a heart brings with itself.)
 This quality is combined with blood restoration from levels which would make defibbing or revival surgery impossible or more difficult and the fact that strange reagent is a chem acquirable very close to roundstart whereas revival requires a bunch of med branch research, plus it too skipping the defib time threshold which can sometimes indeed still matter.
Additionally, considering organ rotting takes around the same five minutes defib memory loss requires, it also is also on the same level disadvantage-wise when used on somebody dead for long enough to have major organ damage.

DeltaFire15:
Also the synthtissue revival / dead heal drawbacks are actually **far less** brutal than what they would have been if it worked in the first place when it was added. It now scales with damage healed as opposed to potency directly, loses borrowed health faster with the same damage per tick, and cannot lead to instant redeaths with more damage than before if splashed.
(Though I'd probably still be up to change the damage split to limit clone to a certain percentile total so people wouldn't have to be around med for too long with clonedam and could just purely use toxchems instead after the initial clonedamage wave wears off.)

Linzolle (approved):

DeltaFire15:
Need to think about some tweaks to the synthtissue damage / borrow logic due to pretty high punishment in most use cases despite already kneecapping the calculation. Do not fullmerge till then.

DeltaFire15:
Should now be in a final state. Synthtissue caps its borrowed health at 250 max and slowly goes from pure clonedamage to pure toxdamage over time (e.g. only needs them in medbay for a bit during the clonedamage phase)
Aditionally, strange reagent heals all failed organs for 5 health, plus the heart for 15.

Issue #15211, opened 2021-09-27
Autolathe Serving Tray Missing


TripleShades:
Issue reported from  Round ID: 28502 (Citadel Station)

Reporting client version: 514.1556

The serving tray is no longer printable from the autolathe, and has been replaced by the cafeteria tray entirely.

PR #15214, opened 2021-09-27
[Fix] Grab bag bug fixing

### LINKED ISSUE #15196: Canister Break Sprite Broken
Issue reported from  Round ID: 28447 (Citadel Station)

Reporting client version: 514.1556

Canisters do not properly turn into their broken sprite when they are broken, instead staying the intact sprite.

ghost:
## About The Pull Request

Fixes #15196, #15211 and restores the spray can sound effect

## Why It's Good For The Game

begone bug

## Changelog
:cl:
fix: Canisters now update their sprite when broken
fix: Autolathe now have both Cafeteria and Serving tray
fix: Spray cans now make the spray sound again
/:cl:


code/modules/atmospherics/machinery/portable/canister.dm
```diff
@@ -235,7 +235,7 @@
 		. += "can-open"
 	if(connected_port)
 		. += "can-connector"
-	var/pressure = air_contents.return_pressure()
+	var/pressure = air_contents?.return_pressure()
```
Putnam3145, line 238: this'll still runtime on the next line, since it returns null and you can't compare null to a number (that I know of)

code/modules/atmospherics/machinery/portable/canister.dm
```diff
@@ -235,7 +235,7 @@
...
 	if(pressure >= 40 * ONE_ATMOSPHERE)
```
Putnam3145, line 239:
```suggestion
	if(pressure && pressure >= 40 * ONE_ATMOSPHERE)
```

code/modules/atmospherics/machinery/portable/canister.dm
ghost: from the DM Doc 
>The operators >, <, >=, and <= test if the left-hand expression is greater than, less than, greater than or equal to, and less than or equal to the right-hand expression. These only apply to numerical expressions. Any other type of data (like null or a text string) is treated like 0.

which is funnily the best behavior here since there's "nothing" inside
and it didn't on my side
and it would only runtime if you try spawning the base object anyway, which i did for debugging (probably happens for crafting as well)

Putnam3145: sure, but that's still a runtime that should be fixed

Putnam3145 (approved):

PR #15220, opened 2021-09-29
Clowncar does not ignore airlocks again

LetterN:
## About The Pull Request
Reverts this regression where the clowncar ignores airlocks


DeltaFire15 (changes requested):
Fix up the inconsistant indentation and this should be good.

buffyuwu:
honk

DeltaFire15 (approved):

PR #15222, opened 2021-09-29
save hours of pain for new engineers with this one simple trick!! click here to find out

jack-avery:
## About The Pull Request

APCs now attempt to encapsulate the area for the turf that the player is on rather than the area that the wall's turf is in.

## Why It's Good For The Game

please stop making me build a wall in the middle of the room to place my apc

Summary of additional testing:
- Two APCs can go on the same wall and represent different areas fine.
- No, APCs cannot be placed in space.
- Placing an APC from where a wall used to be will prioritize the APC area to be the area that the wall was in. e.g.: https://cdn.discordapp.com/attachments/488850419301220352/892845810604187678/unknown.png
- This is a thing: https://cdn.discordapp.com/attachments/488850419301220352/892846805237243984/unknown.png
- APCs still abide by normal APC placement rules (valid area, needs an APC)
- This means it's now impossible to place two APCs such that two terminals will be on the same tile.
- Mentioned by someone, no, you can't place APCs using TK this way since it checks first to make sure you're 1 cardinal tile away (see /obj/item/wallframe/proc/try_build)
- No, wizard fireballs don't collide with APCs placed this way...
- **A result of this is that APCs are _technically_ placed on the floor instead, similar to Box's Central Hall APC**, and you can interact with them from a tile further but literally who cares (https://cdn.discordapp.com/attachments/658596233576841217/892849833524727828/unknown.png uhh)

## Changelog
:cl:
tweak: apc frame placement turf priority
/:cl:

### DIFF
--- a/code/game/objects/items/apc_frame.dm
+++ b/code/game/objects/items/apc_frame.dm
@@ -95,7 +95,7 @@
 /obj/item/wallframe/apc/try_build(turf/on_wall, user)
 	if(!..())
 		return
-	var/turf/T = get_turf(on_wall) //the user is not where it needs to be.
+	var/turf/T = get_turf(user)
 	var/area/A = get_area(T)
 	if(A.get_apc())
 		to_chat(user, "<span class='warning'>This area already has an APC!</span>")


Solaris-Shade:
I am all for anything that will enable players IN GAME to fix mapping issues.
I wounder in what way this could be abused if any.

Plus one in my book.

jack-avery:
> I wounder in what way this could be abused if any.

I tested placing APCs in multiple different areas with different states (space, rooms with APCs, etc.), everything worked as I'd expect it to

DeltaFire15:
I don't think this'd always work as intended since it only touches the part checking if the turf it's placed on is viable, not the actual placing itself. Gonna have to look into it some more somewhen.

jack-avery:
> I don't think this'd always work as intended since it only touches the part checking if the turf it's placed on is viable, not the actual placing itself. Gonna have to look into it some more somewhen.

It still runs the first /obj/item/wallframe/proc/try_build(turf/on_wall, mob/user) as I got "you can't place this here" strings that came from that proc, which runs I assume the initial checks to ensure that the spot itself is viable. /obj/item/wallframe/apc/try_build(turf/on_wall, user) is the additional proc ran for the APC frame that checks APC placement viability (does this area have an APC / does it need one / is this a valid area). I will do some more testing and get back to you on this.

Putnam3145:
APC building is honestly rare enough that this seems "fine" to me even if it's "more jank"

PR #15232, opened 2021-10-02
Removes minesweeper

Putnam3145:
## About The Pull Request

As title, removes minesweeper.

## Why It's Good For The Game

1. Minesweeper is badly coded. It works by iterating through every single tile on the board every single time you click anything, even if you didn't click any part of the board. If the board is large, this used to cause server lag until I added CHECK_TICKs to it, and now instead it's merely horribly unresponsive. This problem is not really tractable.
2. Minesweeper is badly designed. It gives two toys instead of 1 despite being the easiest arcade game by far. It even used to allow you to play the same board twice until I removed that. The fact that it's a well-known game means it's going to be the most attractive to anyone trying to get toys anyway.
3. Emagged minesweeper is the strongest bomb in the game. Each individual mine is a (0,2,3) explosion. Each *mine*. The more mines, the more explosions. A failed emagged minesweeper ends a round guaranteed, for the cost of one (1) emag charge. I don't know who thought this was a good idea.

Problems 2 and 3 are tractable. Problem 1 is not. The advantages do not outweigh the disadvantages; minesweeper is unsalvageable in its current state. 

## Changelog
:cl:
del: Removed minesweeper
/:cl:


Tupinambis:
based

PR #15254, opened 2021-10-08
Dynamic extended: antags allowed, but only later

Putnam3145:
## About The Pull Request

As of right now, voting for "extended" in the dynamic vote prevents dynamic roundstart/latejoin rolls while keeping midrounds. That's it. I considered this good enough for the first go.

This replaces that with something slightly more in-depth. What voting "extended" does now is:

1. Threat level is on average *much* lower.
2. All threat level is allocated to the midround pool, so no roundstarts can roll.
3. No high-impact midround antags can roll (this is just revs and nukies, for now)
4. The first midround/latejoin rolls are delayed, taking twice as long as normal.

## Why It's Good For The Game

Basically, my vision for extended is "the round will definitely last longer, but not be boring". Ideally there will be more ways to ask people if they're bored and have dynamic spice things up, but this is just the beginning.

## Changelog
:cl:
tweak: Dynamic extended rework
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -326,7 +326,7 @@ SUBSYSTEM_DEF(vote)
 					return message_admins("A vote has tried to change the gamemode, but the game has already started. Aborting.")
 				GLOB.master_mode = "dynamic"
 				if(. == "extended")
-					GLOB.dynamic_forced_extended = TRUE
+					GLOB.dynamic_extended = TRUE
 				message_admins("The gamemode has been voted for, and has been changed to: [GLOB.master_mode]")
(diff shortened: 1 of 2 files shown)


TripleShades:
So, hey, apparently these changes make it impossible for admins to tell if extended has rolled, since it used to turn Force Extended to On instead of Off, and Game Info only says Dynamic.

Please maybe... do something about that?

Putnam3145:
> 
> 
> So, hey, apparently these changes make it impossible for admins to tell if extended has rolled, since it used to turn Force Extended to On instead of Off, and Game Info only says Dynamic.
> 
> Please maybe... do something about that?

done, it's a good thing people ping me about this instead of going through intermediaries or whatever's happening here

PR #15257, opened 2021-10-08
Reverts an undocumented change to pirate payoff minimum cost from the dynamic PR

DeltaFire15:
## About The Pull Request
The dynamic PR had a change modifying pirate minimum payoff to 10k instead of 1k, which is a value that is primarily only relevant when cargo is not manned (even 15000 in cargo's budget would be over 10k with the default calculation)
As cargo is not always manned and the popcap of this event is comperatively low (10 legacy modes, 18 dynamic), both Silicons and I agreed this change would not have passed if PRd seperately.

## Why It's Good For The Game
Undocumented changes that have no maintainer okay bad.

## Changelog
:cl:
balance: Reverts an undocumented change to pirate minimum payoff (10k back to 1k)
/:cl:

### DIFF
--- a/code/modules/events/pirates.dm
+++ b/code/modules/events/pirates.dm
@@ -24,7 +24,7 @@
 	var/pirate_type = PIRATES_ROGUES //pick(PIRATES_ROGUES, PIRATES_SILVERSCALES, PIRATES_DUTCHMAN)
 	var/datum/comm_message/threat_msg = new
 	var/payoff = 0
-	var/payoff_min = 10000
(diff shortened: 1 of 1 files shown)


keronshb:
I plan on doing another PR to revert it back to 10k or even 5k if this goes through so it is documented.  Yes an undocumented change is bad, but it's too piss easy to pay off the pirates with a 1k payment.  Way too easy.

Pirates also isn't 10 pop anymore.  In line with other midrounds it's at 24 people.

PR #15260, opened 2021-10-10
Putting Trash and Fire to a box

MrJWhit:
## About The Pull Request

Adds two things throughout boxstation:
-Fire cabinets in public locations.
-Adds trashpiles to maint.

## Why It's Good For The Game

Box has a horrible lack of fire-saftey closets in it, and this should help alleviate it. It does reduce some o2 closets, but, only when it's two o2 closets right next to each other.
And I think trashpiles are an interesting thing to try out.

## Changelog
:cl:
add: Adds more fire-saftey closets and trashpiles to boxstation
/:cl:


MrJWhit:
Also I moved an APC that was annoying me and fixing a grate in atmos.

PR #15267, opened 2021-10-13
Removes hardcoded dynamic

SandPoot:
## About The Pull Request
Title, if you must, a config for the server operator is available.

## Why It's Good For The Game
Not only dynamic please.

## Changelog
:cl:
fix: If a gamemode fails to load too many times, a safe gamemode will be loaded instead.
code: Dynamic is no longer hardcoded, and you can now set a gamemode to be forced in the configs.
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/server.dm
+++ b/code/controllers/configuration/entries/server.dm
@@ -88,3 +88,6 @@
 /datum/config_entry/number/rounds_until_hard_restart
 	config_entry_value = -1
 	min_val = 0
+
+/datum/config_entry/string/force_gamemode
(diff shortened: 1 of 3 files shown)


Iamgoofball:
the point of dynamic is to replace secret, no

SandPoot:
Yes

keronshb:
> 
> 
> the point of dynamic is to replace secret, no

It is and that's what I was told.  But then 4 other people popped up after the entire two week dynamic test and said that wasn't the case, when they all had the time to review it while it was up.

PR #15274, opened 2021-10-16
Removes Dynamic's ability to hijack loneop

keronshb:
## About The Pull Request

Removes Dynamic's ability to hijack loneop.

## Why It's Good For The Game

Lone Ops are already exceptionally rare. Dynamic probably shouldn't be hijacking it away.

## Changelog
:cl:
 tweak: Removes Dynamic's ability to hijack loneop.
/:cl:


Putnam3145:
it's actually good because lone op happens specifically when the disk is kept still for a while and dynamic hijacking it means that the literal one (1) thing it's there for doesn't happen but yeah it is good

PR #15290, opened 2021-10-20
[Ready] Bone Anvil and processable Bone ingots

WanderingFox95:
## About The Pull Request

Requested by @dzahlus and "the gang" which I think includes @keronshb and frankly a good next step for adding to the tribal crafting PR. You can now make an anvil out of bones (which has the same stats as the one cargo can order but is harder to make). You can also turn bones into ingots now, because why mine when you can bone?... okay, bad joke. But still. They're slightly more powerful than titanium, but not as good as adamantium.

## Why It's Good For The Game

It's only logical to extend the uses for bones, sinews and goliath hide further, right?

## Changelog
:cl:
add: bone anvils and bone ingots
imageadd: bone anvil sprites
/:cl:


code/modules/smithing/smithed_items.dm
```diff
@@ -49,6 +49,9 @@
 /obj/item/ingot/iron
 	custom_materials = list(/datum/material/iron=12000)
 
+/obj/item/ingot/bone
+	custom_materials = list(/datum/material/bone=12000)
```
DeltaFire15: Crafting recipe needs 2 bones, material value is effectively 6 bones. Change it to 4000 to avoid material duplication.

code/datums/materials/basemats.dm
```diff
@@ -362,6 +362,7 @@ Unless you know what you're doing, only use the first three numbers. They're in
 	color = "#e3dac9"
 	categories = list(MAT_CATEGORY_RIGID = TRUE, MAT_CATEGORY_BASE_RECIPES = TRUE)
 	sheet_type = /obj/item/stack/sheet/bone
+	strength_modifier = 1.2
```
DeltaFire15: I have no idea how you alloy bones, or *smelt* them once that is somehow done, but if we look past however that works, it should probably be at base damage tier or only slightly above, definitely not over titanium.

code/datums/components/crafting/recipes/recipes_primal.dm
```diff
@@ -295,6 +295,16 @@
...
 	category = CAT_PRIMAL
 
+/datum/crafting_recipe/bone_anvil
+	name = "Bone Anvil"
+	result = /obj/structure/anvil/obtainable/bone
+	time = 200
+	reqs = list(/obj/item/stack/sheet/bone = 4,
+				 /obj/item/stack/sheet/sinew = 2,
+				 /obj/item/stack/sheet/animalhide/goliath_hide = 1)
+	tools = list(/obj/item/weldingtool/experimental/ashwalker, /obj/item/crowbar/ashwalker)
```
DeltaFire15: Kinda weird to have one of the strongest normally obtainable anvils be available for killing like four lavaland enemies total when the alternative is a pretty hefty cargo price. Then again, not strictly against it.

DeltaFire15 (changes requested):

code/datums/components/crafting/recipes/recipes_primal.dm
WanderingFox95: Let's see... In total we have... 2 bones, 2 sinews, 1 regen core for the welding tool, 2 bones, 1 sinew for the crowbar, and then 4 bones, 2 sinews and 1 goliath hide for the anvil. That's 8 bones, 5 sinews, 1 core, 1 hide, 
3 Watchers, 3 Goliaths, 1 Legion. I can up it a little more though, that's not much of an issue.

code/datums/materials/basemats.dm
WanderingFox95: You don't need to fight to get titanium so it feels like a downgrade to fight for the bones, isn't that kind of the opposite of the idea here?
The original idea was making it the opposite of gold, but I did not find a speed stat for gold and couldn't give one to the bones.

code/modules/smithing/smithed_items.dm
WanderingFox95: This one's fair.

code/datums/components/crafting/recipes/recipes_primal.dm
```diff
@@ -295,6 +295,16 @@
...
 	category = CAT_PRIMAL
 
+/datum/crafting_recipe/bone_anvil
+	name = "Bone Anvil"
+	result = /obj/structure/anvil/obtainable/bone
+	time = 200
+	reqs = list(/obj/item/stack/sheet/bone = 4,
+				 /obj/item/stack/sheet/sinew = 2,
+				 /obj/item/stack/sheet/animalhide/goliath_hide = 2)
+	tools = list(/obj/item/weldingtool/experimental/ashwalker, /obj/item/clothing/gloves/tackler/combat/goliath)
```
DeltaFire15: Needing tackler gloves as tool for it just seems weird, feel free to return it to the crowbar or some other simillar tool (maybe just slightly more bone instead?)

DeltaFire15 (changes requested):
All good otherwise imo now.

code/datums/components/crafting/recipes/recipes_primal.dm
WanderingFox95: Added the cutters for the sinew-working, guess that makes more sense. And two more bones.

code/modules/mining/equipment/kinetic_crusher.dm
```diff
@@ -235,9 +235,6 @@
 	icon_state = "crusher-bone"
 	item_state = "crusher0-bone"
 
-/obj/item/kinetic_crusher/glaive/bone/update_icon_state()
```
DeltaFire15: The fix PR that also did this got merged , might want to readd this so you don't accidentally revert it.

DeltaFire15 (approved):

PR #15293, opened 2021-10-21
Makes bodies laying on the ground also act as obstacles to crawling.

Higgin:
## About The Pull Request

Applies the same do_after that crawling under standing people gets to crawling over prone people/bodies. 

## Why It's Good For The Game

The sprite-jumbling scramble that goes on when people are crawling around under each other in the middle of a ground-fight isn't good. It's bad, even. This is a bit more of a reason to want to get back on your feet in a fight because your options suddenly become a lot more limited if you're on the ground in a crowd with other people who've gone down. 

## Changelog
:cl:
tweak: Applies the same delay from crawling under standing people to crawling over prone people.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -116,19 +116,19 @@
 						to_chat(src, "<span class='warning'>[L] is restraining [P], you cannot push past.</span>")
 					return 1
 
-	//CIT CHANGES START HERE - makes it so resting stops you from moving through standing folks without a short delay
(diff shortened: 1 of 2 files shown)


keronshb:
I don't think this is a good balance change.

KrabSpider:
> I don't think this is a good balance change

What could possibly be a better alternative?

keronshb:
> 
> 
> > I don't think this is a good balance change
> 
> What could possibly be a better alternative?

Just not having it in.

PR #15306, opened 2021-10-26
Adjusts Dynamic requirements to be more lienent (less 24/7 traitor and maybe a chance for something else holy shit)

keronshb:
## About The Pull Request

This adjusts the requirements for everything to be considered to spawn for the most part.

## Why It's Good For The Game

Right now we're at 24/7 traitor with a few midrounds for about 99% of the rounds.

The current requirements for most of everything are WAY TOO HIGH considering that the average is suppose to be 50 threat.  And with traitors only requiring 10 across the board, they eat up most of it roundstart.

No I didn't do this in the config because I don't have access to that.  I don't have any idea if these were adjusted already or not but these NEED to be adjusted.

## Changelog
:cl:
tweak: A bunch of requirement adjustments
balance: removes required enemies
balance: Lowers assassination threat threshold 
/:cl:

### DIFF
--- a/code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
+++ b/code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
@@ -63,7 +63,7 @@
 	required_candidates = 1
 	weight = 7
 	cost = 5
-	requirements = list(40,30,20,10,10,10,10,10,10,10)
+	requirements = list(101,40,25,20,15,10,10,10,10,10)
 	repeatable = TRUE
 
 //////////////////////////////////////////////
@@ -159,5 +159,5 @@
 	required_candidates = 1
 	weight = 4
 	cost = 10
-	requirements = list(101,101,101,50,40,10,10,10,10,10)
+	requirements = list(101,101,60,50,40,20,20,15,10,10)
 	repeatable = TRUE
--- a/code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
+++ b/code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
@@ -159,5 +159,5 @@
 	required_candidates = 1
 	weight = 4
 	cost = 10
-	requirements = list(101,101,60,50,40,20,20,15,10,10)
+	requirements = list(101,101,80,50,40,20,20,15,10,10)
 	repeatable = TRUE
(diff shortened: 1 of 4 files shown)


code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
```diff
@@ -159,5 +159,5 @@
 	required_candidates = 1
 	weight = 4
 	cost = 10
-	requirements = list(101,101,101,50,40,10,10,10,10,10)
+	requirements = list(101,101,60,50,40,20,20,15,10,10)
```
DeltaFire15: Not on board with 12pop heretics

code/game/gamemodes/dynamic/dynamic_rulesets_roundstart.dm
```diff
@@ -138,7 +138,7 @@
 	weight = 3
 	cost = 15
 	scaling_cost = 9
-	requirements = list(101,101,101,55,40,25,20,15,10,10)//higher because of 'round end'
+	requirements = list(101,101,60,50,40,20,20,15,10,10)//higher because of 'round end'
```
DeltaFire15: Same here

code/game/gamemodes/dynamic/dynamic_rulesets_roundstart.dm
```diff
@@ -179,7 +179,7 @@
 	required_candidates = 1
 	weight = 3
 	cost = 20
-	requirements = list(101,101,100,80,50,40,30,20,10,10)//100 because of configt, otherwise equal to nukies
+	requirements = list(101,101,80,60,40,20,20,20,10,10)//100 because of configt, otherwise equal to nukies
```
DeltaFire15: Higher possibility for wizard on 12pop instead of it being once in a blood moon on that pop? Big ehh.

keronshb: 100 is extremely unlikely with how things are currently with the dynamic system.  80 is even extremely unlikely.

code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
keronshb: I could put it to 70 or 80 if needed but around 15 pop was the original heretic poplock:

image

keronshb: I put it to 80

code/game/gamemodes/dynamic/dynamic_rulesets_roundstart.dm
keronshb: I put it to 80

Putnam3145:
None of this is going to actually have an effect because the config overrides most of it

PR #15307, opened 2021-10-27
fixes "Makes bodies laying on the ground also act as obstacles to crawling."

silicons:


DeltaFire15:
this one doesn't work because people laying down have density 0. I'm tming mine for now instead, seemed to work.

PR #15244, opened 2021-10-05
TGUI alerts, lists, inputs

SandPoot:
## About The Pull Request
Title, includes all the cbt.
image
image

## Why It's Good For The Game
Does it?

## Changelog
:cl:
tweak: Replaced alerts, input lists, inputs and so on.
/:cl:

PR #15309, opened 2021-10-28
Revert "TGUI alerts, lists, inputs"

LetterN:
Reverts Citadel-Station-13/Citadel-Station-13#15244

this:
 - does not exist on tg
 - is fucking slow (input delay when doing t/any input box is horrible)
 - does not autofocus on the input box
 - does not even have fallbacks for when tgui fails

this is why you do not tgui everything.


silicons:
Well
I didn't want to do this really but this just is causing too many issues.
Sorry @SandPoot.
I'd recommend just PRing in the helpers/whatever is needed but for the most part it's best if we dont tgui literally everything.

PR #15310, opened 2021-10-28
No Frollo you can't burn Jacq in the cremator with hellfire (plus two other grief prevention additions)

keronshb:
## About The Pull Request

Prevents the use of the cremator on Jacq.  She'll simply teleport away.

inb4 salt PR, got it approved:

ss

As per @LetterN I also added a station loving component.  No taking her off station now, the Ash Lizards do not deserve halloween.

I also made Bartholomew indestructible.  Why would you hurt him??

## Why It's Good For The Game

This is an oversight that she can die this way.

## Changelog
:cl:
balance: Jacq can't burn in the cremator anymore
balance: Jacq also can't be cheesed off station
balance: Barth also cannot be destroyed
/:cl:


silicons:
still a salt pr but atleast it's a good salt pr
and now i go back to handling the tgui drama :/

BlueWildrose:
I've also seen instances of Bartholomew, the cauldron cat getting destroyed with acid, if you want to prevent that as well. Mark the cat as indestructible.

code/modules/holiday/halloween/bartholomew.dm
```diff
@@ -8,6 +8,7 @@
 	anchored = TRUE
 	var/items_list = list()
 	speech_span = "spooky"
+	resistance_flags = INDESTRUCTIBLE //Barth is immortal you know
 	var/active = TRUE
 
 /obj/item/barthpot/Destroy()
```
DeltaFire15: Please also delete this part while at it (the Destroy() override), it can cause a recursive garbage del loop and should just be prevented from being destroyed by most stuff instead (Like you effectively did).. or just replace the proc's contests with only a "return QDEL_HINT_LETMELIVE", which bypasses some garbage stuff at least.

keronshb: I'm not following what you mean here.

code/game/objects/structures/morgue.dm
```diff
@@ -247,7 +247,11 @@ GLOBAL_LIST_EMPTY(crematoriums)
 
 		locked = TRUE
 		update_icon()
-
+		for(var/mob/living/simple_animal/jacq/J in conts)
+			visible_message("<b>[src]</b> cackles, <span class='spooky'>\"You'll nae get rid a me that easily!\"</span>")
+			playsound(loc, 'sound/spookoween/ahaha.ogg', 100, 0.25)
+			J.poof()
+			return
```
DeltaFire15: Unlock the cremator when poofing to prevent it bricking.
```suggestion
		for(var/mob/living/simple_animal/jacq/J in conts)
			visible_message("<b>[src]</b> cackles, <span class='spooky'>\"You'll nae get rid a me that easily!\"</span>")
			playsound(loc, 'sound/spookoween/ahaha.ogg', 100, 0.25)
			J.poof()
			locked = FALSE
			update_icon()
			return
```

keronshb: Done

PR #15313, opened 2021-10-28
Fixes dragging things over prone individuals

DeltaFire15:
## About The Pull Request
Alternate PR to #15311 
I don't have an opinion on the feature in itself, but having it be not broken may be useful for gauging the actual interest in it.
Cause of the problems was a var that wasn't typechecked on use and caused it to runtime. This fixes that.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: You can now drag things over prone people again.
/:cl:


shellspeed1:
Ya know, I'm kinda starting to like this. Makes crawling over a bunch of dead bodies feel like a slog through hell. Would be even better if we could use bodies for cover.

PR #15335, opened 2021-11-04
Synth / IPC kinda-buffs - Chemwarfare Edition

DeltaFire15:
## About The Pull Request
This PR adds two new flags to the possible chemical_flags of reagents, REAGENT_ORGANIC_PROCESS and REAGENT_ROBOTIC_PROCESS, which allow a reagent to be processed by an organic- or robotic carbon, respectively. Those with both can be processed by either. If a reagent is incompatible, it will slowly dissipate at a rate of currently 1u / chem tick instead of being processed. This nonetheless does not override reaction_mob or simillar things, since most of those are usually some physical impact related to a chem if present.
~~(Could probably use a chem-master / chemscan info tidbit on who can process a scanned chem)~~
Chem analyzers and ChemMaster machines provide information about who can process a reagent.
Additionally, health analyzers using reagent scan mode (and other things making use of the same proc) will now show reagents in grey if they cannot be processed by the individual they are in.

Now, onto the use for them. Robotpeople, currently, whilst not positively affected by most chems and immune to direct normal toxdamage or drunkness from them, would still suffer side-effects from them. This changes most chemicals to not be processed at all, save for some still needed or logical ones, making robots 99%* immune to chemical warfare or simillar.
This should not influence disease curing or RDS, as those simply check chems in someone, which these still are, though they are generally purged faster.

Chemicals still processed are:
Fermichems - Funky chems which imo they shouldn't be locked out of
Oxygen heal chems - oxydamage is still relevant and it's the only way to deal with it
Cryo tube chems - base chem for wound healing, still no brute / burn heal obviously.
Mutation-related chems (Mutadone / Mutagen)
Mutation toxins (Minus ADV mutation toxin due to their disease immunity)
Clonedamage healing chems - Since they can still suffer that damage type
a few misc healing chems e.g. dizzyness / hallucination / temp ones
Radiation healing and causing chems
Crit-related chems (Epi, Atropine, etc)
Reagent purging chems (Calomel / Charcoal, Pentetic)
Acid - obviously
Most pyrotechnic reagents - Napalm, Thermite, etc; for obvious reasons
Wound-related chems (e.g. Determination and wound clotting chems)
Adminordrazine because adminbuse
Antag-related chems (changeling meth, syndi nanites, clockcult / bloodcult water, holy water, bz metabolites and simillar)
Brain-related chems (Neurine / Mannitol), would be a bit wack without
Other Organ / Surgery / Preservation chems
Robot-intended chems (System Cleaner)
Blood Recovery chems (including blood itself)
Vaccine because it's a viro core reagent
Strange Reagent because it's frankly magic

This is a pretty long list of things they still get affected by, but trust me this is not a lot in the whole scale of things. Most of these are chems from the med chem category that have no or only hardly a non-chem alternative and would be too inhibitive if blocked.

Categories that have basically every chem blocked are:
Drinks / Food
Toxins
Brute- / Burn- / Tox related chems

If you feel like they should or should not have access to a specific chem, or think I missed something, do tell. They could probably live without some of the chems still left in, though some are hardly replacable with non-chem alternatives.

Oh this also fixes System Cleaner not forcing a update, which could lead to some visual inconsistancies, plus makes it somewhat stronger due to it working really slow.

## Why It's Good For The Game
Synthetics as of now chemical-wise have a weird state, where they do not benefit from a lot of chems, but still get affected by a lot of side effects of negative (or even usually positive) ones, which is not the intended state.
This makes them far more resistant to most chemicals, whilst (for now) leaving in the greater part of still-beneficial medical chems as medbay is very chem-focussed and fully locking them out would be rather big of a disadvantage.

## Changelog
:cl:
balance: Synthetics gained quasi-immunity to most chemicals, excluding some core medical chems and a few specific disruptive ones.
add: Health analyzer, ChemMaster and chemical analyzer interactions with the chemical processing flag system.
fix: System Cleaner should now always update health as opposed to before.
balance: system cleaner is now twice as effective.
/:cl:


ghost:
Would romerol fall under the organ chems?

DeltaFire15:
> Would romerol fall under the organ chems?

Romerol indeed still works, yup. Would've been too impactful not to, since it's so high of a tc cost too (and for the same reason why synths / IPC can still get facehugged as opposed to people who get themselves full-metal torsos)

lolman360:
they should be immune to cloneloss too (out of scope ik but)
i dont get why epi/atropine works on them when all other general healing chems dont

TheEaterOfBeans:
> they should be immune to cloneloss too (out of scope ik but) i dont get why epi/atropine works on them when all other general healing chems dont

They should just NOT be cloneable, maybe give robotics a cloner equivalent for robots and only robots but the organic cloner should be useless for them.

DeltaFire15:
> they should be immune to cloneloss too (out of scope ik but) i dont get why epi/atropine works on them when all other general healing chems dont

They can *process* epi / atropine, yes, though the effects on them are limited to their core of being crit-stabilization chems. They obviously still won't heal any brute / burn unless somebody taped a organic limb to them or something

Cloneloss removal would require revisiting a bunch of stuff and would indeed be out of scope here, though in general something that would be good to do.

DeltaFire15:
> They should just NOT be cloneable, maybe give robotics a cloner equivalent for robots and only robots but the organic cloner should be useless for them.

Out of scope, though in general considered. Would need to be a well thought out replacement system to avoid reducing their capabilities in dangerous jobs more.

DeltaFire15:
Now tested and it works:tm:

Additionally, some QoL features added, namely:
- ChemMaster reagent info and chemical analyzers will show who can process a scanned reagent
- Health analyzers in chemscan mode now display invalid chemicals in someone in grey

These tweaks should make it much easier to deal with this when not aware of the exact chems which are invalid (and hey feel free to test around, after all it's not like a not processed chem will harm them, I'm not that evil.)

Should be ready for testmerging.

code/__DEFINES/reagents.dm
```diff
@@ -72,6 +72,11 @@
 #define REAGENT_FORCEONNEW			(1<<5)  //Forces a on_new() call without a data overhead
 #define REAGENT_SNEAKYNAME          (1<<6)  //When inverted, the inverted chem uses the name of the original chem
 #define REAGENT_SPLITRETAINVOL      (1<<7)  //Retains initial volume of chem when splitting
+#define REAGENT_ORGANIC_PROCESS		(1<<8)	//Can be processed by organic carbons - will otherwise slowly dissipate
+#define REAGENT_ROBOTIC_PROCESS		(1<<9)	//Can be processed by robotic carbons - will otherwise slowly dissipate
+
```
Putnam3145, line 77:
```suggestion
#define REAGENT_ALL_PROCESS (REAGENT_ORGANIC_PROCESS | REAGENT_ROBOTIC_PROCESS)
```
there's enough uses of both that this might be reasonable

DeltaFire15: fair

PR #15338, opened 2021-11-05
Adds Clock Cult to the midround pool.

keronshb:
## About The Pull Request

Adds the ability for dynamic to run Clock Cult midround and not just round start.

Note this does **NOT ASK GHOSTS**, it picks people **MIDROUND** to become Clock Cultists.  If you really want a ghost role for this, wait for Marauder/Stand.

I would like to roll this as a test to see if this is a good feature to add for some other round types.

Big thanks to @DeltaFire15 for help fixing some bugs.  I didn't even really write this myself, it was just copying stuff to make it work.

## Why It's Good For The Game

This is going to be a long laundry list so please bear with me.

1) Citadel has a huge issue of latejoiners and exceptionally low round start pop.  Most of the other game modes (Wizard, Nukies, Revs, both Cults) do not have a hell of a chance of rolling round start with how things are currently.  However with dynamic, Wizard and Nukies do have a chance to roll it midround, but they're a bit rare.  Adding Clock Cult to the mix should expand the midround pool a bit more and the variety of rounds, assuming it gets a chance to roll.  So it addresses both the latejoin issues and the population issues.

2) Since this is MIDROUND ONLY, it means people that are observing cannot fish for this or latejoin to get it.  The only way to get this is to be active and in round (and on station).  Don't worry it already accounts for people who are already antag so it won't double dip anyone.  We don't have much of an observer issue anymore, but this should address that issue slightly as it requires you to be IN the round to roll.

3) Since it picks from the midround pool it also has to compete against all other midrounds + latejoins as well.  So while a midround cult could seem overwhelming, we already have midround wiz/nukies AND it has to compete with all the other midrounds as well.  If this test works out well I would want to roll it out for some other modes.

## Changelog
:cl:
add: Lets dynamic pick clock cultists midround
/:cl:


code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
```diff
@@ -382,6 +382,65 @@
...
+
+/datum/dynamic_ruleset/midround/ratvar_awakening/trim_candidates()
+	..()
+	candidates = living_players
+	for(var/mob/living/player as anything in candidates)
+		var/turf/player_turf = get_turf(player)
+		if(!player_turf || !is_station_level(player_turf.z))
+			candidates -= player //no ghost roles
+			continue
+
```
DeltaFire15, line 424: also add a check here for
if(!is_eligible_servant(player))
    candidates -= player
    continue

keronshb: Done

TheObserver-sys:
Only thing I wonder is how we're gonna handle Reebe when doing this. Could probably give the game away if it loads Reebe in the middle of gameplay, since that'll likely cause a lagspike midgame, but if we always preload it, it will give us a nice performance hit in the face.

keronshb:
> 
> 
> Only thing I wonder is how we're gonna handle Reebe when doing this. Could probably give the game away if it loads Reebe in the middle of gameplay, since that'll likely cause a lagspike midgame, but if we always preload it, it will give us a nice performance hit in the face.

I haven't figured out Reebe yet.

One option is to have it always on the admin z level.  But for now it'll be disabled for midround.

code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
```diff
@@ -382,6 +382,70 @@
...
+			break
+		var/mob/living/clock_antag = pick_n_take(candidates)
+		assigned += clock_antag.mind
+	for(var/datum/mind/M in assigned) //add them to the clockwork team
+		add_servant_of_ratvar(M.current)
+		SSticker.mode.equip_servant(M.current)
+		SSticker.mode.greet_servant(M.current)
+		message_admins("[ADMIN_LOOKUPFLW(M.current)] was selected by the [name] ruleset and has been made into a midround clock cultist.")
+		log_game("DYNAMIC: [key_name(M.current)] was selected by the [name] ruleset and has been made into a midround clock cultist.")
+	return ..()
```
DeltaFire15, line 447: Move the load_reebe here please to avoid calling it when execute would abort early.
```suggestion
		log_game("DYNAMIC: [key_name(M.current)] was selected by the [name] ruleset and has been made into a midround clock cultist.")
	load_reebe()
	return ..()
```

DeltaFire15 (approved):
Code looks fine, testmerging for safety.

DeltaFire15:
Eh this seems fine enough and pop is not exactly high enough for it to roll in most cases. If there's anything wack it can always be fixed later.

PR #15348, opened 2021-11-09
Makes Face hugger jambayala give more bites.

shellspeed1:
## About The Pull Request
What it says on the can. Makes the food item have more bites.

## Why It's Good For The Game
Its kind of annoying to make, should get more benefit out of it.

## Changelog
:cl:
balance: More bits from face hugger soup
/:cl:


DeltaFire15 (changes requested):
Basically, this doesn't actually give it more bites or nutrition, but instead only makes you eat it "faster". Bitesize is how much of the reagents (nutriments / w/e) gets transferred per bite, with the food counting as "fully eaten" once it has no reagents left. Increasing reagent count (nutriment primarily) would be the proper way to increase value, preferrably via bonus_reagents as those only affect if you craft it vs. getting it from, say, xenobio.

PR #15375, opened 2021-11-15
Removes being able to access departmental budgets via NTIRN

DeltaFire15:
## About The Pull Request
Turns out people have been able to access departmental accounts other than cargo's with this for a good nine months. 
Apparently an intentional change originally despite most players not even knowing this fact, so I'll count this as balance change.
This removes being able to use budget accounts in the NTIRN tablet / laptop program. Heads can still order or approve things, cargo budget only though.

## Why It's Good For The Game
There effectively being more than 130k credits accessible roundstart in budgets, or more than 500k after the round goes on a while due to departmental passive or active income can absolutely mess up multiple gamemodes / rulesets / antags in full, and can cause major issues with anything else. This amount of money completely sidesteps cargo, without any work required at all. Do I even need to list how much wack stuff you can do with 150k roundstart, or, god forbit, those over 500k at the lower end if dynamic dares roll something like nukies or revs midround?
Budgets throw the already fragile cargo balance completely out the window, and should not exist. That this was even allowed in initially is horrendous.

## Changelog
:cl:
balance: NTIRN can no longer access non-cargo departmental budgets.
/:cl:

### DIFF
--- a/code/modules/modular_computers/file_system/programs/budgetordering.dm
+++ b/code/modules/modular_computers/file_system/programs/budgetordering.dm
@@ -70,7 +70,6 @@
 	if(id_card?.registered_account)
 		if(ACCESS_HEADS in id_card.access)
 			requestonly = FALSE
-			buyer = SSeconomy.get_dep_account(id_card.registered_account.account_job.paycheck_department)
 			can_approve_requests = TRUE
 		else
 			requestonly = TRUE
@@ -225,8 +224,7 @@
 				return
 
 			if(!self_paid && ishuman(usr) && !account)
-				var/obj/item/card/id/id_card = card_slot?.GetID()
-				account = SSeconomy.get_dep_account(id_card?.registered_account?.account_job.paycheck_department)
+				account = SSeconomy.get_dep_account(ACCOUNT_CAR)
 
 			var/turf/T = get_turf(src)
 			var/datum/supply_order/SO = new(pack, name, rank, ckey, reason, account)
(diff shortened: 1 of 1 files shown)


shellspeed1:
Ya know, hot take. So long as the budget exists people are going to try and get into them. It might be better to remove the budget accounts altogether.

DeltaFire15:
> Ya know, hot take. So long as the budget exists people are going to try and get into them. It might be better to remove the budget accounts altogether.

I have no idea why they exist. I personally have been against them from the very beginning, but completely excising them from the code would require quite a bunch more work with how deep they are in some economy sections.

lightcat:
Instead of removing anything that could by slightly overpowered why not instead improve it?

Give a set amount for each department at round start and disable passive gain. Give department's ways to make money by completing certain tasks or maybe a random event. Remove the money given from the science bomb.

So like, 10-20k per department wouldn't actually be that bad.

DeltaFire15:
> So like, 10-20k per department wouldn't actually be that bad.

As I've talked about on discord, I have thought about lowering departmental budgets and decoupling salaries, but unfortunately that still runs into the issue of "too little money - may aswell not be accessible, too much - this current state happens"
50k from the departments would already allow for some major purchases bypassing any time cost from cargo moneymaking, and there's reasons for why servers with budget cards usually just basically go "hey lets pile all this money" after all. Especially during things going south.

Putnam3145:
> Instead of removing anything that could by slightly overpowered why not instead improve it?

we can improve them later, for now their being accessible is a bug

PR #15392, opened 2021-11-22
oil drum now recognises synthetic anthormorphs as synths

timothyteakettle:
## About The Pull Request
title

## Why It's Good For The Game
fixes a bug

## Changelog
:cl:
fix: oil drum now recognises synthetic anthormorphs as synths
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/anthropomorph.dm
+++ b/code/modules/mob/living/carbon/human/species_types/anthropomorph.dm
@@ -48,3 +48,4 @@
 	miss_sound = 'sound/weapons/slashmiss.ogg'
 
 	allowed_limb_ids = list("mammal","aquatic","avian", "human")
+	species_category = "robot"


DeltaFire15:
there's also a is_robotic() helper in is_helpers.dm that does not take them into account, maybe add them to that one aswell while you are at it.

Linzolle:
> there's also a is_robotic() helper in is_helpers.dm that does not take them into account, maybe add them to that one aswell while you are at it.

Would need a snowflake proc to do that considering how race drinks work.

DeltaFire15:
> Would need a snowflake proc to do that considering how race drinks work.

[...] || isspecies(arg, /datum/species/mammal/synthetic) is not really snowflakey, merely maintaining correctness.
Is also unrelated to the original fix this PR does, the line this PR changes does fix what it intends to fix.

PR #15395, opened 2021-11-24
silencers currently dont output any text whatsoever and that's wack

buffyuwu:
## About The Pull Request

downstream thing im pulling upstream here https://github.com/FortunaSS13/Fortuna/pull/469
this just bumps back indentation 1 to make the firing text code reachable for silenced and non-silenced weapons

## Why It's Good For The Game

silencers should have output in chat considering the range is like 3 tiles and well within the audio range anyway

currently you see the text of the impact of the silenced bullets but not the actual firing text from the firer, which if its supposed to mask the shooter by not outputting any text why does it.. output impact text? this seems like a mistake that nobody ever fixed

## Changelog
:cl:
fix: silenced guns now output fire text when fired like any other gun
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -163,11 +163,11 @@
 		playsound(user, fire_sound, 10, TRUE, ignore_walls = FALSE, extrarange = SILENCED_SOUND_EXTRARANGE, falloff_distance = 0)
 	else
 		playsound(user, fire_sound, 50, 1)
-		if(message)
-			if(pointblank)
-				user.visible_message("<span class='danger'>[user] fires [src] point blank at [pbtarget]!</span>", null, null, COMBAT_MESSAGE_RANGE)
-			else
-				user.visible_message("<span class='danger'>[user] fires [src]!</span>", null, null, COMBAT_MESSAGE_RANGE)
+	if(message)
+		if(pointblank)
+			user.visible_message("<span class='danger'>[user] fires [src] point blank at [pbtarget]!</span>", null, null, COMBAT_MESSAGE_RANGE)
+		else
+			user.visible_message("<span class='danger'>[user] fires [src]!</span>", null, null, COMBAT_MESSAGE_RANGE)
 
 /obj/item/gun/emp_act(severity)
 	. = ..()


DeltaFire15:
This is intentional. Suppressors have the intent of not giving a firing message.
Impact text does not give any direct hint at the shooter, shooting does.
additionally the hit text gets changed to be small instead of very large and hitsound gets reduced

DeltaFire15:
image
This is even stated in the variable define itself.

buffyuwu:
> image This is even stated in the variable define itself.

this seems a little silly still considering the range is small for this text to proc anyway that the gun doesn't output anything imo

lolman360:
> why does it.. output impact text?

because it's one of the few stimuli that you can respond to
if your health just magically starts dropping with no visible source it's 1. kinda wack for counterplay and 2. absolutely awful since people wont know whats going on (generally there should be feedback whenever something happens)
impact messages allow someone to identify a suppressed weapon is being used but without firing messages a suppressor doesn't immediately oust the shooter like an unsuppressed gun (as firing text has the shooter's name)

buffyuwu:
> > why does it.. output impact text?
> 
> because it's one of the few stimuli that you can respond to if your health just magically starts dropping with no visible source it's 1. kinda wack for counterplay and 2. absolutely awful since people wont know whats going on (generally there should be feedback whenever something happens) impact messages allow someone to identify a suppressed weapon is being used but without firing messages a suppressor doesn't immediately oust the shooter like an unsuppressed gun (as firing text has the shooter's name)

the range being so low to begin with means you definitely know who is shooting you already, so it not outputting that text seems an oversight here. also

> (as firing text has the shooter's name)

if they arent wearing a mask or making any attempt to conceal their identity yes, but that person 3 tiles away definitely knows who is shooting them already since they're... reaaaally close! the jig is up if you're firing within normal combat message range anyway so why dont we have a message text like any other gun is my point

i think the default combat_message_range is 3

silicons:
this should be 1-2 range tops if this is done at all
silencers are already borderline worthless
no reason to nerf them more

DeltaFire15:
Closing this PR as the current behavior is intended & a silencer nerf is not neccessary. 'Guaranteed no message' is quite a lot more certain than 'maybe no message unless they moved a tile closer' and is quite a lot less warning in either situation.

PR #15398, opened 2021-11-27
Fix few instances of bleeding white blood

Linzolle:
## About The Pull Request
No more cum blood.

## Changelog
:cl:
fix: bloodsplatter effects no longer white in certain cases
fix: the blood in the BDM ruin is no longer white
/:cl:


DeltaFire15 (approved):
Surely this is fine.

PR #15407, opened 2021-12-03
Fixes fireman carry dropping the carried target if passing over a prone individual

DeltaFire15:
## About The Pull Request
Title. Another funky bug introduced by making crawling under people take time, though by a different means.
This makes carrying someone add a trait to them that allows them to pass by mobs, as they'll only attempt the movement if the carrier already succeeded.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Fireman carry no longer drops the carried person if passing over a prone individual.
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -222,6 +222,8 @@
 /// forces us to not render our overlays
 #define TRAIT_HUMAN_NO_RENDER			"human_no_render"
 #define TRAIT_TRASHCAN					"trashcan"
+///Used for fireman carry to have mobe not be dropped when passing by a prone individual.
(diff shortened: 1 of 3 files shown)


keronshb:
Can't we just revert the original PR?  It's a weird balance change in the first place and the person that made it kind of came and went.

DeltaFire15:
> Can't we just revert the original PR? It's a weird balance change in the first place and the person that made it kind of came and went.

I think I've said this before, but I have no strong opinion on the change in itself so I'm just fixing stuff that I find out isn't working well. Haven't really heard much from other Maintainers regarding it either.

Linzolle:
I'm not too into the change personally but it doesn't bother me that much, fix man good

PR #15417, opened 2021-12-08
Fixes Linters

DeltaFire15:
## About The Pull Request
Title. Linters realized some should_not_sleep stuff was sleeping and started screaming. This fixes them.
Additionally, it also failed because some flavor objective .txts were empty, so I put a filler objective into each so it doesn't fail, aswell as flavor objectives not exactly currently being used.

## Why It's Good For The Game
Linters of PRs being able to not fail is kinda neat.

## Changelog
:cl:
fix: Linters should no longer scream.
/:cl:


Putnam3145:
speedmerging without 24 hour delay because i'm already annoyed at it

PR #15425, opened 2021-12-09
Nerfs* passive organ healing, plus some minor stuff

DeltaFire15:
## About The Pull Request
So basically, a big while ago Silicons buffed organ healing because "it would take more than 30 minutes to heal"
This assumed the base organ heal rate (aka % of organ damage that gets healed per life tick) is constant. In reality, organ healing can be up to x5 of its default depending on satiety (added by vitamins on process mainly, very quickly so)
This made it so you'd heal from 1 organ hp to full within 3 minutes if satiated... And here I was wondering why organ damage disappeared so quickly.
Given that medical has multiple ways to deal with damaged organs (chemicals, monkeys, organ printer, robotics organs after earlyish research), weakening the passive heal requiring nothing shouldn't be as big of a deal. Or, you could just grab a nice healthy snack to speed it up. (at full satiety, a completely messed up organ close to failure (1hp) will take 9 minutes to heal now. Base assumption is about half satiety or 15 minutes, as the default healing intends, slower at low satiety)

I also noticed two seperate things, the first being that cybernetic organs still healed despite the on_life comment saying they did not, so I fixed that (IPC / Synth organs are not technically "true" Cybernetic organs codewise, merely snowflaked, so that doesn't apply to them)
Additionally, while looking at the satiety stuff I realized Synthetics currently cannot gain satiety due to not processing food, so I moved it to their stomach slowly increasing their satiety provided they are sufficiently charged (5 minutes of sufficient charge to reach max satiety rn, could be tweaked). This only applies to Synthetics that have them, no cheating satiety as organic.

## Why It's Good For The Game
Organ damage that is not total failure being totally irrelevant is kinda bad. This should make it at least slightly more relevant and makes medbay's options slightly more viable to be used on organs that aren't fully rotten as opposed to now.

Cybernetic organs healing damage despite the comment on the proc saying they don't is a fix I'd say.

Synths being able to gain satiety may be useful given there's a bunch of things that hook into it they're still affected by.

## Changelog
:cl:
balance: Base organ healing rate reduced by 66% (It can still be increased to up to 500% of the base value via sufficient satiety)
fix: Cybernetic organs no longer passively heal damage despite the proc stating they didn't.
balance: Synthetics now gain satiety at sufficient charge (~5 minutes to max satiety from 0)
/:cl:

### DIFF
--- a/code/__DEFINES/DNA.dm
+++ b/code/__DEFINES/DNA.dm
@@ -182,7 +182,7 @@
 
 ////organ defines
 #define STANDARD_ORGAN_THRESHOLD 	100
-#define STANDARD_ORGAN_HEALING 		(1/(15 MINUTES / (2 SECONDS)))
+#define STANDARD_ORGAN_HEALING 		(1/(15 MINUTES / (2 SECONDS))) / 3  //Base organ healing can be amped by a factor of up to x5 via satiety. This assumes it to be somewhat in the upper center of positive satiety as base.
 #define STANDARD_ORGAN_DECAY		(1/(15 MINUTES / (2 SECONDS)))		//designed to fail organs when left to decay for ~15 minutes. 2 SECOND is SSmobs tickrate.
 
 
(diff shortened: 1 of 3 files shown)


Arturlang:
Maybe instead make severe organ damage not heal so you have to get medical help

DeltaFire15:
> Maybe instead make severe organ damage not heal so you have to get medical help

Would be an option, though severe damage starts at like ~45 damage out of 100. This PR doesn't really intend to invert it to make organ damage a pain to deal with as opposed to too easy (as some organ damage above the high threshold (45), especially on dead people, is pretty common)

PR #15430, opened 2021-12-14
Synth stuff : Balancing Nightmare Edition (aka, coolant)

DeltaFire15:
## About The Pull Request
This PR does some pretty high-impact changes to Synthetics which were initally outlined in the design doc, some of which are absolutely a matter of balance discussion and will need tweaking.

Firstly, Synths no longer care about low blood levels (which is rebranded "coolant" for them. Still uses synthetic blood, they can still bleed it via wounds, etc.

Secondly, Synths are now kind-of coldblooded and do not *usually* have active temperature regulation.
Their default cooling efficiency is regulated by the environment (cold at high pressure is higher efficiency) aswell as their coolant level. They also passively heat up, though this passive increase is mitigated / usually nullified in a normal environment with default heat / pressure / coolant levels (read: roundstart situations)
*Currently, this can heat them up to now 250 degrees Celsius if they're unable to cool. Could probably be still set higher. (was 200 initially)
If the place they are in is warmer than themselves, they heat up mirroring coldblooded. If it is colder however, cooling efficiency directly influences how much they cool down.
This is not strictly copy of a -2 trait for them, but instead their lungs (which have been called "ipc cooling system" as fluff for a whiile now so I thought it'd be fitting for that change to be there) have an 'active cooling' toggle, which, if the Synths temperature is too far away from room temp, will burn through coolant to restore the temperature. Coolant drain is modified by the current cooling efficiency, e.g. active cooling in a cold room is far cheaper than in a melting fusion room. Additionally, the effective temperature change from it depends on said efficiency aswell, though it is at least 50% even if it would be less.
*If you are wearing a full set of EVA gear (any spacesuit + any spacehelmet) or full miner gear (miner suit + hood), your environment cooling efficiency automatically is set to 100%, regardless if active cooling or not. Your coolant level can nonetheless reduce it. **Note that miner gear only has this effect on lavaland.** (or well, whatever the z designated as mining z happens to be)

Thirdly, Synths now do not breathe. Their lungs, while technically not required, are used in temperature regulation / are their primary cooling system (they grant the active cooling ability) If in crit, they meltdown and suffer burn damage, as opposed to brute damage like non-robotic nobreath species.
*This means they cannot CPR either - snowflake them and allow it? undecided on this.

Additionally, Synths now are immune to low pressure damage, their cold_damage modifier is now 0.5, while their heat_damage modifier is now 1.2
After some thought on the topic, they also gained 125 degrees of cold damage resistance, which means they'll only start taking damage at pretty low temps (e.g. icethrowers, coldfloods, simillar). This emulates their coolant starting to freeze at such low temps and their systems starting to give out.

***Gameplay Effects:***
What you are probably here for, condensed tl;dr of effects:
- Synths can kind-of spacewalk by default, but will melt in space unless they burn through coolant using active cooling or wear an actual suit (Well, even with active cooling it only limits the overheating, you'll still overheat slowly as of now (though you can sustain it for longer)
- Synths don't breathe
- Synths despise hot environments, which will overheat them even more than normal individuals.
- Synths are pretty resistant to cold (125 degrees C / K more than normal people) and enjoy being in properly cooled environments. (cold damage and simillar at ~-138C and below)
- Synths do not directly suffer from low blood levels, but are more likely to overheat the lower they are due to passive heat generation
- Synths generally behave like coldblooded people, but can use active cooling, burning through coolant (blood) to massively boost temperature regulation. Their preferred environment is pretty much the inverse of coldblooded's though (freezing cold as opposed to nice and warm)
- If you are a Synthetic you have a info thing on the right side of your HUD  which shows your coolant volume, aswell as how efficient your cooling is right now (can be examined for more precise info than the 5 efficiency readout states it has)

**TODO:**

- [x] Blacklist tempchange chem processing (no longer needed)
- [x] Blacklist oxyloss chem processing (no longer needed)
- [x] More QOL readouts & access to coolant levels / current cooling efficiency / etc
- [x] Test everything more
- [ ] Balance:tm: stuff
- [x] Blacklist coldblooded from Synths since it'd be 2 free trait points.

## Why It's Good For The Game
More uniqueness for Synths, further following the original design doc Silicons made ages ago.
(https://hackmd.io/5CTM_2QDTq-05y3ORw1FIA)
In total, this tends towards the buff-ish side of things, making them stronger towards environmental dangers, for a burst of time at least as they'll require active cooling for dangerous situations.

## Changelog
:cl:
balance: Synths no longer take direct damage from low blood. They will however suffer from lower cooling efficiency and be in danger of overheating.
balance: The temperature system of synths is now simillar to the coldblooded trait, though depending on environment and coolant level, with access to short bursts of direct control via active cooling.
balance: Synths no longer breathe. If in crit, they meltdown and suffer burn damage over time.
balance: Synths are now immune to low pressure, resistant to low temp, and vulnerable to high temp.
balance: Synths no longer process oxyloss and temperature adjustment chemicals.
balance: Nanogel is now printable at med- / scilathes and mechfabs after researching the Advanced Robotics node. It does need a bit of diamond though.
tweak: All Synthetics now use blood type S
/:cl:


Linzolle:
This is making me ask, why the hell do synthetics not have standardised blood types. Why are IPCs HF, synths S, and synthetic mammals... have normal blood???

DeltaFire15:
> This is making me ask, why the hell do synthetics not have standardised blood types. Why are IPCs HF, synths S, and synthetic mammals... have normal blood???

.. Did whoever made Synthetic Mammals a /mammal/synthetic type *really* leave their blood type normal??
wack.

TripleShades:
White on white is a little bit hard to read.
Would it be possible to make the text a different color, or the backing a different color?
edit: Perhaps color coding to show if you're good, low, or extra low on coolant (correlating to how blood level alert works when scanned with a medical scanner.)
image
There's also currently no indicator that cooling is on except for the point drain.

It'd be nice to have it spit out what percentage you're at by clicking on the readout indicator.
edit: Also instruction on what the arrows below mean, perhaps.

This arrow indicator is also very confusing as to what it's referring to.
Does it mean we're below a threshold? Which threshold?
image
I noticed it changed to this later when equalizing my temp, and the coolant level began to raise, but that isn't intuitive at all, since to the user it appears both mean "oh coolant is regenerating now."
image
image

DeltaFire15:
> White on white is a little bit hard to read. Would it be possible to make the text a different color[...]

Yes, colorcoding is doable (it actually already kinda is though it's not too visible). I'll also modify it to show percentage instead of volume as percentages (0 - 100) usually are more helpful for thought than "0 - 655" or whatever blood volume normal is.

>There's also currently no indicator that cooling is on except for the point drain.

Fair point, I'll take a look at that one

> [...]Also instruction on what the arrows below mean, perhaps.

They represent the current cooling efficiency, which is effectively information on the *current* situation (examining the thing aka shift + clicking gives you exact values) This is relative info, as cooling efficiency is a very variable thing, but generally double-red is low enough for you to passively overheat (maybe single-red too if you are way too insulated). If you have around single-green the current environment & coolant level together are enough to effectively null passive heat increase (non-nulled passive heat increase does not directly mean you'll overheat, just that you'll have to radiate said heat), while double-green is very rare and means your efficiency is great right now (or, were very very hot and entered a normal room as relative temp matters for environmental efficiency)
so, conditionally, very bad - bad - neutral - good - very good, from double-red to double-green

bunny232:
image
Why do i have a feeling this has something to do with this pr. Noticed it a while after draining my blood alot and using saline/iron to restore.

DeltaFire15:
> image Why do i have a feeling this has something to do with this pr. Noticed it a while after draining my blood alot and using saline/iron to restore.

This PR has not touched any reagent's effects on those that are affected by it.
Salglu provides a temporary buffer of blood volume while also increasing regeneration, which can lead to lower blood volume after it metabolizes completely compared to it still being present.

TripleShades:
Yeah, that's just Salglu doing what it's supposed to. You cannot use it as a direct replacement for blood, it eventually dissipates, so you need to find something to add to your total blood. It's so you have some wiggle-room time to not die/pass out from blood loss. Iron only does so much so fast, so you'd need actual blood transfusion, or a bloody mary.

TheEaterOfBeans:
> image Why do i have a feeling this has something to do with this pr. Noticed it a while after draining my blood alot and using saline/iron to restore.

tfw caught grabbing the blue synth

PR #15436, opened 2021-12-20
Removes tesla/singulo

Putnam3145:
## About The Pull Request

I like to have *fun* when I play the game. Doing the literal exact same thing to set up power every time where any tiny deviation results in a total, round-ending failure that takes five times as long as it needs to for no reason is not fun. Joining as engineer and finding that my future is "do the exact same thing you've done a billion times before with no variation nor even any *chance* of variation" is not fun. These engines are not fun.

## Why It's Good For The Game

Because these engines are not fun, no matter how much people on reddit try to convince you.

(this is mostly a joke i'm just mad we have the tesla in rotation and it gave it to me twice in a fucking row)

## Alternatives you've considered
Adding a 40% chance to remove xenobio on Box and replace it with a nanite lab, plus a 30% chance to remove genetics and replace it with more chemistry, plus a 20% chance to remove all megafauna from lavaland.

## Changelog
:cl:
del: tesla/singulo engine (the objects themselves still exist)
/:cl:


lolman360:
why remove it from cargo

Putnam3145:
you can't do a "quick engine" with the tesla and singulo because, again, they take forever to set up for no reason

and like. have you seen my supermatters? "the exact same thing every time" is the opposite of what goes down, I've done 4 completely different setups my last 4 times, each of them perfectly weathering whatever the hell went on in the meantime. Maybe I'm too "grizzled" and am "too good" at the supermatter now, but surges don't pose much of an issue for me.

And the reason surges can exist is because the supermatter actually has a state between "it's loose and needs to be put down or the round's over" and "it's harmless and providing tons of free power", which is something i didn't even mention here

> Not everyone needs that mechanical stimulation in the funny atmospherics simulator,,,

i'm not sure i can disagree harder? it's a game

lolman360:
no but seriously i get taking it out of rotation it's super boring but then why also remove it from cargo

dapnee:
Putnam, there’s changing one thing cause you don’t like it, and there’s deleting things other people enjoy. These are engines we almost never get that have traitor items that will now stop existing, these engines are more dangerous in terms of how easy they are to sabotage, but wow, low pop and the fact no one’s really allowed to go nuclear cause yeah, that’s not fun. What you’re leaving is three engines that work purely on atmos (exempting the antimatter engine), which is honestly boring.

You’re like, the only person with the will and know how to make a new engine, I’m sure it be more preferable to have something replace these instead of just straight removal.

LetterN:
@silicons tag `Awaiting further discussion/approval`

Auris456852:
Now, I halfway expect this to be a shitpost or some really hard salt, probably the latter. But just for good measure...

First off... Setting either engine should require no more time than it takes to set all the solars, if personal experience serves. Which, isn't much time. It would take me much more time to set any of those pipe mess SM setups if they aren't literally just replacing/adding pumps. Maybe it just feels like a slog because it's boring, but I am otherwise going to have to chalk it up to a difference in experience. A so-called skill issue, if you will.

But now, for the main reason to keep it. Sure, these engines are rather one-dimensional, but that is part of their strength, especially should an engineer wish to dabble more in construction work more than the biggest power number. Sure, it is dangerous if released. But, due to how there is mainly a single way to do things, short of getting fancy with a setup powering itself from the inside or making a pretty pattern with the field gens, a rookie engineer should have no issues grasping and troubleshooting either engine. All that you need, is to make sure all components are secure, field generators are not too far or close, emitters are powered and shooting the generators and that power will be received. It is also much easier to set them both up from scratch, as laying down the pipework, cooling loop and all, is not required.

I am sure an engineer main who specializes in such engines would probably be able to say more, since I don't really play much of that anymore. There is, of course, the fact that it is a really easy end of round option, for reasons not to keep it, and the other engines actually being salvageable if anything goes awry. But I personally advocate for a setup like DeltaStation's, otherwise travesty as it may be, where you can choose between either or.

DeltaFire15:
Very against this, even more so considering this even removes cargo's option to order it for engineers that feel up to it.
Roundstart (aka box) Singularity and tesla take setup time comparable to the supermatter if setup above bogstandard and neither are fully incontainable (Bags of Holding for the Singulo, grounding rods for the Tesla), though the Tesla far more so than its counterpart as a simple grounding-rod array can completely lockdown a released tesla to its general vicinity and drain its energy balls (or even use it as a power source still)
DNMing this unless Silicons overrules me.

keronshb:
We just got them back in rotation on box and... you want to remove them?  People want this shit.

This is also the same shit TG did.

Let's not become TG.

Putnam3145:
both have counterplay and I was paying close attention to the tesla counterplay, but my main update on my opinions vis a vis the tesla that came with that update was "not worth removing completely anymore"

i removed it from cargo mostly because that would minimize the chance of it actually getting merged, since I'm mildly scared that someone would actually merge it

I hope the "alternatives you've considered" entry illustrates my point better

MrsTonedOne:
Risk is fun.

Just because *you* don't find risk fun doesn't mean its not fun.

MrJWhit:
This doesn't even compile.

TheEaterOfBeans:
The real solution is to _make it so all engines are available on every map._

PR #15443, opened 2021-12-27
Prevents SMES from outputting power if in a loop

SandPoot:
## About The Pull Request
Title.

## Why It's Good For The Game
Maybe? putnam thinks this could affect performance positively

## Changelog
:cl:
tweak: SMES will no longer output any power if in a loop.
/:cl:

### DIFF
--- a/code/modules/power/smes.dm
+++ b/code/modules/power/smes.dm
@@ -43,6 +43,8 @@
 	. = ..()
 	if(!terminal)
 		. += "<span class='warning'>This SMES has no power terminal!</span>"
+	if(check_loop())
+		. += "<span class='warning'>This SMES' output loops back to the input, it cannot release any power!</span>"
 
 /obj/machinery/power/smes/Initialize()
 	. = ..()
@@ -265,7 +267,7 @@
 		if(outputting)
 			output_used = min( charge/SMESRATE, output_level)		//limit output to that stored
 
-			if (add_avail(output_used))				// add output to powernet if it exists (smes side)
+			if (!check_loop() && add_avail(output_used))				// add output to powernet if it exists (smes side)
 				charge -= output_used*SMESRATE		// reduce the storage (may be recovered in /restore() if excessive)
 			else
 				outputting = FALSE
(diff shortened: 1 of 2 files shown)


MrJWhit:
This is way to get a 0 TC power sink that's hidden with 4 cables and some gloves. Just connect the power output of the engine to the powernet of the station to instantly cut power.

shellspeed1:
> This is way to get a 0 TC power sink that's hidden with 4 cables and some gloves. Just connect the power output of the engine to the powernet of the station to instantly cut power.

This is really easy to do and hard to catch too

Putnam3145:
> This is way to get a 0 TC power sink that's hidden with 4 cables and some gloves. Just connect the power output of the engine to the powernet of the station to instantly cut power.



> > This is way to get a 0 TC power sink that's hidden with 4 cables and some gloves. Just connect the power output of the engine to the powernet of the station to instantly cut power.
> 
> This is really easy to do and hard to catch too

It is now trivial to catch because the SMES tells you if this is happening. Also, it wouldn't cut off power? I'm pretty sure the actual engine's power output would still be applying.

keronshb:
> > This is way to get a 0 TC power sink that's hidden with 4 cables and some gloves. Just connect the power output of the engine to the powernet of the station to instantly cut power.
> 
> > > This is way to get a 0 TC power sink that's hidden with 4 cables and some gloves. Just connect the power output of the engine to the powernet of the station to instantly cut power.
> > 
> > 
> > This is really easy to do and hard to catch too
> 
> It is now trivial to catch because the SMES tells you if this is happening. Also, it wouldn't cut off power? I'm pretty sure the actual engine's power output would still be applying.

Step 1 do the loop
Step 2 put tile back on
Step 3 laugh as no one can figure this out because no one looks at prs/git/code (seriously, people still don't know the tesla has a counter among other things).

The rest of the station won't be getting power unless the SMES are outputting it or if the engine is wired directly to the grid, which this promotes doing that more now because of this weird SMES nerf.

DeltaFire15:
I do not think this is needed, as using SMES as budget inline batteries, sometimes even manually toggling output / input for better effectiveness, doesn't seem like a bad thing.

silicons:
Not feeling it but only because of what @DeltaFire15 said
We really do want to get power priorities at some point for powernets.

PR #15444, opened 2021-12-28
Literally just makes it so slimes can have (decorative) wings.

TheEaterOfBeans:
## About The Pull Request

yes

## Why It's Good For The Game

customization good

## Changelog
:cl:
add: Slimes can have wings now.
/:cl:

PR #15448, opened 2021-12-30
literally also just makes it so IPCs can have winges

CoreFlare:
## About The Pull Request

gives IPCs wings just like https://github.com/Citadel-Station-13/Citadel-Station-13/pull/15444

## Why It's Good For The Game

i want my ultrakill wings. we have robotic wings but we can't use them on robots??? seems cringe to me

## Changelog
:cl:
add: Humanity is... Ok. Blood is fuel. IPCs have wings.
/:cl:


DeltaFire15:
Those wings are flight-potion only and functional as opposed to decorative, so you would not have customization access to them regardless.

PR #15457, opened 2022-01-04
Big Box Medbay Revamp [Ready to TM]

TripleShades:
## About The Pull Request
It's time to make poor design choices.
Medbay has been nearly completely relaid in an effort to make it fresh.

Surgery bay is now large instead of cramped, and the small storage room to the side has been replaced with a maint area, as its contents were moved into Medbay's storage proper.
Additionally, chairs were added outside of Chemistry, because I couldn't decide what else to put there.
surgery

Medbay's storage room is largely in the same spot, but now the right side leads to the break room.
The mini apothacary has been completely removed after previous discussion made it clear it wasn't really needed.
The patient room has been moved to be closer to Surgery, to help with quicker treatment.
Also, the delivery area has been moved to deliver into storage.
storage

CMO's office is now swapped with Genetics.
The treatment area has been condensed into one central area, and now connects to Cloning, which connects to Genetics still.
(Locker in front of airlock has been moved out of the way but I don't want to take a new picture)
central

Genetics has been moved down, and the in-between to Science has been expanded to mimic a similar hallway that Meta has between Toxins and Circuit Lab.
genetics

Virology has been shifted around, and is no longer directly connected to Medbay.
There is now a through-way through maints, if one bothers to smash down some things, as well as an Auxiliary Medbay Storage.
viro

Here is the current layout, for comparison:
current
current2

In addition, I've fixed a hovering sink in Meta's surgery, and added moth plushies to the patient rooms to have a comfort aid for patients.

## Why It's Good For The Game
The initial attempt I made to change Box's Surgery was very beginner level, and I thought to try and do a better job, which ended up spiraling into this.
I'm hoping that it ends up being well received, since Box is the most popular map, and could do with a fresh coat of paint so to speak.

## Changelog
:cl:
add: (Box) New maint "room" where Medbay's mini-storage used to be
add: Moth plushies to each station's patient room
del: (Box) Apothacary Room
tweak: (Box) Nearly all of Medbay layout
fix: (Meta) Right-side surgery sink was hovering and now it's not
/:cl:

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -35237,6 +35237,7 @@
 /obj/structure/sign/poster/official/no_erp{
 	pixel_x = 32
 	},
+/obj/item/toy/plush/mothplushie,
 /turf/open/floor/plasteel/white,
 /area/medical/patients_rooms/room_a)
 "bIl" = (
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -65646,6 +65646,10 @@
 	},
 /turf/open/floor/plasteel,
 /area/commons/fitness/pool)
+"vBY" = (
+/obj/machinery/portable_atmospherics/canister/oxygen,
+/turf/open/floor/plasteel,
+/area/medical/medbay/central)
 "vCb" = (
 /obj/machinery/rnd/production/techfab/department/service,
 /turf/open/floor/plasteel,
@@ -104931,7 +104935,7 @@ bwz
 brg
 btZ
 bzd
-bxQ
+vBY
 bxQ
 bxQ
 bAv
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -4912,14 +4912,11 @@
 /turf/open/floor/plasteel,
 /area/security/office)
 "ajv" = (
-/obj/machinery/computer/med_data{
+/obj/machinery/atmospherics/pipe/simple/scrubbers/hidden{
 	dir = 4
 	},
-/obj/structure/sign/poster/official/medical_green_cross{
-	pixel_x = -32
-	},
-/turf/open/floor/plasteel/white,
-/area/medical/medbay/central)
+/turf/closed/wall/r_wall,
+/area/command/heads_quarters/cmo)
 "ajw" = (
 /obj/structure/disposalpipe/segment{
 	dir = 6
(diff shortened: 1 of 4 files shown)


The-Sun-In-Splendour:
Why do you need to go through the BREAK ROOM to go through virology

The-Sun-In-Splendour:
Why is surgery reinforced
Why do I need to go through the public viewing to go inside surgery?

TripleShades:
> Why do I need to go through the public viewing to go inside surgery?

That's how it is on Meta and Pubby now.

Surgery being reinforced was something I forgot to fix, so let me do that.

TripleShades:
<snip>
An alternative layout is being considered right now, and may replace the above pictured.

Hatterhat:
maybe move the bloodbank gen closer to treatment, and the bioprinter closer to surgery? as of now they're a bit out of the way and probably a bit of a puta to try and upgrade if goin with BRPED cam shenanigans, and having them closer to more accessible areas would likely help get them used

bunny232:
> maybe move the bloodbank gen closer to treatment, and the bioprinter closer to surgery? as of now they're a bit out of the way and probably a bit of a puta to try and upgrade if goin with BRPED cam shenanigans, and having them closer to more accessible areas would likely help get them used

^
this, bloodbank/bioprinter should be closer to surgery

TripleShades:
Moved the blood bank and limb grower closer to surgery, but still inside of storage.
image

DeltaFire15:
This feels really cramped path-wise now, with only one way from the entrance to each point in med, as opposed to the old circle-ish very open layout in and around the cryo / sleeper area.. I dislike this.

TripleShades:
I do agree that it does feel cramped, which is why I'd like to test it first.
If it goes over well, a similar "crampening" could be done upon science, since there's also a large amount of space.

edit: If this doesn't go over well, I will simply just do the surgery area and rethink down the line any other layout changes to medbay.

TheEaterOfBeans:
Box good, major map changes bad.

TripleShades:
Accidentally deleted the repo but it's okay, I don't think this would have gone over well.
Back to the drawing board.

TheEaterOfBeans:
only thing I like from this is the operating theatre, solely because it doesn't let you get in from maintenance

TripleShades:
> 

Funny you say that, because I was told to add an entry from maints, so I did later on.
I agree that you should be able to enter from maints into the general room of surgery/waiting room.

TheEaterOfBeans:
> > 
> 
> Funny you say that, because I was told to add an entry from maints, so I did later on. I agree that you should be able to enter from maints into the general room of surgery/waiting room.

All I want is for antagonists to NOT be able to hack in and shoot me.

TripleShades:
That's too bad! If I go at it again, the surgery room will either have a direct maints door, or one in the waiting room like the other two I've put out.

PR #15469, opened 2022-01-14
Makes it so robots can heal normally from cores.

TheEaterOfBeans:
## About The Pull Request

Robots suck at mining.
This makes them suck less at mining.

## Why It's Good For The Game

Fast heals for robots is probably a good thing

## Changelog
:cl:
tweak: robots heal normally from legion cores now
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -578,9 +578,6 @@
 /datum/status_effect/regenerative_core/on_apply()
 	. = ..()
 	ADD_TRAIT(owner, TRAIT_IGNOREDAMAGESLOWDOWN, "regenerative_core")
-
-	if(HAS_TRAIT(owner, TRAIT_ROBOTIC_ORGANISM))	//Robots can heal from cores, but only get 1/5th of the healing. They can use this to get past the damage threshhold however, and then regularely heal from there.
-		heal_amount *= 0.2
 	owner.adjustBruteLoss(-heal_amount, only_organic = FALSE)
 	if(!AmBloodsucker(owner))	//use your coffin you lazy bastard
 		owner.adjustFireLoss(-heal_amount, only_organic = FALSE)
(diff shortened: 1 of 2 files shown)


TheEaterOfBeans:
it probably doesn't work but whatever
edit: nevermind it actually works fine

DeltaFire15 (changes requested):
As you may not be aware, the reason for legion cores to have reduced effects on Synthetics is that it can also repair thresholded limbs, effectively making it lavaland nanogel (except better considering instant and nodamageslowdown for a time in addition to its minor heal). I'd rather not have miner cores be simillarely prevalent in that aspect like miners handing out their highly powerful survival pens like candy.

If you want to change this still, either consider merely increasing their multiplier to 0.4 instead of 0.2 which should still be in line, or alternatively remove its ability to heal thresholds if you want them to heal like for organics.

TheEaterOfBeans:
> As you may not be aware, the reason for legion cores to have reduced effects on Synthetics is that it can also repair thresholded limbs, effectively making it lavaland nanogel (except better considering instant and nodamageslowdown for a time in addition to its minor heal). I'd rather not have miner cores be simillarely prevalent in that aspect like miners handing out their highly powerful survival pens like candy.
> 
> If you want to change this still, either consider merely increasing their multiplier to 0.4 instead of 0.2 which should still be in line, or alternatively remove its ability to heal thresholds if you want them to heal like for organics.

I want miner robots to get heals that DON'T need them to stand still so they can fight megafauna.
That's literally it.
I am not competent enough to code it better.

Thresholds shouldn't be a problem on Lavaland.

TheEaterOfBeans:
I would make legion cores decay if you take them off Lavaland or something like that, but I can't because again, I have zero actual coding skill.

edit: okay fuck it I will actually try to make cores unusable off Lavaland

TheEaterOfBeans:
yeah uh I was going to try stealing from nuke disk code after trying to use the suspiciously simple code someone posted in the Discord
but it's coNTROLLED BY "STATIONLOVING" INSTEAD OF ACTUALLY BEING IN THE NUKE DISK'S CODE, AND WHEN I FOUND _THAT_, IT WAS JUST
A WALL
OF CODE
THAT I HAVE NO HOPE OF UNDERSTANDING AND THEREFORE CAN'T STEAL FROM
**PAIN**

PR #15476, opened 2022-01-22
Buffs regen core healing for robots on lavaland, nerfs regen cores when not on lavaland

DeltaFire15:
## About The Pull Request
Regen cores used on Synthetics while on lavaland (or well, the mining z level, usually lavaland) now heal 80% of their base healing instead of 20%. 20 healing (of brute and burn) which ignores thresholds is pretty significant.
Regen cores used on anyone not on lavaland now have a .2 multiplier to healing and only apply ten seconds of slowdown reduction to strongly discourage farming them for onstation combat.

This is an alternate PR to #15469 

## Why It's Good For The Game
Cores used on Synthetics were somewhat weakish, but granting them the full heal previously would have been rather problematic given it is effectively better nanogel effect-wise. Now, they should be more viable for Synthetics on lavaland, while on-station use is discouraged except in specific situations (injured sprinting for a short time, needing a little bit of healing more, simillar)

## Changelog
:cl:
balance: Regenerative cores when used on Synthetics on lavaland now heal for 20 brute + burn instead of 5
balance: Regenerative cores when used now on lavaland now only heal 5 brute + burn, regardless of whom they are used on. Additionally, their damage slowdown negation is reduced to ten seconds.
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -578,8 +578,12 @@
 /datum/status_effect/regenerative_core/on_apply()
 	. = ..()
 	ADD_TRAIT(owner, TRAIT_IGNOREDAMAGESLOWDOWN, "regenerative_core")
-
-	if(HAS_TRAIT(owner, TRAIT_ROBOTIC_ORGANISM))	//Robots can heal from cores, but only get 1/5th of the healing. They can use this to get past the damage threshhold however, and then regularely heal from there.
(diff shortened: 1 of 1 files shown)


silicons (approved):

TheEaterOfBeans:
Make synths heal the full amount if the core is used on Lavaland, they can't use pens so they shouldn't heal less from cores.

PR #15502, opened 2022-02-07
Fungoids R&I Rework

LeadOnTaste:
Reworked fungoids metabolism, providing them some healing from nutriments. Pick light irritation quirk and be carnivorous for immersion. Also slightly buffed their brute resistance for realism, as chitin in cells provides it.

## Changelog
:cl:
add: Added new things
add: Added more things
del: Removed old things
tweak: tweaked a few things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/mushpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/mushpeople.dm
@@ -1,5 +1,5 @@
 /datum/species/mush //mush mush codecuck
-	name = "Anthropomorphic Mushroom"
+	name = "Fungoid"
 	id = SPECIES_MUSHROOM
 	mutant_bodyparts = list("caps" = "Round")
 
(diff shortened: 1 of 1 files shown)


deathride58:
In your description, you're describing a straight buff to shrooms in every way, with no associated explanation, reasoning, or elaborated thought process as to why you're wanting to make those changes. Your intent here appears to be blind power creep, to put it bluntly.

In your code, the effect you've added to nutriment has the exact opposite effect of what you're saying it's supposed to do; it adds damage rather than healing it. Additionally, the code doesn't even compile. As such, this strongly feels to me like a PR made to the wrong repo, as the PR doesn't match up at all with Citadel Main's code or balance expectations. Are you sure you submitted this to the right place?

Linzolle:
what bhijn said.

PR #15514, opened 2022-02-20
adds snake coiling

timothyteakettle:
## About The Pull Request
adds action button if you have the naga taur part
if you use it while grabbing someone, they're moved to your tile and you get a sprite change
that's it
if you rest, lose the taur part, move, or the grab is broken for whatever reason, then it stops and your sprite goes back to normal

this isn't *inherently* sexual so it should be fine right? haha? guys?

sprites are from skyrat i think, so thanks for those 👍 

oh also the offset will automatically update if you turn around so it always looks like they're in the coils

08d45a90f3cf91f57521456b05003b7a

## Why It's Good For The Game
funny snake hug

## Changelog
:cl:
add: adds snake coiling
/:cl:


Linzolle:
might want to consider handling granting the action in addition to removing it in handle_mutant_bodyparts so slimes don't get cucked out of it

desertmote:
I'm the artist. 😄

PR #14369, opened 2021-03-05
Ports ambition objectives system from Skyrat

dzahlus:
NOT THE HUGBOXED VERSION!!!

## About The Pull Request
PORTS: https://github.com/Skyrat-SS13/Skyrat13/pull/2266
I can't believe Silicons made me code, what it says on the tin

## Why It's Good For The Game
Silicons told me to do this, that's why it's good for the game. No seriously though we're having a lot of issues with antags being too boring, lame, what have you, because of their absolute inability to accept the fact that greentext is not what it is about. Hopefully this will change everything as it allows you to set your own objectives modularily. (admins have to approve any objective changes) You still spawn with all objectives and greentext still works the same way it does, but now you can request objective removal/completion, new objectives of your choice and ambitions.

C/P from the original PR

> Adds an ambitions system for antags. It stores a series of player-set information with the purpose of describing what they hope to achieve for the round, to provide better context to their actions and gimmicks at the round end.
> 
> Admins can see and edit the antag's ambitions through the objectives-ambitions panel.
> 
> All additions, edits and deletes are logged. Players have a small 5 seconds cooldown between edits to prevent spam, admins don't.
> 
> The max number of ambitions is set to 5 by default, but it's a config and thus can be changed either during the game or on the configs.txt
> 
> Added a safe lazylist cut macro for convenience.
> 
> Adds some cooldown systems for convenience. (cit already has this)
> 
> Adds a pair of admin macros to give them the choice of teleporting to the incident reported.
> 
> Did minor cleanup on the code.
> 
> Fixes a typo on the blackmail objective. (cit doesn't have this, not included)
> 
> On the same verb/panel, added an interface for players to request modifications, removals or additions of their objectives. The cooldown for using it is 10 minutes, which is pretty steep but it was what was requested of me.
> 
> Admins can access this interface through the Player Panel, which is now in nice dark-mode, instead of blinding white.
> 
> Players get notified when the cooldown has ended, and if they have pending requests that the next one will bwoink the admins.

WIP by the way lol, please check this code and tell me if you see anything that could break shit this is as of yet untested.

EDIT: Tested locally, all functions as intended, it compiles, admin shit works, rolling objectives works etc. removing WIP tag.

## Changelog
:cl:
add: Added ambition objective system
del: Removed old objective generation
tweak: tweaked the entire objective system
refactor: refactored a bunch of shit to work with ambition
config: added configs for ambition system
admin: changes to antag panel for ambition system
/:cl:

PR #15525, opened 2022-02-22
Ambitions from Skyrat

keronshb:
## About The Pull Request

It's this
https://github.com/Citadel-Station-13/Citadel-Station-13/pull/14369

Please, it's been a fucking year just speedmerge this

## Why It's Good For The Game
It was asked, original reasons in the post

## Changelog
:cl:
/:cl:


code/controllers/configuration/entries/general.dm
```diff
@@ -349,3 +349,8 @@
 /datum/config_entry/flag/dynamic_config_enabled
 
 /datum/config_entry/flag/station_name_needs_approval
+
+//ambition end
+	config_entry_value = 5
+/datum/config_entry/number/max_ambitions	// Maximum number of ambitions a mind can store.
```
Linzolle: these two lines are flipped around backwards for some reason

keronshb: ...weird.  Fixed.

Linzolle:
Can confirm this works.

jjpark-kb:
Just prefacing that I mainly play and contribute on Skyrat.

Ambitions was a decent system for antags, so that there were motivations behind what they were doing; however, there might be some performance issues and some discontent antags. Many antags we found dislikeed Ambitions because "RNG chose me to write an essay this shift," so you may find that antags start later than when they usually did.

Currently on Skyrat, we are using a system called OPFOR (Opposing Forces). It is similar to Ambitions, except it is applicable to everyone where people can request to do gimmicks.

If you go with Ambitions, I hope you enjoy it!

keronshb:
> Just prefacing that I mainly play and contribute on Skyrat.
> 
> Ambitions was a decent system for antags, so that there were motivations behind what they were doing; however, there might be some performance issues and some discontent antags. Many antags we found dislikeed Ambitions because "RNG chose me to write an essay this shift," so you may find that antags start later than when they usually did.
> 
> Currently on Skyrat, we are using a system called OPFOR (Opposing Forces). It is similar to Ambitions, except it is applicable to everyone where people can request to do gimmicks.
> 
> If you go with Ambitions, I hope you enjoy it!

Thanks for the heads up!  I'm not too familiar with either system, but I remember people telling me that ambitions went through a few different changes.  I think this might be one of the earlier versions, which I'm not sure what to compare it to.  I definitely want to try out ambitions here to see if they work, and if not so well then maybe I'll check out OPFOR.

Linzolle:
so the only weird thing i've noticed with this actually is the adding and removal of the Objectives and Ambitions verb; it seems to not show up in the IC panel sometimes, and also doesn't get removed when deleting the antag datum through the traitor panel.
Granted, for some reason when you aren't an antag and still have the objectives verb you can only add ambitions.

PR #15526, opened 2022-02-23
adds bartender cartridge for drink recipe searching, updates chemist cartridge to have equivalent functionality for all recipes

timothyteakettle:
## About The Pull Request
title
won't show the secret sauce recipe (i've tried and it does not return the recipe, woo.)
specifically looks at all consumable subtype recipes

works by adding an option that when clicked, shows a textbox to type in
you type a search term, i.e. 'tea' and it finds the first recipe matching it, i.e. 'Long Island Iced Tea'
it's not cap sensitive so you can type 'lOnG iSlAnD' and it'll still find it

it shows reagents and temperature required

## Why It's Good For The Game
nobody maintains the wiki anymore
makes it easy to look for recipes

image

## Changelog
:cl:
add: adds a cartridge for the bartender's PDA that lets them search drink recipes
add: updates chemist cartridge to be able to search all chemical recipes
/:cl:


SandPoot:
Chemistry subtype?

timothyteakettle:
maybe

Linzolle:
this is actually a very neat idea

deathride58:
You could probably do the search on the client via JS; send the client the list, do the search in-browser. This greatly reduces the potential workload the server would have to deal with.

TheObserver-sys:
This is based, and should be standard almost everywhere. The wiki's are rarely maintained, but the game is always up to date.

PR #15546, opened 2022-03-09
Makes supermatter take damage (and not heal) if too near to spaced turfs

Putnam3145:
## About The Pull Request

Yeah this was inevitable. Basically, for each space turf within 2 tiles of the supermatter, it takes a little bit of damage, and if there are *any*, it doesn't heal damage, ever. Could still allow panic repairs via spacing a turf, but won't allow setups based entirely around spacing a turf.

## Why It's Good For The Game

Engineers hate her! See how she used nothing but an RCD and 5 seconds to setup the supermatter with this one weird trick!

## Changelog
:cl:
balance: Supermatter now takes damage when near space tiles
/:cl:


Linzolle:
we must simply make the SM chamber bigger 4Head

Linzolle:
we must simply make the SM chamber bigger 4Head

PR #15570, opened 2022-03-22
Climax with partner now counts mouth

bunny232:
## About The Pull Request
Allows the climax with partner count people with their mouth exposed as a valid target. (I should note I can't test this on local but putnam said it will work.)
2/4 approving maintainers: 
https://discord.com/channels/161574200581029888/336748540690825216/955314725711921183 Lin
https://discord.com/channels/161574200581029888/336748540690825216/955479154277621780 Putnam
other two didn't respond over a 48+ hour period with two pings.

## Why It's Good For The Game
Uhhh... i don't think it is? But breastfeeding is hot af imo.

## Changelog
:cl:
add: The receiving party of a climax with partner action can have their mouth exposed to count as a valid 'hole' 
/:cl:

### DIFF
--- a/code/modules/arousal/arousal.dm
+++ b/code/modules/arousal/arousal.dm
@@ -145,7 +145,7 @@
 			partners -= L
 		if(iscarbon(L))
 			var/mob/living/carbon/C = L
-			if(!C.exposed_genitals.len || !C.is_groin_exposed() || !C.is_chest_exposed() || !C.is_mouth_covered()) //Nothing through_clothing, no proper partner.
+			if(!C.exposed_genitals.len && !C.is_groin_exposed() && !C.is_chest_exposed() && C.is_mouth_covered()) //Nothing through_clothing, no proper partner.
 				partners -= C
 	//NOW the list should only contain correct partners
 	if(!partners.len)
--- a/code/modules/arousal/arousal.dm
+++ b/code/modules/arousal/arousal.dm
@@ -145,7 +145,7 @@
 			partners -= L
 		if(iscarbon(L))
 			var/mob/living/carbon/C = L
-			if(!C.exposed_genitals.len && !C.is_groin_exposed() && !C.is_chest_exposed()) //Nothing through_clothing, no proper partner.
+			if(!C.exposed_genitals.len || !C.is_groin_exposed() || !C.is_chest_exposed() || !C.is_mouth_covered()) //Nothing through_clothing, no proper partner.
 				partners -= C
 	//NOW the list should only contain correct partners
 	if(!partners.len)


Captain277:
Hey my man, you should really change that title to something a little more euphemistic.

code/modules/arousal/arousal.dm
```diff
@@ -145,7 +145,7 @@
 			partners -= L
 		if(iscarbon(L))
 			var/mob/living/carbon/C = L
-			if(!C.exposed_genitals.len && !C.is_groin_exposed() && !C.is_chest_exposed()) //Nothing through_clothing, no proper partner.
+			if(!C.exposed_genitals.len || !C.is_groin_exposed() || !C.is_chest_exposed() || !C.is_mouth_covered()) //Nothing through_clothing, no proper partner.
```
DeltaFire15: This is an inverse conditional check checking for invalid conditions and removing unwanted elements from the list, which means you need to be checking for your condition not applying. Do this instead:
```suggestion
			if(!C.exposed_genitals.len && !C.is_groin_exposed() && !C.is_chest_exposed() && C.is_mouth_covered()) //Nothing through_clothing, no proper partner.
```

DeltaFire15 (changes requested):

keronshb:
Why do we need this.

code/modules/arousal/arousal.dm
Linzolle: this would require them to all be uncovered instead of just one which was talked about in the discord server. also your proposed change says C.is_mouth_covered() instead of !C.is_mouth_covered()

bunny232: ^ what lin said

bunny232:
> Why do we need this.

Horny kitsune wants breastfeeding instead of milking into a cup then feeding it to someone.

code/modules/arousal/arousal.dm
DeltaFire15: Look at the next line of code. This is a check filtering invalid elements which can be seen by the "candidates -=" in the next line, as opposed to a "+="
The current "&&" implementation is already an or, as !(A && B) is the same as (!A || !B).
Any of the parts of the logic expression need to be false for it to not remove the element, aka, at least one of the wanted conditions needs to be the case.

DeltaFire15: This is also the reason why it checks for if it IS covered, because that being the case *as well as* all other things being covered should remove it from the candidates.

Linzolle: I see that now, nevermind.

bunny232: Ah ok my b

PR #15603, opened 2022-04-26
Some weird high-temperature reactions (including: absolute hot)

Putnam3145:
## About The Pull Request

1. At above 2000 kelvins, nitrogen and oxygen start combining to make nitric oxide. This is a "fire" effect, but it's endothermic, i.e. having nitrogen in the air will keep fires from getting *too* hot. However, when nitric oxide gets colder, it reacts with oxygen in the air to make nitryl, so that'll have to be cleaned up from hot fires.
2. At above 2,000,000,000,000 kelvins (2 trillion), all gases break down into quark matter. This can have its temperature and amount of substance increased and decreased as one might expect. Adding more gas to the quark matter will turn it into more quark matter unless you get below 1.9 trillion kelvins, at which point...
3. Quark matter condenses into random gases, based on those gases' heat capacity. This is weighted by inverse of heat capacity, so stimulum is the most common gas here, followed by hydrogen, and so on. Tritium and quark matter are the only blacklisted gases from the list (mostly to prevent further fusion), the rest are possible, including new ones that may be added. After this process is over, the total energy is conserved (at least so far as it's tracked). Getting to two trillion kelvins is actually not terribly easy, in my experience; it takes something like 30 minutes of fusion unless you've got some real crazy stuff going on.

## Why It's Good For The Game

I like weird stuff, mostly. I'm also just feeling like going on a wild and crazy coding binge that may or may not have anything actually get into the game lately? These things I present for consideration more than expectation of merging.

## Changelog
:cl:
add: Nitric oxide (NO), with reactions associated
add: A maximum temperature above which all gases break down into quark matter, which will break down into gases again if it goes below that temperature.
/:cl:

### DIFF
--- a/code/__DEFINES/atmospherics.dm
+++ b/code/__DEFINES/atmospherics.dm
@@ -275,6 +275,7 @@
 #define GAS_PLASMA				"plasma"
 #define GAS_H2O					"water_vapor"
 #define GAS_HYPERNOB			"nob"
+#define GAS_NITRIC				"no"
 #define GAS_NITROUS				"n2o"
 #define GAS_NITRYL				"no2"
 #define GAS_HYDROGEN			"hydrogen"
@@ -289,6 +290,7 @@
 #define GAS_AMMONIA				"ammonia"
 #define GAS_FLUORINE			"fluorine"
 #define GAS_ETHANOL				"ethanol"
+#define GAS_QCD					"qcd"
 
 #define GAS_GROUP_CHEMICALS		"Chemicals"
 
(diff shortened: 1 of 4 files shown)


QRQ-2817:
At what point will gas EVER reach that temperature?? Two TRILLION kelvin??? The SM goes critical at 5000 kelvin. What the hell can even reach that temperature?

Putnam3145:
> At what point will gas EVER reach that temperature?? Two TRILLION kelvin??? The SM goes critical at 5000 kelvin. What the hell can even reach that temperature?

I've managed it with fusion a few times. Also, the supermatter's very, very recoverable at 5000 in my experience, but my experience is *weird*.

PR #15607, opened 2022-04-29
Updates auxmos to 1.0 (plus other atmos stuff with it)

Putnam3145:
## About The Pull Request

**This is not done until https://github.com/Putnam3145/auxmos/pull/43 is merged**

1. Updates auxmos to 1.0.0, fully stable, katmos can be enabled, heat works, all the good stuff.
2. Adds a "smells" system, allowing certain gases to be detected if they're in the air via smell, even if they're invisible. Nitric oxide, ammonia, bromine and (counterfactually) methane all have this feature.
3. Allows gases to use a generic icon and just have their color defined. Nitrogen dioxide (formerly nitryl) was moved to this system.
4. Scrubbers now scrub chems properly. I feel a bit embarrassed about this one.
5. Air alarms now procedurally generate their default TLVs, since before they wouldn't display any gas they weren't hardcoded to show (gross).
6. "Contaminated" air alarm setting now only scrubs out gases flagged DANGEROUS. This flag needs more uses anyway.
7. Water vapor was given the DANGEROUS flag.
8. Hydrogen is now visible; it's not as dangerous as trit, but still flammable, after all.
9. Bromine is visible.

## Why It's Good For The Game

Having a permatestmerge is bad. I tested this as much as I could--mostly by setting things on fire, repeatedly sending and receiving shuttles, all that good stuff--but I might not have gotten everything, so this should probably be testmerged for a little while.

## Changelog
:cl:
add: Smelly gases (Bromine, methane, ammonia, nitric oxide and stimulum in particular)
add: Easier-to-make-visible gases; gases made visible with this system are bromine and hydrogen (plus nitryl, but that was already visible in the same way)
fix: Scrubbers scrub chems properly
tweak: Air alarms now show all gases, not just ones they're hardcoded to show
tweak: Contaminated now only removes "dangerous" gases (this is actually a consequence of the hardcode reduction)
tweak: Water vapor is now counted as DANGEROUS (this makes it included in "contaminated" and adds admin logging when people open water vapor cans)
/:cl:


deathride58:
pogmos

PR #15625, opened 2022-05-16
Allow numbers in character names

Linzolle:
## About The Pull Request
Title.

## Why It's Good For The Game
We have IPCs and synths on this server, why are they not allowed to have numbers in their names.

## Changelog
:cl:
tweak: character names can now have numbers in them
/:cl:


deathride58:
idly wondering how this interacts with nameless chars, what happens if a character's All number

Linzolle:
All number names are still invalid with reject_bad_name, and it shouldnt affect nameless chars? Possible unintended consequence is that you can't have names that start with numbers, so something like "7EVEN" would get trimmed down to "EVEN" the same way that a hyphen is currently invalid at the start of a name.

deathride58:
Ah! That's p neat then

deathride58 (approved):

PR #15636, opened 2022-05-22
Switches transfer vote to approval voting

Putnam3145:
## About The Pull Request

As title. Switches the voting system for transfer votes from plurality to approval.

## Why It's Good For The Game

You might think this does *literally nothing*, since it's a two-option vote. But that's not the case: the raw amount of votes for "Continue playing" matters too. With this method, one could vote for *both* continue playing and initiate transfer, which is a distinct thing from not voting for either, since their vote would be counted for the minimum.

There are times I thought "well, I wouldn't mind it continuing, so I'll abstain unless 2 people vote continue, in which case I'll vote continue", but now you can just do both.

Approval also allows you to unvote completely, which is nice.

## Changelog
:cl:
tweak: Transfer vote is now approval instead of plurality
/:cl:

### DIFF
--- a/code/controllers/subsystem/autotransfer.dm
+++ b/code/controllers/subsystem/autotransfer.dm
@@ -31,7 +31,7 @@ SUBSYSTEM_DEF(autotransfer)
 	if(world.time < targettime)
 		return
 	if(maxvotes == NO_MAXVOTES_CAP || maxvotes > curvotes)
-		SSvote.initiate_vote("transfer","server")
+		SSvote.initiate_vote("transfer","server", votesystem=APPROVAL_VOTING)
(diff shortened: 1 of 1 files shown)


moonheart08:
Would the UX for this not be clearer if it presented a "don't care" option?

Putnam3145:
> Would the UX for this not be clearer if it presented a "don't care" option?

I'm not really sure about that? A "don't care" option wouldn't imply "this counts as continue for the minimum but not for the actual vote", I don't think. Then again, something that *is* explicitly that might be good, I guess, added to the approval vote.

deathride58 (approved):

PR #15638, opened 2022-05-22
fixes the ability to attach multiple tracking beacons to a single mech

deathride58:
## About The Pull Request
This is an exploit that's apparently existed for years, but has gone largely unreported. The exploit in question allows you to completely nuke a mech with just a box of tracking beacons

## Changelog
:cl: Bhijn & Myr
fix: You can no longer attach multiple tracking beacons to a single mech, fixing an exploit where mechs can be nuked with minimal effort.
/:cl:

### DIFF
--- a/code/modules/vehicles/mecha/mecha_control_console.dm
+++ b/code/modules/vehicles/mecha/mecha_control_console.dm
@@ -118,7 +118,7 @@
 	return ..()
 
 /obj/item/mecha_parts/mecha_tracking/try_attach_part(mob/user, obj/vehicle/sealed/mecha/M)
-	for(var/obj/item/I in M.equipment)
+	for(var/obj/item/I in M.contents)
 		if(istype(I, src))
 			to_chat(user, "<span class='warning'>[M] already has a tracking beacon!</span>")
 			return
--- a/code/modules/vehicles/mecha/mecha_control_console.dm
+++ b/code/modules/vehicles/mecha/mecha_control_console.dm
@@ -118,7 +118,12 @@
 	return ..()
 
 /obj/item/mecha_parts/mecha_tracking/try_attach_part(mob/user, obj/vehicle/sealed/mecha/M)
-	if(!..())
+	for(var/obj/item/I in M.equipment)
+		if(istype(I, src))
+			to_chat(user, "<span class='warning'>[M] already has a tracking beacon!</span>")
+			return
+	. = ..()
+	if(!.)
 		return
 	M.trackers += src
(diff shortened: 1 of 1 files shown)


DeltaFire15:
I do not think the issue with beacons is that you can attach multiple, but instead that they can be attached with no time cost at all (running up and clicking the mech once). Being able to attach multiple would not be a problem if the one intending this had to actually put efford into it.

deathride58:
No, the core issue is that you can attach an indefinite amount of beacons to a single mech. Adding a cooldown doesn't fix the underlying issue that someone can just roll up to a mech, spend the time attaching an indefinite number of beacons (keep in mind they're incredibly cheap to produce!) onto it (mechs take time to turn around), then fuck off to a maint console, and remotely detonate it from there, all with no counterplay on the mech pilot's part. Every individual beacon can be shocked to EMP the entire mech, as there's no shared cooldown between EMPs, thus allowing a mech to be destroyed in mere seconds, even though tracking beacon shocks are only intended to give a temporary stun with light amounts of damage on a cooldown. Do you not see the core issue here?

DeltaFire15:
Which would not be as much of an issue if mechs had counterplay aka it involved an attachment delay. A mech simply moving a single tile would cancel any such doafter, making attempts to driveby attach beacons less effective. Even with a single beacon allowed the key aspect of 'run up, instantly attach beacon, leave, EMP mech from another dimension' still exists, even if the burst disruption potential is somewhat lower.
Mechs take as much time to turn as to strafe 8r move in general, which means as long as the delay isn't close to zero it would be a pretty major challenge to achieve.

deathride58:
Again, you don't seem to be understanding the core issue; tracking beacons were never actually intended to allow mechs to be immediately remotely detonated. They're intended largely to, as the label says on the tin, track a mech, and periodically disrupt it in the event that it goes rogue. They also serve a combat purpose; allowing mechs to be tagged in the heat of combat, in turn allowing a remote party to periodically stun it.
A single glance of a mech's status text after having numerous tracking beacons placed on it is more than enough to tell you that tracking beacons are, in fact, not intended to be stacked on a single mech! The output is incredibly spammy.

If the ability to remotely detonate mechs is something you'd think would be good for the game, it'd be far better to instead make a module for that functionality, one that has a do_after for attachment, along with other balance considerations. As it stands right now, multiple tracking beacons on a single mech is a balancing nightmare of an exploit, no matter how you slice it.

Linzolle:
either way it shouldn't be able to have multiple, the fact that it can be attached instantly isn't what is supposed to be addressed with this PR

PR #15642, opened 2022-05-25
Multiplicative blood

deathride58:
## About The Pull Request
So! This PR's hard to explain in words. This PR makes the vast majority of blood visuals use multiplicative blending rather than overlay blending. This results in blood looking significantly less cartoony, which in my opinion, matches much better with the rugged visual aesthetic of this codebase. There's a caveat, though: due to a BYOND bug, this causes blood to completely disappear from the right click menu, despite the blood objects still retaining their mouse opacity.
There's also the potential concern of brighter slimes looking weird with this, as brighter colors are significantly less visible due to how multiplicative blending works.

Without further ado, here's a before and after!

Before: (Features some slight artifacting from manually varediting after taking below screenshot)
image

After: (Note: this is slightly out of date! Screenshot does not include multiplicative blood overlays, only multiplicative blood decals)
image

And here's an image of how the mob overlays look, in which the effect is immediately noticeable! (Note how the color of the uniform shines through significantly better than it does currently)
image

I would've applied multiplicative blending to item blood overlays, as well, but attempting to do so ends up breaking item animations, and my quick prodding reveals that it's incredibly easy to break item animations even further. As a result, item blood overlays are completely unaffected by this PR. I'm far too sober to rip apart item animation code as of writing.

## Changelog
:cl: Bhijn & Myr
tweak: All blood visuals, with the exception of item blood overlays, are now multiplicative!
/:cl:

### DIFF
--- a/code/datums/dna.dm
+++ b/code/datums/dna.dm
@@ -77,6 +77,7 @@
 	new_dna.species = new species.type
 	new_dna.species.say_mod = species.say_mod
 	new_dna.species.exotic_blood_color = species.exotic_blood_color //it can change from the default value
+	new_dna.species.exotic_blood_blend_mode = species.exotic_blood_blend_mode
 	new_dna.species.eye_type = species.eye_type
 	new_dna.species.limbs_id = species.limbs_id || species.id
 	new_dna.real_name = real_name
--- a/code/datums/mutable_appearance.dm
+++ b/code/datums/mutable_appearance.dm
@@ -10,7 +10,7 @@
 						// And yes this does have to be in the constructor, BYOND ignores it if you set it as a normal var
 
 // Helper similar to image()
-/proc/mutable_appearance(icon, icon_state = "", layer = FLOAT_LAYER, plane = FLOAT_PLANE, alpha = 255, appearance_flags = NONE, color = "#FFFFFF")
+/proc/mutable_appearance(icon, icon_state = "", layer = FLOAT_LAYER, plane = FLOAT_PLANE, alpha = 255, appearance_flags = NONE, color = "#FFFFFF", blend_mode = BLEND_DEFAULT)
 	var/mutable_appearance/MA = new()
 	MA.icon = icon
 	MA.icon_state = icon_state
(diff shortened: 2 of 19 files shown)


TripleShades:
As cool as this is, I'm really concerned about visibility of blood trails on certain tiles, which is especially noticable with the slime, as you've already said.
The darker colors also make it much harder to distinguish what species the blood could have come from, which I honestly think is important to know for investigative reasons, though this only seems to be the case for darker blood.

deathride58:
Yea, slimes in particular are a weird one. Personally, I'm torn between simply giving their blood a max luminosity (inheriting the hue and saturation), making blood as a whole revert to overlay blending when above a certain luminosity threshold (the two darker slimes in that screenshot do have decent-looking blood for what it's worth), or digging into blood code to make slimes specifically have overlay-blended blood (and maybe toning the alpha down on their decals for slight visual consistency with other blood decals)

Personally, I've found that the non-static blood colors are distinct enough to properly tell apart with multiplicative blending, even with my color blindness. On low-contrast displays (I took a glance at this PR on my phone, which is my go-to for testing worst-case display scenarios), I've noticed that the static blood colors remain fairly distinguishable (lizard blood vs human blood in particular; they can still be distinguished by hue on awful displays). However, dynamic blood colors, or more specifically, slime blood, is a weird case, in which red/orange slimes can end up with blood looking incredibly similar to human blood (which I suppose is more weight towards the third option of the above, making slimes use overlay blending for blood)

TripleShades:
That's a fair assessment! I think any of those options for slimes would be a good idea, assuming they're the only case it happens for, though it may be good to have it apply to a certain threshold for all blood in the rare case of adding new races (so the second option, if I'm reading this right?)
Ultimately, it's up to you on what to do, as you're the coder!

deathride58:
Went ahead and gave slimes specifically overlay-blended blood! Individual species can define their own exotic blendmodes (if someone wanted to make a species with additive blood, for instance, that'd be doable!)

Here's a screenshot! (ignore the xeno I forgot the station species lacks a unique blood color)
image

Poojawa:
Yey. glad this is still in use, at least. Honk

PR #15645, opened 2022-05-27
[S] properly patches up a freshly rediscovered aimbot exploit

deathride58:
No changelog since this is an exploit patch that doesn't affect good-faith players at all. The code that previously deterred this exploit is one that was particularly overengineered, and affects only a hyperspecific strain of the exploit.

This also removes a warning that's long since gone obsolete. The hyperspecific strain of the exploit that the warning notifies admins of, is one that's long since fallen into complete obscurity, and currently exists primarily to prank less-knowledgeable players into spamming admin logs. Aside from that, clicks are already logged more than well enough to manually investigate anomalous click patterns.

After some testing with higher movespeeds, this doesn't appear to have any impact on good-faith gameplay.

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -101,14 +101,6 @@
 	..()
 
 /client/MouseDrag(src_object,atom/over_object,src_location,over_location,src_control,over_control,params)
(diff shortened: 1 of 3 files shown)


keronshb:
Bless

PR #15653, opened 2022-05-30
changes to how slimes and cold interact

timothyteakettle:
## About The Pull Request
stuff this pr does:
1. removes all snowflake behaviour of slime lungs (less hp and get hurt by cold 25K earlier)
2. makes slimes start taking cold damage 25K earlier in **body temperature**
2.1. you can reduce the weakness by wearing things like winter coats, as this is how bodytemp changes work

for reference your body takes damage when 50K below regular temperature
ice reduces it by 5K / sodas reduce it by 8K (you cant stack these to freeze yourself with how adjust_bodytemp works)
this means slimes will take damage when 25K below regular temperature instead

no you will not die from drinking soda

please testmerge this first to get feedback

## Why It's Good For The Game
slimes are weak to cold, not just cold air!
this changes their weakness to mechanically make more sense

## Changelog
:cl:
add: changes to how slimes and cold interact, no longer cold air, but cold bodytemp
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/jellypeople.dm
@@ -24,6 +24,8 @@
 	coldmod = 6   // = 3x cold damage
 	heatmod = 0.5 // = 1/4x heat damage
 	burnmod = 0.5 // = 1/2x generic burn damage
+	cold_offset = -10 //take cold damage slightly earlier, can be negated by wearing things like wintercoats, space suits, etc
+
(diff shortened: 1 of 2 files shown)


timothyteakettle:
because of body temperature regulation, 25 is probably too low, so i'll experiment with values up to 40

timothyteakettle:
closing while some bodytemperature bugs are present that mean that cold_offset should not be used by species
edit: apparently bodytemperature also does not resemble actual body temperature in any way or form right now so uh yes

PR #15656, opened 2022-05-31
Animation refresh - Fully restores animations broken by BYOND updates, some light animation-related tweaks, and more

deathride58:
## About The Pull Request
So this PR is more or less an audit of major uses of animate(). Within the past year or so, changes to how BYOND handles animations has ended up completely breaking numerous animations in the game. This includes the most common animations in the codebase, such as squishing, jiggling, shaking, and more.

Additionally, some animations had some funky behavior when taking into account the existence of other animations. For instance, the float animation didn't play very well at all with Y offsets, spin animations were interrupted by the grab animation (this is technically a salt pr; my adminbus suffered from this), filteriffic always interrupts playing animations, etc etc.

This PR is largely unatomized, and touches every single broken/wonky animation that I'm personally aware of. I've also done a little bit of code cleanup, such as merging the overridden screenshake proc with the one in the base code (fixing a bug that's been around for 5 years, in which some screenshake sources don't respect the screenshake pref), and allowing the pull animation to support diagonals (though due to pulling being fairly buggy, this currently only applies to the initial grab)

## Why It's Good For The Game
Animations are polish. Polish good

## Changelog
:cl: Bhijn & Myr
fix: The shake animation, jiggle animation, squish animation, jitter animation, and float animation, have all been fixed.
fix: This also means that storage animations are fully working again. It's been a few years!
fix: The pull animation no longer interrupts ongoing animations.
tweak: The pull animation now supports diagonal directions (currently only applies to the initial grab! pull code is spaghetti)
fix: All screenshake sources now fully support the screenshake pref
code: cit_screenshake.dm has been fully merged into the base files.
fix: You can no longer remotely apply fingerprints to storage objects by clicking and dragging it around. Additionally, storage objects now rustle again when clicking and dragging (only took like,, a year to fix)
fix: The pickup animation now properly slides into your spaceman again, instead of just kinda disappearing.
tweak: When pulling a resting mob, they'll now skip the laying animation entirely, fixing a visual oddity where they shrink and grow to flip.
tweak: Alerts are now a little snappier and bouncier, as they've been swapped from sine easing, to back easing.
tweak: Radial menus now use sine easing, rather than linear easing. This just looks a little prettier.
tweak: Runechat now properly respects an object's X offset. Previously, it relied on a hardcoded check for dogborgs.
tweak: The float animation now uses pixel_z instead of pixel_y, fixing issues where the Y offset gets wiped due to the float animation. This also makes floating fully compatible with ctrl+WASD.
/:cl:


TripleShades:
Will this fix borgs sometimes not having their module transformation?

deathride58:
Admittedly haven't looked into that! I know borg TF requires an animated sprite for it, and those that lack sprites for it simply don't play the animation. From my brief testing with borgs for the runechat pixel_x support, it seems like transformations are consistent for skins that have an associated animation

TripleShades:
I mostly notice it on the Loader. Sometimes it'll fire and sometimes it wont. I hope that helps?

PR #15663, opened 2022-06-03
Blockquotes for chat; Prettier examine text and more!

deathride58:
## About The Pull Request
Pictures can speak a thousand words, so here's an image!
image

This PR adds support for the \<blockquote\> element to the chat UI. Several classes have also been granted block-quote specific properties. In my opinion, this looks a lot nicer for displaying combined lists.

## Changelog
:cl: Bhijn & Myr
add: The chat now supports blockquotes!
tweak: Examining, medical scanners, self-checking, and mood checkups, have all been converted to make use of blockquotes.
/:cl:

### DIFF
--- a/code/datums/components/embedded.dm
+++ b/code/datums/components/embedded.dm
@@ -346,9 +346,9 @@
 
 /datum/component/embedded/proc/examineTurf(datum/source, mob/user, list/examine_list)
 	if(harmful)
-		examine_list += "\t <a href='?src=[REF(src)];embedded_object=[REF(weapon)]' class='warning'>There is \a [weapon] embedded in [parent]!</a>"
+		examine_list += "<a href='?src=[REF(src)];embedded_object=[REF(weapon)]' class='warning'>There is \a [weapon] embedded in [parent]!</a>"
 	else
-		examine_list += "\t <a href='?src=[REF(src)];embedded_object=[REF(weapon)]' class='warning'>There is \a [weapon] stuck to [parent]!</a>"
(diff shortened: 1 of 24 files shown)


13spacemen:
for examine, specifically, you may want to look at this
https://github.com/Sandstorm-Station/Sandstorm-Station-13/pull/235

deathride58:
Huh, unaware that downstreams already tackled this concept!

That PR has more or less the same purpose as this one; allowing chat messages to be encapsulated within a box. Due to the differences in the implementation of going about this, these PRs are surprisingly compatible with each other on the CSS side of things, albeit slightly redundant (this PR makes use of the `<blockquote>` tag rather than `<div>`, which carries quite a few benefits.)

That being said though, I do notice that PR makes use of `<hr>` to denote sections. I'm admittedly unsure how that PR looks in the actual chatbox, but I'll likely experiment with how `<hr>` looks with this PR's blockquotes

13spacemen:
z
this is how it looks on SPLURT which is a Sandstorm downstream
I quite like the boxy look personally
@deathride58 
z2

deathride58:
Added support for \<hr\> tags! I haven't given it limited class support like I did for blockquotes, but it looks decent enough since the background color isn't affected by class anyhow (making that happen without adding further tech debt to the entire chat CSS in the process would require quite a bit of elbow grease to reduce the copy-pasting! The chat CSS has a *lot* of room for improvement, especially in regards to actually using SASS's features, though this is out of this PR's scope)

I've also added \<hr\> tags to examine text!

Dark mode:
image

Light mode:
image

PR #15677, opened 2022-06-09
Vocal barks - Fully customizable noises for talking!

deathride58:
## About The Pull Request
This PR's hard to explain in words! Here's a brief preview:

https://user-images.githubusercontent.com/6356337/172796553-fe810e2b-7bf0-4f8b-b29d-1cd631607784.mp4

Basically, this PR adds talking noises! It's all fully customizable, and it even supports genetics (which allows bypassing limitations and obtaining otherwise admin-only barks)! Every few characters (defined by the mob's bark speed) causes a sound to be played (defined by the mob's bark id) with a pitch (defined by the mob's bark pitch) carrying variance (defined by the mob's bark variance!). Out of the box, the majority of bark sounds available are instrument samples, and sounds that work well as musical stabs. This is entirely a stylistic choice designed to create a unique feel for soundbytes usable as barks. This also keeps the overhead and maintenance fairly low, as all that's needed for a bark is a single sound sample.

This PR also comes with a pref that allows you to mute these new sounds entirely. The pref is enabled by default, though! (this also comes with a savefile version bump since the toggle prefs dont automatically set defaults as they're added)

## Changelog
:cl: Bhijn & Myr
add: Talking is now accompanied with fully-customizable sounds! You can modify your character's speech under the speech tab in chargen, and you can toggle these sounds via the "Hear/Silence Vocal Barks" verb under the preferences tab! This also fully supports genetics.
fix: Admittedly an unatomized change, but Cit's DNA blocks can now properly be updated through genetics. Before, these required a complete genetics reset to apply.
/:cl:


SandPoot:
YES! DON'T STARVE SPEECH!

NullDagaf:
Time to look through the code in search of any N-bombs Bob might've left in.

deathride58:
> Time to look through the code in search of any N-bombs Bob might've left in.

Not sure what you're on about; Bobjoga's never had any direct PRs accepted here.
If you want idle banter, take it to the Discord rather than leaving it on random PRs.

deathride58:
> Not sure what you're on about

I've been informed that apparently Bob's server has a feature similar in core concept. But looking at its code, their implementation holds no concern for performance at all (their implementation reads a chat message's string as a list, and performs further string operations on individual characters, both of which are particularly pricy! it also `sleep()`s, which is particularly detrimental to performance. there's additional overhead by use of `playsound()`, which creates redundant `hearers()` calls every time, which is also fairly pricy. since saycode is one of the most expensive processes on every SS13 codebase, looking at their implementation almost gave me a heart attack, and the longer I stare, the more sins I find)

The implementation in this PR is designed with performance as the main focus, and as a result, differs heavily from Bob's implementation. Though you've probably already noticed that if your comment was in good faith.

ChungusGamer666:
> 

Thanks for the constructive criticism. I like your implementation but i will not change mine as i focused mostly on functionality which would be lost had i not repeatedly called playsound, and will happily take the performance loss since my server never, ever got more than 40 players. That said, using sleep was absolutely a terrible unexcusable idea which i'll rectify.
Not sure why Null had to mention it though since this is clearly the same idea but implemented VERY differently.

👍

deathride58:
For posterity's sake, since we were DM'd about it: we fully consent for key parts of the code within this PR (and the followup, #15684) to be used under the MIT license.

If you are porting code from here to SS14, then you don't need to do anything special, as `Space Wizards Federation` suffices as a copyright notice for this (ie: you don't have to worry about it if you're compliant with SS14's license).

If you are here to port to another SS13 codebase that uses something other than the AGPL, we'd prefer that the MIT copyright be attributed to `Bhijn and Myr`, though `Citadel Station Developers` and `Space Wizards Federation` are also acceptable attributions.

This permission cleanly applies to the following code:
- The defines added by this PR and #15684 within `code/__DEFINES/say.dm`
- All bark datums defined within this PR and #15684 (within `code/datums/bark.dm`)
- `/atom/movable/proc/bark()`, as defined within `code/game/say.dm`
- The `if(vocal_bark || vocal_bark_id)` codepath within `/atom/movable/proc/send_speech()`, also defined in `code/game/say.dm`
- The expanded vocal bark handling code for living mobs (which supports yelling), as defined within `code/modules/mob/living/say.dm`. We recommend taking the improvements to this from #15684 
- The additional preference customization code, alongside the bark preview code, as defined within `code/modules/client/preferences.dm`

When porting this PR, whether to SS13 or SS14, please be sure to take care that you aren't lifting more than what this PR and #15684 specifically add. We're only able to relicense what we've added to saycode here, so the rest of saycode is still under the AGPL.

Additionally, for SS14 specifically: do note that the code here is very specifically optimized for use within Byond. You don't need to manually handle playing sounds for specific clients like this code here does; you can just `PlayPredicted()` the bark sounds like any ordinary sound. You're also able to handle this entirely clientside (and we recommend doing so!)

PR #15649, opened 2022-05-28
Deals with a bunch of runtime sources

SandPoot:
## About The Pull Request
Makes a couple runtimes go away.

## Why It's Good For The Game
Less runtimes.

## Changelog
:cl:
fix: Deals with a bunch of runtimes.
/:cl:

PR #15682, opened 2022-06-11
fixes client/New to chat runtimes

silicons:


SandPoot:
#15649

code/modules/client/client_procs.dm
```diff
@@ -422,10 +422,12 @@ GLOBAL_LIST_INIT(blacklisted_builds, list(
...
 		GLOB.host = key
 		world.update_status()
-
+	
 	if(holder)
 		add_admin_verbs()
-		to_chat(src, get_message_output("memo"))
+		var/admin_memo_note = get_message_output("memo")
+		if(admin_memo_note)
+			to_chat(src, get_message_output("memo"))
```
deathride58: why are we running get_message_output a second time here

deathride58:
```suggestion
			to_chat(src, admin_memo_note)
```

deathride58 (changes requested):

code/modules/client/client_procs.dm
silicons: oh true

silicons:
@deathride58

PR #15683, opened 2022-06-11
shims storage to work properly until we can rewrite to atom storage

silicons:
tl;dr
mouse opacity fuckery is disrupting volumetrics
but is required for default mode

because click passinig broke.
oops.


DeltaFire15 (approved):
Helps with the opacity problem which I guess will work out till another storage rework.

PR #15693, opened 2022-06-15
Fixes some minor bits of code

TripleShades:
## About The Pull Request
These are all things I've fixed in a downstream that I've taken forever to port back up to Main.

1. Fixes the growth state icons of the four mutated forms of Cannabis (Death, White/Life, Rainbow, Omega) so that they actually use their own sprites instead of just the default Cannabis. Also lets them use their own death state icons, even though these are all the same in the icon file.
image

2. Fixes a typo in glassworking that made it so teacups could be finalized instead of getting to one of the final stages and having it eat your blowing rod and the incomplete glasswork, as well as making it so that once there, it can actually be turned into a teacup. Also, relatedly, fixes the description on the first step of spouts for glassworking, as it didn't require heat but the cutting tools instead. ALSO, RELATEDLY, fixes it so that trying to make tea cups doesn't get you tea plates/saucers and vice versa.
I am not touching the actual container code, even though I noticed it was broken. I can do that later or someone else can.
image

3. Enables #14696 to actually be in the admin verbs menu, under the Events section.
image

## Why It's Good For The Game
Things working good. Please yell at me if these need to be split up.

## Changelog
:cl:
fix: Cannabis mutation variants now use the proper growth state icons
fix: Picking to make tea cups will now give the right starting glasswork base
fix: Picking to make tea plates will now give the right starting glasswork base
spellcheck: Fixed a state typo in tea cup glassworking
spellcheck: Fixed an incorrect step displayed on spout glassworking
code: Enables the make priority announcement admin verb for admin use
/:cl:

### DIFF
--- a/code/datums/components/crafting/glassware/glassware.dm
+++ b/code/datums/components/crafting/glassware/glassware.dm
@@ -179,7 +179,7 @@
 
 /obj/item/glasswork/glass_base/spouty
 	name = "Glass fodder sheet (spout)"
-	desc = "A set of glass sheets set aside for glass working. This one is ideal for a spouty flask. It needs to heated with something very hot."
+	desc = "A set of glass sheets set aside for glass working. This one is ideal for a spouty flask. It needs to be cut with some glassworking tools."
 	next_step = /obj/item/glasswork/glass_base/spouty_part2
 
 /obj/item/glasswork/glass_base/spouty/attackby(obj/item/I, mob/user, params)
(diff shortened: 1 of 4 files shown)


TripleShades:
Belatedly, the tea saucer is at an ugly angle, and when putting the cup on the same tile as it, they dont even line up.
You cant even fucking have your tea!
I hate this.

PR #15696, opened 2022-06-16
fixes slimes blood processing (see pr description)

timothyteakettle:
## About The Pull Request
TESTMERGE THIS FIRST

it has been tested but it's best for a large species change like this to get some review on balance

currently handle_blood and spec_life for slimes runs at the same time
this means they are regenerating blood twice and taking damage penalties for blood twice
this means they die before even being able to cannibalize a limb and some parts of the code can't even be reached

this makes the following changes:
handle_blood is no longer used for slimes blood regeneration / low blood penalties, meaning they do not instantly die at 0 blood, nor do they regenerate blood from this proc, or take oxyloss from this proc
generic bleeding is still handled via this proc

spec_life continues to handle the regeneration of blood using nutrition for slimes, just using an updated tg version, slimes take brute when they have no blood and regain a small amount of blood (you can have negative blood from toxinlover it seems)

when below a certain amount of blood you will automatically consume limbs to gain 20 blood, (regeneration costs 40), losing your legs first, and upon having no more limbs to lose, you begin to lose oxygen (and thereby die)

----
right now 6u of charcoal or anti toxin can kill you via bloodloss, like you just outright die
with this: 15u of charcoal will crit you via toxin damage and you will then lose all your limbs over the course of about 30 seconds while in crit
12u of charcoal will crit you via toxin damage and you will lose both your legs, then gaining enough blood to be stable blood-wise, but still in crit

also they're regenerating far less blood passively than before, meaning drinking toxin is more important

also: you can no longer go past BLOOD_VOLUME_MAXIMUM from toxin as a slime (otherwise you'd just be able to have so much blood, charcoal/anti-toxin would do nothing to you)

## Why It's Good For The Game
makes slime blood work how it's supposed to work instead of having it be similarly handled in two places

the main change here is just no incredibly fast deaths from anti-toxins, instead it pulls it out to be a longer death that makes use of the limb cannibalism proc

## Changelog
:cl:
add: slimes no longer die from having no blood, properly consume limbs when at low blood, and suffer brute damage and oxygen loss for having no blood and no limbs left to consume respectably, they cannot go past BLOOD_VOLUME_MAXIMUM from consuming toxin
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -363,6 +363,3 @@
 #define MAPPING_HELPER_TRAIT "mapping-helper"
 /// Trait associated with mafia
 #define MAFIA_TRAIT "mafia"
-
-/// snowflake blood process (slimes)
-#define TRAIT_SNOWFLAKE_BLOOD_PROCESS "snowflake_blood_process"
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -363,3 +363,6 @@
 #define MAPPING_HELPER_TRAIT "mapping-helper"
 /// Trait associated with mafia
 #define MAFIA_TRAIT "mafia"
+
+/// snowflake blood process (slimes)
+#define TRAIT_SNOWFLAKE_BLOOD_PROCESS "snowflake_blood_process"
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -38,6 +38,9 @@
 	if(HAS_TRAIT(src, TRAIT_NOMARROW)) //Bloodsuckers don't need to be here.
 		return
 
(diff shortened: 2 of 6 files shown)


timothyteakettle:
before: 6u kills
now: 12u crits + removes both legs, 15u crits + removes all 4 limbs

this means that yes, it is a buff, it affects balance, and should be reviewed with a balance pass
even though yes a single syringe still kills you there's a lot of changes here worth considering

timothyteakettle:
point worth considering: make the maximum blood you can regen to based on species, then make it so roundstart slimes cant drink toxin past like 1.2x the normal starting blood level, while the xenobio slimes can go to maximum blood volume for the full 2k

timothyteakettle:
t-testmerge first?
haha?
it's fine right?
guys?

PR #15709, opened 2022-07-19
Fixes slime people (i don't know what the ill effect but runtimes is)

SandPoot:
## About The Pull Request
Proc calls without arguments, go figure.

## Why It's Good For The Game
More fixes!!!!!!!!!

## Changelog
:cl:
fix: Put arguments into handle_blood (it needs them).
/:cl:


SandPoot:
Okay, hang on.

code/modules/mob/living/carbon/life.dm
```diff
@@ -9,7 +9,7 @@
 		. = FALSE
 	if(!.)
 		return
-	handle_blood()
+	handle_blood(seconds, times_fired)
```
Putnam3145: i know it's out of scope but rename this var to delta_time i beg you (it's not actually necessary though)

Putnam3145 (approved):

code/modules/mob/living/carbon/life.dm
SandPoot: I would need to touch it elsewhere, as i'm just grabbing a proc argument from the mob subsystem

SandPoot: Actually, it should probably be fine yeah, lemme see...

SandPoot:
Rust-g download failed for integration tests.

PR #15714, opened 2022-07-22
Puts public library paintings in Omegastation's library instead of generic empty ones

BlueWildrose:
## About The Pull Request

This adds public library paintings to Omegastation instead of the empty ones.

## Why It's Good For The Game

You can now submit paintings that will persist in later rounds on Omega, and therefore have another meaningful thing to do on that lowpop map.

## Changelog
:cl:
qol: Public library paintings have been added to Omegastation. You can now submit paintings that will appear in later rounds on that station.
/:cl:


TripleShades:
Whoops that's probably my bad? I feel like that's my bad.
The various painting types confused me, so I think I grabbed whatever Box used?

PR #15727, opened 2022-07-28
Enables proper "green shifts" (goals, sec level) in dynamic

Putnam3145:
## About The Pull Request

1. Lower threat levels can have more station goals: below 30, you'll always have at least 2.
2. Modes can now set their own station goal budgets. All of them are 1 for now, except dynamic, which has a different one depending on threat level (extended hardcoded gets every station goal, still).
3. "Peaceful Waypoint" dynamic rounds are greenshifts for all intents and purposes now: they have ever station goal and start in green alert instead of blue.

## Why It's Good For The Game

Well, people keep asking for this. I guess it's for something to keep busy on in long rounds, but, like, I'm an atmos tech main, so, uh, I entertain myself automatically and thus might need more feedback.

## Changelog
:cl:
add: Dynamic greenshifts
code: STATION_GOAL_BUDGET define changed to a gamemode-level var
/:cl:

### DIFF
--- a/code/game/gamemodes/dynamic/dynamic.dm
+++ b/code/game/gamemodes/dynamic/dynamic.dm
@@ -274,10 +274,6 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 				. += "<b>Peaceful Waypoint</b></center><BR>"
 				. += "Your station orbits deep within controlled, core-sector systems and serves as a waypoint for routine traffic through Nanotrasen's trade empire. Due to the combination of high security, interstellar traffic, and low strategic value, it makes any direct threat of violence unlikely. Your primary enemies will be incompetence and bored crewmen: try to organize team-building events to keep staffers interested and productive."
 				green_shift = TRUE
-				for(var/T in subtypesof(/datum/station_goal))
-					var/datum/station_goal/G = new T
-					station_goals += G
-					G.on_report()
 			else
 				. += "<b>Core Territory</b></center><BR>"
 				. += "Your station orbits within reliably mundane, secure space. Although Nanotrasen has a firm grip on security in your region, the valuable resources and strategic position aboard your station make it a potential target for infiltrations. Monitor crew for non-loyal behavior, but expect a relatively tame shift free of large-scale destruction. We expect great things from your station."
--- a/code/game/gamemodes/dynamic/dynamic.dm
+++ b/code/game/gamemodes/dynamic/dynamic.dm
@@ -267,12 +267,13 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 
 /datum/game_mode/dynamic/send_intercept()
 	. = "<b><i>Central Command Status Summary</i></b><hr>"
+	var/green_shift = FALSE
 	switch(round(shown_threat))
 		if(0 to 19)
 			if(!current_players[CURRENT_LIVING_ANTAGS].len)
 				. += "<b>Peaceful Waypoint</b></center><BR>"
 				. += "Your station orbits deep within controlled, core-sector systems and serves as a waypoint for routine traffic through Nanotrasen's trade empire. Due to the combination of high security, interstellar traffic, and low strategic value, it makes any direct threat of violence unlikely. Your primary enemies will be incompetence and bored crewmen: try to organize team-building events to keep staffers interested and productive."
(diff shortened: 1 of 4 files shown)


deathride58:
Having hard tells for the state of the round in the form of the security level is something that has historically lead to quite a lot of genuinely awful player behavior. Notably, in the past, hard tells for green shifts have lead to people going out of their way to self-antag on the basis of "oh but there's no antags", and players refusing to ahelp during blue shifts due to the notion that self-antagging is exclusive to green shifts.

Personally I'm iffy on even having a hard tell present at all like there currently is. However, I think it might be worth experimenting in this regard, as dynamic means that the state of the round isn't necessarily a strict binary, especially during higher pop.

code/game/gamemodes/dynamic/dynamic.dm
```diff
@@ -291,9 +301,11 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 			. += G.get_report()
 
 	print_command_report(., "Central Command Status Summary", announce=FALSE)
-	priority_announce("A summary has been copied and printed to all communications consoles.", "Enemy communication intercepted. Security level elevated.", "intercept")
-	if(GLOB.security_level < SEC_LEVEL_BLUE)
-		set_security_level(SEC_LEVEL_BLUE)
```
deathride58, line 296: The new send_intercept no longer sets the security level at all. The roundstart security level doesn't get set at all with this PR. `raise_security_level()` and `lower_security_level()` are defined in this PR, but they aren't actually used anywhere.

Putnam3145: this was a bug from a completely different PR, fixed in #15813

code/game/gamemodes/dynamic/dynamic.dm
```diff
@@ -257,13 +257,23 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 			return rule.round_result()
 	return ..()
 
+/datum/game_mode/dynamic/generate_station_goals()
+	if(round(shown_threat) < 30)
+		if(length(current_players[CURRENT_LIVING_ANTAGS]))
+			station_goal_budget = 2
+		else if(round(shown_threat < 20))
+			station_goal_budget = INFINITY
+	return ..()
```
DeltaFire15, line 266: Shouldn't be able to be used as an antag check, which this is.

code/game/gamemodes/dynamic/dynamic.dm
```diff
@@ -291,9 +301,11 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
 			. += G.get_report()
 
 	print_command_report(., "Central Command Status Summary", announce=FALSE)
-	priority_announce("A summary has been copied and printed to all communications consoles.", "Enemy communication intercepted. Security level elevated.", "intercept")
-	if(GLOB.security_level < SEC_LEVEL_BLUE)
-		set_security_level(SEC_LEVEL_BLUE)
+	if(green_shift)
```
DeltaFire15, line 304: Should be probability based to allow fake greens / fake blues like our non-dynamic modes have.

code/game/gamemodes/dynamic/dynamic.dm
```diff
@@ -257,13 +257,23 @@ GLOBAL_VAR_INIT(dynamic_forced_threat_level, -1)
...
+	return ..()
+
 /datum/game_mode/dynamic/send_intercept()
 	. = "<b><i>Central Command Status Summary</i></b><hr>"
+	var/green_shift = FALSE
 	switch(round(shown_threat))
 		if(0 to 19)
 			if(!current_players[CURRENT_LIVING_ANTAGS].len)
 				. += "<b>Peaceful Waypoint</b></center><BR>"
 				. += "Your station orbits deep within controlled, core-sector systems and serves as a waypoint for routine traffic through Nanotrasen's trade empire. Due to the combination of high security, interstellar traffic, and low strategic value, it makes any direct threat of violence unlikely. Your primary enemies will be incompetence and bored crewmen: try to organize team-building events to keep staffers interested and productive."
```
DeltaFire15, line 275: Why does this allow roundstart antag metaing?

DeltaFire15 (changes requested):

PR #15744, opened 2022-08-02
Some tweaks to untied shoelaces (+ miner boots minor buff)

Putnam3145:
## About The Pull Request

Instead of harmonic probabilities, the untied shoelaces now generates a budget and subtracts from that budget. If a pair of shoes would be above the budget, it's ignored. The budget is based on the lace time. The average amount of untied shoelaces should now be about 2.5 (rather than the old *e*, i.e. 2.718), although slightly less due to the fact that harder shoes remove more from the budget.

Also, anyone not on the station Z level has a 50% chance of being skipped anyway.

Also, miner work boots have had their lace time increased from 5 seconds to 10 seconds. This reduces their chance of being caught in the event a bit and otherwise does little.

## Why It's Good For The Game

Miners falling over and just friggin' dying because of a random event isn't very fun.

## Changelog
:cl:
tweak: Untied shoelaces probabilities rework
balance: Miner boots are slightly harder to tie/untie now
/:cl:

### DIFF
--- a/code/modules/events/untie_shoes.dm
+++ b/code/modules/events/untie_shoes.dm
@@ -21,7 +21,7 @@
 			continue
 		if(!is_station_level(C.z) && prob(50))
 			continue
-		if(prob(5)
+		if(prob(5))
 			C.shoes.adjust_laces(SHOES_KNOTTED)
 			budget -= C.shoes.lace_time // doubling up on the budget removal on purpose
 		else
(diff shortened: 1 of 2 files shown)


code/modules/events/untie_shoes.dm
```diff
@@ -9,20 +9,23 @@
...
 			continue
 		if (HAS_TRAIT(C,TRAIT_EXEMPT_HEALTH_EVENTS))
 			continue
-		if(!C.shoes || !C.shoes.can_be_tied || C.shoes.tied != SHOES_TIED)
+		if(!C.shoes || !C.shoes.can_be_tied || C.shoes.tied != SHOES_TIED || C.shoes.lace_time > budget)
 			continue
-		if(prob(5))
+		if(!is_station_level(C.z) && prob(50))
+			continue
+		if(prob(5)
```
Linzolle: funny missing closing parentheses

PR #15791, opened 2022-08-17
Removes persistence from glowing goo messes

BlueWildrose:
## About The Pull Request

Title.

## Why It's Good For The Game

I just learned that it's entirely possible for the crew to treat a certain ghost-role antagonist just like revenants only they fire <REDACTED REAGENT> everywhere from a smoke machine instead of salt.

Glowing goo is removed from persistence in case that ends up becoming _the meta_. It hasn't yet, but since I vaguely mentioned it, someone might try it one day now (and probably irradiate unsuspecting crewmembers)

## Changelog
:cl:
balance: Glowing goo is removed from persistence.
/:cl:


BlueWildrose:
hold on I messed something up.

PR #15807, opened 2022-08-21
Dynamic uses antag rep now (please review carefully)

Putnam3145:
## About The Pull Request

Right now, dynamic doesn't use antag rep even if it's on. This allows it to be used, which should reduce the amount of times people get antag in a row. This only applies to roundstart and from-crew midround antags; latejoin and from-ghosts midrounds are still totally random, as latejoins don't exactly have a list to pick from and from-ghost is probably more fair to be completely random, since it's extremely opt-in.

## Why It's Good For The Game

I dunno, main reason *I* don't have antag on is because then I get antag more times in a row than I'd like. Maybe there's some other people like that. Variety's nice.

## Changelog
:cl:
tweak: Dynamic roundstart antags and from-crew midrounds now use antag rep
/:cl:


MrJWhit:
ahelp what's my antag rep.

shellspeed1:
As I said previously, I keep getting antag simply because no one else has it enabled. If you want to see different people as antags people need to turn it on. As is,  bringing reputation back will just eliminate the breathing room we currently had for antags to actually do their thing as the worst people possible (people who know way too fucking much about combat and only care about winning) will be playing heads and sec to maximize their chances at getting antag rolls. People will game the system if possible so its better to just not mess with it.

ghost:
> Maybe there's some other people like that. 

yeah, ive went through that multiple times
in fact it seems that i get antag every first round whenever i play.

DeltaFire15 (changes requested):
What are these changes to the candidate lists needed for? They don't really seem to serve any beneficial purpose, and in case of malf AIs even break things.

PR #15811, opened 2022-08-22
Adds new traitor bundle, Canned Ship. Merge ready, like seriously its been months.

shellspeed1:
## About The Pull Request
After a bit of testing with survival pods I realized it was possible to make one without an area. Taking the next logical step I thought about what could be made if we could designate the area of the survival pod and came up with the idea of a survival pod that was a ship. Behold, the result of this stupid idea, the ship in a can, now with extra fat.

Screenshot (640)

Anyway to summarize, this PR adds a 20 tc traitor bundle. It comes with a ship and a gear choice beacon which provides aa soft syndicate space and a selection in between a particle defender as recommended by .357, a stechkin and friends kit which comes with a spare agent ID needed for the ships airlocks and a space suit, and lastly the pirate kit which comes with a pirate space suit and energy sword. Please note these are not finalized and if deemed unbalanced will be removed. As for the ship itself, it comes with an advanced camera console, a syndicate sleeper, a bag of x4, and a syndicate toolbox with tools needed to assemble the ship.

As for fighting the ship it has several major weaknesses. First, its thrusters are exposed to space making them a target for heavier weaponry or brave assistants with toolboxes. Each thruster removed dramatically slows the ship down making it take longer to travel through space. If all are destroyed the ship can no longer move and is left a sitting duck. Next, the ship is intentionally built to have a gap in its defenses around its nacelles. While not necessarily easy to sneak all the way in, it is possible to cut off the ships power by taking advantage of the fact that its power source is located in said nacelles as well as the valve for one of the ships engines. Disabling both power supplies will leave the ship vulnerable as space travel is extremely power intensive. Lastly, an emp can be used to completely disable the ship as it has a single ACP meaning it cannot jump until its APC regains power and stops being shorted.

As previously noted I am not sure on the balance of this and I am willing to listen to ideas on how to improve this ship or even why it might be a terrible idea to add this. There is one feature I am unsure about how it will turn out altogether and that is custom ships being modifiable after being built. I would appreciate if this could be test merged as well, there were two runtimes occurring with one of them no longer occurring and Kevin providing a fix for the other. Otherwise I've tested everything fairly extensively and haven't found any issues.

## Why It's Good For The Game
We haven't had any new toys added for traitors to play with in a while and this particular one provides an excellent vehicle for several gimmicks while encouraging cooperation and perhaps a bit of backstabbing with your partner. Also did I mention it will help to get a rather unused system used a bit more so we can find more issues with it?

## Changelog
:cl:
add: Meet the syndicates newest product, a ship in a can.
/:cl:


shellspeed1:
Just as a note, the sprite for the ship is temporary, I intend to replace it with an actual can instead of just a survival capsule.

DeltaFire15:
I feel like this adds too much incentive for the good old "spaceborne" traitor that just goes space exploring & sits away from the station like a comms agent except they can teleport around & should be one to interact with the station.
If you have some ideas on how to alleviate this it'd be nice, as I like the concept of this gimmick item, just not the consequences.

PR #15818, opened 2022-08-25
Let's just hard cap QCD research output (per burst)

Putnam3145:
## About The Pull Request

yeah turns out you can make tens of millions at a time through basic (well, as advanced as literally possible, but STILL) fusion methods, so, uh. nah

## Why It's Good For The Game

Atmos shouldn't be able to trivialize science?? Sorry??

## Changelog
:cl:
balance: Quark matter research is now capped at 100k points per decomp burst
/:cl:


deathride58 (approved):
It might be worth it to make it so there's diminishing returns, similar to toxins, but this does break the meta for the timebeing, which is good.

PR #15836, opened 2022-09-01
Botany max yield is now 4, max potency 250

Putnam3145:
## About The Pull Request

Same total ingredients. Might actually be a buff????

## Why It's Good For The Game

image

## Changelog
:cl:
tweak: Botany yield has been reduced and potency increased accordingly
/:cl:

### DIFF
--- a/code/game/objects/effects/glowshroom.dm
+++ b/code/game/objects/effects/glowshroom.dm
@@ -69,8 +69,8 @@
 	if(spread)
 		myseed.potency -= round(myseed.potency * 0.25) // Reduce potency of the little mushie if it's spreading
 	if(mutate_stats) //baby mushrooms have different stats :3
-		myseed.adjust_potency(rand(-4,3))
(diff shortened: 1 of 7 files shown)


shellspeed1:
Honestly, I don't think this will solve the problem of hundreds of items on a tile causing crashes, especially when botany can be scaled up infinitely. There is also a whole issue of plants being capped at how much they can hold reagent wise which makes this a hard nerf too. I think it would be overall better to go  with a rotting mechanism to force plants on the floor to clear up.
Edit: bees are not accounted for here, meaning you are possibly looking at 2 yield without them.

silicons:
we do realize what 250 potency does right

just make items not harvest if the tile is stacked tbh

silicons:
also at a certain point it's an administrative thing of like

"hey please don't intentionally [crash] the server, the funny gets unfunny when no one gets to play"

Putnam3145:
yeah i've got a better thing in mind lol

PR #15840, opened 2022-09-04
Hard caps supermatter gas output

Putnam3145:
## About The Pull Request

1. Plasma and oxygen are now only up to 100 and 200 moles per second, respectively. Before, there was no hard limit on this, and as it turns out, it could add... quick calculation suggests it *should* have been up to 240 per second, but that seems a lot.. Even 100 per second might be a bit much? But it takes quite a lot to get to that point in the first place.
2. The oxygen generation algorithm has been fixed so that it doesn't escalate so rapidly with temperature that plasma cannot keep up. I'm not even sure it was intentional that it was the way it was. 
3. The supermatter can now cool down gas again, like it used to. It hard caps the temperature of the gas it releases back to the environment, so if it gets too hot it will make it cold again. This hard cap lowers with better coolant gases (plasma is most dangerous for this, minimum hard cap possible is 1250).

## Why It's Good For The Game

Runaway supermatter delaminations that do not involve passive vents are WAY too strong.

## Changelog
:cl:
balance: Nerfed runaway supermatter delaminations with various capping tweaks
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -42,7 +42,7 @@
 #define THERMAL_RELEASE_MODIFIER 350         //Higher == more heat released during reaction, not to be confused with the above values
 #define THERMAL_RELEASE_CAP_MODIFIER 250     //Higher == lower cap on how much heat can be released per tick--currently 1.3x old value
 #define PLASMA_RELEASE_MODIFIER 750        //Higher == less plasma released by reaction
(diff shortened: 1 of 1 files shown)


deathride58 (approved):
Okay, yeah. Playing around with this on local, this is a bit of a nerf to hellburns, but it *does* make delams significantly less awful to deal with. As one of the few people who know how to do hellburns, this is an acceptable tradeoff.

PR #15841, opened 2022-09-05
Some dynamic midround tweaking

Putnam3145:
## About The Pull Request

1. Makes dynamic's midround antag picking respect the midround antag pref. Yeah, turns out it just... doesn't?? It just ignores it?? Yikes.
2. From-ghost midrounds that fail now just try to roll another midround in 60 seconds instead of, uh, trying to make a midround traitor and then making another traitor _if it succeeds_(??)
3. Midround on-station blob has been fixed. Yeah, we've had it this whole time, it was just broken, programmed to check if you have an antag type that does not exist. Whoops!
4. Adds a midround from-station wizard. Lower weight by default than the from-ghosts wizard, since it's a bit more obtrusive. Asks everyone if they actually want to be a wizard first. Might add an event that asks this question anyway, as a goof.
5. Selecting "never for this round" when dynamic asks for an antag actually works now

## Why It's Good For The Game

I mean, the first two are baffling. The last two are questionable, but worst case scenario they can be disabled.

## Changelog
:cl:
add: Midround wizard from crew
tweak: Dynamic midround antags now respect the midround antag setting
tweak: From-ghost midround failure no longer results in either two or zero traitors rolling (??)
tweak: "Never for this round" works for dynamic antags now
fix: Midround blob no longer uses an invalid antag flag (that nobody has on by its nature)
/:cl:

### DIFF
--- a/code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
+++ b/code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
@@ -15,6 +15,7 @@
 	var/restrict_ghost_roles = TRUE
 	/// What mob type the ruleset is restricted to.
 	var/required_type = /mob/living/carbon/human
+	var/should_use_midround_pref = TRUE
 	var/list/living_players = list()
 	var/list/living_antags = list()
 	var/list/dead_players = list()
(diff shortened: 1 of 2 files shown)


DeltaFire15:
Seems mostly fine. Disliking wizard midrounds as char picked from stationside individuals, so merging this with that one's weight set to 0.

DeltaFire15 (approved):

PR #15857, opened 2022-09-18
Readds the premium ka for borgs

shellspeed1:
## About The Pull Request
Readds the premium KA for borgs as a 8000 point mining vendor reward to match the premium ka for miners. My goal in doing this is to add at least some level of progression outside of the tech webs for borgs because miners have come to significantly outclass the capabilities of most mining borgs.

## Why It's Good For The Game
Currently miners have access to insane powerful armors, extremely powerful mining equipment with sonic jackhammers, advanced plasma cutters which borgs get a worse version of, forged pickaxes, the rose gold pickaxe, premium kas which outclass the borgs current ka, and the massive strength that is the premium melees which allow parrying which makes quite a bit of fauna barely even dangerous, especially with parrying which borgs can't do. In order to give borgs a slight leg up and make mining feel like less of a shitty experience, I would like to reintroduce the premium KA which gives borgs a ranged option which is useful without having to change their armor or trying to bring back the far stronger premium melee upgrades. Having it be 8k mining points too instead of just a buildable tech also means that borgs can't just instantly outclass miners. Additionally, the 8k mining points mean that borgs that want to valid hunt have to either go and spend time mining in advance, mine while other people deal with the antag or find someone to buy the upgrade for them which isn't very likely with miners.  Not that a KA is really useful against most threats without a traitor providing assistance due to the pressurized nature of most space stations. KAs are significantly weaker in such environments and if a borg started draining atmos or blowing holes in the station to fight people regardless of its laws it would probable be bonked. If we were to instead use the kinetic melees which can do damage in pressurized environments we would likely see far more valid hunting which makes these method idea. Lastly, this method of giving borgs a leg up comes with obvious trade offs as the KA cannot be good to be good at mining or combat at the same time meaning that distinct trade offs need to be made when using this item, especially since it replaces the basic KA that borgs have. 

TLDR: This pr is good because it gives borgs a bit of progression and makes the power scaling difference in between miners and borgs significantly less without resorting to stronger buffs which can cause worse issues. Also gives borgs something to work for.,

## Changelog
:cl:
balance: The borg premium KA has returned as an 8000 point reward in the mining vendor.
/:cl:


timothyteakettle:
i want no power scaling for borgs
specifically: i think mining borgs should have nothing but the regular KA

PR #15862, opened 2022-09-23
adds multiline /me emotes

silicons:
ayo

## why this is good for the game

roleplay

## changelog

:cl: 
add: multiline me emotes
/:cl:

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -53,6 +53,36 @@
 /proc/sanitize(t,list/repl_chars = null)
 	return html_encode(sanitize_simple(t,repl_chars))
 
+//Used for preprocessing entered text
+/proc/sanitize_rp(var/input, var/max_length = MAX_MESSAGE_LEN, var/encode = 1, var/trim = 1, var/extra = 1)
+	if(!input)
+		return
+
+	if(max_length)
+		input = copytext(input,1,max_length)
+
+	if(extra)
+		var/temp_input = replace_characters(input, list("\n"="  ","\t"=" "))//one character is replaced by two
+		if(length_char(input) < (length_char(temp_input) - 12)) //12 is the number of linebreaks allowed per message
+			input = replace_characters(temp_input,list("  "=" "))//replace again, this time the double spaces with single ones
+
+	if(encode)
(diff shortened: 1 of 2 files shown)


code/__HELPERS/text.dm
```diff
@@ -53,6 +53,36 @@
 /proc/sanitize(t,list/repl_chars = null)
 	return html_encode(sanitize_simple(t,repl_chars))
 
+//Used for preprocessing entered text
+/proc/sanitize_rp(var/input, var/max_length = MAX_MESSAGE_LEN, var/encode = 1, var/trim = 1, var/extra = 1)
```
Zirok-BYOND: I'd rename this to sanitize_multiline, and specifically annotate in a comment that, unlike sanitize, this does **not** strip newlines

code/__HELPERS/text.dm
```diff
@@ -53,6 +53,36 @@
...
+	if(!input)
+		return
+
+	if(max_length)
+		input = copytext(input,1,max_length)
```
Zirok-BYOND: I think if we need to trim we should be be loud rather than silently dropping parts of a long emote. Additionally, we're calling copytext here with MAX_MESSAGE_LEN, and then the caller does it **again** to what we return, which seems fairly redundant?

code/modules/mob/say.dm
```diff
@@ -54,7 +54,7 @@
 		to_chat(usr, "<span class='danger'>^^^----- The preceeding message has been DISCARDED for being over the maximum length of [MAX_MESSAGE_LEN]. It has NOT been sent! -----^^^</span>")
 		return
 
-	message = trim(copytext_char(sanitize(message), 1, MAX_MESSAGE_LEN))
+	message = trim(copytext_char(sanitize_rp(message), 1, MAX_MESSAGE_LEN))
```
Zirok-BYOND: I suppose this can simply be `message = sanitize_rp(message)`, though I'd definitely add comments that sanitize_rp trims and limits length. Likewise to sanitize_rp to callers.

silicons:
should work better now

deathride58 (approved):
fuck it; ship it

PR #15863, opened 2022-09-27
Reworks RBMK to respect gas properties, use auxgm in general, use the same units

Putnam3145:
## About The Pull Request

1. Rather than simply calculating the "heat delta" by literally checking the temperature difference ignoring the actual amount of coolant (!!!), the RBMK now uses the pre-existing temperature_share function
2. All instances of PSI have been replaced with kPa. We're not wasting cycles on converting to and from units we don't use elsewhere in the server for literally no reason except to confuse you. Like. Why.
3. All instances of storing temperature in celsius have been replaced with kelvins. See above. At least this one's additive instead of multiplicative.
4. RBMK power output has been halved. 20% is 800 kW at this point, now it's 40% for that. 20% is still enough to power the whole damn station, even, which might still be too much.
5. All the hardcoded get_moles calls have been auxgmified, same way the supermatter is.
6. Supermatter's auxgm integration has been mildly reworked. All properties for RBMK and supermatter gas stuff are now stored in their respective files, rather than being in the big gas_types.dm file. DM lets me do this, I'm using it, sorry.
7. RBMK no longer spams admins with "not enough fuel" messages.
8. Fuel rod radioactivity is now permanent (but cannot contaminate)

## Why It's Good For The Game

Ehh, remains to be seen. Perhaps keeping it simple and hardcoded is for the better, but I sorta doubt it (otherwise I wouldn't have opened this).

## Changelog
:cl:
tweak: RBMK coolant now respects heat capacity and mole amount (!)
tweak: RBMK now uses kelvins and kPa internally instead of celsius and PSI
tweak: RBMK now uses PSI for display instead of celsius (seriously why. WHY)
refactor: Supermatter/RBMK gas info now lives in the supermatter/RBMK files
/:cl:

### DIFF
--- a/code/modules/power/reactor/rbmk.dm
+++ b/code/modules/power/reactor/rbmk.dm
@@ -4,28 +4,85 @@
 #define MODERATOR_INPUT_GATE airs[2]
 #define COOLANT_OUTPUT_GATE airs[3]
 
-#define RBMK_TEMPERATURE_OPERATING 640 //Celsius
-#define RBMK_TEMPERATURE_CRITICAL 810 //At this point the entire ship/station is alerted to a meltdown. This may need altering
-#define RBMK_TEMPERATURE_MELTDOWN 900
+#define RBMK_TEMPERATURE_OPERATING 640+T0C
+#define RBMK_TEMPERATURE_CRITICAL 810+T0C //At this point the entire ship/station is alerted to a meltdown. This may need altering
+#define RBMK_TEMPERATURE_MELTDOWN 900+T0C
 
 #define RBMK_NO_COOLANT_TOLERANCE 5 //How many process()ing ticks the reactor can sustain without coolant before slowly taking damage
 
-#define RBMK_PRESSURE_OPERATING 1000 //PSI
-#define RBMK_PRESSURE_CRITICAL 1469.59 //PSI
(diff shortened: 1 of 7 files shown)


code/modules/power/reactor/rbmk.dm
```diff
@@ -4,28 +4,85 @@
 #define MODERATOR_INPUT_GATE airs[2]
 #define COOLANT_OUTPUT_GATE airs[3]
 
-#define RBMK_TEMPERATURE_OPERATING 640 //Celsius
-#define RBMK_TEMPERATURE_CRITICAL 810 //At this point the entire ship/station is alerted to a meltdown. This may need altering
-#define RBMK_TEMPERATURE_MELTDOWN 900
+#define RBMK_TEMPERATURE_OPERATING 640+T0C
+#define RBMK_TEMPERATURE_CRITICAL 810+T0C //At this point the entire ship/station is alerted to a meltdown. This may need altering
+#define RBMK_TEMPERATURE_MELTDOWN 900+T0C
```
Linzolle, line 9: i know this is like a super old PR that i am just now getting to look at but. shouldn't this be _minus_ T0C because the numbers were originally presented in celsius?

DeltaFire15:
Merge conflicting & unchanged since a year.

PR #15868, opened 2022-10-06
misc clockcult balance changes

KeRSedChaplain:
## About The Pull Request
Three small changes to three seperate parts of the mode.

--Cyborg scripture: Allows engineering and service borgs to summon obelisks and mania motors (specific to the engineering module) without a second invoker
--Clockwork armor: Gives the armor 50 laser/energy protection, with the war variable adding another 15
--Clockwork mech: Makes it tankier, but slightly slower, and removes it's internal damage threshold. Also reduces the cost to summon by a considerably amount, from 75kw down to 40kw. 

## Why It's Good For The Game
--Cyborg scripture: More incentive to convert cyborgs. Not much else to say about that.
--Clockwork armor: -25 laser/energy armor as it originally had was idiotic and 0 armor wasn't much better considering the station's main (and arguably strongest) armament are mostly laser weaponry.
--Clockwork mech: Neovgre, quite frankly, is very underwhelming. It has less resistances than the clockwork armor prior to this PR from damage sources, and about 60% of the time the pilot would die to an internal fire and defeat the main drawback of the mech. To weigh out the armor buffs, it's now one level slower. As for the scripture's cost, the cult only gets 80kw maximum they can store, and the ark itself costs 70kw of it. As far as I can tell, this went under the radar during when I was balancing the mode to be station-based instead of focusing around a seperate Z level.

## Changelog
:cl:

balance: lets engineering/service borgs use structure summoning scripture without having to have another servant
balance: gives clockwork armor 50 (65 with the buffed version) laser/energy protection, none from -25 was still a dumb choice
balance: made Neovgre tankier and cheaper, but slower. No more dying due to an internal fire that you can't fix yourself!

/:cl:

### DIFF
--- a/code/modules/antagonists/clockcult/clock_items/clockwork_armor.dm
+++ b/code/modules/antagonists/clockcult/clock_items/clockwork_armor.dm
@@ -8,7 +8,7 @@
 	resistance_flags = FIRE_PROOF | ACID_PROOF
 	flags_inv = HIDEEARS|HIDEHAIR|HIDEFACE|HIDESNOUT
 	mutantrace_variation = STYLE_MUZZLE
-	armor = list(MELEE = 50, BULLET = 70, LASER = 0, ENERGY = 0, BOMB = 60, BIO = 0, RAD = 0, FIRE = 100, ACID = 100, MAGIC = 60, WOUND = 65)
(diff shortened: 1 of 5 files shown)


githubuser4141:
Why would their mech need such absurdly high armor, it's on-par with the seraph (deathsquad) mech now. Mechs don't care about penetration, so this thing's going to take tickle damage from almost any source

PR #15869, opened 2022-10-07
Grant Arachnid Abilities to Driders

JerryWester:
## About The Pull Request

Self-explanatory, allows other races to use "spin web" and "spin cocoon" with the drider taur option

## Why It's Good For The Game

One would assume that half-spider-half-humans would be just as capable of spinning webs as humanoid spiders. That aside, I have no real justification for the change other than "driders hot, web bondage hot"

## Changelog
:cl:
add: `STYLE_ARACHNID_TAURIC` taur mode and added to Drider taur type
tweak: move removal of arachnid abilities from `/datum/species/arachnid` to its parent's class
tweak: grant arachnid abilities if Drider taur type is used on any other race
tweak: slime customization accounts for the ability to change/remove taur legs
/:cl:

### DIFF
--- a/code/__DEFINES/inventory.dm
+++ b/code/__DEFINES/inventory.dm
@@ -103,11 +103,12 @@
 #define STYLE_SNEK_TAURIC		(1<<2) //taur-friendly suits
 #define STYLE_PAW_TAURIC		(1<<3)
 #define STYLE_HOOF_TAURIC		(1<<4)
-#define STYLE_ALL_TAURIC		(STYLE_SNEK_TAURIC|STYLE_PAW_TAURIC|STYLE_HOOF_TAURIC)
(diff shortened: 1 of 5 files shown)


KeRSedChaplain:
honestly quite based :garfieldsmoking:

timothyteakettle:
this gives a mutation ability with no downside to anyone who uses a specific taur option
no

arachnids take more brute damage as a downside to their abilities
play an arachnid with drider or look into an arachnid sprite change

silicons:
unbalanced but i think balance is the least of our worries atm

JerryWester:
I realized pretty soon after I made the PR that this implementation was pretty poor. In all honesty, I was kind of hoping this would just slip past the radar and end up in PR hell, where it would remain open until inevitably closed. I made an edit downstream that I'm much more happy with (still imperfect, but preferable to this). For this PR, I'll just revert and close.

Another issue cropped up in that downstream when it got testmerged, the concern of preference issues. While I'd love to deal with that, it's equally an issue with the arachnid race. Might consider implementing a pref toggle for that in the future.

PR #15877, opened 2022-10-26
Lowering the Transfer Vote Threshold.

IHOPMommyLich:
## About The Pull Request
As a palliative solution, lowering the threshold should enable the people who are currently playing to run into the problem of forced transfers a bit less.

## Why It's Good For The Game
A lowered threshold fits better with our current playerbase.

## Changelog
:cl:
tweak: Lowers the base threshold from 3 to 1.
/:cl:


TripleShades:
There's no reason to lower it to 1, as it might as well not be enabled at all at that point.

PR #15889, opened 2022-12-06
Makes RBMK not early return with a prob(80) when there's food on it

Putnam3145:
## About The Pull Request

Title.

## Why It's Good For The Game

It's a bit stupid??

## Changelog
:cl:
fix: prob(80) return should be prob(80) continue
/:cl:


Shadowtail117 (changes requested):
The fix looks good, but probably a good idea to revise the changelog to actually say what the fix is. I.e., "Fixed RMBK not actually cooking things properly."

PR #15890, opened 2022-12-08
Separate character max age into max generated and max input

LeDrascol:
## About The Pull Request
Adds a method to set maximum character age _input_ separately from _randomly generated_ maximum character age. When using the character age setting dialog, and when saving that value, the code will reference AGE_MAX_INPUT instead of AGE_MAX. By default; these two definitions are identical.

Originally requested as a method to remove maximum age for S.P.L.U.R.T. [in this PR](https://github.com/SPLURT-Station/S.P.L.U.R.T-Station-13/pull/618), but deemed too difficult to introduce as modular.

## Why It's Good For The Game
Allows raising or unrestricting the maximum character age a user can set _without_ causing age to randomly generate in the new range.

## Changelog
:cl:
code: Added definition for maximum character age input
/:cl:

### DIFF
--- a/code/modules/client/preferences_savefile.dm
+++ b/code/modules/client/preferences_savefile.dm
@@ -905,7 +905,7 @@ SAVEFILE UPDATING/VERSIONING - 'Simplified', or rather, more coder-friendly ~Car
 	shirt_color = sanitize_hexcolor(shirt_color, 6, FALSE, initial(shirt_color))
 	socks = sanitize_inlist(socks, GLOB.socks_list)
 	socks_color = sanitize_hexcolor(socks_color, 6, FALSE, initial(socks_color))
-	age = sanitize_integer(age, AGE_MIN, SHORT_REAL_LIMIT, initial(age))
+	age = sanitize_integer(age, AGE_MIN, AGE_MAX_INPUT, initial(age))
 	hair_color = sanitize_hexcolor(hair_color, 6, FALSE)
 	facial_hair_color = sanitize_hexcolor(facial_hair_color, 6, FALSE)
 	grad_style = sanitize_inlist(grad_style, GLOB.hair_gradients_list, "None")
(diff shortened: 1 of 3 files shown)


code/modules/client/preferences_savefile.dm
```diff
@@ -905,7 +905,7 @@ SAVEFILE UPDATING/VERSIONING - 'Simplified', or rather, more coder-friendly ~Car
 	shirt_color = sanitize_hexcolor(shirt_color, 6, FALSE, initial(shirt_color))
 	socks = sanitize_inlist(socks, GLOB.socks_list)
 	socks_color = sanitize_hexcolor(socks_color, 6, FALSE, initial(socks_color))
-	age = sanitize_integer(age, AGE_MIN, AGE_MAX, initial(age))
+	age = sanitize_integer(age, AGE_MIN, INFINITY, initial(age))
```
Vizoplez: Time to overflow the age boundary!

LeDrascol: Reducing the value from INFINITY to SHORT_REAL_LIMIT should resolve this issue.

timothyteakettle:
if you want a value like this to be changed for downstream then making the value change upstream **is not the way to do it**

make the value you wish to change configurable upstream such that you can then introduce it as a modular change downstream

in this case it looks like you just want to separate the age for random characters and the maximum input value

you can do this by making a config option to have no limit on the input given, or have two separate defines
a config option would be preferred personally

LeDrascol:
> you can do this by making a config option to have no limit on the input given, or have two separate defines a config option would be preferred personally

Thank you for this suggestion. I hadn't considered that possibility, and agree that it's a better method to use.

Vizoplez:
> if you want a value like this to be changed for downstream then making the value change upstream **is not the way to do it**
> 
> make the value you wish to change configurable upstream such that you can then introduce it as a modular change downstream
> 
> in this case it looks like you just want to separate the age for random characters and the maximum input value
> 
> you can do this by making a config option to have no limit on the input given, or have two separate defines a config option would be preferred personally

Their downstream are complaining to make it upstream.

Major L i guess.

PR #15891, opened 2022-12-11
Updated Delta Station medical's surgery room 

Mazen-AG:
Heyo,

Remolded the medical surgery room and added another surgery table and operating computer.

## About The Pull Request

So for medical in all the other stations it mostly has two surgery tables and two surgery computers in the surgery room area, which is a good start.
Except delta station which had only one surgery table and computer, So I remolded the surgery room to fit two and added more tools to fit the theme of the theater surgery 

Before: 
image

After:
image

How it looks in-game:
image

## Why It's Good For The Game

Medical is one of the most crucial parts of the station, the influx of bodies can start very early and while engineers are busy setting up power doctors may receive many bodies to work on before any engineer is free to remodel or add more surgery equipment to medical, so two tables and computers should make it easier until medical can get more assistance later on if the influx of bodies keep increasing 

## Changelog

:cl:
add: One surgery table with an operating computer next to it
add: Extra surgery tools in the surgery room now that there are to tables to operate on
tweak: moved equipment around to make it more fitting with the new design
qol: moved the AI displays one tile to the side to allow for defibrators mounts near the tables
/:cl:

### DIFF
--- a/_maps/map_files/Deltastation/DeltaStation2.dmm
+++ b/_maps/map_files/Deltastation/DeltaStation2.dmm
@@ -54037,10 +54037,10 @@
 /obj/item/clothing/gloves/color/latex{
 	step_y = 10
 	},
-/obj/item/clothing/mask/surgical{
-	step_y = -10
-	},
 /obj/item/clothing/suit/apron/surgical,
+/obj/item/clothing/mask/surgical{
+	step_y = -6
+	},
 /turf/open/floor/plasteel/white,
 /area/medical/surgery)
 "dsP" = (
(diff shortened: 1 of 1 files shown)


Linzolle:
instead of a giant pile of tools, wouldn't it be easier just to put 2 surgery dufflebags like the other maps have?

Mazen-AG:
@Linzolle wanted to keep the Deltastation's idea of having the tools on display but if you think that just the dufflebags would suffice then I will probably change it
Edit: Changed it into a surgical tool case because it looks way better and is much easier

Mazen-AG:
Changed the surgery tool display to two surgical tool cases

BlueWildrose:
I've been meaning to get around to this but I've never had time to do so. Thank you for this.

PR #15896, opened 2022-12-17
Fix init_process not preventing processing

LeDrascol:
## About The Pull Request
Prevents machines from running faster when set to both **not** init_process and **not** speed_process. Changes mineral machine init_process to use this fix.

Related to a Sandstorm machine issue here: https://github.com/Sandstorm-Station/Sandstorm-Station-13/pull/276

## Why It's Good For The Game
Setting a machine to not start when constructed should actually do that.

## Changelog
:cl:
fix: Fixed non-autostart machines autostarting
/:cl:

### DIFF
--- a/code/modules/mining/machine_processing.dm
+++ b/code/modules/mining/machine_processing.dm
@@ -121,12 +121,12 @@
 	icon_state = "furnace"
 	density = TRUE
 	needs_item_input = TRUE
+	init_process = TRUE
 	var/obj/machinery/mineral/CONSOLE = null
 	var/on = FALSE
 	var/datum/material/selected_material = null
 	var/selected_alloy = null
 	var/datum/techweb/stored_research
-	init_process = TRUE
 
 /obj/machinery/mineral/processing_unit/Initialize(mapload)
 	. = ..()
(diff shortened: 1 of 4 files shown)


code/modules/mining/machine_processing.dm
```diff
@@ -4,7 +4,7 @@
 
 /obj/machinery/mineral
 	speed_process = TRUE
-	init_process = FALSE
+	init_process = TRUE // Must be true after init_process fix
```
SandPoot: I really think this shouldn't be touched

SandPoot:
```suggestion
	init_process = FALSE
```
leave it

SandPoot (changes requested):

code/modules/mining/machine_processing.dm
LeDrascol: This is a required change to prevent mineral machines from no longer processing. The current machines require `speed_process = TRUE` and `init_process = FALSE` due to an error that is being resolved.

SandPoot (approved):

PR #15899, opened 2022-12-17
[PORT] Abductor surgical tools resprite

LeDrascol:
## About The Pull Request
Ports https://github.com/tgstation/tgstation/pull/63127 to this codebase. Old sprites have "_old" appended to the name.

https://gyazo.com/8e57f49e78c2960e7642375fe8decb90

## Why It's Good For The Game
Quoting the original PR by FernandoJ8:

1. Fulfill my cripping addiction to high-tech hard light designs
2. Make the abductors tools fit better with the rest of the abductor equipment, which is quite colorful in comparison to the old tools
3. Make the tools easier to distinguish and less samey between themselves. We've all been there, it's often hard to remember which is the scalpel and which is the cautery as a glance, this fixes that.

## Changelog
:cl:
imageadd: Resprites all of the abductors surgical tools
/:cl:


SandPoot:
I don't think we'd need to keep the old sprites... right?

LeDrascol:
> I don't think we'd need to keep the old sprites... right?

Most likely not. I can remove them from the file if that'd be better.

SandPoot (approved):
It looks generally nice and removes (somewhat) the confusion of not knowing which one does what

PR #15911, opened 2022-12-26
prescription kits, aka the eyewear accessibility update

Hatterhat:
## About The Pull Request
adds the Prescription Lens Kit, which allows any eyewear to be made prescription/vision_correcting

## Why It's Good For The Game
accessibility's a bitch

## Changelog
:cl:
add: Nanotrasen has decided to phase out the old prescription HUD designs, citing redundancy; instead, new prescription lens kits have begun to be rolled out. Check your local techfab or protolathe after Advanced Biotechnology has been researched.
/:cl:


code/modules/clothing/glasses/_glasses.dm
```diff
@@ -27,6 +27,11 @@
 	if(glass_colour_type && ishuman(user))
 		. += "<span class='notice'>Alt-click to toggle its colors.</span>"
 
+/obj/item/clothing/glasses/proc/prescribe()
+	vision_correction = TRUE
+	name = "prescription [src]"
```
Linzolle: i would do "prescription [initial(name)]" here

Hatterhat: i'm just slapping "prescription" on the name because of the nonzero chance someone might have snowflake non-prescription glasses they want to prescription up

Linzolle: yeah i can see that, it would probably just be better to use initial(name) to get the name here instead of src

Hatterhat: think if it was initial(name) it'd get the base name? could do src.name or something idk

PR #15918, opened 2023-01-04
[PORT] Non-global heirloom definitions

LeDrascol:
## About The Pull Request
Port of the current TG heirloom system.

Heirlooms are now stored in job_types and species_types instead of a global list. Items used as heirlooms match their modern TG counterparts when possible.

Also fixes the ancient first aid kit, which is used as a heirloom for three jobs.

## Why It's Good For The Game
Makes assigning and updating heirlooms much easier.

## Changelog
:cl:
tweak: Updated job-specific heirlooms (too many to list)
fix: Fixed the ancient medkit being identical to the normal medkit
code: Heirlooms are listed per-job and per-species instead of globally
/:cl:


timothyteakettle:
this is a better way of organising heirlooms overall and brings us more up to date with tg

PR #15920, opened 2023-01-05
[PORT] Implement ID card age

LeDrascol:
## About The Pull Request
Ports the ID card age display mechanics from current TG.

References to minors and underage crew have been removed. Tightly controlled Nanotransen stations may "enforce" stricter substance usage standards without implying their crew are children.

- Added registered age text when closely examining an ID
- Added an MLDA restriction indicator when presenting an ID
- - Added **unimplemented** ID card age change system for modular computers
- Added minimum legal drinking age (20)
- - Replaces AGE_MINOR with AGE_MIN_DRINKING
- - Replaces narcd_underages with narcd_cantdrink
- Implemented vending machine age restriction check
- - Does not mark any items as restricted

## Why It's Good For The Game
Age flavor text is currently unused outside of security terminals. This change enables non-security crew to use the character creator's age setting.

## Changelog
:cl:
add: Added registered age text when closely examining an ID
add: Added alcohol prohibited indicator (as "MLDA") when presenting an ID
code: Implemented vending machine MLDA restrictions (currently restricts nothing)
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -291,6 +291,7 @@
 #define AGE_MIN					18	// youngest a character can be // CITADEL EDIT - 17 --> 18
 #define AGE_MAX					85	// oldest a character can be randomly generated
 #define AGE_MAX_INPUT			85	// oldest a character's age can be manually set
+#define AGE_MIN_DRINKING		20 	// Youngest a character can legally drink and smoke
 #define WIZARD_AGE_MIN			30	// youngest a wizard can be
 #define APPRENTICE_AGE_MIN		29	// youngest an apprentice can be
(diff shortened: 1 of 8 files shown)


timothyteakettle:
i'm not too sure about the system for reporting to sec if someone under 20 tries to buy alcohol but i feel like it's such a small niche thing that if anyone tries to be annoying with it, it'd be sorted out in-game rather than through code, so this change overall seems fine, especially the new define names and the fact the vending machine behaviour is actually implemented for this

LeDrascol:
> i'm not too sure about the system for reporting to sec if someone under 20 tries to buy alcohol

I agree, but implemented it since the code already existed in a commented-out state. Without further edits to mark items as restricted, it shouldn't do anything.

DeltaFire15:
I do not foresee this to have any beneficial impact.

PR #15931, opened 2023-01-09
[PORT] Updated /tg/station emags

LeDrascol:
## About The Pull Request
Main changes:
- Backports changes to the Cryptographic Sequencer (emag)
- - Cost reduced from 6 TC to 4 TC
- - Can't open airlocks
- - Unlimited charges
- Introduces the Airlock Authentication Override Card (doormag)
- Removes the Electromagnet Charging Device
- Moves emags to a unique file

Additional changes:
- Any syndicate kit containing an emag now also contains a doormag
- Doormag cost reduced from 4 TC to 3 TC to maintain cost parity with original emag
- Does NOT include the Syndicate Jaws of Life

## Why It's Good For The Game
- Prevents the Cryptographic Sequencer from being a "jack of all trades" item
- Makes the standalone item less dangerous overall, reducing possible abuse from alternative sources (ex. Black Market)
- Improves feature parity with TG

## Changelog
:cl:
add: Added Authentication Override Card
del: Removed Electromagnet Charging Device
tweak: Cryptographic Sequencer no longer consumes charges
tweak: Cryptographic Sequencer no longer opens airlocks
tweak: Cryptographic Sequencer cost reduced from 6 TC to 4 TC
/:cl:


deathride58:
The removal of emag charges is the part I'm most uncertain of here. The reason emag charges were added here (Not sure if TG ever had emag charges to begin with? Which makes the statement of unlimited charges being a feature backport somewhat confusing) is due to the sheer amount of abuse that unlimited emag charges had. Notably, unlimited emag charges means that if one object/door is emagged, then it's safe to assume that *anything* on the station can also be emagged, and in practice, this means mostly *everything* ends up emagged when a traitor buys one. This also makes it particularly annoying to deal with from the side of non-antag crewmembers, as being able to emag an unlimited number of objects greatly limits the ability to investigate what's being emagged and why, in turn making it pretty hard to deduce who's doing the emagging without witnessing it during the act itself. Needless to say, unlimited emag charges had a tendency to be *heavily* abused here, as it was far easier to greentext by emagging as many things as possible. Additionally, emags being unlimited is directly antithetical to the type of environment this codebase is going for, as it allows a single traitor to end up causing *way* too much chaos for such little investment, even if their ability to open airlocks were removed entirely (as was experimented with prior to the charge system being added to emags).

Emag charges add a layer of decision-making to emagging. Instead of the pattern that antags used to have where anyone buying an emag simply emagged everything they came across, antags have to think carefully about the effects of what they're emagging, though the limited amount of charges means there's definitely some leeway. This has also lead to emagging being reasonably plausible to investigate, as things that end up emagged usually end up being related to a traitor's goals in some way, shape, or form, with traitors often having to use up the rest of their charges on unrelated objects to obfuscate their trail. Additionally, the limited emag charges means there's significantly less need to nerf various emag effects (as TG has done over the years), as emagging something has an actual cost instead of being something you can do infinitely. Emag charges also limits the potential harm that can be caused by a nonantag crewmember ending up with an emag in some way, in turn limiting the harm of opportunistic griefing. Emags having limited charges also has the effect of encouraging antags to learn each department's airlock wires, as doing so allows conserving emag charges, consequently making hacking a fairly valuable skill to have, while also reducing the rampancy of emagged airlocks.

That being said, authentication override cards with their recharging mechanic, combined with the presence of limited emag charges, and emag charges retaining their ability to open airlocks, could probably make for a fairly interesting dynamic. Emag rushes would still be as viable as they are now, but those seeking to conserve emag charges would have a choice between spending the extra TC to unlock up to 3 airlocks before having to wait, or cheaping out and instead sticking with learning departmental wires to open doors the hard way.

LeDrascol:
Thank you for the extensive feedback on this matter. I agree that unlimited emags charges is a dubious idea at best, and am in favor of keeping them limited in some way.

An alternative could be to give them a similar behavior to the doormag; Reduced uses with a recharge timer. This would still technically allow hacking an unlimited number of things, but much slower.

> That being said, authentication override cards with their recharging mechanic, combined with the presence of limited emag charges, and emag charges retaining their ability to open airlocks, could probably make for a fairly interesting dynamic.

Would allowing both cards to open airlocks also require reverting the TC price changes, or the inclusion of doormags in kits that had emags? Both of these changes were made to maintain functional parity while using a non-airlock emag.

DeltaFire15:
I do not think the /tg/ version does things better than our finite charges-based approach.

PR #15936, opened 2023-01-12
prescription lens kit fix

Hatterhat:
## About The Pull Request
title. used the wrong attack proc? dunno, it worked when i tested it maybe i forgot to push the actual fix
## Why It's Good For The Game
make thing work

## Changelog
lol oops


timothyteakettle (changes requested):
currently this means you use the glasses on the kit which is not in the order you would normally apply things

two options:
1. override attackby on glasses (see: accessories)
2. override afterattack on the prescription kit (see: durathread kit)

timothyteakettle (approved):

PR #15948, opened 2023-01-22
[DO NOT MERGE] Harddel unit test

KrissKr0ss:
## About The Pull Request
Ports over https://github.com/tgstation/tgstation/pull/59612

```
Adds a unit test that runs at the very end of all other tests. It creates and then instantly destroys one of each type path. Some are blacklisted because of various reasons.

This won't catch anywhere near everything, but it'll help a lot with regressions, and catches a lot of runtimes too. The really nasty stuff is things that sleep as a part of initialize, since we don't init, wait, then qdel since it's a lot more prone to error

Oh and I went through and fixed all the errors I could, alongside adding descriptive errors on the stuff that's just gonna runtime in init.

This is gonna cause some unit test failures, but hopefully it'll just catch the simple stuff, which should be easy for people to fix.
```

## Why It's Good For The Game

[The Guide](https://github.com/Citadel-Station-13/Citadel-Station-13/blob/147df0e8fc9282b30fa8df11af88132ec0f855b5/.github/guides/HARDDELETES.md)

This'll allow us to prevent hard delete regression, and it'll catch a good amount of general errors too since we're guaranteed to hit the init and destroy procs of nearly every typepath, which should help prevent runtime spam from say, spooky material not cleaning up properly or whatever.

## Changelog
:cl:
code: adds a unit test that looks for harddels
/:cl:


Mothblocks:
image

image

DeltaFire15 (commented):
Still failing its own added test for the time being.

PR #15962, opened 2023-01-31
Contextual screentips -- Screentips now show you what items/objects can do

SandPoot:
## About The Pull Request
***Port of https://github.com/tgstation/tgstation/pull/64502***
image
dreamseeker_OgKW6xm5Kx
dreamseeker_BOj9ueD8ji

No checking for corrections necessary anymore, however, testing it a bit should still be good.

## Why It's Good For The Game
Provides an easy system to use for teaching through gameplay what objects can do. Does not require the use of shift-click, which requires the player already be inquisitive on what something does. This will fail in cases such as new players who shouldn't have to shift-click everything they see, and for cases where an alternate behavior is not apparent, such as firelocks (I already know LMB with a crowbar opens it, why would I ever examine it?). Also helpful for cases where we change existing behavior, such as fire alarms, where I remember Zxaber creating huge admin notices telling people what the new things did.

## Changelog
:cl:
qol: Hovering over specific objects, over hovering with specific items, will now tell you what they can do. For example, hovering a firelock with a crowbar will tell you can both left-click to hold open, and right-click to keep open. The preference for screentips has been changed from on/off to on/off/only with tips. If you had your preference on off before, it has been changed to "only with tips". That way, it's not as intrusive as when it's always on, but you can still pick up information when it can actually be provided. Try to give it a shot, and don't disable it right away.
/:cl:


timothyteakettle:
note: testmerge first

SandPoot:
All done and waiting for review

PR #15968, opened 2023-02-08
[Port-TG] Add defines for byond-tracy support 

Phoenix4O4:
Not intended for use in production, and we do not ship a copy of byond-tracy. It is extremely easy to compile yourself, but if you're someone interesting enough to play around with this then let me know and I can just give you a build.
I'm going to be using this for init profiling.
image

**Port** [Add defines for byond-tracy support #70931](https://github.com/tgstation/tgstation/pull/70931)

## Why It's Good For The Game
Metrics

## Changelog
Developer stuff


tools/tracy_replay/replay.py
DeltaFire15, line 1: Shouldn't this file not be in this PR if you are only adding the integration and not the tool itself (like the ported PR)?

code/_compile_options.dm
```diff
@@ -32,6 +32,13 @@
 
 #endif //ifdef REFERENCE_TRACKING
 
+// If this is uncommented, we do a single run though of the game setup and tear down process with unit tests in between
+// #define UNIT_TESTS
```
DeltaFire15, line 36: duplicate define located in l. 53. Remove the old one if you want to move it here.

tools/tracy_replay/replay.py
4310V343k: replay.py is already [deprecated](https://github.com/ParadiseSS13/Paradise/blob/bf5a82365a3bfb42628a4b6670d867248956eeb3/tools/tracy_replay/README.MD) in favour of https://github.com/AffectedArc07/ParaTracyReplay, so yeah, it shouldn't be

SandPoot:
I'm not bothering with this anymore, i compiled the file, set up defines and all and still nothing.

PR #15987, opened 2023-07-03
Jukebox bugs

KrissKr0ss:
## About The Pull Request
Fixes the bug with some jukeboxes where for some reason the proc will halt after calling activate_music() and only run the rest after a song has stopped playing. I have no idea as to why this happens

## Why It's Good For The Game
I believe this problem can cause lag when adding a lot of songs to the queue

## Changelog
:cl:
fix: fixes a bug when adding songs to a jukebox queue
/:cl:

### DIFF
--- a/code/game/machinery/dance_machine.dm
+++ b/code/game/machinery/dance_machine.dm
@@ -121,7 +121,7 @@
 			if(!allowed(usr))
 				return
 			if(!active && !playing)
-				activate_music()
+				INVOKE_ASYNC(src, .proc/activate_music)
 			else
 				stop = 0
 			return TRUE
@@ -152,7 +152,7 @@
 			if(active)
 				say("[selectedtrack.song_name] has been added to the queue.")
 			else if(!playing)
-				activate_music()
+				INVOKE_ASYNC(src, .proc/activate_music)
 			playsound(src, 'sound/machines/ping.ogg', 50, TRUE)
 			queuecooldown = world.time + (3 SECONDS)
 			return TRUE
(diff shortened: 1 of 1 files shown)


deathride58:
The lag from jukeboxes comes from byond itself (it doesn't even matter if you give the sound the stream flag; clients receiving sounds is *always* synchronous on the client's end, and blocks the client's main thread whether you want it to or not. Meanwhile, asset sending is always asynchronous on the server's end, but this tends to bloat the tick due to how expensive asset sending can be, and large files/large amounts of files can *massively* bloat the tick, or in some cases, oversaturate the server's upload bandwidth), and activate_music() does not block code execution in a meaningful way due to how jukebox code is designed.

 The only costs associated with having a lot of songs in the jukebox queue are that of a singular list insertion and a few bytes allocated in memory to keep a ref of a jukebox song datum in that list. To be frank, this PR doesn't fix any bugs up here, given that this PR doesn't affect the control flow in a meaningful way. You can test this yourself on local by putting an excessively large audio file (500+ MB is ideal for gauging how it performs in the worst case scenario) in your jukebox directory.

As an aside, to alleviate the lag from jukeboxes, it's ideal to keep sounds as small as feasibly possible. Through testing, I've found that 2 MB is the sweet spot maximum for jukebox audio size, as it keeps the clientside lagspike fairly reasonable (it's less than half a second for those with bad-but-otherwise-playable internet connections), and simply doesn't starve the tick at high playercounts (under 2 MB, it carries about the same cost as a client changing to a z-level they don't have cached). In theory, this could be avoidable entirely by including jukebox audio in the RSC, but byond currently doesn't provide a good solution for doing so (as the only way to do so is to hardcode each individual file, one by one).

As for this PR, I'd like to see proof of testing here, as this doesn't appear to be doing anything meaningful.

DeltaFire15:
Normal jukeboxes do not seem to have any sleeping calls (only the discoball does), which are the only thing that INVOKE_ASYNC helps against.

PR #15991, opened 2023-07-19
Adds a couple hints to things

SandPoot:
## About The Pull Request
Title.

## Why It's Good For The Game
This could show some players what they can do and even obscure things.

## Changelog
:cl:
add: Adds many hints to various objects.
/:cl:

### DIFF
--- a/code/modules/clothing/under/_under.dm
+++ b/code/modules/clothing/under/_under.dm
@@ -99,12 +99,13 @@
 		has_sensor = BROKEN_SENSORS
 		sensordamage = sensormaxintegrity
 
-/obj/item/clothing/under/New()
+/obj/item/clothing/under/Initialize(mapload)
 	if(sensor_flags & SENSOR_RANDOM)
 		//make the sensor mode favor higher levels, except coords.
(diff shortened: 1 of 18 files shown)


code/modules/clothing/under/_under.dm
```diff
@@ -104,6 +104,7 @@
 		//make the sensor mode favor higher levels, except coords.
 		sensor_mode = pick(SENSOR_OFF, SENSOR_LIVING, SENSOR_LIVING, SENSOR_VITALS, SENSOR_VITALS, SENSOR_VITALS, SENSOR_COORDS, SENSOR_COORDS)
 	sensor_mode_intended = sensor_mode
+	register_context()
```
Linzolle: is this runtiming because it's under New instead of Initialize?

Linzolle: oh wait you changed it from initialize... that's weird that it would still be doing that. does it need to be under its own initialize seperate from the sensor shit?

SandPoot: It was already new, i changed to initialize then back, how the runtime happens doesn't make sense.

Linzolle: i mean yeah you tried to as initialize and it didn't work, which is weird .....

SandPoot:
@Linzolle i think it's fixed now

PR #16020, opened 2023-10-28
[Testmerge ONLY] Revert accessories autoequip

BlueWildrose:
players need to change loadout during ball. don't have time to bugfix.


BlueWildrose:
Fixed on behalf of #16022

PR #16027, opened 2023-11-13
Some modsuit updates

DeltaFire15:
## About The Pull Request
Ports https://github.com/tgstation/tgstation/pull/63749 (minus waddle modules which is commented because we don't have the element, or some few other module changes like to antag modules because we do not have those ported).

Additionally does some updating of my own, such a reverting a skyrat decision to make helmets retractable while the suit is active (contrary to the point of modsuits), aswell as making all modsuit helmets internals-compatible like hardsuits when sealed, and making modsuits count as viable suitlinking targets for Synthetics while active.

(sprites the ported PR adds were already ported in the previous modsuit PR and thus aren't referred to in the CL credits)

## Why It's Good For The Game
Modsuits cool, outdated code not cool.

## Changelog
:cl: Fikou, descriptions by Nerevar and Halcyon, additional modifications by DeltaFire15
add: (Port) you can see extended descriptions of modsuit themes by examining them closely.
add: (Port) All MODsuits now have an extended description viewable by examining them closely.
balance: (Port) MODsuits no longer slow you down when undeployed in backpack form (instead applying partial slowdown depending on parts deployed).
balance: (Port) The Research MODsuit type is now slightly faster.
balance: (Port) The  Prototype MODsuit type is now slightly faster but less energy efficient.
fix: (Port) MODsuit chestplates no longer inherit spacesuit slowdown, making you much much slower than intended.
balance: MODsuit helmets are once again not retractable while active.
balance: All MODsuit helmets can now be used as internals connectors while sealed (active).
balance: MODsuits are now valid suitlink targets for Synthetics, provided they are deployed and active.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -713,6 +713,12 @@ GLOBAL_LIST_INIT(ballmer_windows_me_msg, list("Yo man, what if, we like, uh, put
 	if(istype(head_item, /obj/item/clothing/head/helmet/space) && istype(suit_item, /obj/item/clothing/suit/space))
 		return 1
 
+	if(istype(head_item, /obj/item/clothing/head/mod) && istype(suit_item, /obj/item/clothing/suit/mod))
+		var/obj/item/clothing/suit/mod/modsuit = suit_item
+		var/obj/item/mod/control/mod_control = modsuit.mod
+		if(mod_control && mod_control.active)
+			return 1
+
 	if(T && is_mining_level(T.z) && istype(head_item, /obj/item/clothing/head/hooded/explorer) && istype(suit_item, /obj/item/clothing/suit/hooded/explorer))
 		return 1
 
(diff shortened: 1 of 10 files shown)


SandPoot (commented):
I dunno about the re-removal of helmet toggling while active, suppose both i and other folks used it a bunch.

DeltaFire15:
> I dunno about the re-removal of helmet toggling while active, suppose both i and other folks used it a bunch.

You are not supposed to be able to wear helmets together with a ~~Flight-~~ MODsuit, which this allows (and with no disadvantage at all at that, since all MOD-components are only related to the mod core itself). Besides that it'd also mess with the updated part-based slowdown logic (not the most critical since it'd be somewhat easy to fix and not a main consideration of my change).
Secondarily, you are also intended to not be as easily able to consume / inject things with an engaged suit, which there is a specific module for (that has its own tradeoffs). Toggling helmets does this for free, and much more efficiently.

deathride58 (approved):

PR #16043, opened 2023-12-13
Make stacks show how much there are in inventory

SandPoot:
## About The Pull Request
Items that stack will always show their quantity if not on a /turf, such as being inside a bag who is on the floor.
This
***From [TGMC/8566](https://github.com/tgstation/TerraGov-Marine-Corps/pull/8566)***

## Why It's Good For The Game
Easier inventory management.

## Changelog
:cl:
qol: stacks now show how many items there are when in your inventory.
/:cl:


deathride58 (approved):

code/game/objects/items/stacks/stack.dm
```diff
@@ -132,6 +132,14 @@
 	else
 		icon_state = "[initial(icon_state)]_3"
 
+/obj/item/stack/update_overlays()
+	. = ..()
+	if(isturf(loc))
+		return
```
BlueWildrose, line 138: Doesn't ore stacks make use of overlays for their own stack system? Wouldn't this interfere with that?

BlueWildrose (commented):
Just a minor concern I got.

code/game/objects/items/stacks/stack.dm
SandPoot: This is a legitimate issue but also storage code sucks, i'm stumped and i hate it.

deathride58: This should probably just be `return .`

From there that should Just Work

```suggestion
		return .
```

SandPoot: Not actually a problem, `return` always returns what's on `.`

SandPoot:
Reminder that this isn't good only because of storages that show numbers for stacks.

deathride58 (approved):

PR #16075, opened 2024-04-13
throttles mouse movement calls

deathride58:
So right now it's possible to DoS a citmain-running server by simply cranking your client framerate up as far as your computer can handle, turning on combat mode, and shaking the ever-living Fuck out of your mouse. This causes Dreamseeker to send the server a mouse movement command for *every single pixel your mouse moves*

That is, needless to say, not ideal!

Even moreso considering `MouseMove()` has some distinctly non-negligible costs associated with it (most notably: a signal call).

So this PR makes the cost of `MouseMove()` far more consistent (scaling far more predictably with playercounts) with one simple trick: shrimply throttling the proc.

Judging from a quick test at both 0fps (server tickrate locked) and 100fps, it seems like this works.

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -92,7 +92,11 @@
 	. = 1
 
 //Please don't roast me too hard
+//Oh don't worry, We Will.
 /client/MouseMove(object, location, control, params)
+	if(next_mousemove > world.time)
+		return
+	next_mousemove = world.time + world.tick_lag
 	mouseParams = params
 	mouse_location_ref = WEAKREF(location)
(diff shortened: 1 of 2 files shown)


code/_onclick/drag_drop.dm
```diff
@@ -92,7 +92,11 @@
 	. = 1
 
 //Please don't roast me too hard
+//Oh don't worry, We Will.
 /client/MouseMove(object, location, control, params)
+	if(next_mousemove > world.time)
```
Poojawa:
```suggestion
	if(prob(80))
```

PR #16077, opened 2024-04-13
Upstream a number of fixes from downstream

liz-lavenza:
## About The Pull Request
Since apparently you all think we don't want to do work to fix issues...
image
And badmouth us while taking our fixes...
image
I made this PR to prove that we *can* contribute upstream, if we have good reasons to. Not that the attitude of certain people here makes me want to do so, but I'm doing this to extend an olive branch of sorts.

## Why It's Good For The Game
Ports `set_stat()` and `COMSIG_MOB_STATCHANGE` from TG.
Fixes the following issues:
- `disease_cough`, `disease_sneeze`, `disease_infect`, `spawn_induction_package`, `slam`, `scavenge`, `coffer`, and `riot` cooldown actions need to be refactored to use `Activate()` instead of `Trigger()`, and to not check their parent's return value.
- The BSA control computer, the nanite cloud controller computer, the cyborg integrated welding tool, computer batteries, the cyborg mounted laser, the scattershot laser rifle, the instakill rifle, the red instakill rifle, the blue instakill rifle, and the miniature egun all have invalid overlay states.
- Actions that use `AB_CHECK_CONSCIOUS` need to update their icon on `COMSIG_MOB_STATCHANGE`, which requires porting `COMSIG_MOB_STATCHANGE`.
- Any closet that uses door icon state auto-setting is missing its door.

## Changelog
:cl: liz-lavenza (Porting from S.P.L.U.R.T.)
fix: Fixes closets missing their doors.
fix: Fixes mobs not updating their iconstate on status change.
fix: Fixes invalid overlays for the BSA control computer, the nanite cloud controller computer, the cyborg integrated welding tool, computer batteries, the cyborg mounted laser, the scattershot laser rifle, the instakill rifle, the red instakill rifle, the blue instakill rifle, and the miniature egun.
fix: Fixes various actions relying on legacy code.
/:cl:


code/modules/mob/living/simple_animal/hostile/regalrat.dm
```diff
@@ -131,9 +129,10 @@
 	button_icon_state = "riot"
 	background_icon_state = "bg_clock"
 	cooldown_time = 80
+	check_flags = AB_CHECK_CONSCIOUS
 	///Checks to see if there are any nearby mice. Does not count Rats.
 
-/datum/action/cooldown/riot/Trigger()
+/datum/action/cooldown/riot/Activate()
 	. = ..()
```
SandPoot:
```suggestion
```

code/modules/mob/living/simple_animal/hostile/regalrat.dm
```diff
@@ -131,9 +129,10 @@
...
 	if(!. || owner.stat != CONSCIOUS)
```
SandPoot:
```suggestion
	if(owner.stat != CONSCIOUS)
```

SandPoot (approved):

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -76,7 +77,7 @@
 		. += "[icon_door_override ? icon_door : icon_state]_open"
 		return
 
-	if(icon_door)
+	if(has_door_icon)
 		. += "[icon_door || icon_state]_door"
```
deathride58, line 81: This `if()` isn't necessary since `icon_door` is null by default
```suggestion
	. += "[icon_door || icon_state]_door"
```

deathride58 (changes requested):
Most of the PR looks good; just a single thing stands out

code/game/objects/structures/crates_lockers/closets.dm
SandPoot: It turns out that whoever added the icon_door variable didn't set it up on every closet and now the overlay isn't being added, can be checked on main.

liz-lavenza: That is incorrect. The `cardboard_door` state for the cardboard box 'closet' does not (and should not) exist. The original `if(icon_door)` was added by SandPoot to mask this error, but it also broke anywhere that relied on an auto-set icon_state. This new `if(has_door_icon)` I added *disables adding the door icon entirely* for the cardboard box (and anything else that could possibly need it), rather than just clumsily skipping it if `icon_door` is unset.

SandPoot, the point of using `icon_door || icon_state` is to make it so that it doesn't have to be set on every closet. `icon_door` is specifically for closets whose door icon state differs than their base icon state. (This is not to be confused with `icon_door_override`, which is used when the *open* door icon state differs from the *closed* door icon state, I believe?)

deathride58: When an iconstate is missing from the DMI (such as for `cardboard_door`), Byond simply uses whatever iconstate in the icon has a null name. In the case of lockers, there's no iconstate with a null name, and thus `cardboard_door` will display nothing. Cardboard boxes should definitely be converted to the newer separated door format, though.

That being said, it feels a little overkill to have this check in place for a single edge-case, but the reasoning is fair enough, all things considered.

deathride58 (approved):

code/game/objects/structures/crates_lockers/closets.dm
liz-lavenza: The existing fix was worse, and variables with their default value don't take up any memory cost per-object. I'd rather have overkill that works than overkill that doesn't.

PR #16106, opened 2024-10-15
Chem grinder change

sheepishgoat:
## About The Pull Request
Modifiers the chem grinders in two different ways:

- Chem grinders can now have their input direction changed with CTRL+SHIFT+CLICK
- Chem grinders now extract any reagents from inside of the things that they grind.

## Why It's Good For The Game
This is worthwhile because it makes the chem grinders less obtuse to use.

## Changelog
:cl:
fix: chem grinders now extract any reagents from inside of the things that they grind.
qol: Chem grinders can now have their input direction changed with CTRL+SHIFT+CLICK
/:cl:


lolman360:
grind functionality is ok, but why the input changes? you could always rotate them so they could input from any direction, so this change just makes it easier for you to accidentally grind stuff instead of deliberately input it

sheepishgoat:
> grind functionality is ok, but why the input changes? you could always rotate them so they could input from any direction, so this change just makes it easier for you to accidentally grind stuff instead of deliberately input it

Now that I am thinking about it,  there is probably a better way to do this.

There we go.

deathride58 (approved):

deathride58:
Factorio gaming

PR #16111, opened 2024-10-25
Removes mob emote spam in chat

sheepishgoat:
## About The Pull Request
Does what the title says. Emotes from mobs without a client that are far away will no longer be displayed in chat for ghosts.

## Why It's Good For The Game
This makes observing and keeping track of things so much more bearable.

## Changelog
:cl:
qol: Mobs without clients no longer send emotes to ghosts, unless they are close.
/:cl:

### DIFF
--- a/code/datums/emotes.dm
+++ b/code/datums/emotes.dm
@@ -57,12 +57,13 @@
 	user.log_message(msg, LOG_EMOTE)
 	msg = "<span class='emote'><b>[user]</b> [msg]</span>"
 
-	for(var/mob/M in GLOB.dead_mob_list)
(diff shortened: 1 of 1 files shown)


code/datums/emotes.dm
```diff
@@ -58,7 +58,7 @@
 	msg = "<span class='emote'><b>[user]</b> [msg]</span>"
 
 	for(var/mob/M in GLOB.dead_mob_list)
-		if(!M.client || isnewplayer(M))
```
SandPoot, line 61: Don't check this for each dead, do it outside the loop

SandPoot (approved):
Yes

PR #16126, opened 2025-03-04
More action updates

SandPoot:
## About The Pull Request
Title.

## Why It's Good For The Game
It's cool.

## Changelog
:cl:
fix: Inventory checks properly for showing if you can wear something.
fix: Action drop points no longer linger on screen.
refactor: Touched action code a lot.
qol: Item based actions can be granted in specific slots now.
qol: Observers can see players' action buttons.
/:cl:


code/__DEFINES/dcs/signals/signals_object.dm
```diff
@@ -0,0 +1,96 @@
+// /obj/item signals
```
silicons, line 1: nitpick - for all of the dcs files, it's best to have signals be like this
```
/// text
#define
```

or atleast like this
`#define ... //! text`

this way langserv comments it

not really too important but just as a note

SandPoot: ah, my bad, if you could tell i was reorganizing stuff, moving the object/item stuff into its own file

